PR #37, opened 2018-03-16
SWIFT-41, SWIFT-42

kmahar:
Since I ended up having to edit `MongoCursor` anyway I lumped in the cursor error handling stuff with this. 

For the optional pointers, I've added a `unwrapX` method to each type that should be used rather than directly accessing the pointers. We might want to think about doing some kind of checking in `Document` where we are keeping around these forced `UnsafeMutablePointer<bson_t>!`s, but I think it probably makes more sense to hold off on that until the `Document` as`struct` refactor is merged in. 

For the cursor errors, I added a `getError` function as well as a way to iterate the cursor `nextOrError` that does throw, if you really want to do that instead of checking at the end. Both of these feel kinda weird to me though so let me know if you have any better ideas

### DIFF
--- a/Sources/MongoSwift/Cursor.swift
+++ b/Sources/MongoSwift/Cursor.swift
@@ -6,8 +6,7 @@ public class MongoCursor: Sequence, IteratorProtocol {
     private var _client: MongoClient?
 
     /**
-     *  Initializes a new MongoCursor instance. This assumes the caller has checked
-     *  the validity of the cursor pointer and the client's internal pointer.
+     * Initializes a new MongoCursor instance, not meant to be instantiated directly
      */
     internal init(fromCursor: OpaquePointer, withClient: MongoClient) {
         self._cursor = fromCursor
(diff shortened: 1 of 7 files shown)


mbroadst:
@kmahar sorry, but can we actually split this into two PRs? It's confusing to focus on the cursor changes with all the unwrapping stuff

Sources/MongoSwift/Cursor.swift
```diff
@@ -24,19 +24,26 @@ public class MongoCursor: Sequence, IteratorProtocol {
...
         self._cursor = nil
-        self._client = nil
     }
 
     /**
      * Returns the next document in this cursor, or nil
      */
     public func next() -> Document? {
+        do {
+            _ = try unwrapCursor()
```
kmahar: while we should technically save and use this later on in the function rather than the direct accesses to `self._cursor`, I wrote it this way to avoid a big diff from wrapping everything in a `do/catch`since the function can't throw... the state of this function will be pretty temporary considering the changes I have now separated out to deal with cursor errors

kmahar:
OK I have removed the cursor error changes

kmahar:
closing per our conversation earlier, I'll just put the collection name + db name stuff in a separate PR

PR #41, opened 2018-03-21
Add missing convenience initializers

kmahar:
This adds an `IntType` protocol which `Int`, `Int32`, and `Int64` conform to. This will allow users to pass in any of the three types.
Also added a few options initializers in places we forgot them the first time around.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -12,19 +12,19 @@
       },
       {
         "package": "libbson",
-        "repositoryURL": "ssh://git@github.com/10gen/swift-bson",
+        "repositoryURL": "https://github.com/mongodb/swift-bson",
         "state": {
           "branch": "master",
-          "revision": "725977b1c0d3a9f7938d0bec0522edc6cdc95f83",
+          "revision": "077a71336aead18a26223b6c3c07f94478c7921a",
           "version": null
         }
       },
       {
         "package": "libmongoc",
-        "repositoryURL": "ssh://git@github.com/10gen/swift-mongoc",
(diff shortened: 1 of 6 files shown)


kmahar:
re: our conversation earlier: this was mildly useful to me in a few places in the command monitoring tests where `libbson` parsed values to the wrong numeric types and I was trying to pass values directly from a document into an `options` struct (for example [here](https://github.com/mongodb/mongo-swift-driver/pull/42/files#diff-362ef17b0f12333242e708d4546705fcR148), where `batchSize` comes out as an `Int64` but `FindOptions` requires an `Int32`.) (I think it also would have made crud tests slightly easier.)

however, I am not sure how generally useful it is to users, because people probably just type out the options and enter the numbers as literals most of the time. under that assumption we also shouldn't change the initializers to take `Int`s for every number though, because then we have to do extra work to convert the values to the correct types. whereas, with the state of things before this change, we get that typing for free because `Int`, `Int32`, and `Int64` are all `ExpressibleByIntegerLiteral`. 

so as I think about it more I'm not sure having this built in makes a lot of sense, and I ended up having to do so much hacky special-casing in the command monitoring tests anyway that this doesn't present a significant improvement. and this was mostly a find+replace job so I'm not too attached to the work lol. let me know what you think, either way we can keep the convenience initializers we were missing.

mbroadst:
@kmahar it sounds like we should revert the `IntType` change and just add the convenience ctors here, let's do that.

kmahar:
kk, updated

PR #42, opened 2018-03-22
SWIFT-36, 37: Implement idiomatic support for command monitoring and corresponding spec tests

kmahar:
(This uses the changes from the other PR regarding integers)

This adds support for command monitoring via `Notification`s and the default `NotificationCenter`. Also implements the corresponding spec tests. I encountered a number of odd edge cases there which I have attempted to document

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -1,0 +1,197 @@
+import Foundation
+import libmongoc
+
+/// A struct representing a server connection, consisting of a host and port.
+public struct ConnectionId {
+    let host: String
+    let port: UInt16
+
+    internal init(_ hostList: UnsafePointer<mongoc_host_list_t>) {
+        var hostData = hostList.pointee
+        self.host = withUnsafeBytes(of: &hostData.host) { (rawPtr) -> String in
+            let ptr = rawPtr.baseAddress!.assumingMemoryBound(to: CChar.self)
+            return String(cString: ptr)
+        }
+        self.port = hostData.port
+    }
+}
+
+/// An event published when a command starts. The event is stored under the key `event`
+/// in the `userInfo` property of `Notification`s posted under the name .commandStarted.
+public struct CommandStartedEvent {
+    /// An internal opaque pointer to a mongoc_apm_command_started_t 
+    internal let _event: OpaquePointer
+
+    /// An internal initializer for a CommandStartedEvent
+    internal init(_ event: OpaquePointer) {
+        self._event = event
+    }
+
+    /// The command.
+    var command: Document {
+        let commandData = UnsafeMutablePointer(mutating: mongoc_apm_command_started_get_command(self._event)!)
+        return Document(fromPointer: commandData)
+    }
+
+    /// The database name.
+    var databaseName: String { return String(cString: mongoc_apm_command_started_get_database_name(self._event)) }
+
+    /// The command name.
+    var commandName: String { return String(cString: mongoc_apm_command_started_get_command_name(self._event)) }
+
+    /// The driver generated request id.
+    var requestId: Int64 { return mongoc_apm_command_started_get_request_id(self._event) }
+
+    /// The driver generated operation id. This is used to link events together such
+    /// as bulk write operations.
+    var operationId: Int64? { return mongoc_apm_command_started_get_operation_id(self._event) }
+
+    /// The connection id for the command.
+    var connectionId: ConnectionId { return ConnectionId(mongoc_apm_command_started_get_host(self._event)) }
+}
+
+/// An event published when a command succeed. The event is stored under the key `event`
+/// in the `userInfo` property of `Notification`s posted under the name .commandSucceeded.
+public struct CommandSucceededEvent {
+    /// An internal opaque pointer to a mongoc_apm_command_succeeded_t 
+    internal let _event: OpaquePointer
+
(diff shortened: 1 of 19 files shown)


Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -0,0 +1,432 @@
...
+
+                // 1. Setup the specified DB and collection with provided data
+                let db = try client.db(testFile.databaseName)
+                try db.drop() // In case last test run failed, drop to clear out DB
+                let collection = try db.createCollection(testFile.collectionName)
+                _ = try collection.insertMany(testFile.data)
+
+                // 2. Add an observer that looks for all events
+                var expectedEvents = test.expectations
+                let observer = center.addObserver(forName: nil, object: nil, queue: nil) { (notif) in
```
kmahar: FYI, `forName` could be `.commandSucceeded`, `.commandStarted`, etc. if you only want particular named events

Sources/MongoSwift/APM.swift
```diff
@@ -0,0 +1,206 @@
...
+     *  If no specific event types are provided, all events will be posted. 
+     *
+     *  Calling this function will reset all previously enabled events - i.e.
+     *  calling
+     *      client.enableNotifications(forEvents: [.commandStarted])
+     *      client.enableNotifications(forEvents: [.commandSucceeded])
+     *
+     *  will result in only posting notifications for .commandSucceeded events.
+     */
+    public func enableNotifications(forEvents events: [MongoEvent] =
```
mbroadst: nit: `enableCommandMonitoring` ? "Notifications" sounds pretty generic

Sources/MongoSwift/APM.swift
```diff
@@ -0,0 +1,206 @@
...
+}
+
+/// An internal callback that will be set for "command started" events if the user 
+/// enables those notifications. This function generates a new `Notification` and posts 
+/// it to the default NotificationCenter for the app. 
+internal func commandStarted(_event: OpaquePointer?) {
+    guard let event = _event else { return }
+    let eventStruct = CommandStartedEvent(event)
+    let notification = Notification(name: .commandStarted, userInfo: ["event": eventStruct])
+    NotificationCenter.default.post(notification)
```
mbroadst: we should explore ways to support non-default `NotificationCenter`s. One way might be to accept a `withNotificationCenter: NotificationCenter` that defaults to `NotificationCenter.default` in `enableCommandMonitoring` (I'm now starting to think that should be an imperative `monitorCommands([ ... ], withNotificationCenter: ... )` ? ).  Inside that method you can register a lambda that uses the provided `NotificationCenter`

Sources/MongoSwift/APM.swift
```diff
@@ -0,0 +1,206 @@
...
+}
+
+/// An event published when a command starts. The event is stored under the key `event`
+/// in the `userInfo` property of `Notification`s posted under the name .commandStarted.
+public struct CommandStartedEvent {
+    /// An internal opaque pointer to a mongoc_apm_command_started_t 
+    internal let _event: OpaquePointer
+
+    /// An internal initializer for a CommandStartedEvent
+    internal init(_ event: OpaquePointer) {
```
mbroadst: Rather than storing the opaque pointer here, I think you should "fill out" the struct with the data retrieved from the pointer. It might be unlikely that someone else deletes the pointer, but it _might_ happen.

Also what are the ownership rules for this pointer? Are we responsible for deleting it when we're done with it? If so, then we are presently leaking memory.

Sources/MongoSwift/APM.swift
```diff
@@ -0,0 +1,228 @@
...
+     *  Notifications can only be enabled for a single NotificationCenter at a time.
+     *
+     *  Calling this function will reset all previously enabled events - i.e.
+     *  calling
+     *      client.enableNotifications(forEvents: [.commandStarted])
+     *      client.enableNotifications(forEvents: [.commandSucceeded])
+     *
+     *  will result in only posting notifications for .commandSucceeded events.
+     */
+    public func enableCommandMonitoring(
```
kmahar: I'm wondering if we should rename this to also encompass SDAM monitoring - because the callbacks are set all at once for command monitoring + SDAM monitoring, we'll need the logic to create the `mongoc_apm_callbacks_t` in a single place. `enableApplicationPerformanceMonitoring`? or maybe just `enableMonitoring`?

Sources/MongoSwift/APM.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    /// The failure, represented as a MongoError.
+    let failure: MongoError
+
+    /// The client generated request id.
+    let requestId: Int64
+
+    /// The driver generated operation id. This is used to link events together such
+    /// as bulk write operations.
+    let operationId: Int64?
```
mbroadst: is this really optional?

mbroadst: ditto for `CommandSucceededEvent`

kmahar: I took the optionality from the defined interfaces in the spec - it sounded like it would only be set for certain operation types.  but it appears that `libmongoc` always returns a value so I'll make it non-optional.

Sources/MongoSwift/APM.swift
kmahar: oops missed this comment before I merged, I'll rename it in the soon-to-come SDAM sketch PR

PR #43, opened 2018-03-26
Add custom measurement code for benchmarks, and script to parse results

kmahar:
This adds an extension to `XCTestCase` to provide custom time measurement code. The time is calculated by taking the difference in system uptime (the same way as in the built-in `measure` functionality.)

While by default benchmarks will now use 100 iterations, I have adjusted the number  in various cases better match the spec requirements that:
- iterations should loop for at least 1 minute cumulative execution time
- iterations should stop after 100 iterations or 5 minutes cumulative execution time, whichever is shorter

Note that this makes the total benchmark execution time far longer, as there are now 13 benchmarks each running for a minimum of 1 minute. 

This also adds a python script to read benchmark output from stdin and store it in a dictionary like this:
`{'testRunCommand': {'score': 0.236, 'time': 0.677}, 'testDeepDecoding': {'score': 11.6672, 'time': 1.683}, ...}`

Where the time is median time for the task and the "score" is task size in MB / median time. 

Right now that is just printed out at the end, TBD how exactly we need to structure/provide info for plotting to Evergreen.

### DIFF
--- a/Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
+++ b/Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
@@ -1,0 +1,33 @@
+import Foundation
+import XCTest
+
+extension XCTestCase {
+
+    /// Measure the time for a single execution of operation
+    func measureTime(_ operation: () throws -> Void) throws -> Double {
+        let startTime = ProcessInfo.processInfo.systemUptime
+        try operation()
+        let timeElapsed = ProcessInfo.processInfo.systemUptime - startTime
+        return Double(timeElapsed)
+    }
+
+    /// Measure the median time for iterations executions of operation
+    func measureOp(_ operation: () throws -> Void, iterations: Int = 100) throws -> Double {
+        var results = [Double]()
+        for _ in 1...iterations {
+            results.append(try measureTime(operation))
+        }
+        return median(results)
+    }
+
+    /// Compute the median of an array of doubles 
+    func median(_ input: [Double]) -> Double {
+        return input.sorted(by: <)[input.count / 2]
+    }
+
+    func printResults(time: Double, size: Double) {
+        let roundedTime = Double(floor(time * 1000) / 1000)
+        let roundedScore = Double(floor(size / time * 10000) / 10000)
+        print("Results for \(self.name): median time \(roundedTime) seconds, score \(roundedScore) MB/s")
+    }
+}
--- a/Tests/MongoSwiftBenchmarks/BsonBenchmarks.swift
+++ b/Tests/MongoSwiftBenchmarks/BsonBenchmarks.swift
@@ -4,8 +4,11 @@ import XCTest
 
 let basePath = "Tests/Specs/benchmarking/data/"
 let flatBsonFile = URL(fileURLWithPath: basePath + "flat_bson.json")
+let flatSize = 75.31
 let deepBsonFile = URL(fileURLWithPath: basePath + "deep_bson.json")
+let deepSize = 19.64
 let fullBsonFile = URL(fileURLWithPath: basePath + "full_bson.json")
+let fullSize = 57.34
 
 final class BsonBenchmarkTests: XCTestCase {
     static var allTests: [(String, (BsonBenchmarkTests) -> () throws -> Void)] {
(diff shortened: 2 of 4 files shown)


Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import XCTest
+
+extension XCTestCase {
+
+    /// Measure the time for a single execution of operation
+    func measureTime(_ operation: () throws -> Void) throws -> Double {
+        let startTime = ProcessInfo.processInfo.systemUptime
```
mbroadst: Is this monotonic and/or high resolution time? It looks like [from this post](https://stackoverflow.com/questions/24755558/measure-elapsed-time-in-swift) we could also be using `DispatchTime.now()`. Are there alternatives?

mbroadst: I think that _might_ be OSX/iOS specific though, GCD probably isn't provided by the swift implementation for e.g. linux

mbroadst: that post is interesting, it looks like there's also straight up:
```
let start = Date()
try operation()
return start.timeIntervalSinceNow
```

Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    /// Compute the median of an array of doubles 
+    func median(_ input: [Double]) -> Double {
+        return input.sorted(by: <)[input.count / 2]
+    }
+
+    func printResults(time: Double, size: Double) {
+        let roundedTime = Double(floor(time * 1000) / 1000)
+        let roundedScore = Double(floor(size / time * 10000) / 10000)
+        print("Results for \(self.name): median time \(roundedTime) seconds, score \(roundedScore) MB/s")
```
mbroadst: If you're already printing this out do we even need the python script? Can't we just print out JSON right here?

Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
kmahar: I didn't investigate this much and just chose what `XCTest` uses for measuring time ([here](https://github.com/apple/swift-corelibs-xctest/blob/01e6126d1cd1c5a81e7a1695487c09d57261577c/Sources/XCTest/Private/WallClockTimeMetric.swift#L58)). this returns a [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval) (which is really just a `Double`) and has "sub-millisecond precision". looks like `timeIntervalSinceNow` also returns a `TimeInterval`. though I am looking at the [source](https://github.com/apple/swift-corelibs-foundation/blob/04b903a6091562a72846f21b048a966532a37f94/Foundation/Date.swift#L93) for `timeIntervalSinceNow` and it appears to use `CFAbsoluteTimeGetCurrent()`, which is stated to not necessarily be monotonic on repeated calls in its [docs](https://developer.apple.com/documentation/corefoundation/1543542-cfabsolutetimegetcurrent). (but this is a single call for each operation so maybe that's ok?) 

yeah, the `DispatchTime` thing is odd, the [documentation](https://developer.apple.com/documentation/dispatch/dispatchtime) talks about behavior on Apple platforms but its unclear what happens if you try to use it on Linux... though I see a [Swift ticket](https://bugs.swift.org/browse/SR-5706) about some fixes to it, so it looks like it would be available? 

what do you think?

Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
kmahar: so [here's the script](https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/run-perf-tests.sh) the Python driver uses for running perf tests on Evergreen, and [here's](https://github.com/mongodb/mongo-python-driver/blob/master/test/performance/perf_test.py) the actual `perf_test.py` file. it looks like they keep this global `result_data` list and at the end dump that all out to a `results.json` file. 

then, looking at their [perf project](https://evergreen.mongodb.com/task/mongo_python_driver_perf_perf_tests_perf_3.0_standalone_f847be3fcb1c75801576748128b0440de144304f_18_03_22_15_47_34) (see log at bottom) they do some `attach.results` and `json.send` commands using that file. I'm not sure how that evergreen magic works, looks like there are related commands defined in their [`config.yml`](https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/config.yml#L341) and [`perf.yml`](https://github.com/mongodb/mongo-python-driver/blob/15f62db70c4e9e81f9fdd8075df5f6d71143c23c/.evergreen/perf.yml#L130).

SO, based on all that, I've been operating under the assumption that we need to output all the results together at the end in a single file, meaning that we can't just individually print out JSON results interspersed with the other test output and have evergreen understand it.

so it seems to me like we do need the python script, and probably what should be happening in there is that (rather than just printing `benchmarks` as I do now) we dump it out to a JSON file which we then somehow tell evergreen to use for plots.

mbroadst: well we don't want to force developers to have a random file on the system each time they run the benchmark tests, I think what you have here is actually probably fine, I was just wondering why we needed the extra step

Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
kmahar: 👍

Tests/MongoSwiftBenchmarks/BenchmarkUtils.swift
kmahar: well, you could just run `make benchmark` and not pipe it if you don't want this stuff, but regardless it's more flexible/less work to just pipe `benchmark.py`'s output to another file so ya let's just leave it as printing. 
I guess we should really be printing `json.dumps(benchmarks)` though so it's properly formatted, but I think I'll have to do some more finagling with the output structure to make evergreen understand it anyway, so we can worry about that when we actually have automated running the benchmarks.
we should think about what exactly we want to plot - python does ops/second, or we could just plot the times or scores, or I guess we can probably plot multiple of them

PR #44, opened 2018-03-26
Add a getError method and a way to iterate+throw for cursors

kmahar:
This adds a method to get the error that occured while iterating the cursor, if any. Also adds a `nextOrError` throwing method for iterating the cursor if you don't want to have to remember to check the cursor at the end. this still feels kind of weird so if you have any other ideas let me know... but at least for now this is better than just printing out the error message

### DIFF
--- a/Sources/MongoSwift/Cursor.swift
+++ b/Sources/MongoSwift/Cursor.swift
@@ -34,6 +34,41 @@ public class MongoCursor: Sequence, IteratorProtocol {
     }
 
     /**
+     * Returns the next document in this cursor, or nil. Throws an error if one
+     * occurs. (Compared to next(), which returns nil and requires manually checking
+     * for an error afterward.)
+     *
+     */
+    func nextOrError() throws -> Document? {
+        let out = UnsafeMutablePointer<UnsafePointer<bson_t>?>.allocate(capacity: 1)
+        defer {
+            out.deinitialize(count: 1)
+            out.deallocate(capacity: 1)
+        }
+
+        if !mongoc_cursor_next(self._cursor, out) {
+            var error = bson_error_t()
+            if mongoc_cursor_error(self._cursor, &error) {
+                throw MongoError.invalidCursor(message: toErrorString(error))
+            }
+            return nil
+        }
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/Cursor.swift
```diff
@@ -34,24 +34,50 @@ public class MongoCursor: Sequence, IteratorProtocol {
...
             return nil
         }
+        return Document(fromPointer: UnsafeMutablePointer(mutating: out.pointee!))
+    }
 
+    /**
+     *  Returns the error that occurred while iterating this cursor, if one exists. 
+     *  This function should be called to check for errors after next() returns nil. 
+     */
+    func getError() -> Error? {
```
mbroadst: is it more idiomatic to have a computed variable rather than a getter here? `getError` feels very Java to me

Sources/MongoSwift/Cursor.swift
```diff
@@ -34,24 +34,50 @@ public class MongoCursor: Sequence, IteratorProtocol {
...
 
     /**
-     * Returns the next document in this cursor, or nil
+     * Returns the next document in this cursor, or nil. Throws an error if one
+     * occurs. (Compared to next(), which returns nil and requires manually checking
+     * for an error afterward.)
+     *
      */
-    public func next() -> Document? {
+    func nextOrError() throws -> Document? {
```
mbroadst: can't you reuse `next` in the body of this function instead of duplicating the code?

Sources/MongoSwift/Cursor.swift
kmahar: I agree, sgtm

Sources/MongoSwift/Cursor.swift
kmahar: yeah good point, can also use the `error` property to get the error, there was a lot of duplication here lol

kmahar:
yeah, examples seem like they would be useful, or at the very least we can put a cursor example in the README... I'm not sure what is the most effective way to get people to actually look at it

PR #46, opened 2018-03-28
Sketch data structures and functions for SDAM monitoring

kmahar:
This adds `structs` to represent SDAM monitoring events (as defined in the spec [here](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst)), as well as for `ServerDescription` and `TopologyDescription`, from the spec [here](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst). 

There are a couple places I slightly deviated from the spec for `ServerDescription` and `TopologyDescription`, I'll point them out in comments. 

Additionally sketches out the callbacks we'll use, and adds extra enum cases where needed.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -42,7 +42,7 @@ public struct ServerDescription {
     /// The hostname or IP and the port number that the client connects to. Note that this is not the
     /// server's ismaster.me field, in the case that the server reports an address different from the 
     /// address the client uses.
-    let connectionId: ConnectionId = ConnectionId()
+    let connectionId: ConnectionId
 
     /// The last error related to this server.
     let error: MongoError? = nil
@@ -122,7 +122,7 @@ public struct TopologyDescription {
     let maxElectionId: ObjectId? = nil
 
     /// The servers comprising this topology. By default, a single server at localhost:270107.
-    let servers: [ServerDescription] = [ServerDescription()]
+    let servers: [ServerDescription] = [ServerDescription(connectionId: ConnectionId())]
 
     /// For single-threaded clients, indicates whether the topology must be re-scanned.
     let stale: Bool = false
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/APM.swift
```diff
@@ -14,6 +14,133 @@ public struct ConnectionId {
...
+    let maxElectionId: ObjectId? = nil
+
+    /// The servers comprising this topology. By default, a single server at localhost:270107.
+    let servers: [ServerDescription] = [ServerDescription(connectionId: ConnectionId())]
+
+    /// For single-threaded clients, indicates whether the topology must be re-scanned.
+    let stale: Bool = false
+
+    /// Exists if any server's wire protocol version range is incompatible with the client's.
+    let compatibilityError: MongoError? = nil
```
kmahar: they suggest a bool indicating an error and then a second string property containing the error message, but this seems cleaner to me

Sources/MongoSwift/APM.swift
```diff
@@ -14,6 +14,133 @@ public struct ConnectionId {
...
+    case rsPrimary = "RSPrimary"
+    case rsSecondary = "RSSecondary"
+    case rsArbiter = "RSArbiter"
+    case rsOther = "RSOther"
+    case rsGhost = "RSGhost"
+    case unknown = "Unknown"
+}
+
+/// A struct describing a mongod or mongos process.
+public struct ServerDescription {
```
kmahar: libmongoc `mongoc_server_description_t`s also have an `id` available ([docs](http://mongoc.org/libmongoc/current/mongoc_server_description_id.html)), which isn't part of the spec here but we could add it... I'm not really sure what its purpose is though

kmahar:
closing as I ended up just combining this with the ticket to fill out the sketch on a diff branch

PR #47, opened 2018-04-03
Add support for SDAM monitoring

kmahar:
Still TBD exactly what we want the API for enabling and disabling monitoring to look like. I've included what I think is at least a clearer interface than having `enableMonitoring` behave differently depending on if it's been called already, but as we talked about yesterday, it's not really clear why someone would want to be switching monitoring on and off repeatedly. anyway, while we figure that out there is a lot else here to review.

I've added a basic test based on [this](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/monitoring/standalone.json) spec test. basically checking that the events come in the correct order and contain the values/descriptions we expect. I think the rest of the tests would require a mock server, or support for our running tests with something other than a standalone mongod, so 
I am not sure how far down that rabbit hole we want to go.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -42,7 +42,7 @@ public struct ServerDescription {
     /// The hostname or IP and the port number that the client connects to. Note that this is not the
     /// server's ismaster.me field, in the case that the server reports an address different from the 
     /// address the client uses.
-    let connectionId: ConnectionId = ConnectionId()
+    let connectionId: ConnectionId
 
     /// The last error related to this server.
     let error: MongoError? = nil
@@ -122,7 +122,7 @@ public struct TopologyDescription {
     let maxElectionId: ObjectId? = nil
 
     /// The servers comprising this topology. By default, a single server at localhost:270107.
-    let servers: [ServerDescription] = [ServerDescription()]
+    let servers: [ServerDescription] = [ServerDescription(connectionId: ConnectionId())]
 
     /// For single-threaded clients, indicates whether the topology must be re-scanned.
     let stale: Bool = false
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -13,7 +13,7 @@ public protocol MongoEvent {
 
 /// A protocol for monitoring events to implement, indicating that they can be initialized from an OpaquePointer
 /// to the corresponding libmongoc type.
-internal protocol InitializableFromOpaquePointer {
+private protocol InitializableFromOpaquePointer {
     /// Initializes the event from an OpaquePointer.
     init(_ event: OpaquePointer)
 }
@@ -46,9 +46,8 @@ public struct CommandStartedEvent: MongoEvent, InitializableFromOpaquePointer {
     /// The connection id for the command.
     let connectionId: ConnectionId
 
-    /// An internal initializer for creating a CommandStartedEvent from an OpaquePointer to a
-    /// mongoc_apm_command_started_t
-    internal init(_ event: OpaquePointer) {
+    /// Initializes a CommandStartedEvent from an OpaquePointer to a mongoc_apm_command_started_t
+    fileprivate init(_ event: OpaquePointer) {
         let commandData = UnsafeMutablePointer(mutating: mongoc_apm_command_started_get_command(event)!)
         self.command = Document(fromPointer: commandData)
         self.databaseName = String(cString: mongoc_apm_command_started_get_database_name(event))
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/APM.swift
```diff
@@ -14,11 +14,247 @@ public struct ConnectionId {
...
+    /// The logicalSessionTimeoutMinutes value for this server.
+    var logicalSessionTimeoutMinutes: Int64?
+
+    /// An internal initializer to create a `ServerDescription` with just a ConnectionId.
+    internal init(connectionId: ConnectionId) {
+        self.connectionId = connectionId
+    }
+
+    /// An internal function to handle parsing isMaster and setting ServerDescription attributes appropriately.
+    internal mutating func parseIsMaster(_ isMaster: Document) {
```
kmahar: note: I broke this out into its own function because if we eventually choose to implement all the SDAM monitoring spec tests we'd need to write a separate `ServerDescription` init that directly takes in a `ConnectionId` and an isMaster `Document`

Sources/MongoSwift/APM.swift
```diff
@@ -14,11 +14,247 @@ public struct ConnectionId {
...
+}
+
+/// A struct describing the state of a MongoDB deployment: its type (standalone, replica set, or sharded), 
+/// which servers are up, what type of servers they are, which is primary, and so on.
+public struct TopologyDescription {
+    /// The type of this topology. 
+    let type: TopologyType
+
+    /// The replica set name. 
+    var setName: String? { return self.servers[0].setName }
```
kmahar: while this is always going to be the same value right now (same goes for `logicalSessionTimeoutMinutes`) and could technically just be a set property, in the future we would want to be dynamically calculating it as servers are added/removed

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -467,6 +467,10 @@ class ObjectId: BsonValue, Equatable, CustomStringConvertible {
         bson_oid_copy(from, &self.oid)
     }
 
+    internal init(from: bson_oid_t) {
+        self.oid = from
```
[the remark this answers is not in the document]
mbroadst: also side comment: why is `ObjectId` a class again?

Sources/MongoSwift/APM.swift
```diff
@@ -1,24 +1,32 @@
...
-    internal init(_ hostList: UnsafePointer<mongoc_host_list_t>) {
-        var hostData = hostList.pointee
-        self.host = withUnsafeBytes(of: &hostData.host) { (rawPtr) -> String in
-            let ptr = rawPtr.baseAddress!.assumingMemoryBound(to: CChar.self)
-            return String(cString: ptr)
-        }
-        self.port = hostData.port
-    }
+/// A protocol for monitoring events to implement, specifying their type and name.
+public protocol MongoEvent {
```
kmahar: I changed this to public so that it is actually useful to users. for example, they might want to store events in an array of `MongoEvent`. 
the static `eventName` property will also allow them to do something like
```swift
let observer = center.addObserver(forName: CommandStartedEvent.eventName, object: nil, queue: nil) { (notif) in
    // do stuff
}
```

and then added a separate protocol for the internal initializer so we can still post everything through `postNotification`, which requires that initializer exists

Sources/MongoSwift/Client.swift
```diff
@@ -1,13 +1,23 @@
 import Foundation
 import libmongoc
 
-public struct ClientOptions {
+public struct ClientOptions: BsonEncodable {
```
kmahar: we don't actually send these options now anyway but I added the `encode` method so that we don't forget `eventMonitoring` should be omitted

Sources/MongoSwift/APM.swift
```diff
@@ -120,109 +357,344 @@ public struct CommandFailedEvent {
...
-     *  will result in only posting notifications for .commandSucceeded events.
+     *  Whatever set of events are selected in the first call to this function will remain
+     *  the monitored events for this client for its lifetime, persisting across any
+     *  number of calls to enableMonitoring and disableMonitoring. The NotificationCenter
+     *  may be changed in future calls to reenableMonitoring.
      */
-    public func enableCommandMonitoring(
-        forEvents events: [MongoEvent] = [.commandStarted, .commandSucceeded, .commandFailed],
-        usingCenter center: NotificationCenter = NotificationCenter.default) {
+    public func initializeMonitoring(forEvents type: MongoEventType?,
```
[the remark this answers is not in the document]
mbroadst: I think that’s fine. Also, do enums in swift have an integer value? The convention in e.g. c++ is that we could provide bitwise ORs of multiple enums if you wanted to specify more complicated choices. So you would say `enableMonitoring(.commandMonitoring | .sdamMonitoring)` and you could even store it as a single int

Sources/MongoSwift/APM.swift
mbroadst: very nice!

mbroadst: very nice!

mbroadst: very nice!

Sources/MongoSwift/BSON/BsonValue.swift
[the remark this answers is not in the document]
mbroadst: That sounds reasonable to me. Either way let’s document it thoroughly so neither of us has to remember again :)

PR #49, opened 2018-04-10
SWIFT-69 Basic decoder implementation 

kmahar:
Ok, here is phase one of this. 

It supports decoding a `Document` (which you could create by hand, or from data using our various initializers) into a `Decodable` `struct` or `class`. 

I've tried to stay as close as possible to how the `JSONEncoder` works, and a lot of this is essentially copy-paste from there - plus introducing some generic functions to reduce the ridiculous amount of code duplication... lol. 

So I think phase two will be that we mark all our custom types as `Decodable` and add `init(from decoder: Decoder)` for them.

I have various comments to make inline, but one large point on the numbers stuff we've been discussing: I realized that our conversation about how you'd store various non-BSON numbers was more about how to properly encode numbers to BSON without loss, vs. how to properly *decode* them from BSON. Take for example the value `42`: you could store that in BSON as an `Int32`, `Int64`, or `Double`, and you can represent it exactly in pretty much every available numeric type in Swift (or at least the ones we're required to be able to decode to.) It seems to me like we should allow the value to have been stored as any BSON number type, and as long as we can convert that BSON number to the desired type without losing any info, we should do it. 
Otherwise, we are placing implicit constraints on how you encode your data. If you encoded it using our encoder then we can control the type mapping (i.e. deterministically always store an `Int8` as `Int32`, etc.), but given that you can create a `Document` from raw BSON data I don't think we can assume in reverse that a value someone wants to parse to an `Int8` will always be encoded as an `Int32`. So I propose that we write `init(from:BsonValue)`s for each of the non-BSON numbers, and in there we can switch through the possible number types and convert where possible, doing range checks as necessary etc.

### DIFF
--- a/Sources/MongoSwift/BSON/Codec.swift
+++ b/Sources/MongoSwift/BSON/Codec.swift
@@ -333,3 +333,602 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
         preconditionFailure("Value \(value) with type \(type(of: value)) didn't match any expected types")
     }
 }
+
+/// `BsonDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
+public class BsonDecoder {
+
+    /// Contextual user-provided information for use during decoding.
+    open var userInfo: [CodingUserInfoKey: Any] = [:]
+
+    /// Options set on the top-level decoder to pass down the decoding hierarchy.
+    fileprivate struct _Options {
+        let userInfo: [CodingUserInfoKey: Any]
+    }
+
+    /// The options set on the top-level decoder.
+    fileprivate var options: _Options {
+        return _Options(userInfo: userInfo)
+    }
+
+    /// Initializes `self`.
+    public init() {}
+
+    /// Decodes a top-level value of the given type from the given BSON representation.
+    ///
+    /// - parameter type: The type of the value to decode.
+    /// - parameter data: The BSON document to decode from.
+    /// - returns: A value of the requested type.
+    /// - throws: An error if any value throws an error during decoding.
+    public func decode<T: Decodable>(_ type: T.Type, from data: Document) throws -> T {
+        let _decoder = _BsonDecoder(referencing: data, options: self.options)
+        _decoder.storage.push(container: data)
+        defer { _decoder.storage.popContainer() }
+        return try type.init(from: _decoder)
+    }
+}
+
+private class _BsonDecoder: Decoder {
+
+    /// The decoder's storage.
+    fileprivate var storage: _BsonDecodingStorage
+
+    /// Options set on the top-level decoder.
+    fileprivate let options: BsonDecoder._Options
+
+    /// The path to the current point in decoding.
+    fileprivate(set) public var codingPath: [CodingKey]
+
+    /// Contextual user-provided information for use during decoding.
+    public var userInfo: [CodingUserInfoKey: Any] {
+        return self.options.userInfo
+    }
+
+    /// Initializes `self` with the given top-level container and options.
+    fileprivate init(referencing container: BsonValue?, at codingPath: [CodingKey] = [], options: BsonDecoder._Options) {
+        self.storage = _BsonDecodingStorage()
+        self.storage.push(container: container)
+        self.codingPath = codingPath
+        self.options = options
+    }
+
+    // Returns the data stored in this decoder as represented in a container keyed by the given key type.
+    public func container<Key>(keyedBy type: Key.Type) throws -> KeyedDecodingContainer<Key> {
+        guard self.storage.topContainer != nil  else {
+            throw DecodingError.valueNotFound(KeyedDecodingContainer<Key>.self,
+                                  DecodingError.Context(codingPath: self.codingPath,
+                                                        debugDescription: "Cannot get keyed decoding container -- found null value instead."))
+        }
+
+        guard let topContainer = self.storage.topContainer as? Document else {
+            throw DecodingError._typeMismatch(at: self.codingPath, expectation: Document.self, reality: self.storage.topContainer)
+        }
+
+        let container = _BsonKeyedDecodingContainer<Key>(referencing: self, wrapping: topContainer)
+        return KeyedDecodingContainer(container)
+    }
+
+    // Returns the data stored in this decoder as represented in a container appropriate for holding a single primitive value.
+    public func singleValueContainer() throws -> SingleValueDecodingContainer {
+        return self
+    }
+
+    // Returns the data stored in this decoder as represented in a container appropriate for holding values with no keys.
+    public func unkeyedContainer() throws -> UnkeyedDecodingContainer {
+        guard self.storage.topContainer != nil else {
+            throw DecodingError.valueNotFound(UnkeyedDecodingContainer.self,
+                                  DecodingError.Context(codingPath: self.codingPath,
+                                                        debugDescription: "Cannot get unkeyed decoding container -- found null value instead."))
+        }
+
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/BSON/Codec.swift
```diff
@@ -333,3 +333,580 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
...
+    }
+}
+
+/// Just a temporary error to make code compile until we implement everything.
+struct UnimplementedError: LocalizedError {
+    public var errorDescription: String? { return "Unimplemented" }
+}
+
+/// A protocol for types that are not BSON types but require decoding support. 
+private protocol NonBsonValue {
```
kmahar: perhaps not the best name lol, but basically I'm trying to capture that it's a primitive type we need to be able to decode to but it isn't a `BsonValue`

Sources/MongoSwift/BSON/Codec.swift
```diff
@@ -333,3 +333,580 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
...
+    /// Initializes `self`.
+    public init() {}
+
+    /// Decodes a top-level value of the given type from the given BSON representation.
+    ///
+    /// - parameter type: The type of the value to decode.
+    /// - parameter data: The BSON document to decode from.
+    /// - returns: A value of the requested type.
+    /// - throws: An error if any value throws an error during decoding.
+    public func decode<T: Decodable>(_ type: T.Type, from data: Document) throws -> T {
```
kmahar: was debating whether we should add a`decode` method that directly takes in `Data`, or just let people make the document themselves with `Document.init(fromBSON:)`. what do you think?

Sources/MongoSwift/BSON/Codec.swift
```diff
@@ -333,3 +333,580 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
...
 }
+
+/// `BsonDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
+public class BsonDecoder {
+
+    /// Contextual user-provided information for use during decoding.
+    open var userInfo: [CodingUserInfoKey: Any] = [:]
+
+    /// Options set on the top-level decoder to pass down the decoding hierarchy.
+    fileprivate struct _Options {
```
kmahar: unclear what we would use this or `userInfo` for right now, but eventually if we supported some different decoding strategies or something we could store that info in here. (and the `Decoder` protocol requires a `userInfo` property anyway)

Sources/MongoSwift/BSON/Codec.swift
```diff
@@ -333,3 +333,580 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
...
+
+        return _BsonUnkeyedDecodingContainer(referencing: self, wrapping: arr)
+    }
+}
+
+// Storage for a _BsonDecoder.
+private struct _BsonDecodingStorage {
+
+    /// The container stack, consisting of `BsonValue?`s. 
+    private(set) fileprivate var containers: [BsonValue?] = []
```
kmahar: one place I diverged from the `JSONEncoder` is using optionals/nils rather than `NSNull` to represent null values... this felt more intuitive to me because we can specify that if a container exists it's definitely a `BsonValue` rather than `Any`... but can you think of any advantage to using `NSNull` instead?

Sources/MongoSwift/BSON/Codec.swift
```diff
@@ -333,3 +333,580 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
...
+        return KeyedDecodingContainer(container)
+    }
+
+    // Returns the data stored in this decoder as represented in a container appropriate for holding a single primitive value.
+    public func singleValueContainer() throws -> SingleValueDecodingContainer {
+        return self
+    }
+
+    // Returns the data stored in this decoder as represented in a container appropriate for holding values with no keys.
+    public func unkeyedContainer() throws -> UnkeyedDecodingContainer {
```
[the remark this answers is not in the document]
mbroadst: I've refrained from commenting below further on this, but it looks like this change could potentially reduce the code by a large margin

Sources/MongoSwift/BSON/Codec.swift
```diff
@@ -333,3 +333,580 @@ private func getBsonValue(_ value: Any) -> BsonValue? {
...
+    public var errorDescription: String? { return "Unimplemented" }
+}
+
+/// A protocol for types that are not BSON types but require decoding support. 
+private protocol NonBsonValue {
+    init(from: BsonValue) throws
+}
+
+extension Int8: NonBsonValue {
+    init(from: BsonValue) throws {
```
mbroadst: Not sure how to directly comment on your proposal in the description, but these stub implementations seem to be the proposal realized? Either way, I agree with the approach laid out there!

Sources/MongoSwift/BSON/Codec.swift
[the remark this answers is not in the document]
kmahar: So I’ve been looking into this a bit more, and I’m not sure we actually want to get rid of these. 

Consider the following struct: 
```swift
struct SimpleStruct: Decodable, Equatable {
    let ints: [Int]
}
```

The way that gets decoded via the synthesized method is basically:
```swift
init(from decoder: Decoder) throws {
    // get top level keyed container
    let container = decoder.container() 
    // decode what's stored under "ints" to an [Int]
    self.ints = try container.decode([Int].self, forKey: "ints") 
}
```

Under the hood, that `container.decode<T: Decodable>(_ type: T.Type, forKey: Key)` call is pushing the value stored under `ints` onto the container stack, and then calling `Array`'s `init(from decoder: Decoder)` method:
```swift
self.storage.push(container: value)
defer { self.storage.popContainer() }
return try T(from: self)
```

 I *think* that `init` ends up being something like: 
```swift
extension Array {
    init(from decoder: Decoder) throws {
        // get top level unkeyed container
        let container = decoder.unkeyedContainer() 

        var data = [/* some Decodable type */]()
    	while !container.isAtEnd {
            self.append(try container.decode(/* some Decodable type */))
	}

    }
}
```

And then `container.decode(Element.Type.self)`, similarly to the previous decode call, ends up pushing each element of the array onto the stack one by one and then calling `Int`’s `init(from decoder: Decoder)` with each of them, which requests a `singleValueContainer`. 

The one weird thing about this that I’m not sure on is that one would *think* the `container.decode(Element.Type.self)` call made in there would get routed to the `decode(_ type: Int.Type)` method instead of  `decode<T: Decodable>(_ type: T.Type)`  when we’re decoding an `[Int]`, which would allow us to just return the value directly rather than pushing onto stack and then retrieving it.  The only thing I can figure there is that the `Array.init(from decoder: Decoder)` method uses generics somehow and just knows elements are `Decodable`  ,but doesn’t know their specific types. 

So anyway, though I’m still not totally sure what’s going on, it appears to me that both `singleValueContainer` and `unkeyedContainer` s actually still end up getting used with some basic struct configurations we probably want to support.

Sources/MongoSwift/BSON/Codec.swift
kmahar: agreed, I'll figure out how to restructure all this stuff with the filename `+` or whatever 👍

Sources/MongoSwift/BSON/Codec.swift
kmahar: yup, put them here so things would compile but wanted to hold off on implementing til we discussed - sounds good

Sources/MongoSwift/BSON/Codec.swift
kmahar: 👍

PR #50, opened 2018-04-13
SWIFT-68 Make building and testing work both in command line and Xcode, and update README

kmahar:
Changes:

1. `make` no longer generates an `.xcodeproj` file. Discovered by accident that don't need it for building/testing from command line, just `swift build` will do, so may as well remove it to speed up the build process, unless you had some reason for putting it there? 
2. Add a `project` target to the `Makefile`, which generates the `.xcodeproj` file and then edits it using the ruby script to add corresponding `.json` files to the correct targets. Since we have two separate targets, this required handling each folder in `Tests/Specs` separately... perhaps we should consider moving all the non-benchmark specs into a subdirectory so we can just add all the files from that at once, and not have to add a line to the ruby script every time we add more specs.
3. Add methods to each test target that figure out whether we're running from the command line or `Xcode` and return the appropriate path to look for specs in. While ideally this method would be shared, I'm not sure there's a clean way to do that since they are separate targets. The methods are slightly different anyway - the one for benchmarks is a class method as it needs to work without an instance of `XCTestCase`.
4. Update `README` accordingly 

Can you please test that the following works for you to make sure I didn't miss anything?
1. `make clean`
2. Checkout this branch
3. Run `make` and then `make test` to verify typical command line usage still works 
4. `make clean` again
5. Assuming you don't already have the `xcodeproj` gem installed: run `make project`, which depending on your permissions will either install the gem or fail and give you an error message about installing it yourself. If you do already have it installed then skip this step. 
6. Run `make project` again, this time it should work
7. Open the `.xcodeproj` file in Xcode
8. Build in Xcode and then run `MongoSwiftTests`
9. Try running one of the benchmark tests; if it doesn't fail immediately then we're good

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,12 @@ The official [MongoDB](https://www.mongodb.com/) driver for Swift.
     - [FIRST: Install the MongoDB C Driver](#first-install-the-mongodb-c-driver)
     -  [NEXT: Install the Driver Using Swift Package Manager](#next-install-the-driver-using-swift-package-manager)
     - [OR: Install the Driver Using CocoaPods](#or-install-the-driver-using-cocoapods)
+- [Building](#building)
+    - [From the command line](#from-the-command-line)
+    - [In Xcode](#in-xcode)
+- [Testing](#testing)
+    - [From the command line](#from-the-command-line-1)
+    - [In Xcode](#in-xcode-1)
 - [Example Usage](#example-usage)
     - [Connect to MongoDB and Create a Collection](#connect-to-mongodb-and-create-a-collection)
     - [Create and Insert a Document](#create-and-insert-a-document)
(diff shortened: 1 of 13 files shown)


mbroadst:
Ran through all the steps, everything checked out with `make test`, almost all tests ran in xcode as well but `testMonitoring` failed with this message: "failed - Notification name = NSWillBecomeMultiThreadedNotification, object = nil, userInfo = nil did not contain an event"

mbroadst:
Just ran it again and it worked fine, so I think it's just a flakey test. Great job here!

README.md
```diff
@@ -85,6 +90,38 @@ end
 
 Finally, run `pod install` to install your project's dependencies. 
 
+## Building
```
mbroadst: This whole section is great, thanks for taking the time with it

README.md
```diff
@@ -85,6 +90,38 @@ end
...
+1. Install the Ruby gem `xcodeproj` with `gem install xcodeproj` (you may need to `sudo`)
+2. Run `make project`
+3. You're ready to go! Open `MongoSwift.xcodeproj` with it and build and test as normal.
+
+Why is this necessary? The project requires a customized "copy resources" build phase to include various test `.json` files. By default, this phase is not included when you run `swift package generate-xcodeproj`. So `make project` first generates the project, and then uses `xcodeproj` to manually add the files to the appropriate targets (see `add_json_files.rb`). 
+
+## Running Tests
+**NOTE**: `ClientTests`, `CollectionTests`, `CommandMonitoringTests`, `CrudTests`, and `DatabaseTests` all require a mongod instance to be running on the default host/port, `localhost:27017`. The remainder of the tests are for the BSON library, and should succeed regardless of whether a mongod is running.
+
+Additionally, please note that each benchmark test runs for a minimum of 1 minute and therefore **the entire benchmark suite will take around 20-30 minutes to complete**.
```
mbroadst: lol

Tests/MongoSwiftBenchmarks/DocumentBenchmarks.swift
```diff
@@ -3,11 +3,11 @@ import Foundation
 import XCTest
 import Nimble
 
-let tweetFile = URL(fileURLWithPath: basePath + "tweet.json")
+let tweetFile = URL(fileURLWithPath: basePath + "/tweet.json")
```
mbroadst: nits for the future, but let's use string interpolation as much as possible

Tests/MongoSwiftTests/ClientTests.swift
```diff
@@ -14,8 +14,7 @@ final class ClientTests: XCTestCase {
     func testListDatabases() throws {
         let client = try MongoClient()
         let databases = try client.listDatabases(options: ListDatabasesOptions(nameOnly: true))
-        let expectedDbs: [Document] = [["name": "admin"], ["name": "config"], ["name": "local"]]
-        expect(Array(databases) as [Document]).to(equal(expectedDbs))
+        expect((Array(databases) as [Document]).count).to(beGreaterThan(0))
```
mbroadst: thanks for finally handling this, sigh

README.md
kmahar: 🙂

Tests/MongoSwiftBenchmarks/DocumentBenchmarks.swift
kmahar: 👍

PR #53, opened 2018-04-18
SWIFT-70 Make OpaquePointers to libmongoc types optional to prevent freeing unset values

kmahar:
While the pointers were already technically optional because we created them using [`init?(bitPattern: Int)`](https://developer.apple.com/documentation/swift/opaquepointer/1688016-init), initializing with `bitPattern: 1` never fails. This meant that in cases where assigning a real pointer failed for whatever reason, we would try to free a value at `0x0000000000000001`, causing libmongoc to crash. 

This change makes it so that if we fail to set the pointer to a real value, we won't try to free it.

 I also moved the nil-ing out of `self._client`s to the top of each `deinit` so we do it even if we return early because the main pointer is nil. though actually I'm not sure if we even really need to nil out the client at all - shouldn't ARC take care of it and know whether to call `MongoClient.deinit` or not?

### DIFF
--- a/Sources/MongoSwift/Client.swift
+++ b/Sources/MongoSwift/Client.swift
@@ -41,7 +41,7 @@ public struct ListDatabasesOptions: BsonEncodable {
 
 // A MongoDB Client
 public class MongoClient {
-    internal var _client = OpaquePointer(bitPattern: 1)
+    internal var _client: OpaquePointer?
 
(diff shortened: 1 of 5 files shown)


kmahar:
I don't mind the explicitness, just wanted to make sure I wasn't missing something :) but yes, I'll look into it anyway, figuring out how to check for leaks and doing so throughout the code is on my todo list

PR #54, opened 2018-04-19
SWIFT-29 Implement ReadConcern type

kmahar:
- Implement `ReadConcern` type wrapping a `mongoc_read_concern_t`
- Allow `ReadConcern` to be specified when creating a `MongoClient`, `MongoDatabase`, or `MongoCollection` and for commands that support it 
- Add a new `skipFields` property to the `BsonEncodable` protocol allowing user to specify field names that should be omitted when using the default `encode(to encoder: BsonEncoder)` implementation. This is useful since we want to append `ReadConcern`s to options documents differently than the other fields.

### DIFF
--- a/Sources/MongoSwift/Client.swift
+++ b/Sources/MongoSwift/Client.swift
@@ -9,16 +9,21 @@ public struct ClientOptions: BsonEncodable {
     /// command and server discovery and monitoring events.
     public let eventMonitoring: Bool
 
-    /// Convenience initializer allowing retryWrites to be omitted or optional
-    public init(retryWrites: Bool? = nil, eventMonitoring: Bool = false) {
+    /// Specifies a ReadConcern to use for the client. If one is not specified,
+    /// the server's default read concern will be used.
+    let readConcern: ReadConcern?
+
+    /// Convenience initializer allowing any/all to be omitted or optional
+    public init(eventMonitoring: Bool = false, readConcern: ReadConcern? = nil, retryWrites: Bool? = nil) {
         self.retryWrites = retryWrites
         self.eventMonitoring = eventMonitoring
+        self.readConcern = readConcern
     }
 
-    /// Custom `encode`, because we don't want to actually send the `eventMonitoring` option
-    public func encode(to encoder: BsonEncoder) throws {
-        try encoder.encode(retryWrites, forKey: "retryWrites")
-    }
+    /// `eventMonitoring` is a field that we set on the MongoClient, and `readConcern`
+    /// is used to set a default read concern for the client after it's created, so neither
+    /// of them should be encoded with the client options.
+    public var skipFields: [String] { return ["eventMonitoring", "readConcern"] }
 }
 
 public struct ListDatabasesOptions: BsonEncodable {
@@ -39,6 +44,12 @@ public struct ListDatabasesOptions: BsonEncodable {
     }
 }
 
+public struct DatabaseOptions {
+    /// A read concern to set on the retrieved database. If one is not specified,
+    /// the database will inherit the client's read concern. 
+    let readConcern: ReadConcern?
+}
+
 // A MongoDB Client
 public class MongoClient {
     internal var _client: OpaquePointer?
@@ -48,6 +59,13 @@ public class MongoClient {
 
     /// If command and/or server monitoring is enabled, indicates what event types notifications will be posted for.
     internal var monitoringEventTypes: [MongoEventType]?
+
+    /// The read concern set on this client.
+    public var readConcern: ReadConcern {
+        // per libmongoc docs, we don't need to handle freeing this ourselves
+        let readConcern = mongoc_client_get_read_concern(self._client)
+        return ReadConcern(readConcern)
+    }
 
     /**
      * Create a new client connection to a MongoDB server
@@ -67,10 +85,12 @@ public class MongoClient {
             throw MongoError.invalidClient()
         }
 
-        // if the user passed in this option, set up all the monitoring callbacks
-        if let opts = options, opts.eventMonitoring {
-            self.initializeMonitoring()
+        // if a readConcern is provided, set it on the client
+        if let rc = options?.readConcern {
+            mongoc_client_set_read_concern(self._client, rc._readConcern)
         }
+
+        if options?.eventMonitoring == true { self.initializeMonitoring() }
     }
 
     /**
@@ -140,13 +160,19 @@ public class MongoClient {
      *
      * - Parameters:
      *   - name: the name of the database to retrieve
+     *   - options: Optional settings
      *
      * - Returns: a `MongoDatabase` corresponding to the provided database name
      */
-    public func db(_ name: String) throws -> MongoDatabase {
+    public func db(_ name: String, options: DatabaseOptions? = nil) throws -> MongoDatabase {
         guard let db = mongoc_client_get_database(self._client, name) else {
             throw MongoError.invalidClient()
         }
+
+        if let rc = options?.readConcern {
+            mongoc_database_set_read_concern(db, rc._readConcern)
+        }
+
         return MongoDatabase(fromDatabase: db, withClient: self)
     }
 }
--- a/Sources/MongoSwift/Client.swift
+++ b/Sources/MongoSwift/Client.swift
@@ -60,11 +60,13 @@ public class MongoClient {
     /// If command and/or server monitoring is enabled, indicates what event types notifications will be posted for.
     internal var monitoringEventTypes: [MongoEventType]?
 
-    /// The read concern set on this client.
-    public var readConcern: ReadConcern {
+    /// The read concern set on this client, or nil if one is not set.
+    public var readConcern: ReadConcern? {
         // per libmongoc docs, we don't need to handle freeing this ourselves
         let readConcern = mongoc_client_get_read_concern(self._client)
-        return ReadConcern(readConcern)
+        let rcObj = ReadConcern(readConcern)
+        if rcObj.isDefault { return nil }
+        return rcObj
     }
 
     /**
--- a/Sources/MongoSwift/Collection.swift
+++ b/Sources/MongoSwift/Collection.swift
@@ -25,16 +25,23 @@ public struct AggregateOptions: BsonEncodable {
     /// The index to use for the aggregation. The hint does not apply to $lookup and $graphLookup stages.
     // let hint: Optional<(String | Document)>
 
+    /// A ReadConcern to use for this operation. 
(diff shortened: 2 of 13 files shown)


Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -0,0 +1,91 @@
...
+        if !mongoc_read_concern_append(self._readConcern, doc.data) {
+            throw MongoError.readConcernError(message: "Error appending readconcern to document \(doc)")
+        }
+    }
+
+    /// Since we have to follow certain rules about whether to include or omit a ReadConcern,
+    /// this function handles obeying those, factoring in the ReadConcern, if any, for
+    /// whatever object is calling this function. It returns a final options Document for the
+    /// calling function to use, or nil if the Document ends up being empty.
+    internal static func append(_ readConcern: ReadConcern?, to opts: Document?, callerRC: ReadConcern?) throws -> Document? {
```
kmahar: I was initially hoping that we could achieve appending the `ReadConcern` by just making `ReadConcern` a `BsonEncodable` and handling the logic in its `encode` function, but given that whether we should encode it depends on the context we're calling it from (i.e. whether the user has set a `ReadConcern` already), I wasn't sure how it would be possible to build that logic in

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -0,0 +1,218 @@
...
+    func testConnectionStrings() throws {
+    	let csPath = "\(self.getSpecsPath())/read-write-concern/tests/connection-string"
+    	let testFiles = try FileManager.default.contentsOfDirectory(atPath: csPath).filter { $0.hasSuffix(".json") }
+    	for filename in testFiles {
+    		let testFilePath = URL(fileURLWithPath: "\(csPath)/\(filename)")
+            let asDocument = try Document(fromJSONFile: testFilePath)
+            let tests: [Document] = try asDocument.get("tests")
+            for test in tests {
+	            let description: String = try test.get("description")
+                if description == "wtimeoutMS as an invalid number" { continue }
```
kmahar: I'm waiting to hear back from Kevin/Jesse about this case (see msgs in #mongoc) - it looks to me like the C driver just ignores a negative wtimeout value rather than throwing an error. but I don't see them skipping this case in their test runner for this so I'm not totally sure we should be skipping it

kmahar: to update, it looks like due a [bug](https://github.com/mongodb/mongo-c-driver/blob/master/tests/test-mongoc-read-write-concern.c#L86) in their test runner they stop executing after the first invalid case (`return` instead of `continue`), so this case wasn't actually being checked, and with that fix they don't pass it. so we will have to skip this until the C driver has a fix. will update with a C driver ticket number later on

Sources/MongoSwift/Collection.swift
```diff
@@ -25,16 +25,23 @@ public struct AggregateOptions: BsonEncodable {
     /// The index to use for the aggregation. The hint does not apply to $lookup and $graphLookup stages.
     // let hint: Optional<(String | Document)>
 
+    /// A ReadConcern to use for this operation. 
+    let readConcern: ReadConcern?
```
mbroadst: could we put `readConcern` and `writeConcern` in a base class instead of explicitly defining it for every options type? `ReadConcernable` 😄

mbroadst: Hmm, I guess it makes it hard to add it to the field-wise initializer :thinking: But it might make it so you don't need `skipKeys` anymore - you could ignore fields brought in by known decorators, leaving just the base struct as the codable type for serialization. Worth a quick check I guess?

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -0,0 +1,213 @@
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+import libmongoc
+
+final class ReadWriteConcernTests: XCTestCase {
+	static var allTests: [(String, (ReadWriteConcernTests) -> () throws -> Void)] {
+        return [
+        	("testReadConcernType", testReadConcernType),
```
mbroadst: weird spacing

Sources/MongoSwift/Client.swift
```diff
@@ -49,6 +60,13 @@ public class MongoClient {
     /// If command and/or server monitoring is enabled, indicates what event types notifications will be posted for.
     internal var monitoringEventTypes: [MongoEventType]?
 
+    /// The read concern set on this client.
+    public var readConcern: ReadConcern {
```
mbroadst: why not `public var readConcern: ReadConcern?` ? I wouldn't expect the client to have a default readConcern with a nil level (tested below)

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -0,0 +1,213 @@
...
+
+    	// expect that a DB created from this client inherits its unset RC 
+    	let db1 = try client1.db("test")
+    	expect(db1.readConcern.level).to(beNil())
+
+    	// expect that a DB created from this client can override the client's unset RC
+    	let db2 = try client1.db("test", options: DatabaseOptions(readConcern: try ReadConcern(.majority)))
+    	expect(db2.readConcern.level).to(equal("majority"))
+
+    	client1.close()
```
mbroadst: do these need to be called explicitly? If you're concerned about memory the deinitializer will be called later

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -0,0 +1,104 @@
...
+    /// Initialize a new ReadConcern from a String corresponding to a read concern level.
+    public init(_ level: String) throws {
+        self._readConcern = mongoc_read_concern_new()
+        if !mongoc_read_concern_set_level(self._readConcern, level) {
+            throw MongoError.readConcernError(message: "Failed to set read concern level to '\(level)'")
+        }
+    }
+
+    /// Initialize a new empty ReadConcern.
+    public init() {
```
mbroadst: does it make sense to make a copy initializer for read concerns? like `init(fromReadConcern: ReadConcern)`

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -0,0 +1,104 @@
...
+        return self.asDocument.description
+    }
+
+    /// Initialize a new ReadConcern from a ReadConcernLevel.
+    public convenience init(_ level: ReadConcernLevel) throws {
+        try self.init(level.rawValue)
+    }
+
+    /// Initialize a new ReadConcern from a String corresponding to a read concern level.
+    public init(_ level: String) throws {
```
mbroadst: hmm, looking at the code in the tests below, I think we should stop these from throwing. I think this looks unintuitive to me:
```
let options1 = RunCommandOptions(readConcern: try ReadConcern(.local))`
```
As far as I've read the mongoc docs it looks like returning false will already emit a log error. As an alternative, I propose that we add an extra default level for a `ReadConcern`: `invalid`. In the case of an error, set the level to `invalid` (or it could just be nil as well now that I think about it). When the `ReadConcern` is eventually used in some I/O operation we should throw an error _there_

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -0,0 +1,213 @@
...
+
+        expect(try coll.distinct(fieldName: "a",
+        	options: DistinctOptions(readConcern: try ReadConcern(.local)))).toNot(throwError())
+    }
+
+    func testConnectionStrings() throws {
+    	let csPath = "\(self.getSpecsPath())/read-write-concern/tests/connection-string"
+    	let testFiles = try FileManager.default.contentsOfDirectory(atPath: csPath).filter { $0.hasSuffix(".json") }
+    	for filename in testFiles {
+    		let testFilePath = URL(fileURLWithPath: "\(csPath)/\(filename)")
```
mbroadst: more weird space, there are a few below as well so I'll stop. Maybe turn on auto formatting in sublime?

kmahar: ugh I think this happened because I copy pasted some code from the `WriteConcern` branch on GitHub into this file - I'd realized a cleaner way to do the tests as I finished filling them out there but I guess that introduced a mix of tabs and spaces. anyway, yes, found a sublime plugin to autoconvert tabs to spaces on save for the future :)

Sources/MongoSwift/ReadWriteConcern.swift
kmahar: sure, added

Sources/MongoSwift/ReadWriteConcern.swift
kmahar: yeah I did feel like that was kind of odd myself as I wrote the tests. 

now that I look at `mongoc_read_concern_set_level`'s [source](https://github.com/mongodb/mongo-c-driver/blob/865e11a3498240dbd927e537dc3f212049f901fb/src/mongoc/mongoc-read-concern.c#L88), I don't think it ever actually returns `false`? from the [docs](http://mongoc.org/libmongoc/current/mongoc_read_concern_set_level.html) it looks like that used to happen if the ReadConcern was "frozen" which no longer happens as of v1.9. I think as long as we pass in a string (which we always will) it should return true. and if the string does not represent a valid RC, mongoc doesn't throw til we try to use it in an operation. 

so I think we can just ignore the result of this function, so let's get rid of the error. we could alternatively use a `precondition` to check that it's true but that's maybe overkill

Sources/MongoSwift/Collection.swift
kmahar: we'd then have to convert every options type to a `class`, `struct`s have no inheritance 😢 
I tried to think about how to do it with a `protocol` + an `extension` of that `protocol` as well, but you can't define stored properties in protocol extensions (which is what we would want for the `readConcern` property), only computed properties and methods.

Sources/MongoSwift/Client.swift
mbroadst: Well those languages don't have idiomatic support for optionals, but I also don't feel very strongly about this. To me it seemed like "Why would I ever check client.readConcern.level == nil", to which the answer might be: "Have I set a client-level read concern, or is it the server default?" In which case, checking whether the optional has been set seems to convey that very same info in clearer, more idiomatic fashion.

kmahar: yeah, I guess having to also check for an unset level does sound annoying, will change it!

PR #55, opened 2018-05-01
SWIFT-30 Implement WriteConcern type

kmahar:
This is on top of the `ReadConcern` implementation so I don't expect you to review it til that's merged in, but just moving this into review for the sake of our tracking where I'm at work-wise. 

Anyway, pretty straightforward, this adds a `WriteConcern` type and adds support for specifying WC at the client, database, collection, and operation level.

### DIFF
--- a/Sources/MongoSwift/Collection.swift
+++ b/Sources/MongoSwift/Collection.swift
@@ -26,13 +26,16 @@ public struct AggregateOptions: Encodable {
     /// The index to use for the aggregation. The hint does not apply to $lookup and $graphLookup stages.
     // let hint: Optional<(String | Document)>
 
-    /// A `ReadConcern` to use for this operation. 
+    /// A `ReadConcern` to use in read stages of this operation.
     let readConcern: ReadConcern?
+
+    /// A `WriteConcern` to use in `$out` stages of this operation.
+    let writeConcern: WriteConcern?
 
     /// Convenience initializer allowing any/all parameters to be optional
     public init(allowDiskUse: Bool? = nil, batchSize: Int32? = nil, bypassDocumentValidation: Bool? = nil,
                 collation: Document? = nil, comment: String? = nil, maxTimeMS: Int64? = nil,
-                readConcern: ReadConcern? = nil) {
+                readConcern: ReadConcern? = nil, writeConcern: WriteConcern? = nil) {
         self.allowDiskUse = allowDiskUse
         self.batchSize = batchSize
         self.bypassDocumentValidation = bypassDocumentValidation
@@ -40,9 +43,10 @@ public struct AggregateOptions: Encodable {
         self.comment = comment
         self.maxTimeMS = maxTimeMS
         self.readConcern = readConcern
+        self.writeConcern = writeConcern
     }
 
-    // Encode everything except readConcern
+    // Encode everything except readConcern and writeConcern
     private enum CodingKeys: String, CodingKey {
         case allowDiskUse, batchSize, bypassDocumentValidation,
             collation, maxTimeMS, comment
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/Collection.swift
```diff
@@ -604,10 +675,11 @@ public class MongoCollection<T: Codable> {
      */
     public func aggregate(_ pipeline: [Document], options: AggregateOptions? = nil) throws -> MongoCursor<Document> {
         let encoder = BsonEncoder()
-        let opts = try ReadConcern.append(options?.readConcern, to: try encoder.encode(options), callerRC: self.readConcern)
+        let withRC = try ReadConcern.append(options?.readConcern, to: try encoder.encode(options), callerRC: self.readConcern)
```
mbroadst: might this look a little cleaner if it was like:
```
let opts = try encoder.encode(options)
try ReadConcern.append(options?.readConcern, to: opts, callerRC: self.readConcern)
try WriteConcern.append(options?.writeConcern, to: opts, callerWC: self.writeConcern)
```

Sources/MongoSwift/Database.swift
```diff
@@ -136,6 +156,15 @@ public class MongoDatabase {
         return rcObj
     }
 
+    /// The `WriteConcern` set on this database, or `nil` if one is not set.
+    public var writeConcern: WriteConcern? {
+        // per libmongoc docs, we don't need to handle freeing this ourselves
```
mbroadst: feels like the third time I've seen a block of code that looks _very_ similar

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -115,5 +115,134 @@ public class ReadConcern: Equatable, CustomStringConvertible {
         mongoc_read_concern_destroy(readConcern)
         self._readConcern = nil
     }
+}
+
+/// A class to represent a MongoDB write concern.
+public class WriteConcern: Equatable, CustomStringConvertible {
```
mbroadst: nit: for debugability would it make sense to make it describable too, and use the `asDocument` representation so you can call `print` on it?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -115,5 +115,134 @@ public class ReadConcern: Equatable, CustomStringConvertible {
...
+
+    /// Initializes a new WriteConcern by copying a mongoc_write_concern_t.
+    /// The caller is responsible for freeing the original mongoc_write_concern_t.
+    internal init(_ writeConcern: OpaquePointer?) {
+        self._writeConcern = mongoc_write_concern_copy(writeConcern)
+    }
+
+    /// Sets the journal value on this writeConcern
+    private func setJournal(_ journal: Bool?) {
+        if let journal = journal { mongoc_write_concern_set_journal(self._writeConcern, journal) }
```
mbroadst: is it useful to have these as separately defined methods? Looks like they can just be embedded in the initializer, unless you want to make the public variables have getters/setters

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -79,11 +125,53 @@ final class ReadWriteConcernTests: XCTestCase {
         client3.close()
     }
 
+    func testClientWriteConcern() throws {
+        // create a client with no options and check its RC
+        let client1 = try MongoClient()
```
mbroadst: I know we've spoken about this before, but `close` is called in the clients deinitializer so the calls to that are somewhat superfluous. Also, visually if you broke these up into sub-scopes, it might be easier to scan which test is which:
```
  { 
       // create a client with w: 1 and check its WC
       let client2 = try MongoClient(options: ClientOptions(writeConcern: WriteConcern(w: 1)))
       // although w:1 is default, if it is explicitly provided it should be set
       expect(client2.writeConcern?.w).to(equal(1))

       // expect that a DB created from this client inherits its WC
       let db3 = try client2.db("test")
       expect(db3.writeConcern?.w).to(equal(1))

       // expect that a DB created from this client can override the client's WC
       let db4 = try client2.db("test", options: DatabaseOptions(writeConcern: WriteConcern(w: 2)))
       expect(db4.writeConcern?.w).to(equal(2))
  }
```

kmahar:
needs a lot of refactoring, will reopen when it's actually ready for review

PR #56, opened 2018-05-03
SWIFT-85: Add a custom RegularExpression type

kmahar:
As we discussed earlier, this is necessary if we want all BSON types to be `Codable`.


Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -536,39 +573,33 @@ extension NSRegularExpression: BsonValue {
...
             preconditionFailure("Failed to retrieve regular expression options")
         }
         let optionsString = String(cString: stringOptions)
-        let opts = NSRegularExpression.optionsFromString(optionsString)
+
+        return RegularExpression(pattern: patternString, options: optionsString)
+    }
+
+    // Creates an NSRegularExpression with the specified pattern and options.
+    public var asNSRegularExpression: NSRegularExpression {
```
mbroadst: `.nsRegularExpression`?

PR #57, opened 2018-05-07
SWIFT-21 Complete BsonDecoder implementation

kmahar:
Various comments on this which I will make inline, but the gist is that this makes all of our `BsonValue` types `Decodable`, adds support for converting between the different number types, and add support for decoding from JSON, including just single JSON values.

So I believe this covers all of SWIFT-21 + SWIFT-86 + SWIFT-87


Sources/MongoSwift/BSON/BsonDecoder.swift
```diff
@@ -154,14 +178,22 @@ extension _BsonDecoder {
...
-        return try T(from: unwrapped)
+
+        guard let primitive = T(from: unwrapped) else {
+            throw DecodingError._numberMismatch(at: self.codingPath, expectation: type, reality: value)
+        }
+        return primitive
     }
 
     fileprivate func unbox<T: Decodable>(_ value: BsonValue?, as type: T.Type) throws -> T? {
+        // if the data is already stored as the correct type in the document, then we can short-circuit
```
kmahar: this check is to support the case we discussed in SWIFT-86 where we are given extended JSON input, and so the value is already decoded to the right type and doesn't need to be initialized from the decoder.

Sources/MongoSwift/BSON/BsonDecoder.swift
```diff
@@ -352,7 +384,7 @@ private struct _BsonUnkeyedDecodingContainer: UnkeyedDecodingContainer {
     }
 
     /// Decodes a BsonValue type from this container.
-    private mutating func decodeBsonValue<T: BsonValue>(_ type: T.Type) throws -> T {
+    private mutating func decodeBsonType<T: BsonValue>(_ type: T.Type) throws -> T {
```
kmahar: this is to be consistent with the naming for the keyed container

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -159,28 +159,33 @@ extension Array: BsonValue {
...
 /// A struct to represent the BSON Binary type.
-public struct Binary: BsonValue, Equatable {
+public struct Binary: BsonValue, Equatable, Decodable {
+
     public var bsonType: BsonType { return .binary }
-    var data: Data
-    var subtype: BsonSubtype
+
+    /// The binary data.
+    public let data: Data
```
kmahar: bunch of properties in our `BsonValue` structs were not public but should be, oops 😳

Sources/MongoSwift/BSON/CodableNumber.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+
+/// A protocol for numbers that require encoding/decoding support but are not necessarily BSON types.
+internal protocol CodableNumber {
```
kmahar: as I mentioned in person the other day, it seems like "primitive" is not really the best way to describe a protocol that only contains number types, so hopefully this is more clear. I guess it's internal anyway so naming is not crucial

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -63,19 +63,35 @@ final class CodecTests: XCTestCase {
     struct BasicStruct: Decodable, Equatable {
         let int: Int
         let string: String
+
+        public static func == (lhs: BasicStruct, rhs: BasicStruct) -> Bool {
```
kmahar: added so tests still pass in swift 4.0 without synthesized `Equatable`

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -134,4 +154,318 @@ final class CodecTests: XCTestCase {
...
+            "dec": Decimal128("1.2E+10"),
+            "minkey": MinKey(),
+            "maxkey": MaxKey(),
+            "regex": RegularExpression(pattern: "^abc", options: "imx")
+        ]
+
+        let res2 = try decoder.decode(AllBsonTypes.self, from: doc2)
+        expect(res2).to(equal(expected))
+
+        let base64 = "//8="
```
kmahar: for some reason my syntax highlighter can't handle this being inline...

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -134,4 +154,318 @@ final class CodecTests: XCTestCase {
...
+        let extjson = """
```
kmahar: testing SWIFT-86 behavior

Sources/MongoSwift/BSON/BsonDecoder.swift
```diff
@@ -525,4 +557,127 @@ private extension DecodingError {
...
+            var output = Document()
+            for key in container.allKeys {
+                let k = key.stringValue
+                output[k] = try Document.recursivelyDecodeKeyed(key: key, container: container)
+            }
+            self = output
+        }
+    }
+
+    /// Switch through all possible BSON types (a document can contain any) and try recursively decoding the value
```
mbroadst: does it make sense to actually maintain a list of BSON types, and to iterate through those programmatically for the following guarded let statements? (e.g. `[Binary.self, Double.self, String.self...]`) This would have the benefit of making it very clear which types directly map to the spec, and also seem to cut down on boilerplate. @

Sources/MongoSwift/BSON/BsonDecoder.swift
```diff
@@ -525,4 +557,127 @@ private extension DecodingError {
...
+    /// stored under `key` as that type from the provided keyed container.
+    private static func recursivelyDecodeKeyed(key: _BsonKey, container: KeyedDecodingContainer<_BsonKey>) throws -> BsonValue {
+        let k = key.stringValue
+        if let value = try? container.decode(Double.self, forKey: key) {
+            return value
+        } else if let value = try? container.decode(String.self, forKey: key) {
+            return value
+        // this will recursively call Document.init(from: decoder Decoder)
+        } else if let value = try? container.decode(Document.self, forKey: key) {
```
mbroadst: It seems like the complex cases should be tried last since they have non-trivial implementations (recursive init calls for `Document` as well as for the nested containers).

Sources/MongoSwift/BSON/BsonDecoder.swift
```diff
@@ -525,4 +557,127 @@ private extension DecodingError {
...
+            return value
+        } else {
+            throw MongoError.typeError(message: "Encountered a value in an keyed container under key \(k) that could not be decoded to any BSON type")
+        }
+    }
+
+    /// Switch through all possible BSON types (a document can contain any) and try recursively decoding the next value
+    /// as that type from the provided unkeyed container.
+    private static func recursivelyDecodeUnkeyed(container: inout UnkeyedDecodingContainer) throws -> BsonValue {
+        if let value = try? container.decode(Double.self) {
```
mbroadst: seems like this can be merged with the `keyed` variant above, by just passing `nil` for the key? Am I missing something?

Sources/MongoSwift/BSON/BsonValue.swift
mbroadst: we keep finding more of these! I wonder if it's possible to make a sample program that we ship with the driver that actively uses all types from outside the package in order to catch issues like this earlier? Or are there any linting tools we can use for this purpose?

Sources/MongoSwift/BSON/CodableNumber.swift
mbroadst: 👍

Tests/MongoSwiftTests/CodecTests.swift
mbroadst: :+1:

mbroadst: Is this the issue Jason was running into the other day?

Sources/MongoSwift/BSON/BsonValue.swift
kmahar: Hmm, I'll look into it... Now that we have docs (almost), one easy thing to do is to look at the generated docs for a type and see if the properties and methods you expect show up there since docs are only generated for public stuff.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -460,10 +469,16 @@ public struct CodeWithScope: BsonValue {
...
+
+    public static func == (lhs: CodeWithScope, rhs: CodeWithScope) -> Bool {
+        return lhs.code == rhs.code && lhs.scope == rhs.scope
+    }
 }
 
 /// A struct to represent the BSON MaxKey type.
-public struct MaxKey: BsonValue, Equatable {
+public struct MaxKey: BsonValue, Equatable, Decodable {
+    private var maxKey = 1
```
kmahar: I added this private var to deal with the fact that anything was parseable as a MinKey or MaxKey

Sources/MongoSwift/BSON/BsonDecoder.swift
mbroadst: :+1:

Sources/MongoSwift/BSON/BsonValue.swift
mbroadst: :+1:

PR #58, opened 2018-05-14
SWIFT-77 Rewrite BsonEncoder to conform to Encoder protocol 

kmahar:
This is all on top the `BsonDecoder` PR so hold off on reviewing til that's in.

To make the review process easier for switching to the new encoder, I've written this implementation in its own file and temporarily named it `BSONEncoder`(instead of `BsonEncoder`). The old encoder still exists and all of the options types are encoded with it. 
In a second PR to follow this one, I will remove the old encoder and replace all usages of it with this new one, in the process making all of our options types `Encodable`.

### DIFF
--- a/Sources/MongoSwift/BSON/BsonDecoder.swift
+++ b/Sources/MongoSwift/BSON/BsonDecoder.swift
@@ -75,10 +75,10 @@ public class BsonDecoder {
 }
 
 /// An internal class to actually implement the `Decoder` protocol.
-private class _BsonDecoder: Decoder {
+internal class _BsonDecoder: Decoder {
 
     /// The decoder's storage.
-    fileprivate var storage: _BsonDecodingStorage
+    internal var storage: _BsonDecodingStorage
 
     /// Options set on the top-level decoder.
     fileprivate let options: BsonDecoder._Options
@@ -148,7 +148,7 @@ private class _BsonDecoder: Decoder {
 }
 
 // Storage for a _BsonDecoder.
-private struct _BsonDecodingStorage {
+internal struct _BsonDecodingStorage {
 
     /// The container stack, consisting of `BsonValue?`s. 
     private(set) fileprivate var containers: [BsonValue?] = []
@@ -160,7 +160,7 @@ private struct _BsonDecodingStorage {
     fileprivate var count: Int { return self.containers.count }
 
     /// The container at the top of the stack.
-    fileprivate var topContainer: BsonValue? {
+    internal var topContainer: BsonValue? {
         precondition(self.containers.count > 0, "Empty container stack.")
         return self.containers.last!
     }
@@ -566,7 +566,7 @@ internal struct _BsonKey: CodingKey {
     internal static let `super` = _BsonKey(stringValue: "super")!
 }
 
-private extension DecodingError {
+internal extension DecodingError {
     static func _typeMismatch(at path: [CodingKey], expectation: Any.Type, reality: BsonValue?) -> DecodingError {
         let description = "Expected to decode \(expectation) but found \(type(of: reality)) instead."
         return .typeMismatch(expectation, Context(codingPath: path, debugDescription: description))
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/BSON/BsonDecoder.swift
```diff
@@ -576,18 +576,12 @@ private extension DecodingError {
...
                          "but found value \(String(describing: reality)) of type \(type(of: reality)) instead."
         return .typeMismatch(expectation, Context(codingPath: path, debugDescription: description))
-
     }
 }
 
 /// This needs to be in this file to access some fileprivate decoder properties
 extension Document: Decodable {
 
-    static let decodeToTypes: [Decodable.Type] = [Double.self, String.self, Binary.self, ObjectId.self,
```
kmahar: accidentally left this in the decoder while testing your proposed approach, oops

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    fileprivate func box<T: Encodable>(_ value: T) throws -> BsonValue? {
+
+        // if it's already a BsonValue, just return it, unless if it is an 
+        // array. technically [Any] is a BsonValue, but we can only use this
+        // short-circuiting if all the elements are actually BsonValues.
+        if let bsonValue = value as? BsonValue, !(bsonValue is [Any]) {
+            return bsonValue
+        } else if let bsonArray = value as? [BsonValue?] {
```
mbroadst: Why `[BsonValue?]` and not `[BsonValue]`?

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+        self.codingPath = codingPath
+        self.container = container
+    }
+
+    public mutating func encodeNil(forKey key: Key) throws { self.container[key.stringValue] = nil }
+    public mutating func encode(_ value: Bool, forKey key: Key) throws { self.container[key.stringValue] = value }
+    public mutating func encode(_ value: Int, forKey key: Key) throws { self.container[key.stringValue] = value }
+    public mutating func encode(_ value: Int8, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: Int16, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: Int32, forKey key: Key) throws { self.container[key.stringValue] = value }
```
mbroadst: Why aren't these two using `encodeNumber`?

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+    public mutating func encode(_ value: Int64, forKey key: Key) throws { self.container[key.stringValue] = value }
+    public mutating func encode(_ value: UInt, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: UInt8, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: UInt16, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: UInt32, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: UInt64, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: String, forKey key: Key) throws { self.container[key.stringValue] = value }
+    public mutating func encode(_ value: Float, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
+    public mutating func encode(_ value: Double, forKey key: Key) throws { self.container[key.stringValue] = value }
```
mbroadst: same here

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+
+    fileprivate func add(_ value: BsonValue?) {
+        array.append(value)
+    }
+
+    var count: Int { return array.count }
+
+    /// Converts self to a `Document` where keys "0", "1", etc.
+    /// correspond to array indices. 
+    func asDocument() -> Document {
```
mbroadst: Hm. Do we want to encode arrays this way in bson? It seems less efficient than just storing the array as a single value. Or is this even used anywhere?

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+}
+
+/// A private class wrapping a Swift dictionary so we can pass it by reference
+/// for encoder storage purposes. We use this rather than NSMutableDictionary 
+/// because it allows us to preserve Swift type information.
+private class MutableDictionary: BsonValue {
+
+    var bsonType: BsonType { return .document }
+
+    // rather than using a dictionary, do this so we preserve key orders
```
mbroadst: nice

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+
+private extension EncodingError {
+    static func _numberError<T: CodableNumber>(at path: [CodingKey], value: T) -> EncodingError {
+        let description = "Value \(String(describing: value)) of type \(type(of: value)) cannot be " +
+                            "exactly represented by a BSON number type (Int, Int32, Int64 or Double)."
+        return .invalidValue(value, Context(codingPath: path, debugDescription: description))
+    }
+}
+
+extension Document: Encodable {
```
mbroadst: I think this should be moved to like `Document+Codable.swift` or something

Sources/MongoSwift/BSON/BsonEncoderNew.swift
```diff
@@ -0,0 +1,647 @@
...
+            }
+        case let val as Binary:
+            try container.encode(val, forKey: k)
+        case let val as Bool:
+            try container.encode(val, forKey: k)
+        case let val as Date:
+            try container.encode(val, forKey: k)
+        case let val as Decimal128:
+            try container.encode(val, forKey: k)
+        case let val as Document:
```
mbroadst: same comment as Decoder: let's defer the complex cases to the end of the switch

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -423,7 +423,7 @@ extension Int64: BsonValue {
 }
 
 /// A struct to represent the BSON Code and CodeWithScope types.
-public struct CodeWithScope: BsonValue, Decodable, Equatable {
+public struct CodeWithScope: BsonValue, Codable, Equatable {
```
mbroadst: nit: everything else seems to be using the form: `BsonValue, Equatable, Codable`

Sources/MongoSwift/BSON/BsonEncoderNew.swift
kmahar: I'm not sure what you mean about storing the array as a single value? isn't this exactly what a BSON array is - a document with special keys?

either way this isn't actually used anywhere right now - `MutableArray` and `MutableDictionary` here are slightly modified versions of my original types of the same name that the old encoder uses, so this was a carry-over from that code, where it is used in a few places.

however, I think we may want to keep this function around. this PR does not include support for encoding arrays at the top level as Stitch requested (I made [SWIFT-95](https://jira.mongodb.org/browse/SWIFT-95) for that). 
in the case where the top level container is a `MutableArray`, we will need a way to convert it to a `Document` for the encoder to return. so this would be used in `encode<T: Encodable>(_ value: T) throws -> Document` (where I already use `.asDocument()` on a `MutableDictionary`.) 
were we to use the MongoKitten approach I suggested yesterday where `Document`s have an internal `isArray` flag, we'd set that to `true` for the `Document` returned here. but I guess we should figure out exactly how we want to implement that

Sources/MongoSwift/BSON/BsonEncoderNew.swift
kmahar: 👍

Sources/MongoSwift/BSON/BsonValue.swift
kmahar: fixed

Sources/MongoSwift/BSON/BsonEncoderNew.swift
kmahar: moved, and also pulled in the `extension Document: Decodable` code. note this required changing some types and properties from `private`/`fileprivate` to `internal`.

Sources/MongoSwift/BSON/BsonEncoderNew.swift
kmahar: this would come up in the case someone was encoding an array of optionals that had a nil in it. for example
```swift
struct MyStruct: Encodable {
    let ints: [Int?]
}

let s = MyStruct(ints: [1, 2, nil])
```

Sources/MongoSwift/BSON/BsonEncoderNew.swift
mbroadst: so according to the bson spec, an array is encoded as:`"\x04" e_name document` where a document is `document | ::= | int32 e_list "\x00"` and an element is `e_list | ::= | element e_list`. So that means that arrays are essentially stored in an ordered list, not as key value pairs where the keys are the indices. What this method is doing is the latter, encoding a `MutableArray` as a document, rather than, say a single document with a placeholder value (like you use for the stitch hacks). This approach is certainly serializable, but is somewhat inefficient in particular for very large arrays. The MongoKitten approach would also work just fine imho.

mbroadst: For now, let's remove it from the file and reintroduce when it's actually used after having related design conversation.

PR #59, opened 2018-05-14
SWIFT-88 Convert all options types to use the new BsonEncoder

kmahar:
Switches to the new `BsonEncoder`, making all options types `Encodable`. On top of other PRs so don't bother reviewing til those are committed


Sources/MongoSwift/BSON/BsonEncoder.swift
```diff
@@ -1,348 +1,557 @@
 import Foundation
```
kmahar: FYI I would just ignore this file's entire diff... all I did was rename `BsonEncoderNew.swift` -> `BsonEncoder.swift`

kmahar: actually you should look at `encodeIfPresent`, that was a new addition

kmahar:
Ok, after this is merged in, will rename `BsonEncoderNew.swift` -> `BsonEncoder.swift`. leaving original name now so the diff makes more sense

Sources/MongoSwift/Collection.swift
```diff
@@ -386,20 +400,43 @@ public struct IndexModel: BsonEncodable {
...
-            try encoder.encode(self.defaultName, forKey: "name")
-        }
+    // Encode own data as well as nested options data
+    private enum CodingKeys: String, CodingKey {
+        case key, background, expireAfter, name, sparse, storageEngine, unique,
+            version, defaultLanguage, languageOverride, textVersion, weights,
+            sphereVersion, bits, max, min, bucketSize, partialFilterExpression, collation
+    }
+
+    public func encode(to encoder: Encoder) throws {
```
kmahar: this is annoying, but we have to send the data with the options stored on the level as the index keys, so I couldn't figure out a nicer way to do this

PR #62, opened 2018-05-24
Add matcher functions to handle inconsistent key ordering

kmahar:


Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -484,17 +484,20 @@ final class CodecTests: XCTestCase {
         """.data(using: .utf8)!
 
         let expected: Document = [
```
kmahar: I reordered this to make it match the order of the JSON we parse, since before it was just hard-coded in the expected order

PR #64, opened 2018-05-29
SWIFT-63 Add support for hint and cursorType options to Collection API methods

kmahar:
This adds support for providing a cursor type to `find`, and an index hint to `find`, `aggregate`, and `count`, encoding using the strategies we discussed earlier.


Sources/MongoSwift/Collection.swift
```diff
@@ -1,5 +1,23 @@
 import libmongoc
 
+/// An index to "hint" or force MongoDB to use when performing a query.
+public enum Hint: Encodable {
```
kmahar: open to any better naming suggestions for these cases

jmikola: The [`find` docs](https://docs.mongodb.com/manual/reference/command/find/) refer to both forms as an "index specification", although [`cursor.hint()`](https://docs.mongodb.com/manual/reference/method/cursor.hint/) refers only to the document as an "index specification". The only alternative I might consider is "indexKeys", but I'm comfortable with using "indexSpec" as it should be quite clear to users.

Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -176,8 +175,13 @@ private struct CMTest {
             if let max = modifiers?["$maxTimeMS"] as? Int {
                 maxTime = Int64(max)
             }
+            var hint: Hint?
+            if let hintDoc = modifiers?["$hint"] as? Document {
```
jmikola: Historically, the [`$hint` modifier](https://docs.mongodb.com/manual/reference/operator/meta/hint/#metaOp._S_hint) has accepted either a string or document.

A more general question might be why we're supporting modifiers at all in Swift. Isn't the driver new enough that we should only need to concern ourself with the `find` command's API?

Sources/MongoSwift/Collection.swift
```diff
@@ -191,14 +214,19 @@ public struct FindOptions: Encodable {
...
+                comment: String? = nil, cursorType: CursorType? = nil, hint: Hint? = nil, limit: Int64? = nil,
+                max: Document? = nil, maxAwaitTimeMS: Int64? = nil, maxScan: Int64? = nil, maxTimeMS: Int64? = nil,
+                min: Document? = nil, noCursorTimeout: Bool? = nil, projection: Document? = nil,
+                readConcern: ReadConcern? = nil, returnKey: Bool? = nil, showRecordId: Bool? = nil, skip: Int64? = nil,
+                sort: Document? = nil) {
         self.allowPartialResults = allowPartialResults
         self.batchSize = batchSize
         self.collation = collation
         self.comment = comment
+        self.cursorType = cursorType
```
jmikola: Would it make sense to note (either here or in the declaration above) that we're only storing this for the benefit of user debugging?

kmahar: yes, I will add it, for our own sake if no one else's

Tests/MongoSwiftTests/CommandMonitoringTests.swift
kmahar: there is only one case that uses modifiers and it stores the hint as a `doc`, so I didn't bother checking for a string, as I assume no more cases with modifiers will be added in the future.

we don't support providing a modifiers document, but our `FindOptions` contains all the fields one would have put in modifiers, so rather than skipping the test I parse them out to to the correct fields for the sake of more test coverage

jmikola: Oh, I didn't realize this was in a test file and thought the driver itself was actually supporting the deprecated "modifiers" option. Nevermind!

mbroadst:
@kmahar looks like this one got away from me. What's the current status of this PR, does it need to be updated per other recent changes?

kmahar:
@mbroadst I just rebased on master. I don't think any updates are needed, so should be ready for review

Tests/MongoSwiftTests/CollectionTests.swift
```diff
@@ -298,4 +298,26 @@ final class CollectionTests: XCTestCase {
...
+    }
+
+    func testEncodeHint() throws {
+        let encoder = BsonEncoder()
+
+        let stringHint = AggregateOptions(hint: .indexName("hi"))
+        expect(try encoder.encode(stringHint)).to(equal(["hint": "hi"]))
+
+        let docHint = AggregateOptions(hint: .indexSpec(["hi": 1]))
+        expect(try encoder.encode(docHint)).to(equal(["hint": ["hi": 1] as Document]))
```
jmikola: This casts `["hi": 1]` as a Document, but don't you also want `["hint": ["hi": 1] as Document]` cast to a Document? Is that omitted because you're not testing that AggregateOptions itself is encoded to a Document here (and/or just doing a less strict equality check)?

kmahar: so the compiler is able to figure out that the whole thing (`["hint": ["hi": 1] as Document]`) is a `Document` because I'm comparing it to the result of a function that always returns a `Document` (`encode`). 

the reason I have to say `as Document` for a nested doc is that the compiler *doesn't* know that the only `BsonValue` that is `ExpressibleByDictionaryLiteral` is a `Document`. so it considers just saying `["hint": ["hi": 1]]` to be ambiguous

PR #66, opened 2018-05-31
SWIFT-108 Refactor how ReadConcerns are encoded

kmahar:


Sources/MongoSwift/Options.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
```
kmahar: kind of a lame name for this file but it seemed like this stuff should be grouped together somewhere... so let me know if you have any opinions

kmahar:
closing this as we were doing way too much work, will open new PR shortly with much simplified solution

PR #69, opened 2018-06-06
SWIFT-73 Add @discardableResult to non-void functions with negligible return values

kmahar:
cc @jsflax

The compiler generates a warning if you don't assign the return value of a non-void function to something. Adding the `@discardableResult` attribute tells the compiler to ignore that. [docs on Swift attributes](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html)

This is useful in cases where a user may not care about the results, and saves them having to assign the return values to `_`. You can see several examples of this in the tests I modified. 

Let me know if you think I missed any methods that have negligible results, or if you think we should still be forcing users to acknowledge the return values of any of the methods I modified.

### DIFF
--- a/Sources/MongoSwift/Collection.swift
+++ b/Sources/MongoSwift/Collection.swift
@@ -701,6 +701,7 @@ public class MongoCollection<T: Codable> {
      * - Returns: The optional result of attempting to perform the insert. If the `WriteConcern`
      *            is unacknowledged, `nil` is returned.
      */
+    @discardableResult
     public func insertOne(_ value: CollectionType, options: InsertOneOptions? = nil) throws -> InsertOneResult? {
         let encoder = BsonEncoder()
         let document = try encoder.encode(value)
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/Collection.swift
```diff
@@ -701,6 +701,7 @@ public class MongoCollection<T: Codable> {
      * - Returns: The optional result of attempting to perform the insert. If the `WriteConcern`
```
mbroadst: are there methods elsewhere we want to apply this? e.g. `Database.runCommand` maybe?

jsflax:
Nice! 👍

Sources/MongoSwift/Collection.swift
kmahar: added there - looking through the other API methods I think that's it, everything else is a read command so presumably you'd always want the results

mbroadst: 👍 LGTM

PR #70, opened 2018-06-07
Fix mutation bug with `Document.merge`

kmahar:
We want copy-on-write behavior for `Document`s, and we implemented this already for when values are set by using `isKnownUniquelyReferenced` on the backing `DocumentStorage`. 

However, I forgot to add the same check for this new mutating function, `merge`, which can lead to some unexpected behavior like this:

```swift
let a: Document = ["x": 1]
var b = a
try b.merge(["y": 2])
// a is now  ["x": 1, "y": 2]
```

This fixes that issue by performing the same `isKnownUniquelyReferenced` check before merging in a new document.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -200,7 +200,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
 
         set(newValue) {
-            self.checkIsUniquelyReferenced()
+            self.copyStorageIfRequired()
 
             guard let value = newValue else {
                 if !bson_append_null(self.data, key, Int32(key.count)) {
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -255,10 +245,27 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
+    /// of the underlying `bson_t` and lets the copy/copies keep the original.
+    /// This allows us to provide value semantics for `Document`s. 
+    /// This happens if someone copies a document and modifies it.
+    /// For example: 
+    ///  let doc1: Document = ["a": 1]
+    ///  var doc2 = doc1
+    ///  doc2["b"] = 2
+    /// Therefore, this function should be called just before we are about to
+    /// modify a document - either by setting a value or merging in another doc.
+    private mutating func checkIsUniquelyReferenced() {
```
mbroadst: nit: this is doing more than just checking if its uniquely referenced, I think it sounds more like `copyStorageIfRequired` or something

kmahar: 👍 fixed

PR #71, opened 2018-06-07
SWIFT-108 Refactor how ReadConcerns are encoded

kmahar:
Ugh... more things that should be public but weren't. I think we should enable [this linter rule](https://github.com/realm/SwiftLint/blob/master/Rules.md#explicit-acl) requiring explicit access control levels? We have a bunch of minor linter violations I have been meaning to address anyway so I just opened [SWIFT-115](https://jira.mongodb.org/browse/SWIFT-115) to deal with that. 

Anyway, this gets rid of the special logic for encoding `ReadConcern`s as libmongoc will already take care of that for us, and makes `ReadConcern` `Codable` so we can encode it with the rest of the options. 

Also did some refactoring of `ReadConcern` to pull out a couple of the protocols into extensions. unfortunately we can't do this with `Codable` because `required init`s for classes have to be in the class definition.

### DIFF
--- a/Sources/MongoSwift/Collection.swift
+++ b/Sources/MongoSwift/Collection.swift
@@ -27,7 +27,7 @@ public struct AggregateOptions: Encodable {
     // let hint: Optional<(String | Document)>
 
     /// A `ReadConcern` to use for this operation. 
-    let readConcern: ReadConcern?
+    public let readConcern: ReadConcern?
 
     /// Convenience initializer allowing any/all parameters to be optional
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/Collection.swift
```diff
@@ -41,12 +41,6 @@ public struct AggregateOptions: Encodable {
         self.maxTimeMS = maxTimeMS
         self.readConcern = readConcern
     }
-
-    // Encode everything except readConcern
```
mbroadst: so nice to drop these

PR #72, opened 2018-06-08
SWIFT-30 Implement WriteConcern type 

kmahar:
Main changes since the last time you looked at this @mbroadst -
- no more manually appending `WriteConcern`s
- `WriteConcern` now `Codable`
- I switched `w` to an `enum` to better reflect the three different options for `w` values: tag, number, and "majority"

### DIFF
--- a/Sources/MongoSwift/Collection.swift
+++ b/Sources/MongoSwift/Collection.swift
@@ -26,13 +26,16 @@ public struct AggregateOptions: Encodable {
     /// The index to use for the aggregation. The hint does not apply to $lookup and $graphLookup stages.
     // let hint: Optional<(String | Document)>
 
-    /// A `ReadConcern` to use for this operation. 
+    /// A `ReadConcern` to use in read stages of this operation.
     public let readConcern: ReadConcern?
+
+    /// A `WriteConcern` to use in `$out` stages of this operation.
+    let writeConcern: WriteConcern?
 
     /// Convenience initializer allowing any/all parameters to be optional
     public init(allowDiskUse: Bool? = nil, batchSize: Int32? = nil, bypassDocumentValidation: Bool? = nil,
                 collation: Document? = nil, comment: String? = nil, maxTimeMS: Int64? = nil,
-                readConcern: ReadConcern? = nil) {
+                readConcern: ReadConcern? = nil, writeConcern: WriteConcern? = nil) {
         self.allowDiskUse = allowDiskUse
         self.batchSize = batchSize
         self.bypassDocumentValidation = bypassDocumentValidation
@@ -40,6 +43,7 @@ public struct AggregateOptions: Encodable {
         self.comment = comment
         self.maxTimeMS = maxTimeMS
         self.readConcern = readConcern
+        self.writeConcern = writeConcern
     }
 }
 
@@ -220,9 +224,18 @@ public struct InsertOneOptions: Encodable {
     /// If true, allows the write to opt-out of document level validation.
     public let bypassDocumentValidation: Bool?
 
+    /// An optional WriteConcern to use for the command.
+    let writeConcern: WriteConcern?
+
     /// Convenience initializer allowing bypassDocumentValidation to be omitted or optional
-    public init(bypassDocumentValidation: Bool? = nil) {
+    public init(bypassDocumentValidation: Bool? = nil, writeConcern: WriteConcern? = nil) {
         self.bypassDocumentValidation = bypassDocumentValidation
+        self.writeConcern = writeConcern
+    }
+
+    // Encode everything except writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case bypassDocumentValidation
     }
 }
 
@@ -236,10 +249,19 @@ public struct InsertManyOptions: Encodable {
     /// Defaults to true.
     public var ordered: Bool = true
 
+    /// An optional WriteConcern to use for the command.
+    let writeConcern: WriteConcern?
+
     /// Convenience initializer allowing any/all parameters to be omitted or optional
-    public init(bypassDocumentValidation: Bool? = nil, ordered: Bool? = true) {
+    public init(bypassDocumentValidation: Bool? = nil, ordered: Bool? = true, writeConcern: WriteConcern? = nil) {
         self.bypassDocumentValidation = bypassDocumentValidation
         if let o = ordered { self.ordered = o }
+        self.writeConcern = writeConcern
+    }
+
+    // Encode everything except writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case bypassDocumentValidation, ordered
     }
 }
 
@@ -257,13 +279,22 @@ public struct UpdateOptions: Encodable {
     /// When true, creates a new document if no document matches the query.
     public let upsert: Bool?
 
+    /// An optional WriteConcern to use for the command.
+    let writeConcern: WriteConcern?
+
     /// Convenience initializer allowing any/all parameters to be optional
     public init(arrayFilters: [Document]? = nil, bypassDocumentValidation: Bool? = nil, collation: Document? = nil,
-                upsert: Bool? = nil) {
+                upsert: Bool? = nil, writeConcern: WriteConcern? = nil) {
         self.arrayFilters = arrayFilters
         self.bypassDocumentValidation = bypassDocumentValidation
         self.collation = collation
         self.upsert = upsert
+        self.writeConcern = writeConcern
+    }
+
+    // Encode everything except writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case arrayFilters, bypassDocumentValidation, collation, upsert
     }
 }
 
@@ -278,11 +309,21 @@ public struct ReplaceOptions: Encodable {
     /// When true, creates a new document if no document matches the query.
     public let upsert: Bool?
 
+    /// An optional WriteConcern to use for the command.
+    let writeConcern: WriteConcern?
+
     /// Convenience initializer allowing any/all parameters to be optional
-    public init(bypassDocumentValidation: Bool? = nil, collation: Document? = nil, upsert: Bool? = nil) {
+    public init(bypassDocumentValidation: Bool? = nil, collation: Document? = nil, upsert: Bool? = nil,
+                writeConcern: WriteConcern? = nil) {
         self.bypassDocumentValidation = bypassDocumentValidation
         self.collation = collation
         self.upsert = upsert
+        self.writeConcern = writeConcern
+    }
+
+    // Encode everything except writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case bypassDocumentValidation, collation, upsert
     }
 }
 
@@ -291,9 +332,18 @@ public struct DeleteOptions: Encodable {
     /// Specifies a collation.
     public let collation: Document?
 
+    /// An optional WriteConcern to use for the command.
+    let writeConcern: WriteConcern?
+
      /// Convenience initializer allowing collation to be omitted or optional
-    public init(collation: Document? = nil) {
+    public init(collation: Document? = nil, writeConcern: WriteConcern? = nil) {
         self.collation = collation
+        self.writeConcern = writeConcern
+    }
+
+    // Encode everything except writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case collation
     }
 }
 
@@ -497,6 +547,18 @@ public struct IndexOptions: Encodable {
     }
 }
 
+/// Options to use when creating a new index on a `MongoCollection`.
+public struct CreateIndexOptions {
+    /// An optional `WriteConcern` to use for the command
+    let writeConcern: WriteConcern?
+}
+
+/// Options to use when dropping an index from a `MongoCollection`.
+public struct DropIndexOptions {
+    /// An optional `WriteConcern` to use for the command
+    let writeConcern: WriteConcern?
+}
+
 /// A MongoDB collection.
 public class MongoCollection<T: Codable> {
     private var _collection: OpaquePointer?
@@ -524,6 +586,15 @@ public class MongoCollection<T: Codable> {
         let rcObj = ReadConcern(from: readConcern)
         if rcObj.isDefault { return nil }
         return rcObj
+    }
+
+    /// The `WriteConcern` set on this collection, or nil if one is not set.
+    public var writeConcern: WriteConcern? {
+        // per libmongoc docs, we don't need to handle freeing this ourselves
+        let writeConcern = mongoc_collection_get_write_concern(self._collection)
+        let wcObj = WriteConcern(writeConcern)
+        if wcObj.isDefault { return nil }
+        return wcObj
     }
 
     /// Initializes a new `MongoCollection` instance, not meant to be instantiated directly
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/Database.swift
```diff
@@ -96,10 +105,11 @@ public struct CreateCollectionOptions: Encodable {
         self.validationLevel = validationLevel
         self.validator = validator
         self.viewOn = viewOn
+        self.writeConcern = writeConcern
     }
 
-    // Encode everything except `readConcern`. We skip it because we don't actually
-    // send it with the initial command, we just set it on the collection after its creation.
+    // Encode everything except `readConcern` and 'writeConcern`. We skip them because we don't actually
+    // send them with the initial command, we just set them on the collection after its creation.
```
jmikola: I'm confused. Does the `createCollection()` helper return a MongoCollection object and _that_ is the only place these options are applied?

The [`create` command](https://docs.mongodb.com/manual/reference/command/create/) is a [command that writes](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#other-commands-that-write) and supports a `writeConcern` option on its own, so I would expect that option to be used for executing the command.

Could we work around this by removing the notion of soliciting read and write concern objects just to apply them to the return value? I'd argue in favor of removing the syntactic sugar to have `createCollection` return a MongoCollection. This is going to get more ambiguous if we need to also take a read preference down the line ([SWIFT-31](https://jira.mongodb.org/browse/SWIFT-31)).

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+        /// Specifies that a majoirty of nodes should acknowledge the write.
+        case majority
+
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.singleValueContainer()
+            if let wTag = try? container.decode(String.self) {
+                self = .tag(wTag)
+            } else {
+                let wNumber = try container.decode(Int32.self)
+                if wNumber == MONGOC_WRITE_CONCERN_W_MAJORITY {
```
jmikola: Is it worth handling other [numeric constants](https://github.com/mongodb/mongo-c-driver/blob/1.10.2/src/libmongoc/src/mongoc/mongoc-write-concern.h#L32) here?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+                return true
+            default:
+                return false
+            }
+        }
+    }
+
+    /// Indicates the `W` value for this `WriteConcern`.
+    public var w: W {
+        if let wTag = mongoc_write_concern_get_wtag(self._writeConcern) {
```
jmikola: Would it make sense to implement logic similar to `MONGOC_WRITE_CONCERN_W_MAJORITY` and instead check whether `mongoc_write_concern_get_w()` returns `MONGOC_WRITE_CONCERN_W_TAG`?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+        return mongoc_write_concern_get_journal(self._writeConcern)
+    }
+
+    /// If the write concern is not satisfied within this timeout (in milliseconds),
+    /// the operation will return an error. The value MUST be greater than or equal to 0.
+    public var wtimeoutMS: Int32? {
+        return mongoc_write_concern_get_wtimeout(self._writeConcern)
+    }
+
+    /// Indicates whether vbnm this is an acknowledged write concern.
```
jmikola: Is "vbnm" a typo?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+        return mongoc_write_concern_is_acknowledged(self._writeConcern)
+    }
+
+    /// Indicates whether this is the default write concern.
+    public var isDefault: Bool {
+        return mongoc_write_concern_is_default(self._writeConcern)
+    }
+
+    /// Indicates whether the combination of values set on this `WriteConcern` is valid.
+    public var isValid: Bool {
```
jmikola: Is there a need to expose this? PHPC doesn't have a use for it. I realize it's public, but AFAIK libmongoc uses it internally to sanity check a write concern before sending it in a server command. Are there plans to do so in Swift when you manually build a command?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+                mongoc_write_concern_set_wtag(self._writeConcern, wTag)
+            case .majority:
+                mongoc_write_concern_set_w(self._writeConcern, MONGOC_WRITE_CONCERN_W_MAJORITY)
+            }
+        }
+    }
+
+    /// Initializes a new `WriteConcern` by copying a `mongoc_write_concern_t`.
+    /// The caller is responsible for freeing the original `mongoc_write_concern_t`.
+    internal init(_ writeConcern: OpaquePointer?) {
```
jmikola: What uses this initializer, that the caller is responsible for freeing?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+        self._writeConcern = mongoc_write_concern_copy(writeConcern)
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case w, j, wtimeout
```
jmikola: PHPC also doesn't expose the deprecated `fsync` option in our API, but it's still possible for a WriteConcern to have `fsync` set via the libmongoc URI. Is that worth supporting?

If this BSON encoding is used to ferry the WriteConcern's `mongoc_write_concern_t` object into a `bson_t` to pass to a "with_opts" function in libmongoc, you may want to handle it. PHPC is able to use libmongoc's [`mongoc_write_concern_append()`](https://github.com/mongodb/mongo-c-driver/blob/1.10.2/src/libmongoc/src/mongoc/mongoc-write-concern.c#L476) function, which handles `fsync` for us through its internal BSON conversion.

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+    }
+
+    public required convenience init(from decoder: Decoder) throws {
```
jmikola: What does "required convenience" imply?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,166 @@ extension ReadConcern: Equatable {
...
+    }
+}
+
+/// An extension of `WriteConcern` to make it `CustomStringConvertible`.
+extension WriteConcern: CustomStringConvertible {
+    public var description: String {
+        if let encoded = try? BsonEncoder().encode(self).description {
+            return encoded
+        }
+        return ""
```
jmikola: Should an empty write concern be "{}", or is this return case for a `nil` value?

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -2,15 +2,49 @@
 import Nimble
 import XCTest
 
-import libmongoc
+extension WriteConcern {
+    /// Initialize a new ReadConcern from a Document.
```
jmikola: "new WriteConcern"

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -2,15 +2,49 @@
...
+    fileprivate convenience init(_ doc: Document) {
+        // can be stored under either "j" or "journal"
```
jmikola: Why is this the case? Is it for handling inconsistent spec tests?

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -2,15 +2,49 @@
...
+        }
+
+        var wToUse: W? = nil
+        if let wtag = doc["w"] as? String {
+            wToUse = wtag == "majority" ? .majority : .tag(wtag)
+        } else if let w = doc["w"] as? Int {
+            wToUse = .number(Int32(w))
+        }
+
+        // can be stored under either "wtimeout" or "wtimeoutMS"
```
jmikola: Is this also due to spec test inconsistencies?

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -170,27 +166,40 @@ public class WriteConcern: Codable {
...
+        switch number {
+        case MONGOC_WRITE_CONCERN_W_DEFAULT:
+            return nil
+        case MONGOC_WRITE_CONCERN_W_MAJORITY:
+            return .majority
+        case MONGOC_WRITE_CONCERN_W_TAG:
+            if let wTag = mongoc_write_concern_get_wtag(self._writeConcern) {
+                return .tag(String(cString: wTag))
+            }
+            fallthrough
```
kmahar: if for some reason this is `MONGOC_WRITE_CONCERN_W_TAG` but we don't get a `wtag` back, just fall back to returning `.number` with the number

Sources/MongoSwift/ReadWriteConcern.swift
kmahar: I opened [SWIFT-121](https://jira.mongodb.org/browse/SWIFT-121) about this

Sources/MongoSwift/ReadWriteConcern.swift
jmikola: Ah, I didn't consider that it might have been used for spec tests. PHPC relies on libmongoc for those, although we do have our own tests for errors thrown during construction of an invalid WriteConcern (e.g. out-of-range `w` value).

I noticed that `WriteConcern.init` (with the `journal`, `w`, and `wtimeoutMS` args) doesn't raise an error for invalid values. Would that make more sense?

Sources/MongoSwift/ReadWriteConcern.swift
jmikola: In [Error Handling: Converting Errors to Optional Values](https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html), does `try?` assignment catch and ignore any error thrown -- essentially converting it to a `nil` for assignment? If there is a real error, wouldn't you want to propagate that to the user?

I'm not sure that an error string in place of `""` would be preferable, as the user still has to examine that string and infer that an error occurred.

> if we are certain that the encoding will never fail given a successfully initialized WriteConcern, we could also just trigger a preconditionFailure here

This seems like a reasonable thing to do. If the WriteConcern is valid, we should be able to assert that it can be encoded successfully. Any error during an encode would be seriously unexpected IMO.

Sources/MongoSwift/Database.swift
kmahar: yup, that's what I'm proposing. `createCollection` takes in a `WriteConcern` still, but we use it for the command rather than setting it on the collection. and it returns the newly created collection with inherited RC/WC/RP, and if the user wants to specify something else, they need to do
```swift
_ = try db.createCollection("coll1")
let coll1 = try db.collection("coll1", options: CollectionOptions(readConcern: ..., writeConcern: ...))
```

this is kinda weird, but I'm not sure how else we'd do it without breaking all of our tests and also probably breaking stitch code.

jmikola: That works for me.

Sources/MongoSwift/ReadWriteConcern.swift
kmahar: I considered doing that, but then anytime a user creates a `WriteConcern` they need to use `try`:
```swift
let wc = try WriteConcern(w: 1)
```
which seems kind of annoying, and is asymmetric to how `ReadConcern`s are created. I seem to vaguely recall that in a past version of this PR I was originally doing this and @mbroadst wanted to avoid it. but that was quite a while ago, so I can't say for sure. 

checking validity at the time of encoding (rather than creation time) would prevent us from making the switch to a preconditionFailure in the `description` property though.... hmm.

Sources/MongoSwift/ReadWriteConcern.swift
kmahar: yes, basically, it ignores the error. which I did under the assumption that there would never actually be a relevant encoding error. however, that would not be true if we check validity at the time of encoding rather than time of initialization. 

if we *were* to throw in the `WriteConcern` initializer for an invalid combination of inputs, we could write this as follows:
```swift
public var description: String {
    do {
        return try BsonEncoder().encode(self).description
    } catch {
        preconditionFailure("Error encoding WriteConcern: \(error)")
    }
}
```

if we throw at encoding time, then we need to do something like
```swift
public var description: String {
    do {
        return try BsonEncoder().encode(self).description
    } catch MongoError.writeConcernError {
        // do whatever want to do here for an invalid WriteConcern
    } catch {
        // this is a very unexpected error since the WriteConcern was valid
        preconditionFailure("Error encoding WriteConcern: \(error)")
    }
}
```

kmahar: I'm not really sure what is better 🤔

Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -114,3 +113,175 @@ extension ReadConcern: Equatable {
...
 }
+
+/// A class to represent a MongoDB write concern.
+public class WriteConcern: Codable {
+
+    /// A pointer to a mongoc_write_concern_t
+    internal var _writeConcern: OpaquePointer?
+
+    /// An option to request acknowledgement that the write operation has propagated to specified mongod instances.
+    public enum W: Codable, Equatable {
```
jmikola: Ah, another question that came up. Why is this declared within WriteConcern when ReadConcernLevel is declared outside of ReadConcern?

kmahar: https://jira.mongodb.org/browse/SWIFT-123

Sources/MongoSwift/ReadWriteConcern.swift
mbroadst: My general feeling is that this is a value class and throwing a `try` in there feels unintuitive. We have a few options then:
 - only allow users to use predefined values, so there are no invalid values
 - use preconditions in the initializer, and hard fail the app when someone screws this up
 - throw from the initializer.

I care about this like 3/10, but not having a throwing initializer "felt" more natural to me

Sources/MongoSwift/ReadWriteConcern.swift
mbroadst: Let's not expose deprecated API unless someone explicitly asks for it

kmahar:
Ok, the initializer that takes in params now throws if `!self.isValid`. I have made `isValid` private now, as it would now be impossible for an invalid `WriteConcern` to exist. 
note that `deinit` gets called even if the `init` throws, so we don't need any separate cleanup code in that case.

Sources/MongoSwift/Collection.swift
```diff
@@ -919,46 +975,52 @@ public class MongoCollection<T: Codable> {
...
-        return try _dropIndexes(keys: model.keys)
+    public func dropIndex(_ model: IndexModel, options: DropIndexOptions? = nil) throws -> Document {
+        return try _dropIndexes(keys: model.keys, options: options)
     }
 
     /**
      * Drops all indexes in the collection.
+     * 
+     * - Parameters:
+    *   - writeConcern: An optional `WriteConcern` to use for the command
```
mbroadst: nit: spacing

kmahar: 👍

PR #73, opened 2018-06-11
SWIFT-72 hooks for libmongoc init and cleanup

jmikola:
https://jira.mongodb.org/browse/SWIFT-72

### DIFF
--- a/Sources/MongoSwift/MongoSwift.swift
+++ b/Sources/MongoSwift/MongoSwift.swift
@@ -1,4 +1,40 @@
+import libmongoc
+
 /// Returns a `String` indicating the version of the driver.
 public func version() -> String {
     return "testing"
+}
+
+/// A utility class for libmongoc initialization and cleanup.
+final public class MongoSwift {
+    final class MongocInitializer {
+        static let shared = MongocInitializer()
+
+        private init() {
+            mongoc_init()
+        }
+    }
+
+    /**
+     * Initializes libmongoc.
+     *
+     * This function should be called once at the start of the application
+     * before interacting with the driver.
+     */
+    public static func initialize() {
+        _ = MongocInitializer.shared
+    }
+
+    /**
+     * Cleans up libmongoc.
+     *
+     * This function should be called once at the end of the application. Users
+     * should not interact with the driver after calling this function.
+     */
+    public static func cleanup() {
+        /* Note: ideally, this would be called from MongocInitializer's deinit,
+         * but Swift does not currently handle deinitialization of singletons.
+         * See: https://bugs.swift.org/browse/SR-2500 */
+        mongoc_cleanup()
+    }
 }
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -1,5 +1,11 @@
 import XCTest
+import MongoSwift
 import MongoSwiftTests
+
+/* Ensure libmongoc is initialized. Since XCTMain never returns, we have no
+ * opportunity to cleanup libmongoc (either explicitly or with a deinit). This
+ * may appear as a memory leak. */
+MongoSwift.initialize()
 
 var tests = [XCTestCaseEntry]()
 tests += MongoSwiftTests.allTests()
--- a/Tests/MongoSwiftTests/TestSetup.swift
+++ b/Tests/MongoSwiftTests/TestSetup.swift
@@ -1,0 +1,13 @@
+import Foundation
+import MongoSwift
+
+/// This should be assigned to NSPrincipalClass in the project's plist file
+class TestSetup: NSObject {
+    override init() {
+        MongoSwift.initialize()
+    }
+
+    deinit {
+        MongoSwift.cleanup()
+    }
+}
--- a/Tests/MongoSwiftBenchmarks/BenchmarkSetup.swift
+++ b/Tests/MongoSwiftBenchmarks/BenchmarkSetup.swift
@@ -1,0 +1,13 @@
+import Foundation
+import MongoSwift
+
+/// This should be assigned to NSPrincipalClass in the project's plist file
+class BenchmarkSetup: NSObject {
+    override init() {
+        MongoSwift.initialize()
+    }
+
+    deinit {
+        MongoSwift.cleanup()
+    }
+}


Tests/MongoSwiftTests/TestSetup.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+import MongoSwift
+
+/// This should be assigned to NSPrincipalClass in the project's plist file
+class TestSetup: NSObject {
```
jmikola: @kmahar: I believe we'll want this class referenced from https://github.com/mongodb/mongo-swift-driver/commit/0bb9776fba22f9e5abbec43c4f23bf61b0a6be9c#diff-47062118b9b95971fa9ba0ca24e2270eR22 (or whatever is used to generate that).

For context, see the final section of https://stackoverflow.com/a/41895011/162228.

Sources/MongoSwift/MongoSwift.swift
```diff
@@ -1,4 +1,40 @@
...
 public func version() -> String {
     return "testing"
 }
+
+/// A utility class for libmongoc initialization and cleanup.
+final public class MongoSwift {
+    final class MongocInitializer {
+        static let shared = MongocInitializer()
+
+        private init() {
```
jmikola: A private initializer was advised in [this write-up](https://krakendev.io/blog/the-right-way-to-write-a-singleton). That said, I'm not sure if MongocInitializer is accessible from outside of the MongoSwift class, so maybe that isn't a concern?

Sources/MongoSwift/MongoSwift.swift
```diff
@@ -1,4 +1,40 @@
...
+
 /// Returns a `String` indicating the version of the driver.
 public func version() -> String {
     return "testing"
 }
+
+/// A utility class for libmongoc initialization and cleanup.
+final public class MongoSwift {
+    final class MongocInitializer {
+        static let shared = MongocInitializer()
```
jmikola: This [Stack Overflow answer](https://stackoverflow.com/a/24073016/162228) discusses that this is a thread-safe approach for declaring and initializing a singleton.

Sources/MongoSwift/MongoSwift.swift
```diff
@@ -1,4 +1,40 @@
...
+    }
+
+    /**
+     * Initializes libmongoc.
+     *
+     * This function should be called once at the start of the application
+     * before interacting with the driver.
+     */
+    public static func initialize() {
+        _ = MongocInitializer.shared
```
jmikola: Discarding the result here as we don't actually need a reference to it. Accessing it is enough to trigger lazy initialization of the singleton (i.e. a call to `mongoc_init()`).

Tests/LinuxMain.swift
```diff
@@ -1,6 +1,12 @@
 import XCTest
+import MongoSwift
 import MongoSwiftTests
 
+/* Ensure libmongoc is initialized. Since XCTMain never returns, we have no
+ * opportunity to cleanup libmongoc (either explicitly or with a deinit). This
+ * may appear as a memory leak. */
+MongoSwift.initialize()
```
jmikola: Since Swift won't actually deinit a singleton (see: [SR-2500](https://bugs.swift.org/browse/SR-2500)), we need to explicitly call `mongoc_cleanup()` from a separate context. Two options as I see it:

 * Ask users to call an explicit function, as I'm doing here.
 * Make MongoSwift an instantiable class, which constructs the singleton (i.e. calls `mongoc_init()`) in `MongoSwift.init` and calls `mongoc_cleanup()` from `MongoSwift.deinit` (note: the singleton is never deinitialized).

I'm not a fan of the "instantiable class" approach as it puts the onus on the user to keep a MongoSwift object alive for the duration of their application. Failure to do so could result in a premature call to `mongoc_cleanup()`, which could cause curious failures and other headaches for users.

Sources/MongoSwift/MongoSwift.swift
kmahar: 👍

Tests/MongoSwiftTests/TestSetup.swift
kmahar: So this seems reasonable for ensuring things happen correctly when running the tests in Xcode. And we should be able to edit the .plist files using the ruby gem I mentioned earlier. 

However, now that I think about it, I am not sure that the .plist file is actually considered at all when you run the tests from the command line, as you can run tests there without having ever generated the .xcodeproj. 

So we may need an alternate approach in that case...

Sources/MongoSwift/MongoSwift.swift
jmikola: > so this sounds like multiple calls to `MongoSwift.initialize()` would be safe to make, correct?

Correct. I wouldn't object calling this from `MongoClient.init` as we can possibly avoid the asymmetry altogether if we simply advise users to call `MongoSwift.cleanup()` but allow them to make the choice to do so for themselves.

The side effect of never calling `mongoc_cleanup()` is a potential memory leak if you were tracing the application with Valgrind or some equivalent. If the process is on its way out, one could argue that there's really no need to cleanup. This jives with what I noted in `LinuxMain.swift`, where we really don't have a choice about not cleaning up since `XCTMain()` never returns.

Tests/LinuxMain.swift
```diff
@@ -1,6 +1,12 @@
 import XCTest
+import MongoSwift
 import MongoSwiftTests
 
+/* Ensure libmongoc is initialized. Since XCTMain never returns, we have no
+ * opportunity to cleanup libmongoc (either explicitly or with a deinit). This
```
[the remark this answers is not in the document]
jmikola: It doesn't seem to work with a manual `exit()`, which is presumably what `XCTMain()` is doing.

Consider:

```
defer { print("BYE"); }
print("OK");
```

This prints "OK" and "BYE"  as you might expect. If we add an `exit(0);` on the third line, only "OK" is printed.

Tests/MongoSwiftTests/TestSetup.swift
jmikola: Does this mean that there are effectively three variations for running tests?

 1. Xcode in Darwin, which uses the plist file.
 2. `swift test` in Darwin, which automatically collects tests.
 3. `swift test` in Linux, which requires `LinuxMain.swift` to manually collect tests.

While this approach theoretically ensures that cleanup happens via `deinit`, I'll note my other argument that cleanup really isn't essential since any "leaked" memory will be freed when the process exits. If we go the route of simply calling `MongoSwift.initialize()` from `MongoClient` (as you suggested [here](https://github.com/mongodb/mongo-swift-driver/pull/73#discussion_r194542780)) then we can likely avoid any test setup for all three variants.

kmahar: yes, those are the three. given all these testing variations, I think I like the calling `MongoSwift.initialize()` from `MongoClient` approach best, let's see what @mbroadst thinks.

jmikola: @mbroadst: To clarify, we're not suggesting calling `mongoc_init()` directly from `MongoClient.init` and relying purely on libmongoc to ensure at-most-once behavior. I'd still plan to use the Swift singleton and leave the original `MongoSwift.initialize()` and `MongoSwift.cleanup()` functions in place; however, `MongoClient.init` could call `MongoSwift.initialize()` each time in case a user forgets (or doesn't want to) call it themselves. I think we could still suggest that user's manually call both methods at the start/end of their application, respectively.

mbroadst: After much contemplation I think the only way forward is to make initialization explicit. It's very tempting to bury the initialization call in the initializer for `MongoClient`, however it really ties our hands for future versions of the driver and language (perhaps Swift will support deinitializing singletons in the future, for instance). The only way to provide a consistent experience here is to provide explicit initialize/cleanup methods, and document them appropriately.

Sources/MongoSwift/MongoSwift.swift
mbroadst: yeah I think `private` is indeed the way to go here

PR #74, opened 2018-06-14
Move `ReadConcernLevel` inside of `ReadConcern`

kmahar:
Quick fix for some inconsistency pointed out by @jmikola.

### DIFF
--- a/Sources/MongoSwift/ReadWriteConcern.swift
+++ b/Sources/MongoSwift/ReadWriteConcern.swift
@@ -1,21 +1,21 @@
 import libmongoc
 
-/// An enumeration of possible ReadConcern levels.
-public enum ReadConcernLevel: String {
-    /// See https://docs.mongodb.com/manual/reference/read-concern-local/
-    case local
-    /// See https://docs.mongodb.com/manual/reference/read-concern-available/
-    case available
-    /// See https://docs.mongodb.com/manual/reference/read-concern-majority/
-    case majority
-    /// See https://docs.mongodb.com/manual/reference/read-concern-linearizable/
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/ReadWriteConcern.swift
```diff
@@ -33,8 +33,8 @@ public class ReadConcern: Codable {
         return mongoc_read_concern_is_default(self._readConcern)
     }
 
-    /// Initialize a new `ReadConcern` from a `ReadConcernLevel`.
-    public convenience init(_ level: ReadConcernLevel) {
+    /// Initialize a new `ReadConcern` from a `ReadConcern.Level`.
```
jmikola: If a user imports `MongoSwift`, I assume `ReadConcern` is then accessible and they will need to access `Level` as `ReadConcern.Level`, correct?

Did we have any tests that needed to be updated, or are those all using the `init()` from a raw string value?

kmahar: Yes, that's correct. If they ever needed to explicitly use the type, they would have to use the full `ReadConcern.Level`.

However, for the vast majority of times users are interacting with this type (passing it into a `ReadConcern` initializer), the compiler just infers the type and its name is not needed. So both before and after, 
```swift
```
works, and this is how we're interacting with this type throughout the tests. 

The only time you'd need the type name would be if you were saving a level in a variable somewhere - for example
```swift
let level: ReadConcernLevel = .majority
```
would now be
```swift
let level: ReadConcern.Level = .majority
```

We don't do that anywhere though, so no changes needed.

PR #77, opened 2018-06-18
SWIFT-104 Support encoding and decoding untyped BsonValues

kmahar:
cc @adamchel 

This adds a `Codable` `AnyBsonValue` type that wraps another type conforming to `BsonValue`. This allows users to encode/decode things that they know are _some_ `BsonValue` type, but aren't sure which.

This will be useful to us in making some of our results structs that contain `BsonValue`s (for example, [`InsertOneResult`](https://github.com/mongodb/mongo-swift-driver/blob/a331a8c8dd1748d913dda133f21a4a9ece16a62d/Sources/MongoSwift/Collection.swift#L301)) `Decodable`. 

To implement this, I ended up repurposing the code for `decoding` `Document`s to be more general, and I was also able to cut out a lot of it (i.e. the variations for different container types) in the process. 

There are some limitations if this is used with anything besides our own encoder/decoder, which I have tried to document where appropriate.


Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -319,27 +321,6 @@ final class CodecTests: XCTestCase {
     /// Test decoding/encoding to all possible BSON types
     func testBsonValues() throws {
 
-        // decode from fully de-structured 
-        let doc1: Document = [
```
kmahar: this is removed because I changed the `Document.init(from:)` method to always short-circuit if the decoder being used is a `BsonDecoder`. Basically, this is the way that a non-BSON encoder would have encoded a `Document`.

Since this seems like an unlikely use case, I think we should opt for the vast performance improvement of short-circuiting instead.

Sources/MongoSwift/BSON/BsonEncoder.swift
```diff
@@ -507,16 +509,24 @@ private class MutableArray: BsonValue {
 
     var count: Int { return array.count }
 
+    func insert(_ value: BsonValue?, at index: Int) {
```
jmikola: I realize the method isn't new (just relocated in the diff) and the class isn't implementing a protocol, but is this a method related to some Swift array protocol?

Sources/MongoSwift/BSON/BsonEncoder.swift
```diff
@@ -507,16 +509,24 @@ private class MutableArray: BsonValue {
...
+        self.array.insert(value, at: index)
+    }
+
     func encode(to data: UnsafeMutablePointer<bson_t>, forKey key: String) throws {
         try self.array.encode(to: data, forKey: key)
     }
 
+    /// methods required by the BsonValue protocol that we don't actually need/use. MutableArray
```
jmikola: Should these methods be annotated with `required`, or is that only used for initializers?

Sources/MongoSwift/BSON/Document+Codable.swift
```diff
@@ -1,219 +1,56 @@
 import Foundation
 
+/// An extension of `Document` to implement the `Codable` protocol.
 extension Document: Codable {
 	public func encode(to encoder: Encoder) throws {
+		// if we're using a `BsonEncoder`, we can just short-circuit
```
jmikola: FYI: This entire file appears to be indented with tab characters instead of four spaces.

Sources/MongoSwift/BSON/Document+Codable.swift
```diff
@@ -1,219 +1,56 @@
...
-			try container.encode(val)
-		case nil:
-			 try container.encodeNil()
-		case let val as [BsonValue]:
-			var nested = container.nestedUnkeyedContainer()
-			for v in val {
-				try Document.recursivelyEncodeUnkeyed(v, to: &nested)
+			let key = _BsonKey(stringValue: k)!
+			if let val = v {
+				try container.encode(AnyBsonValue(val), forKey: key)
```
jmikola: Would it make sense to allow `AnyBsonValue` to be constructed with `nil`? That would simply this conditional.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -509,4 +486,208 @@ final class CodecTests: XCTestCase {
...
+
+    struct AnyBsonStruct: Codable {
+        let x: AnyBsonValue
+
+        init(_ x: BsonValue) {
+            self.x = AnyBsonValue(x)
+        }
+    }
+
+    // test encoding/decoding AnyBsonValues with BsonEncoder and Decoder
```
jmikola: I realize it's not an AnyBsonValue, but is there a test case for `nil` (and/or do we need one)?

Tests/MongoSwiftTests/CodecTests.swift
jmikola: I assume this refers to https://github.com/mongodb/mongo-swift-driver/pull/77/files#diff-cd1c06a2751f196995f7a1679f3b0803R15, but why does that require this test case to be removed?

Is the short-circuit behavior being tested elsewhere?

mbroadst: I think this is fine, how often will people be constructing documents this way?

Sources/MongoSwift/BSON/BsonEncoder.swift
```diff
@@ -36,13 +36,16 @@ public class BsonEncoder {
                     debugDescription: "Top-level \(T.self) did not encode any values."))
         }
 
-        guard let dict = topLevel as? MutableDictionary else {
+        switch topLevel {
+        case let dict as MutableDictionary:
+            return dict.asDocument()
+        case let doc as Document:
+            return doc
```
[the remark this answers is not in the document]
mbroadst: I think you should check this as a special case before going through the call to `box`, this is definitely a special case and there's no need to go through all that other logic if we know immediately what we want to do here.

Sources/MongoSwift/BSON/Document+Codable.swift
mbroadst: :+1: to making it feel like a `BsonValue`

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -486,10 +463,10 @@ final class CodecTests: XCTestCase {
 
         let expected: Document = [
             "name": "Durian",
-            "points": 600.0,
```
mbroadst: hm, this looks suspiciously like we don't support floating points anymore? 🤔

kmahar: so this is just with a decoder other than our own, and it's because of the behavior I described in the comment [here](https://github.com/mongodb/mongo-swift-driver/pull/77/files#diff-cd1c06a2751f196995f7a1679f3b0803R48) - basically, since 600.0 can be accurately represented as an `Int`, in the JSONDecoder ends up getting decoded as one regardless of whether the `.0` is there.

Sources/MongoSwift/BSON/BsonEncoder.swift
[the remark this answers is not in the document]
kmahar: 👍 done

Tests/MongoSwiftTests/CodecTests.swift
kmahar: yeah, I'm not really sure what the use case would be

Sources/MongoSwift/BSON/AnyBsonValue.swift
```diff
@@ -0,0 +1,100 @@
...
+    /// 1) This method does *not* support initializing an `AnyBsonValue` wrapping
+    /// a `Date`. This is because, in non-BSON formats, `Date`s are encoded
+    /// as other types such as `Double` or `String`. We have no way of knowing 
+    /// which type is the intended one when decoding to a `Document`, as `Document`s 
+    /// can contain any `BsonValue` type, so for simplicity we always go with a 
+    /// `Double` or a `String` over a `Date`.
+    /// 2) Numeric values will be attempted to be decoded in the following
+    /// order of types: `Int`, `Int32`, `Int64`, `Double`. The first one
+    /// that can successfully represent the value with no loss of precision will 
+    /// be used.
```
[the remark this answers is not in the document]
jmikola: Noting the edge cases, will this require some Extended JSON spec tests to be skipped?

Tests/MongoSwiftTests/CodecTests.swift
jmikola: Would it make sense to test an actual Double value that can't be represented as an integer in this case (e.g. 600.5), or is that not terribly important for this test?

Sources/MongoSwift/BSON/AnyBsonValue.swift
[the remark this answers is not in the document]
kmahar: we don't use the `JSONDecoder` for extended JSON parsing - `Document` has a `fromJSON` initializer that call's libbson's `bson_new_from_json` method

Tests/MongoSwiftTests/CodecTests.swift
kmahar: added a double value

kmahar:
@jmikola @mbroadst I think you both already LGTMed this but lmk if you want to discuss my point above about arrays and codable any more before I merge this ?

jmikola:
Satisfied with what we discussed in person. I believe we concluded there'd be a tracking ticket for future improvement on the array codable stuff, but this PR is LGTM.

kmahar:
I opened [SWIFT-141](https://jira.mongodb.org/browse/SWIFT-141)

PR #78, opened 2018-06-20
SWIFT-97 Append MongoSwift info to mongoc handshake

kmahar:
I tested the behavior of this by calling `MongoSwift.initialize()` and then opening a `MongoClient`, and then looking at the mongod's log.

Before, the output when opening a new client was:
`{ driver: { name: "mongoc", version: "1.9.3" }, os: { type: "Darwin", name: "macOS", version: "17.4.0", architecture: "x86_64" }, platform: "cfg=0xd69a65 posix=200112 stdc=201112 CC=clang 9.0.0 (clang-900.0.39.2) CFLAGS="" LDFLAGS=""" }`

Now, it is:
`{ driver: { name: "mongoc / MongoSwift", version: "1.9.3 / 0.0.2" }, os: { type: "Darwin", name: "macOS", version: "17.4.0", architecture: "x86_64" }, platform: "cfg=0xd69a65 posix=200112 stdc=201112 CC=clang 9.0.0 (clang-900.0.39.2) CFLAGS="" LDFLAGS=""" }`


Sources/MongoSwift/MongoSwift.swift
```diff
@@ -1,17 +1,18 @@
...
+    public static func version() -> String {
+        return "0.0.2"
+    }
+
     final class MongocInitializer {
         static let shared = MongocInitializer()
 
         private init() {
             mongoc_init()
+            mongoc_handshake_data_append("MongoSwift", MongoSwift.version(), nil)
```
kmahar: the third argument here is for any platform info.

Sources/MongoSwift/MongoSwift.swift
```diff
@@ -1,17 +1,18 @@
...
-/// Returns a `String` indicating the version of the driver.
-public func version() -> String {
-    return "testing"
-}
-
 /// A utility class for libmongoc initialization and cleanup.
 final public class MongoSwift {
+    /// Returns a `String` indicating the version of the driver.
+    public static func version() -> String {
+        return "0.0.2"
```
mbroadst: hm this looks very easy to forget to update 😄 Perhaps we should come up with some sort of release script that will ensure this is always correct

mbroadst: Also, could we potentially just make this a floating global constant? I originally created this function as a total skeleton to just have _something_ in the file

kmahar: yeah, we could write something checking that it matches the Cocoapods file I guess? 

I think it's kind of nice to namespace it within the `MongoSwift` class, since we would probably name it something like "MongoSwiftVersionString" anyway. I don't have very strong feelings though.... if we do end up keeping it on the class though, I realized it should be a computed property rather than a function

PR #79, opened 2018-06-21
Rename files and split up MongoCollection into extensions

kmahar:
This diff is not very readable, but here is what I did:

renamed
- `Client.swift` -> `MongoClient.swift`
- `Collection.swift` -> `MongoCollection.swift`
- `Database.swift` -> `MongoDatabase.swift`
- `Cursor.swift` -> `MongoCursor.swift`

Then I split up `MongoCollection` code into a base definition in `MongoCollection.swift`, with two extensions , one for CRUD (`MongoCollection+Crud.swift`) and one for index management (`MongoCollection+Indexes.swift`). 
I also moved all the CRUD results and options into their own file, `MongoCollection+CrudTypes.swift`. let me know if you can think of a better name for that file...


jmikola:
If we're going the route of ensuring consistency between class names and files, I'd love to see read and write concerns split up as well (not specifically in this PR).

> I also moved all the CRUD results and options into their own file, MongoCollection+CrudTypes.swift

I assume this was done as the CRUD extension was plenty lone with the methods alone. Would it make sense to instead segregate by components within CRUD? For instance: CRUD+FindAndModify with just those related methods and options in their own file. I do think it's helpful to keep code options and the methods that use them near each other if possible (speaking as someone that often reads code for documentation).

kmahar:
I am fine with splitting up RC/WC as well. 

> I assume this was done as the CRUD extension was plenty lone with the methods alone. Would it make sense to instead segregate by components within CRUD? For instance: CRUD+FindAndModify with just those related methods and options in their own file.

Yes, that was why I did it that way, but your proposal sounds reasonable too, I see the value in keeping the options near the methods. will rearrange

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -0,0 +1,307 @@
...
+        var error = bson_error_t()
+
+        if !mongoc_collection_write_command_with_opts(self._collection, command.data, opts?.data, nil, &error) {
+            throw MongoError.commandError(message: toErrorString(error))
+        }
+
+        return forModels.map { $0.options?.name ?? $0.defaultName }
+    }
+
+     /**
```
jmikola: Indentation looks off here.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -0,0 +1,307 @@
...
+    @discardableResult
+    public func dropIndex(_ model: IndexModel, options: DropIndexOptions? = nil) throws -> Document {
+        return try _dropIndexes(keys: model.keys, options: options)
+    }
+
+    /**
+     * Drops all indexes in the collection.
+     * 
+     * - Parameters:
+    *    - writeConcern: An optional `WriteConcern` to use for the command
```
jmikola: Indentation is off here.

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -0,0 +1,372 @@
+import libmongoc
+
+/// An extension of `MongoCollection` encapsulating read operations.
+extension MongoCollection {
+	/**
```
jmikola: This file looks to be using tab indents instead of four spaces used elsewhere.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -0,0 +1,349 @@
+import libmongoc
+
+/// An extension of `MongoCollection` encapsulating write operations.
+extension MongoCollection {
+	/**
```
jmikola: This file is also using tabs.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -0,0 +1,349 @@
...
+		var error = bson_error_t()
+		if !mongoc_collection_delete_many(
+			self._collection, filter.data, opts?.data, reply.data, &error) {
+			throw MongoError.commandError(message: toErrorString(error))
+		}
+		return DeleteResult(from: reply)
+	}
+}
+
+/// Write command options structs
```
jmikola: If `///` is used for documentation, what happens to a floating comment like this? Does it get attached to the subsequent element (`InsertOneOptions` in this case)?

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -0,0 +1,349 @@
...
+	public let writeConcern: WriteConcern?
+
+	 /// Convenience initializer allowing collation to be omitted or optional
+	public init(collation: Document? = nil, writeConcern: WriteConcern? = nil) {
+		self.collation = collation
+		self.writeConcern = writeConcern
+	}
+}
+
+/// Write command results structs
```
jmikola: Similar question with this comment.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -0,0 +1,64 @@
...
+    public var name: String {
+        return String(cString: mongoc_collection_get_name(self._collection))
+    }
+
+    /// The `ReadConcern` set on this collection, or `nil` if one is not set.
+    public var readConcern: ReadConcern? {
+        // per libmongoc docs, we don't need to handle freeing this ourselves
+        let readConcern = mongoc_collection_get_read_concern(self._collection)
+        let rcObj = ReadConcern(from: readConcern)
+        if rcObj.isDefault { return nil }
```
jmikola: Would it make sense to use a single `return` with a ternary for this kind of thing? Totally unrelated to the PR -- just a general coding style question.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -0,0 +1,64 @@
...
+    /// Initializes a new `MongoCollection` instance, not meant to be instantiated directly
+    internal init(fromCollection: OpaquePointer, withClient: MongoClient) {
+        self._collection = fromCollection
+        self._client = withClient
+    }
+
+    /// Deinitializes a `MongoCollection`, cleaning up the internal `mongoc_collection_t`
+    deinit {
+        self._client = nil
+        guard let collection = self._collection else {
```
jmikola: Unrelated to the PR, but a general question: what are the circumstances where `deinit` would be invoked but `self._collection` is already `nil`?

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -403,7 +402,7 @@ final class ReadWriteConcernTests: XCTestCase {
                         expect(try encoder.encode(rc)).to(equal(expected))
                     }
                 } else if let wcToUse = test["writeConcern"] as? Document {
-                    
+
```
jmikola: Should this line just be deleted?

PR #80, opened 2018-06-21
Move BsonSubtype into Binary

kmahar:


Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -187,6 +169,24 @@ public struct Binary: BsonValue, Equatable, Codable {
     /// The binary subtype for this data.
     public let subtype: BsonSubtype
 
+    /// Subtypes for BSON Binary values.
+    public enum BsonSubtype: Int, Codable {
```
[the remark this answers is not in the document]
kmahar: 👍

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -185,16 +167,34 @@ public struct Binary: BsonValue, Equatable, Codable {
...
+        /// Binary (old)
+        binaryDeprecated,
+        /// UUID (old)
+        uuidDeprecated,
+        /// UUID
+        uuid,
+        /// MD5
+        md5,
+        /// User defined
+        user
```
jmikola: "user" may be a bit ambiguous. "userDefined" may be clearer.

Separate note: although http://bsonspec.org/spec.html does define a "user defined" type as 0x80, technically everything in the range of 0x80 to 0xFF is reserved for user-defined usage.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -205,7 +205,7 @@ public struct Binary: BsonValue, Equatable, Codable {
     /// Initializes a `Binary` instance from a `Data` object and a `UInt32` subtype.
     internal init(data: Data, subtype: UInt32) {
         self.data = data
-        self.subtype = BsonSubtype(rawValue: Int(subtype))!
+        self.subtype = Subtype(rawValue: Int(subtype))!
```
jmikola: This may be outside the scope of this PR, but is it possible to validate the sub-type as an unsigned 8-bit integer? See https://jira.mongodb.org/browse/PHPC-593.

Also, does constructing a Subtype enum here cause problems if users use a value other than one of the pre-defined constants? We should allow all values within the range of 0-255 (i.e. 0x00-0xFF).

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -185,16 +167,34 @@ public struct Binary: BsonValue, Equatable, Codable {
...
+        uuid,
+        /// MD5
+        md5,
+        /// User defined
+        user
+    }
 
     /// Initializes a Binary instance of the specified subtype using provided `Data`.
-    public init(data: Data, subtype: BsonSubtype) {
+    public init(data: Data, subtype: Subtype) {
```
jmikola: Also outside the scope of this PR, but you may want to validate that "data" is 16-bytes if either of the UUID sub-types is used. Some third-party tooling depends on that even though the server does not care. IMO an exception is helpful to warn users about misusing these sub-types. For instance, some were storing the string representation of a UUID (hexadecimals and dashes) instead of an actual binary value with the UUID sub-type, which is technically incorrect.

See https://jira.mongodb.org/browse/PHPC-895 and the linked GitHub issue for context.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -185,27 +165,75 @@ public struct Binary: BsonValue, Equatable, Codable {
...
+        /// Binary (old)
+        binaryDeprecated,
+        /// UUID (old)
+        uuidDeprecated,
+        /// UUID (RFC 4122)
+        uuid,
+        /// MD5
+        md5,
+        /// User defined
+        userDefined(UInt8)
```
jmikola: Will this be a catch-all for anything outside of the range of zero (`generic`) to five (`md5`)? The spec says that all values in the range of 0x80 to 0xFF are user-defined, but the actual `userDefined` constant is a single 0x80 value.

We should also ensure that _any_ unsigned 8-bit value is accepted as a sub-type. To my knowledge, no driver restricts the sub-type byte to 0x00-0x05 and 0x80-0xFF. While 0x06-0x7F are effectively undefined in the spec, there may very well be BSON data out there that uses those values.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -185,27 +165,75 @@ public struct Binary: BsonValue, Equatable, Codable {
...
+            case 0: self = .generic
+            case 1: self = .function
+            case 2: self = .binaryDeprecated
+            case 3: self = .uuidDeprecated
+            case 4: self = .uuid
+            case 5: self = .md5
+            case 0x80...0xFF:
+                self = .userDefined(rawValue)
+            default:
+                return nil
```
jmikola: Per my earlier comment, I think we still want to accept any unsigned 8-bit value and only consider 0x80 to be `userDefined` (as the constant is defined in other drivers).

Sources/MongoSwift/BSON/BsonValue.swift
kmahar: ah, ok, I misunderstood the spec. so I think we could express this in a couple of different ways.

option 1:
```swift
enum Subtype: UInt8 {
    case generic = 0,
    function,
    binaryDeprecated,
    uuidDeprecated,
    md5,
    userDefined = 0x80
}
```

and then in `Binary` we just store a `UInt8` and allow the user to initialize with either a `Subtype` or some custom `UInt8`. 

option 2:
```swift
enum Subtype: RawRepresentable, Codable {
    case generic = 0,
    function,
    binaryDeprecated,
    uuidDeprecated,
    md5,
    userDefined = 0x80,
    custom(UInt8)
}
```

So basically then we would continue storing a `Subtype` on `Binary`, and allow initialization from either a `Subtype` or a `UInt8`, lumping together all custom values under `.custom` (could be named something else.) 

I don't really have a preference between those two. what do you think?

mbroadst: Let's go with option 1 above

jmikola: The first option (define enums for documented constants but allow any unsigned 8-bit integer) is consistent with other drivers I've looked at. I vote for that.

kmahar:
Ok, I have switched to the proposed option 1. I also refactored the initializers a bit so we now allow any combo of (base64 or Data) + (Subtype or UInt8).

kmahar:
@mbroadst any other thoughts on this or good to go?

PR #83, opened 2018-07-02
SWIFT-137 Sketch remainder of CRUD API

kmahar:
This adds unimplemented methods for bulkWrite, find and modify ops, and the new count API, plus associated structs.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -1,0 +1,134 @@
+import libmongoc
+
+/// An extension of `MongoCollection` encapsulating bulk write operations.
+extension MongoCollection {
+    /**
+     * Execute multiple write operations.
+     *
+     * - Parameters:
+     *   - requests: a `[WriteModel]` containing the writes to perform
+     *   - options: optional `BulkWriteOptions` to use while executing the operation
+     *
+     * - Returns: a `BulkWriteResult`, or `nil` if the write concern is unacknowledged.
+     *
+     * - Throws:
+     *   - `MongoError.invalidArgument` if `requests` is empty
+     *   - `MongoError.bulkWriteError` if any error occurs while performing the writes
+     */
+    @discardableResult
+    public func bulkWrite(_ requests: [WriteModel], options: BulkWriteOptions? = nil) throws -> BulkWriteResult? {
+        throw MongoError.commandError(message: "Unimplemented command")
+    }
+}
+
+/// Options to use when performing a bulk write operation on a `MongoCollection`.
+public struct BulkWriteOptions: Encodable {
+    /// If `true` (the default), when an insert fails, return without performing the remaining writes.
+    /// If `false`, when a write fails, continue with the remaining writes, if any.
+    public var ordered: Bool = true
+
+    /// If `true`, allows the write to opt-out of document level validation.
+    public let bypassDocumentValidation: Bool?
+}
+
+/// A protocol indicating write types that can be batched together using `MongoCollection.bulkWrite`.
+public protocol WriteModel: Encodable {}
+
+/// A model for an `insertOne` command.
+public struct InsertOneModel: WriteModel {
+    /// The `Document` to insert.
+    public let document: Document
+}
+
+/// A model for a `deleteOne` command.
+public struct DeleteOneModel: WriteModel {
+    /// A `Document` representing the match criteria.
+    public let filter: Document
+
+    /// Specifies a collation to use.
+    public let collation: Document?
+}
+
+/// A model for a `deleteMany` command.
+public struct DeleteManyModel: WriteModel {
+    /// A `Document` representing the match criteria.
+    public let filter: Document
+
+    /// Specifies a collation to use.
+    public let collation: Document?
+}
+
+/// A model for a `replaceOne` command.
+public struct ReplaceOneModel: WriteModel {
+    /// A `Document` representing the match criteria.
+    public let filter: Document
+
+    /// The `Document` with which to replace the matched document.
+    public let replacement: Document
+
+    /// Specifies a collation to use.
+    public let collation: Document?
+
+    /// When `true`, creates a new document if no document matches the query.
+    public let upsert: Bool?
+}
+
+/// A model for an `updateOne` command.
+public struct UpdateOneModel: WriteModel {
+    /// A `Document` representing the match criteria.
+    public let filter: Document
+
+    /// A `Document` containing update operators.
+    public let update: Document
+
+    /// A set of filters specifying to which array elements an update should apply.
+    public let arrayFilters: [Document]?
+
+    /// Specifies a collation to use.
+    public let collation: Document?
+
+    /// When `true`, creates a new document if no document matches the query.
+    public let upsert: Bool?
+}
+
+/// A model for an `updateMany` command.
+public struct UpdateManyModel: WriteModel {
+    /// A `Document` representing the match criteria.
+    public let filter: Document
+
+    /// A `Document` containing update operators.
+    public let update: Document
+
+    /// A set of filters specifying to which array elements an update should apply.
+    public let arrayFilters: [Document]?
+
+    /// Specifies a collation to use.
+    public let collation: Document?
+
+    /// When `true`, creates a new document if no document matches the query.
+    public let upsert: Bool?
+}
+
+/// The result of a bulk write operation on a `MongoCollection`.
+public struct BulkWriteResult: Decodable {
+    /// Number of documents inserted.
+    public let insertedCount: Int
+
+    /// Map of the index of the operation to the id of the inserted document.
+    public let insertedIds: [Int: AnyBsonValue]
+
+    /// Number of documents matched for update.
+    public let matchedCount: Int
+
+    /// Number of documents modified.
+    public let modifiedCount: Int
+
+    /// Number of documents deleted.
+    public let deletedCount: Int
+
+    /// Number of documents upserted.
+    public let upsertedCount: Int
+
+    /// Map of the index of the operation to the id of the upserted document.
+    public let upsertedIds: [Int: AnyBsonValue]
+}
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -1,0 +1,150 @@
+import libmongoc
+
+/// An extension of `MongoCollection` encapsulating find and modify operations.
+extension MongoCollection {
+    /**
+     * Finds a single document and deletes it, returning the original.
+     *
+     * - Parameters:
+     *   - filter: `Document` representing the match criteria
+     *   - options: Optional `FindOneAndDeleteOptions` to use when executing the command
+     *
+     * - Returns: The deleted document, represented as a `CollectionType`, or `nil` if no document was deleted.
+     * - Throws: 
+     *   - `MongoError.commandError` if there are any errors executing the command.
+     *   - A `DecodingError` if the deleted document cannot be decoded to a `CollectionType` value
+     */
+    @discardableResult
+    public func findOneAndDelete(_ filter: Document, options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
+        throw MongoError.commandError(message: "Unimplemented command")
+    }
+
+    /**
+     * Finds a single document and replaces it, returning either the original or the replaced document.
+     * 
+     * - Parameters:
+     *   - filter: `Document` representing the match criteria
+     *   - replacement: a `CollectionType` to replace the found document
+     *   - options: Optional `FindOneAndReplaceOptions` to use when executing the command
+     *
+     * - Returns: A `CollectionType`, representing either the original document or its replacement,
+     *      depending on selected options, or `nil` if there was no match.
+     * - Throws: 
+     *   - `MongoError.commandError` if there are any errors executing the command.
+     *   - An `EncodingError` if `replacement` cannot be encoded to a `Document`
+     *   - A `DecodingError` if the replaced document cannot be decoded to a `CollectionType` value
+     */
+    @discardableResult
+    public func findOneAndReplace(filter: Document, replacement: CollectionType,
+                                  options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
+        throw MongoError.commandError(message: "Unimplemented command")
+    }
+
+    /**
+     * Finds a single document and updates it, returning either the original or the updated document.
+     * 
+     * - Parameters:
+     *   - filter: `Document` representing the match criteria
+     *   - update: a `Document` containing updates to apply
+     *   - options: Optional `FindOneAndUpdateOptions` to use when executing the command
+     *
+     * - Returns: A `CollectionType` representing either the original or updated document,
+     *      depending on selected options, or `nil` if there was no match.
+     * - Throws: 
+     *   - `MongoError.commandError` if there are any errors executing the command.
+     *   - A `DecodingError` if the updated document cannot be decoded to a `CollectionType` value
+     */
+    @discardableResult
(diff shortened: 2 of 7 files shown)


Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -66,6 +67,31 @@ extension MongoCollection {
...
+     * Counts the number of documents in this collection matching the provided filter.
+     *
+     * - Parameters:
+     *   - filter: a `Document`, the filter that documents must match in order to be counted
+     *   - options: Optional `CountOptions` to use when executing the command
+     *
+     * - Returns: The count of the documents that matched the filter
+     */
+    public func countDocuments(_ filter: Document = [:], options: CountOptions? = nil) throws -> Int {
+        throw MongoError.commandError(message: "Unimplemented command")
```
jmikola: Should this have a todo linked to a JIRA ticket? Likewise for `estimatedDocumentCount` below.

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -66,6 +67,31 @@ extension MongoCollection {
...
+    /**
+     * Counts the number of documents in this collection matching the provided filter.
+     *
+     * - Parameters:
+     *   - filter: a `Document`, the filter that documents must match in order to be counted
+     *   - options: Optional `CountOptions` to use when executing the command
+     *
+     * - Returns: The count of the documents that matched the filter
+     */
+    public func countDocuments(_ filter: Document = [:], options: CountOptions? = nil) throws -> Int {
```
jmikola: `CountDocumentsOptions` instead of `CountOptions`?

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -0,0 +1,127 @@
...
+    public let sort: Document?
+
+    /// When `true`, creates a new document if no document matches the query.
+    public let upsert: Bool?
+}
+
+/// Options to use when executing a `findOneAndUpdate` command on a `MongoCollection`. 
+public struct FindOneAndUpdateOptions {
+    /// A set of filters specifying to which array elements an update should apply.
+    public let arrayFilters: [Document]?
```
jmikola: AFAICT, FindOneAndUpdateOptions is basically FindOneAndReplaceOptions with the addition of this option. Since Swift doesn't allow struct inheritance, are we stuck with copy-pasting everything?

The same thought also occurred while looking at EstimatedDocumentCountOptions, which only contains a common `maxTimeMS` member.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -0,0 +1,127 @@
...
+
+/// Options to use when executing a `findOneAndDelete` command on a `MongoCollection`. 
+public struct FindOneAndDeleteOptions {
+    /// Specifies a collation to use.
+    public let collation: Document?
+
+    /// The maximum amount of time to allow the query to run.
+    public let maxTimeMS: Int64?
+
+    /// Limits the fields to return for all matching documents.
```
jmikola: Small nit, but "documents" can refer to "document" for all descriptions in the `findAndModify` option structs since these operations never work with multiple documents.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Finds a single document and deletes it, returning the original.
+     *
+     * - Parameters:
+     *   - filter: `Document` representing the match criteria
+     *   - options: Optional `FindOneAndDeleteOptions` to use when executing the command
+     *
+     * - Returns: The deleted document, or `nil` if no document was deleted.
+     */
+    @discardableResult
```
jmikola: I considered that discarding the result doesn't make sense, as users should probably use the single-statement methods (e.g. `deleteOne`); however, the fact that `findAndModify` supports a `sort` option justifies why someone may want to use these methods while still not caring about the return value. :+1:

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -0,0 +1,131 @@
+import libmongoc
+
+/// An extension of `MongoCollection` encapsulating bulk write operations.
+extension MongoCollection {
+    /**
+     * Sends a batch of writes to the server at the same time.
```
jmikola: Not sure if it's worth changing this, but "at the same time" may not be technically correct since operations will be sent in batches. For PHPLIB, we simply say "Executes multiple write operations" and the logic for splitting requests into batches is discussed elsewhere.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * - Parameters:
+     *   - requests: an `[WriteModel]` containing the writes to perform
```
jmikola: I think "a" is fine here, unless you were actually writing "an array of `WriteModel`s".

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -0,0 +1,131 @@
...
+     *   - options: optional `BulkWriteOptions` to use while executing the operation
+     * - Returns: a `BulkWriteResult`
+     * - Throws: an error if `requests` is empty, or if any error occurs while performing the writes
```
jmikola: Should the error type ``MongoError.typeError` be noted here? I noticed that was done in for `Document.get` and I did the same in ReadPreference docs.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -0,0 +1,131 @@
...
+     */
+    @discardableResult
+    public func bulkWrite(_ requests: [WriteModel], options: BulkWriteOptions? = nil) throws -> BulkWriteResult {
+        throw MongoError.commandError(message: "Unimplemented command")
+    }
+}
+
+/// Options to use when performing a bulk write operation on a `MongoCollection`.
+public struct BulkWriteOptions {
+    /// If `true`, when an insert fails, return without performing the remaining writes.
```
jmikola: Should this just use `/** ... */`?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Specifies a collation to use.
+    public let collation: Document?
+
+    /// When `true`, creates a new document if no document matches the query.
+    public let upsert: Bool?
+}
+
+/// The result of a bulk write operation on a `MongoCollection`.
+public struct BulkWriteResult: Decodable {
```
jmikola: Does it make sense to provide accessors for write errors, write concern errors, and whether the write was acknowledged? PHP does that with [WriteResult](http://php.net/manual/en/class.mongodb-driver-writeresult.php).

If a bulk write execution results in an error, libmongoc will create an error message for us but users should still have a programmatic API to inspect the errors. This is particularly relevant if there are both write and write concern errors.

On a separate note, users may want to know whether a write was acknowledged as some fields should not be accessible otherwise (or they should be `nil`).

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -0,0 +1,131 @@
...
+    /// Number of documents inserted.
+    public let insertedCount: Int
+
+    /// Map of the index of the operation to the id of the inserted document.
+    public let insertedIds: [Int: AnyBsonValue]
+
+    /// Number of documents matched for update.
+    public let matchedCount: Int
```
jmikola: Related to the above comment, the update and deleted counts originate from the server and won't be available if the write concern was unacknowledged. I think everything apart from `insertedIds` (which can be generated entirely client-side) should be optional.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: given our conversation earlier, not totally sure how we want to handle this....
would it make sense to throw from  `MongoCollection.bulkWrite` if there is a write error or write concern error? that would mean if we actually get to the point of returning a `BulkWriteResult`, there are no errors.

Sources/MongoSwift/MongoCollection+Read.swift
jmikola: `typealias` seems best so we can be consistent about the option structs matching the method names. The CRUD spec edits for new count APIs were done hastily, so I think that explains the oversight.

A more general question may be whether it makes sense to use option structs for these things instead of just adding optional arguments to the methods themselves. It may be worth taking a look at the Python and Ruby drivers for inspiration, as both of them use keyword args AFAIK. PHP doesn't have that, so we take all options in a dictionary arg (comparable to these option structs I suppose).

Sources/MongoSwift/MongoCollection+BulkWrite.swift
jmikola: I think it's easier to stuff a single write result struct (derived from the libmongoc reply document) onto an error object than extra the count/ID fields into their own "result" object and split out the `writeErrors` array field and `writeConcernError` document field.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -0,0 +1,140 @@
+import libmongoc
+
+/// An extension of `MongoCollection` encapsulating find and modify operations.
+extension MongoCollection {
```
kmahar: note that I have modified all the find and modify commands to support using the `MongoCollection`'s `CollectionType`.

Sources/MongoSwift/MongoCollection+Read.swift
jmikola: > I would be open to adding some helper methods that take in the params directly in cases where there are only a few, but I'm not sure where we would draw the line on that.

These are good points. I'm cool with sticking to option structs and not deviating so we can avoid having to decide where a line is. Let's just keep everything consistent.

Folks can always create their options separately to avoid the first example with a super long line of code.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
jmikola: > so I think for an unacknowledged write we should just return nil. we are already doing (or supposed to be doing) that in some other places in the driver, and since we have optionals I think it's more idiomatic.

If the Swift driver has no intention of reporting inserted IDs, I suppose `BulkWriteResult?` makes sensein place of a `BulkWriteResult` consisting entirely of unset optionals and a single `isAcknowledged = false` property.

I personally think it's useful for drivers to report inserted IDs, since those are often generated by the driver when a document without an `_id` field is inserted; however, a reasonable counter-argument is that any users that actually care about the inserted ID could simply set `_id` themselves before the insert operation.

> IMO it's kind of confusing to users if there's an error field when there doesn't need to be - they might think they need to manually check each BulkWriteResult they get back to see if writeErrors or writeConcernErrors have anything in them.

I think that `bulkWriteError` should still have a message and/or code. libmongoc is already going to provide a `bson_error_t` for us and we can use that as we normally do. The only extra behavior is that we also want to stuff the `bson_t` reply into the error as well.

One thought is that BulkWriteResult may need to be optional. Consider the scenario where an unacknowledged write encounters a network error. libmongoc is still going to fill out a `bson_error_t` even if there isn't anything to report in the `bson_t` reply.

> it does require a bit of work to split up the info, but I think our BSON decoder will automate some of that, and it saves users having to go down two nested levels (error -> result struct -> error fields) to actually see the errors.

I'm personally in favor of BulkWriteResult matching the `bson_t` reply provided by libmongoc, but if that complicates things I won't object. I imagine not splitting out the `writeErrors` and `writeConcernError` fields might prevent us from being able to re-use the same BulkWriteResult struct for both the successful `bulkWrite()` return value and the `bulkWriteError` type. In that case, let's aim for re-use.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
jmikola: Ah, I just remembered that you do have `insertedIds` here. If that's going to stay, then I think you should probably provide it for unacknowledged writes as well.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: > I think that bulkWriteError should still have a message and/or code. libmongoc is already going to provide a bson_error_t for us and we can use that as we normally do. The only extra behavior is that we also want to stuff the bson_t reply into the error as well.
One thought is that BulkWriteResult may need to be optional. Consider the scenario where an unacknowledged write encounters a network error. libmongoc is still going to fill out a bson_error_t even if there isn't anything to report in the bson_t reply.

Ok, so how about this?

```swift
case writeError(code: Int32, message: String)
indirect case bulkWriteError(code: Int32, message: String, result: BulkWriteResult?, writeErrors: [MongoError], writeConcernError: Document?)
```

Adds a code and message to `bulkWriteError`, and makes the `BulkWriteResult` optional (network error case). I also changed `writeErrors` to actually be an array of error structs rather than just documents (added `indirect` keyword to support this nested structure).

Re: the inserted IDs and what `bulkWrite` returns - I don't have very strong feelings on what we do here. but I will note that we took the "no `acknowledged` field, just return `nil` if unacknowledged" approach for all the existing CRUD write ops, and in some of these cases, the optionally returned struct contains inserted IDs. I do agree that it's kind of odd to only provide the IDs when the write concern is acknowledged, since they exist regardless. 
I don't really care what we go with as long as we make the behavior consistent, possibly by modifying the existing write ops. @mbroadst did the initial CRUD API sketch so maybe he has opinions on this?

jmikola: > I also changed writeErrors to actually be an array of error structs rather than just documents (added indirect keyword to support this nested structure).

I think `writeConcernError` should also be an indirect case, too (assuming we have one for a write concern error).

kmahar: ^ just updated my comment

jmikola: Should `writeErrors: [MongoError]` be `[MongoError]?`, or would you ensure that `writeErrors` is an empty array even if the only bulk write error was the `writeConcernError`?

kmahar: I was imagining we'd just use an empty array, but thinking about it more, I think making it an optional is better. it's nicer to be able to say like `if let writeErrs = error.writeErrors { ... }` instead of `if errors.writeErrors.count > 0 { ... }`.

 also saves us creating a bunch of unused arrays

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -128,7 +128,7 @@ public class WriteConcern: Codable {
             let journalStr = String(describing: journal)
             let wStr = String(describing: w)
             let timeoutStr = String(describing: wtimeoutMS)
-            throw MongoError.writeConcernError(message:
+            throw MongoError.writeConcernError(code: 0, message:
```
jmikola: I assume this is something we can clear up with [SWIFT-145](https://jira.mongodb.org/browse/SWIFT-145)? Ideally, we should use something like an invalidArgumentError here and save writeConcernError for an actual runtime exception.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: let's discuss the bulkWrite / unacknowledged stuff in triage as this is blocking further progress on our CRUD work

jmikola: We concluded to keep the existing insertOne/insertMany behavior and have operations return an optional result object (i.e. `nil` for `{w:0}`).

PR #85, opened 2018-07-02
SWIFT-31: Implement ReadPreference type

jmikola:
https://jira.mongodb.org/browse/SWIFT-31

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -113,6 +113,20 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      * - Returns: a new `Document`
      */
     public init(arrayLiteral elements: BsonValue?...) {
+        self.init(elements)
+    }
+
+    /**
+     * Initializes a `Document` using an array where the values are optional
+     * `BsonValue`s. Values are stored under a string of their index in the
+     * array.
+     *
+     * - Parameters:
+     *   - elements: a `[BsonValue?]`
+     *
+     * - Returns: a new `Document`
+     */
+    internal init(_ elements: [BsonValue?]) {
         self.storage = DocumentStorage()
         for (i, elt) in elements.enumerated() {
             self[String(i)] = elt
--- a/Sources/MongoSwift/ReadPreference.swift
+++ b/Sources/MongoSwift/ReadPreference.swift
@@ -1,0 +1,164 @@
+import libmongoc
+
+/// A class to represent a MongoDB read preference.
+final public class ReadPreference {
+
+    /// An enumeration of possible ReadPreference modes.
+    public enum Mode: String {
+        /// See https://docs.mongodb.com/manual/reference/read-preference/
+        case primary
+        case primaryPreferred
+        case secondary
+        case secondaryPreferred
+        case nearest
+
+        internal var readMode: mongoc_read_mode_t {
+            switch self {
+            case .primary:
+                return MONGOC_READ_PRIMARY
+            case .primaryPreferred:
+                return MONGOC_READ_PRIMARY_PREFERRED
+            case .secondary:
+                return MONGOC_READ_SECONDARY
+            case .secondaryPreferred:
+                return MONGOC_READ_SECONDARY_PREFERRED
+            case .nearest:
+                return MONGOC_READ_NEAREST
+            }
+        }
+
+        internal init? (readMode: mongoc_read_mode_t) {
+            switch readMode {
+            case MONGOC_READ_PRIMARY:
+                self = .primary
+            case MONGOC_READ_PRIMARY_PREFERRED:
+                self = .primaryPreferred
+            case MONGOC_READ_SECONDARY:
+                self = .secondary
+            case MONGOC_READ_SECONDARY_PREFERRED:
+                self = .secondaryPreferred
+            case MONGOC_READ_NEAREST:
+                self = .nearest
+            default:
+                return nil
+            }
+        }
+    }
+
+    /// A pointer to a `mongoc_read_prefs_t`.
+    internal var _readPreference: OpaquePointer?
+
+    /// The mode of this `ReadPreference`
+    public var mode: Mode {
+        let readMode = mongoc_read_prefs_get_mode(self._readPreference)
+
+        guard let mode = Mode(readMode: readMode) else {
+            preconditionFailure("Unexpected read preference mode: \(readMode)")
+        }
+
+        return mode
+    }
+
+    /// The tags of this `ReadPreference`
+    public var tagSets: [Document] {
+        guard let bson = mongoc_read_prefs_get_tags(self._readPreference) else {
+            return []
+        }
+
+        let wrapped = Document(fromPointer: bson)
+
+        return wrapped.values as! [Document]
+    }
+
+    /// The maxStalenessSeconds of this `ReadPreference`
+    public var maxStalenessSeconds: Int64? {
+        let maxStalenessSeconds = mongoc_read_prefs_get_max_staleness_seconds(self._readPreference)
+
+        return maxStalenessSeconds == MONGOC_NO_MAX_STALENESS ? nil : maxStalenessSeconds
+    }
+
+    /**
+     * Initializes a `ReadPreference` from a `Mode`.
+     *
+     * - Parameters:
+     *   - mode: a `Mode`
+     *
+     * - Returns: a new `ReadPreference`
+     */
+    public init(_ mode: Mode) {
+        self._readPreference = mongoc_read_prefs_new(mode.readMode)
+    }
+
+    /**
+     * Initializes a `ReadPreference`.
+     *
+     * - Parameters:
+     *   - mode: a `Mode`
+     *   - tagSets: an optional `[Document]`
+     *   - maxStalenessSeconds: an optional Int64
+     *
+     * - Returns: a new `ReadPreference`
+     *
+     * - Throws:
+     *   - A `MongoError.readPreferenceError` if mode is primary and tagSets is non-empty
+     *   - A `MongoError.readPreferenceError` if maxStalenessSeconds non-nil and < 90
+     */
+    public init(_ mode: Mode, tagSets: [Document]? = nil, maxStalenessSeconds: Int64? = nil) throws {
+        self._readPreference = mongoc_read_prefs_new(mode.readMode)
+
+        if let tagSets = tagSets {
+            guard mode != .primary || tagSets.isEmpty else {
+                throw MongoError.readPreferenceError(message: "tagSets may not be used with primary mode")
+            }
+
+            let encoder = BsonEncoder()
+            let tags = try encoder.encode(Document(tagSets))
+            mongoc_read_prefs_set_tags(self._readPreference, tags.data)
+        }
+
+        if let maxStalenessSeconds = maxStalenessSeconds {
+            guard maxStalenessSeconds >= MONGOC_SMALLEST_MAX_STALENESS_SECONDS else {
+                throw MongoError.readPreferenceError(message: "Expected maxStalenessSeconds to be >= \(MONGOC_SMALLEST_MAX_STALENESS_SECONDS), \(maxStalenessSeconds) given")
+            }
+
+            mongoc_read_prefs_set_max_staleness_seconds(self._readPreference, maxStalenessSeconds);
+        }
+    }
+
+    /// Initializes a new `ReadPreference` by copying an existing `ReadPreference`.
+    public init(from readPreference: ReadPreference) {
+        self._readPreference = mongoc_read_prefs_copy(readPreference._readPreference)
+    }
+
+    /// Initializes a new `ReadPreference` by copying a `mongoc_read_prefs_t`.
+    /// The caller is responsible for freeing the original `mongoc_read_prefs_t`.
+    internal init(from readPreference: OpaquePointer?) {
+        self._readPreference = mongoc_read_prefs_copy(readPreference)
+    }
+
+    /// Cleans up the internal `mongoc_read_prefs_t`.
+    deinit {
+        guard let readPreference = self._readPreference else { return }
+        mongoc_read_prefs_destroy(readPreference)
+        self._readPreference = nil
+    }
+}
+
+/// An extension of `ReadPreference` to make it `Equatable`.
+extension ReadPreference: Equatable {
+    public static func == (lhs: ReadPreference, rhs: ReadPreference) -> Bool {
+        if lhs.mode != rhs.mode {
+            return false
+        }
+
+        if lhs.tagSets != rhs.tagSets {
+            return false
+        }
+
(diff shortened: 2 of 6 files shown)


Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,167 @@
...
+            case MONGOC_READ_PRIMARY_PREFERRED:
+                self = .primaryPreferred
+            case MONGOC_READ_SECONDARY:
+                self = .secondary
+            case MONGOC_READ_SECONDARY_PREFERRED:
+                self = .secondaryPreferred
+            case MONGOC_READ_NEAREST:
+                self = .nearest
+            default:
+                return nil
```
jmikola: This was necessary since Swift doesn't translate a C enum directly to a Swift enum, so we have to refer to the C enum values as constants. That said, `nil` return value here will translate to a `preconditionFailure` in the `mode` accessor.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -113,6 +113,20 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      * - Returns: a new `Document`
      */
     public init(arrayLiteral elements: BsonValue?...) {
+        self.init(elements)
```
jmikola: Note: `convenience` was not used here as it's only for classes and Document is a struct.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Initializes a `ReadPreference` from a `Mode`.
+     *
+     * - Parameters:
+     *   - mode: a `Mode`
+     *
+     * - Returns: a new `ReadPreference`
+     */
+    public init(_ mode: Mode) {
+        self._readPreference = mongoc_read_prefs_new(mode.readMode)
```
jmikola: Since initialization from a mode enum will never throw, I think this constructor is helpful.

I did consider making this a `convenience` for the long-form constructor. I believe that would just require changing this to:

```swift
try! self.init(mode)
```

I originally expected I might need to pass `nil` as one of the two optional arguments (`tagSets` or `maxStalenessSeconds`), but Swift seemed to infer that I must be calling the other initializer instead of making a recursive call to this initializer. I'm not sure if that's due to `convenience` or some other clever delegation logic.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Initializes a new `ReadPreference` by copying an existing `ReadPreference`.
+    public init(from readPreference: ReadPreference) {
+        self._readPreference = mongoc_read_prefs_copy(readPreference._readPreference)
+    }
+
+    /// Initializes a new `ReadPreference` by copying a `mongoc_read_prefs_t`.
+    /// The caller is responsible for freeing the original `mongoc_read_prefs_t`.
+    internal init(from readPreference: OpaquePointer?) {
+        self._readPreference = mongoc_read_prefs_copy(readPreference)
```
jmikola: I don't yet have a way to test this, but I expect it will be tested when we support ReadPreference accessors on the core classes (e.g. MongoClient).

Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -0,0 +1,89 @@
...
+        expect(ReadPreference(.primary)).toNot(equal(ReadPreference(.secondary)))
+        expect(ReadPreference(.primary)).toNot(equal(ReadPreference(.secondaryPreferred)))
+        expect(ReadPreference(.primary)).toNot(equal(ReadPreference(.nearest)))
+
+        expect(try ReadPreference(.secondary, tagSets: nil))
+            .to(equal(ReadPreference(.secondary)))
+        expect(try ReadPreference(.secondary, tagSets: []))
+            .to(equal(try! ReadPreference(.secondary, tagSets: [])))
+        expect(try ReadPreference(.secondary, tagSets: [["dc": "east"], []]))
+            .to(equal(try! ReadPreference(.secondary, tagSets: [["dc": "east"], []])))
```
jmikola: I think it makes sense to `try!` here, since I should guarantee that the expected value is well-formed. If it wasn't, I'd prefer a loud error rather than simply a non-equal comparison.

Do you agree this is preferable to `try` and `try?`, the latter of which might end up comparing against an expected `nil`?

jmikola:
Note: I've yet to add ReadPreference to any of the core classes, although it looks like that is expected (based on the corresponding read and write concern PRs). Will continue with that, but feel free to review the base ReadPreference class and tests for now.

kmahar:
Sounds good, I am also fine with you creating separate tickets for that work if you prefer. 

It looks like the tests are not compiling though, so I will wait on those til they are working.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,164 @@
+import libmongoc
+
+/// A class to represent a MongoDB read preference.
+final public class ReadPreference {
+
+    /// An enumeration of possible ReadPreference modes.
+    public enum Mode: String {
+        /// See https://docs.mongodb.com/manual/reference/read-preference/
```
kmahar: let's move this comment up one line. with its current position, it will only show up in the docs under the `.primary` case, but it should be associated with the `Mode` type as a whole. or we could even put it on the `ReadPreference` class itself I suppose. 

you could also list it as a "see also", like [here](https://github.com/mongodb/mongo-swift-driver/blob/c3e9c02711fcab5a88d39156ecb1df2cf1575e3f/Sources/MongoSwift/MongoClient.swift#L106), which then shows up nicely in the [docs](https://mongodb.github.io/mongo-swift-driver/Classes/MongoClient.html#/s:10MongoSwift0A6ClientCACSS16connectionString_AA0C7OptionsVSg7optionstKcfc). 
though given that we have no info on the modes here, maybe the "also" part doesn't make sense... I don't have strong feelings beyond moving it from this spot to somewhere more useful

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,164 @@
...
+    deinit {
+        guard let readPreference = self._readPreference else { return }
+        mongoc_read_prefs_destroy(readPreference)
+        self._readPreference = nil
+    }
+}
+
+/// An extension of `ReadPreference` to make it `Equatable`.
+extension ReadPreference: Equatable {
+    public static func == (lhs: ReadPreference, rhs: ReadPreference) -> Bool {
```
kmahar: I have a slight stylistic preference to just write these like
```swift
return lhs.mode == rhs.mode
    && lhs.tagSets == rhs.tagSets
    && lhs.maxStalenessSeconds == rhs.maxStalenessSeconds
```

was there some reason for writing it like this instead?

Sources/MongoSwift/ReadPreference.swift
jmikola: On second thought, Swift actually didn't like `try! self.init(mode)`. I'm not sure what I was looking at when I posted the last comment.

This will remain as-is, as passing `nil` named arguments just to coerce selection of the other initializer seems a hackish.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,164 @@
...
+     * - Parameters:
+     *   - mode: a `Mode`
+     *
+     * - Returns: a new `ReadPreference`
+     */
+    public init(_ mode: Mode) {
+        self._readPreference = mongoc_read_prefs_new(mode.readMode)
+    }
+
+    /**
```
kmahar: in this comment:
Int64 -> `Int64`
primary -> `.primary`
tagSets -> `tagSets`

Sources/MongoSwift/ReadPreference.swift
kmahar: yes, since this is a one-liner I think it's fine to just have it as a separate/non-`convenience` init.

Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -0,0 +1,89 @@
...
+        ]
+    }
+
+    override func setUp() {
+        self.continueAfterFailure = false
+    }
+
+    func testReadPreferenceMode() {
+        let primary = ReadPreference(.primary)
+        expect(primary.mode).to(equal(.primary))
```
jmikola: Travis [does not like this](https://travis-ci.org/mongodb/mongo-swift-driver/builds/399310940#L1932):

```
/home/travis/build/mongodb/mongo-swift-driver/Tests/MongoSwiftTests/ReadPreferenceTests.swift:20:33: error: ambiguous reference to member 'equal'
        expect(primary.mode).to(equal(.primary))
                                ^~~~~
```

Swift 4.1 didn't seem to have a problem with this, but I assume that's due to some subtle improvement since 4.0. Will change so that `equal()` refers to the full enum (e.g. `ReadPreference.Mode.primary`).

jmikola:
> I am also fine with you creating separate tickets for that work if you prefer.

Aye, that seems preferable. I created [SWIFT-143](https://jira.mongodb.org/browse/SWIFT-143).

kmahar:
we are missing some good prefer / ReadPreference joke opportunities here

I can't figure out how to reply on your other comment since it's now outdated, but that error is surprising, I would have thought the compiler could figure out the type based on what it's being compared to... huh

kmahar:
I assume we will also need to add `ReadPreference` to commands? are there any associated spec tests?

Sources/MongoSwift/ReadPreference.swift
jmikola: I just wasn't sure what the syntax was preferred for multi-line conditionals (haven't setup `swift-lint` locally) and I'm used to enumerating simple guard statements in PHPLIB. I'll change this to one condition.

jmikola:
> I assume we will also need to add ReadPreference to commands? are there any associated spec tests?

I don't believe so. Pre-operation read preference, read concern, and write concern are not required by any spec as some drivers expect all operations on a collection (or other core class) to use the class-level option. This is why drivers have methods that clone a collection with new options (e.g. `withOptions()`).

Sources/MongoSwift/ReadPreference.swift
jmikola: I think a "SeeAlso" at the class level makes the most sense. Will do.

Sources/MongoSwift/ReadPreference.swift
jmikola: If `///` is for documentation and this is internal, wouldn't a non-documentation comment would be preferred?

jmikola: I've opted (no pun intended) to move the precondition failure here, since we only call this once from the `mode` accessor. That let's us remove the optional behavior of this initializer and simplifies the accessor.

Sources/MongoSwift/ReadPreference.swift
jmikola: If `tagSets` should be monospaced as a parameter name, I assume `mode` and ``maxStalenessSeconds` should as well.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -0,0 +1,163 @@
...
+        guard let mode = Mode(readMode: readMode) else {
+            preconditionFailure("Unexpected read preference mode: \(readMode)")
+        }
+
+        return mode
+    }
+
+    /// The tags of this `ReadPreference`
+    public var tagSets: [Document] {
+        guard let bson = mongoc_read_prefs_get_tags(self._readPreference) else {
```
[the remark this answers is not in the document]
jmikola: It shouldn't. In fact, if its argument was null we'd probably get a seg fault first (assuming `BSON_ASSERT()` was disabled in a release build).

Would it be preferable to replace the `guard` with a forced optional unwrapping (i.e. `!`)?

Sources/MongoSwift/ReadPreference.swift
kmahar: jazzy decides what to document based on access control level - it defaults to a minimum ACL of public, so private and internal things are skipped. 

so I usually will use documentation comments on internal stuff anyway to be consistent stylistically (and it would be convenient if we decided we wanted to generate a separate docset of our internal types in the future.) 

in any case, that sounds reasonable to me!

kmahar:
> I don't believe so. Pre-operation read preference, read concern, and write concern are not required by any spec as some drivers expect all operations on a collection (or other core class) to use the class-level option. This is why drivers have methods that clone a collection with new options (e.g. withOptions()).

I see. my concern is that since we already added RC and WC at the command level, not supporting RP there seems somewhat inconsistent. but tbh I don't have a great feel for how people actually use RPs compared to RC and WC, so maybe this isn't that important.

jmikola:
"I don't believe so" in https://github.com/mongodb/mongo-swift-driver/pull/85#issuecomment-401952818 was only referring to the question of spec tests. Sorry for not clarifying.

If operations support RC and WC options, we should absolutely support read preference as well. I've amended [SWIFT-143](https://jira.mongodb.org/browse/SWIFT-143) to cover that.

Sources/MongoSwift/ReadPreference.swift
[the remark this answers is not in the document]
kmahar: I think a `preconditionFailure` would be better, then we can provide an error message

Sources/MongoSwift/ReadPreference.swift
kmahar: ah yes, missed those. 

basically, if you look at how params show up in the docs ([here](https://mongodb.github.io/mongo-swift-driver/Classes/MongoDatabase.html#/s:10MongoSwift0A8DatabaseC10collectionAA0A10CollectionCyxGSS_xm8withTypeAA0E7OptionsVSg7optionstKs9DecodableRzs9EncodableRzlF) is one random example), it doesn't matter if there are backticks around the name in the parameter list because that get specially formatted, but anywhere else that a parameter, type etc is mentioned it's nice to have backticks so we get the monospace text.

Sources/MongoSwift/BSON/Document.swift
jmikola: The bottom example of https://developer.apple.com/documentation/swift/expressiblebyarrayliteral shows an array literal initializer delegating to another `init` method:

```swift
extension OrderedSet: ExpressibleByArrayLiteral {
    init(arrayLiteral: Element...) {
        self.init()
        for element in arrayLiteral {
            self.append(element)
        }
    }
}
```

That documentation goes into some detail about why an Array isn't accepted in place of an array literal, but makes no claims about the opposite. I think we're in the clear with this.

The official docs for [`Set.init(arrayLiteral)`](https://developer.apple.com/documentation/swift/set/1539368-init) also show it annotated as a `convenience` initializer, from which we can reasonably conclude that it delegates to another `init` method.

kmahar: 👍

PR #89, opened 2018-07-08
SWIFT-142 Remove dictionary init for document

kmahar:
this removes the non-literal dictionary `init`, as well as our single usage of it in the tests.


Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -21,7 +21,7 @@ final class CodecTests: XCTestCase {
     }
 
     func testEncodeListDatabasesOptions() throws {
-        let options = ListDatabasesOptions(filter: Document(["a": 10]), nameOnly: true, session: ClientSession())
+        let options = ListDatabasesOptions(filter: ["a": 10], nameOnly: true, session: ClientSession())
```
mbroadst: cool, you don't even need an `as`!

kmahar: yup, since the `filter` param is always a `Document` the compiler infers it 😀

PR #92, opened 2018-07-09
[WIP] SWIFT-132: Error on unsupported BSON type in nextBsonValue()

jmikola:
https://jira.mongodb.org/browse/SWIFT-132

Note: I'm opening this PR for discussion alongside [SWIFT-155](https://jira.mongodb.org/browse/SWIFT-155). For reasons explained in SWIFT-132, we cannot `throw` from `nextBsonValue()`. A precondition error may be reasonable if we validate BSON before initializing a Document from raw `bson_t` data. In that case, the Document initializer can throw long before `nextBsonValue()` would ever be called.

### DIFF
--- a/Sources/MongoSwift/BSON/BsonValue.swift
+++ b/Sources/MongoSwift/BSON/BsonValue.swift
@@ -50,15 +50,17 @@ public enum BsonType: Int {
     maxKey
 }
 
-internal let BsonTypeMap: [UInt32: BsonValue.Type] = [
+internal let BsonTypeMap: [UInt32: BsonValue.Type?] = [
     0x01: Double.self,
     0x02: String.self,
     0x03: Document.self,
     0x04: [BsonValue].self,
     0x05: Binary.self,
+    0x06: nil,
     0x07: ObjectId.self,
     0x08: Bool.self,
     0x09: Date.self,
+    0x0a: nil,
     0x0b: RegularExpression.self,
     0x0c: DBPointer.self,
     0x0d: CodeWithScope.self,
@@ -73,8 +75,15 @@ internal let BsonTypeMap: [UInt32: BsonValue.Type] = [
 ]
 
 internal func nextBsonValue(iter: inout bson_iter_t) -> BsonValue? {
-    let type = bson_iter_type(&iter)
-    guard let typeToReturn = BsonTypeMap[type.rawValue] else { return nil }
+    let type = bson_iter_type(&iter).rawValue
+    guard let index = BsonTypeMap.index(forKey: type) else {
+        let hex = String(type, radix: 16)
+        let key = String(cString: bson_iter_key(&iter))
+        preconditionFailure("Detected unknown BSON type \"\(hex)\" for fieldname \"\(key)\". Are you using the latest driver version?")
+    }
+    guard let typeToReturn = BsonTypeMap[index].value else {
+        return nil
+    }
     return typeToReturn.from(iter: &iter)
 }
 


Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -73,8 +75,15 @@ internal let BsonTypeMap: [UInt32: BsonValue.Type] = [
 ]
 
 internal func nextBsonValue(iter: inout bson_iter_t) -> BsonValue? {
-    let type = bson_iter_type(&iter)
-    guard let typeToReturn = BsonTypeMap[type.rawValue] else { return nil }
+    let type = bson_iter_type(&iter).rawValue
+    guard let index = BsonTypeMap.index(forKey: type) else {
+        let hex = String(type, radix: 16)
+        let key = String(cString: bson_iter_key(&iter))
+        preconditionFailure("Detected unknown BSON type \"\(hex)\" for fieldname \"\(key)\". Are you using the latest driver version?")
```
jmikola: See: https://jira.mongodb.org/browse/DRIVERS-281

mbroadst:
@jmikola I've read SWIFT-155, and I think we have a few options here. Outside of rewriting the entire BSON layer in pure Swift (an admirable goal, but one we should certainly plan for and probably not attempt right now), I think hwe should consider adding in the ability to opt-in to validation similar to libmongoc. The current API here is potentially dangerous, you are correct, but it also provides the best performance profile. 

I think we could either provide a custom initializer (f.ex `withValidation: true`), or an outright new type (`ValidatedDocument`), to address the issues you've brought up. Given one of these methods is provided to users, then I think adding a precondition here is absolutely fine, as it guards users who do not opt-in to validation. What do you think?

jmikola:
> The current API here is potentially dangerous, you are correct, but it also provides the best performance profile.

After closer examination while writing up [SWIFT-155](https://jira.mongodb.org/browse/SWIFT-155) I believe the only danger is appending duplicate keys. I originally expected there was going to be larger issues due to something like `doc["a"]["b"] = "foo"` before I realized that subscript chaining is unsupported. While appending duplicates is still a WTF (and can lead to effectively "missing" data, since the first key wins in a lookup), we can at least clearly document `subscript.set`.

> Given one of these methods is provided to users, then I think adding a precondition here is absolutely fine, as it guards users who do not opt-in to validation. What do you think?

I'm amenable to letting users opt-in to validation (or perhaps opt-out if we enable it by default), but I still have reservations about using a precondition failure here. If we encounter corruption, it's likely completely outside of the user's control (e.g. corrupted network traffic). Aborting the entire application seems a bit much, but I'll concede that it may be safer than simply returning `nil`.

That may support the idea of enabling validation by default and allowing users to opt-in to the current behavior of maximum performance at the risk of an unexpected (and unrecoverable) precondition failure.

kmahar:
I think adding a validating initializer or an alternate type sounds reasonable. I think a preconditionFailure in the subscript getter is preferable to returning nil. we should just document everything well.

To throw out some more ideas I've considered in the past: given that subscript prevents us from throwing, it could make sense to add alternative throwing getters and setters for documents, documenting `subscript.get` as something rather unsafe and only to be used with documents where data corruption. and `subscript.set` as something to only be used where dupe keys are not a concern. We could have something like:

```swift
/// subscript for working with documents known to be valid (either constructed from Swift values
///  or by parsing JSON, or from raw BSON with `validate` on.) preconditionFailure if there is an error. 
public subscript(key: String) -> BsonValue? {
    /// retrieves a value from the document. assumes the document is constructed from valid BSON
    get { ... }
   /// sets a value on the document. assumes the key is not already present in the document. 
    set { ... }
}

/// (this is the function currently called "get". we should change it to *not* use `subscript.get` 
/// internally so it won't preconditionFail on corrupt BSON)
/// retrieves the value stored under `key` and attempts to cast it to type `T`. 
//// throws:
/// - if we have some problem with the bson iter (corrupt BSON) 
/// - if the key is not present in the document
/// - if the value stored under key cannot be cast to type `T` 
public func getTyped<T: BsonValue>(_ key: String) throws -> T { 
    // this just calls `get` and then casts its return value to type `T` 
}

/// retrieves the value stored under `key`
//// throws:
/// - if we have some problem with the bson iter (corrupt BSON) 
/// - if the key is not present in the document
public func get(_ key: String) throws -> BsonValue? { ... }

/// sets the value for `key` to `value`. if `checkIfExists == true`, 
/// confirms that the key does not exist before setting the value.
/// throws if 
///  - appending the value to the bson_t fails
/// - `checkIfKeyExists == true` and either the BSON cannot be parsed or `key` is already present in the document
/// (note that there is a performance penalty for using that option)
public func set(_ value: BsonValue?, forKey key: String, checkIfKeyExists: Bool = false) throws { ... }
```

This gives users a couple different options when they construct from raw BSON:
- validate initially and then use subscript to get and set
- do not validate initially, and then use the throwing getter, which will fail if we have corrupt data and can't successfully iterate the BSON

this would also alleviate the duplicate key issue with the option to check if the key is present before setting another value for it. 

we should also document that you need to create a new doc if you want to remove a key. we might consider adding a `filter` method that returns a new doc to support this.

kmahar:
we should also probably audit everywhere we use subscripts in the BSON lib itself to make sure nothing sketchy is going on

jmikola:
> To throw out some more ideas I've considered in the past: given that subscript prevents us from throwing, it could make sense to add alternative throwing getters and setters for documents, documenting subscript.get as something rather unsafe and only to be used with documents where data corruption isn't an issue.

This made me think of a new idea. Independent of whether we decide to take an optional `validate` option in the Document initializer, we can simply expose a `validate() throws` method on Document that users can call on their own. If that succeeds, they can trust that `subscript.get` will never raise an error. It would also allow us to expose additional options from [`bson_validate_with_error()`](http://mongoc.org/libbson/current/bson_validate_with_error.html) if users would like stricter validation (e.g. no dots in field names).

mbroadst:
@kmahar @jmikola can you capture this conversation in a ticket so we don't lose track?

kmahar:
I wrote a summary on [SWIFT-159](https://jira.mongodb.org/browse/SWIFT-159), feel free to edit

jmikola:
Closing this, as [SWIFT-132](https://jira.mongodb.org/browse/SWIFT-132) is obsolete in light of https://github.com/mongodb/mongo-swift-driver/commit/40e067d2619ca55b2d59f57aff9389832f6944a4 for [SWIFT-110](https://jira.mongodb.org/browse/SWIFT-110).

PR #93, opened 2018-07-10
SWIFT-146: Implement findAndModify API

kmahar:
This fills out the API sketch and implements the corresponding CRUD tests. I plan on adding more test cases tomorrow in `MongoCollectionTests` to check that nothing catastrophic happens when using the options the CRUD tests don't cover, but I figured I would put up what I have so far.

I think the options encoding solution is pretty good: I've created a `private class FindAndModifyOptions` wrapping a `mongoc_find_and_modify_opts_t`. Each of the Swift options `struct`s conforms to the `FindAndModifyOptionsConvertible` protocol, which means it knows how to represent itself as a `FindAndModifyOptions`. The private `findAndModify` method that we route all the public API methods through can then just do the conversion right before it sends the command.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -19,7 +19,8 @@ extension MongoCollection {
     @discardableResult
     public func findOneAndDelete(_ filter: Document,
                                  options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
-        return try self.findAndModify(filter: filter, options: options)
+        let opts = options ?? FindOneAndDeleteOptions()
+        return try self.findAndModify(filter: filter, options: opts)
     }
 
     /**
@@ -73,7 +74,6 @@ extension MongoCollection {
                                options: FindAndModifyOptionsConvertible? = nil) throws -> CollectionType? {
 
         var opts = try options?.asOpts()
-
         if let update = update {
             // if no options were provided but we need to set update, create them
             opts = opts ?? FindAndModifyOptions()
@@ -291,7 +291,7 @@ private class FindAndModifyOptions {
         if let coll = collation { extra["collation"] = coll }
 
         // note: mongoc_find_and_modify_opts_set_max_time_ms() takes in a 
-        // uint32_t, but it should be a positive 64-bit integer. thus, we
+        // uint32_t, but it should be a positive 64-bit integer, so we
         // set maxTimeMS by directly appending it instead. see CDRIVER-1329
         if let maxTime = maxTimeMS {
             guard maxTime > 0 else {
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -89,9 +89,9 @@ extension MongoCollection {
             throw MongoError.commandError(message: toErrorString(error))
         }
 
-        if reply == [:] { return nil }
+        guard let value = reply["value"] as? Document else { return nil }
 
-        return try BsonDecoder().decode(CollectionType.self, from: reply)
+        return try BsonDecoder().decode(CollectionType.self, from: value)
     }
 }
 
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -10,7 +10,9 @@ extension MongoCollection {
      *   - options: Optional `FindOneAndDeleteOptions` to use when executing the command
      *
      * - Returns: The deleted document, represented as a `CollectionType`, or `nil` if no document was deleted.
+     *
      * - Throws: 
+     *   - `MongoError.invalidArgument` if any of the provided options are invalid
      *   - `MongoError.commandError` if there are any errors executing the command.
      *   - A `DecodingError` if the deleted document cannot be decoded to a `CollectionType` value
      */
@@ -30,7 +32,9 @@ extension MongoCollection {
      *
      * - Returns: A `CollectionType`, representing either the original document or its replacement,
      *      depending on selected options, or `nil` if there was no match.
+     *
      * - Throws: 
+     *   - `MongoError.invalidArgument` if any of the provided options are invalid
      *   - `MongoError.commandError` if there are any errors executing the command.
      *   - An `EncodingError` if `replacement` cannot be encoded to a `Document`
      *   - A `DecodingError` if the replaced document cannot be decoded to a `CollectionType` value
@@ -52,7 +56,9 @@ extension MongoCollection {
      *
      * - Returns: A `CollectionType` representing either the original or updated document,
      *      depending on selected options, or `nil` if there was no match.
-     * - Throws: 
+     *
+     * - Throws:
+     *   - `MongoError.invalidArgument` if any of the provided options are invalid
      *   - `MongoError.commandError` if there are any errors executing the command.
      *   - A `DecodingError` if the updated document cannot be decoded to a `CollectionType` value
      */
@@ -93,12 +99,11 @@ extension MongoCollection {
 public enum ReturnDocument {
     /// Indicates to return the document before the update, replacement, or insert occured.
     case before
-
     ///  Indicates to return the document after the update, replacement, or insert occured.
     case after
 }
 
-/// Indicates that an option type can be represented as a `FindAndModifyOptions`
+/// Indicates that an options type can be represented as a `FindAndModifyOptions`
 private protocol FindAndModifyOptionsConvertible {
     /// Converts `self` to a `FindAndModifyOptions`
     func asOpts() throws -> FindAndModifyOptions
@@ -284,13 +289,26 @@ private class FindAndModifyOptions {
         var extra = Document()
         if let filters = arrayFilters { extra["arrayFilters"] = filters }
         if let coll = collation { extra["collation"] = coll }
+
         // note: mongoc_find_and_modify_opts_set_max_time_ms() takes in a 
         // uint32_t, but it should be a positive 64-bit integer. thus, we
         // set maxTimeMS by directly appending it instead. see CDRIVER-1329
-        if let maxTime = maxTimeMS { extra["maxTimeMS"] = maxTime }
-        if let wc = writeConcern { extra["writeConcern"] = try BsonEncoder().encode(wc) }
+        if let maxTime = maxTimeMS {
+            guard maxTime > 0 else {
+                throw MongoError.invalidArgument(message: "maxTimeMS must be positive, but got value \(maxTime)")
+            }
+            extra["maxTimeMS"] = maxTime 
+        }
 
-        if extra != [:], !mongoc_find_and_modify_opts_append(self._options, extra.data) {
+        if let wc = writeConcern {
+            do {
+                extra["writeConcern"] = try BsonEncoder().encode(wc) 
+            } catch {
+                throw MongoError.invalidArgument(message: "Error encoding WriteConcern \(wc): \(error)")
+            }
+        }
+
+        if extra.keys.count > 0, !mongoc_find_and_modify_opts_append(self._options, extra.data) {
             throw MongoError.invalidArgument(message: "Error appending extra fields \(extra)")
         }
     }
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -15,8 +15,9 @@ extension MongoCollection {
      *   - A `DecodingError` if the deleted document cannot be decoded to a `CollectionType` value
      */
     @discardableResult
-    public func findOneAndDelete(_ filter: Document, options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
-        throw MongoError.commandError(message: "Unimplemented command")
+    public func findOneAndDelete(_ filter: Document,
+                                 options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
+        return try self.findAndModify(filter: filter, options: options)
     }
 
     /**
@@ -36,8 +37,9 @@ extension MongoCollection {
      */
     @discardableResult
     public func findOneAndReplace(filter: Document, replacement: CollectionType,
-                                  options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
-        throw MongoError.commandError(message: "Unimplemented command")
+                                  options: FindOneAndReplaceOptions? = nil) throws -> CollectionType? {
+        let update = try BsonEncoder().encode(replacement)
+        return try self.findAndModify(filter: filter, update: update, options: options)
     }
 
     /**
@@ -57,25 +59,53 @@ extension MongoCollection {
     @discardableResult
     public func findOneAndUpdate(filter: Document, update: Document,
                                  options: FindOneAndUpdateOptions? = nil) throws -> CollectionType? {
-        throw MongoError.commandError(message: "Unimplemented command")
+        return try self.findAndModify(filter: filter, update: update, options: options)
+    }
+
+    /// A private helper method for findAndModify operations to use
+    private func findAndModify(filter: Document, update: Document? = nil,
+                               options: FindAndModifyOptionsConvertible? = nil) throws -> CollectionType? {
+
+        var opts = try options?.asOpts()
+
+        if let update = update {
+            // if no options were provided but we need to set update, create them
+            opts = opts ?? FindAndModifyOptions()
+            try opts!.setUpdate(update)
+        }
+
+        let reply = Document()
+        var error = bson_error_t()
+
+        if !mongoc_collection_find_and_modify_with_opts(self._collection,
+        filter.data, opts?._options, reply.data, &error) {
+            // TODO SWIFT-144: replace with more descriptive error type(s)
+            throw MongoError.commandError(message: toErrorString(error))
+        }
+
+        if reply == [:] { return nil }
+
+        return try BsonDecoder().decode(CollectionType.self, from: reply)
     }
 }
 
 /// Indicates which document to return in a find and modify operation.
-public enum ReturnDocument: Encodable {
+public enum ReturnDocument {
     /// Indicates to return the document before the update, replacement, or insert occured.
     case before
 
     ///  Indicates to return the document after the update, replacement, or insert occured.
     case after
+}
 
-    public func encode(to encoder: Encoder) throws {
-        // fill in later on
-    }
+/// Indicates that an option type can be represented as a `FindAndModifyOptions`
+private protocol FindAndModifyOptionsConvertible {
+    /// Converts `self` to a `FindAndModifyOptions`
+    func asOpts() throws -> FindAndModifyOptions
 }
 
 /// Options to use when executing a `findOneAndDelete` command on a `MongoCollection`. 
-public struct FindOneAndDeleteOptions: Encodable {
+public struct FindOneAndDeleteOptions: FindAndModifyOptionsConvertible {
     /// Specifies a collation to use.
     public let collation: Document?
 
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -90,10 +120,25 @@ public struct FindOneAndDeleteOptions: Encodable {
...
     /// An optional `WriteConcern` to use for the command.
     public let writeConcern: WriteConcern?
+
+    fileprivate func asOpts() throws -> FindAndModifyOptions {
+        return try FindAndModifyOptions(collation: collation, maxTimeMS: maxTimeMS, projection: projection,
+                                        remove: true, sort: sort, writeConcern: writeConcern)
+    }
+
+    /// Convenience initializer allowing any/all parameters to be omitted/optional
+    public init(collation: Document? = nil, maxTimeMS: Int64? = nil, projection: Document? = nil, sort: Document? = nil,
```
kmahar: really wish we had a way to auto-generate these initializers.... seems like it would be easy enough to build into Swift as some protocol using reflection

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -147,4 +212,100 @@ public struct FindOneAndUpdateOptions: Encodable {
...
+
+        if let sort = sort, !mongoc_find_and_modify_opts_set_sort(self._options, sort.data) {
+            throw MongoError.invalidArgument(message: "Error setting sort to \(sort)")
+        }
+
+        // build an "extra" document of fields without their own setters
+        var extra = Document()
+        if let filters = arrayFilters { extra["arrayFilters"] = filters }
+        if let coll = collation { extra["collation"] = coll }
+        // note: mongoc_find_and_modify_opts_set_max_time_ms() takes in a
```
kmahar: @mbroadst - @jmikola and I discussed in person, but see the C driver ticket + discussion in #mongoc if you want more info

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -158,11 +149,20 @@ private class CrudTest {
     let args: Document
     let result: BsonValue?
     let collection: Document?
+
+    var arrayFilters: [Document]? { return self.args["arrayFilters"] as? [Document] }
+    var batchSize: Int32? { if let b = self.args["batchSize"] as? Int { return Int32(b) } else { return nil } }
```
kmahar: (just moved this up to make properties alphabetically sorted)

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -276,7 +286,7 @@ private class DistinctTest: CrudTest {
 private class FindTest: CrudTest {
     override func execute(usingCollection coll: MongoCollection<Document>) throws {
         let filter: Document = try self.args.get("filter")
-        let options = FindOptions(batchSize: batchSize, collation: collation, limit: self.limit,
+        let options = FindOptions(batchSize: self.batchSize, collation: self.collation, limit: self.limit,
```
kmahar: happened to notice we weren't using `self`s here, which are unrequired but nice for clarity (and used everywhere else), so added them

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -147,4 +212,100 @@ public struct FindOneAndUpdateOptions: Encodable {
...
+        }
+
+        // build an "extra" document of fields without their own setters
+        var extra = Document()
+        if let filters = arrayFilters { extra["arrayFilters"] = filters }
+        if let coll = collation { extra["collation"] = coll }
+        // note: mongoc_find_and_modify_opts_set_max_time_ms() takes in a 
+        // uint32_t, but it should be a positive 64-bit integer. thus, we
+        // set maxTimeMS by directly appending it instead. see CDRIVER-1329
+        if let maxTime = maxTimeMS { extra["maxTimeMS"] = maxTime }
```
kmahar: should we bother to validate that this is > 0, or wait for a complaint from further down the stack?

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -147,4 +212,100 @@ public struct FindOneAndUpdateOptions: Encodable {
...
+
+    // swiftlint:disable:next cyclomatic_complexity
+    init(arrayFilters: [Document]? = nil, bypassDocumentValidation: Bool? = nil, collation: Document?,
+         maxTimeMS: Int64?, projection: Document?, remove: Bool? = nil, returnDocument: ReturnDocument? = nil,
+         sort: Document?, upsert: Bool? = nil, writeConcern: WriteConcern?) throws {
+        self._options = mongoc_find_and_modify_opts_new()
+
+        if let bypass = bypassDocumentValidation,
+        !mongoc_find_and_modify_opts_set_bypass_document_validation(self._options, bypass) {
+            throw MongoError.invalidArgument(message: "Error setting bypassDocumentValidation to \(bypass)")
```
kmahar: I don't really know if these errors are necessary, which is why I have held off on documenting `invalidArgument` errors being thrown from the API methods. 
looking at the source code, all these setters only return false if the provided value is nil, which we can guarantee is never going to happen since we unwrap values first.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
jmikola: According to https://medium.com/@sergueivinnitskii/easy-struct-initialization-in-swift-8ee46b8d84d5, structs do get a default initializer; however, an explicit initializer is required if you want the convenience of optional properties defaulting to `nil` if unspecified. I suppose that's what you're pining for here? :)

Sources/MongoSwift/MongoCollection+FindAndModify.swift
jmikola: [`mongoc_collection_find_and_modify_with_opts()` ignores non-positive values](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-collection.c#L3258) when handling the named option. That doesn't apply here, since you're merging directly into the command document. In that case, you can trust that users will receive a server error (at least on 4.0):

```
> db.foo.runCommand({ping:1,maxTimeMS:1})
{ "ok" : 1 }
> db.foo.runCommand({ping:1,maxTimeMS:0})
{ "ok" : 1 }
> db.foo.runCommand({ping:1,maxTimeMS:-5})
{
	"ok" : 0,
	"errmsg" : "maxTimeMS is out of range",
	"code" : 2,
	"codeName" : "BadValue"
}
```

Earlier server versions could very well ignore it altogether. I've not tested them.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
kmahar: yup, that's what I meant - we have these for every single options struct and they feel very boilerplate

Sources/MongoSwift/MongoCollection+FindAndModify.swift
jmikola: > all these setters only return false if the provided value is nil, 

Are you referring to the `BSON_ASSERT()` on the `options` struct? I believe that either aborts or is compiled out in release builds. Otherwise, [this setter](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-find-and-modify.c#L153) should always return `true`.

jmikola: Unless you were referring to the [field](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-find-and-modify.c#L104) and general [append](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-find-and-modify.c#L194) setters. I don't see the harm in being defensive and checking for errors, though. We generally do this in PHPC when appending options, even though things are only liable to fall if memory cannot be allocated (in which case we have bigger problems).

kmahar: yeah, I was thinking of the ones for fields, sort, etc. 
but sounds good, let's be defensive, and we can also do the maxTimeMS > 0 check. and then document that an `invalidArgument` error is thrown if any options are invalid.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
kmahar: 👍

kmahar:
Ok, more tests added for using options like `maxTimeMS` and `bypassDocumentValidation` not covered in the crud spec tests. I'm not validating the behavior of the options, just checking that when they are provided nothing bad happens.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -29,15 +32,18 @@ extension MongoCollection {
...
      *   - `MongoError.commandError` if there are any errors executing the command.
      *   - An `EncodingError` if `replacement` cannot be encoded to a `Document`
      *   - A `DecodingError` if the replaced document cannot be decoded to a `CollectionType` value
      */
     @discardableResult
     public func findOneAndReplace(filter: Document, replacement: CollectionType,
-                                  options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
-        throw MongoError.commandError(message: "Unimplemented command")
+                                  options: FindOneAndReplaceOptions? = nil) throws -> CollectionType? {
+        let update = try BsonEncoder().encode(replacement)
```
jmikola: Why does `update` need to be encoded? Is this just for the off chance that it's not a Document so that we can later access `Document.data`? Looking at `BsonEncoder.encode()`, passing it a Document appears to be a no-op.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -50,32 +56,61 @@ extension MongoCollection {
...
+    /// A private helper method for findAndModify operations to use
+    private func findAndModify(filter: Document, update: Document? = nil,
+                               options: FindAndModifyOptionsConvertible? = nil) throws -> CollectionType? {
+
+        var opts = try options?.asOpts()
+
+        if let update = update {
+            // if no options were provided but we need to set update, create them
+            opts = opts ?? FindAndModifyOptions()
+            try opts!.setUpdate(update)
```
jmikola: Is `opts!` necessary here? I would expect the null coalescing on the previous line guarantees that it's an object (also since `FindAndModifyOptions.init` doesn't throw).

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -50,32 +56,61 @@ extension MongoCollection {
...
+        let reply = Document()
+        var error = bson_error_t()
+
+        if !mongoc_collection_find_and_modify_with_opts(self._collection,
+        filter.data, opts?._options, reply.data, &error) {
+            // TODO SWIFT-144: replace with more descriptive error type(s)
+            throw MongoError.commandError(message: toErrorString(error))
+        }
+
+        guard let value = reply["value"] as? Document else { return nil }
```
jmikola: Just realized that `mongoc_collection_find_and_modify_with_opts()` doesn't unwrap the `value` field for us. Oh well.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -50,32 +56,61 @@ extension MongoCollection {
...
-
     ///  Indicates to return the document after the update, replacement, or insert occured.
     case after
+}
 
-    public func encode(to encoder: Encoder) throws {
-        // fill in later on
-    }
+/// Indicates that an options type can be represented as a `FindAndModifyOptions`
+private protocol FindAndModifyOptionsConvertible {
```
jmikola: What does a private protocol entail? Wouldn't `internal` make more sense here to limit it to structs/classes within this file? While implementing the bulk write models, I was experimenting with:

```swift
internal protocol WriteModel {
    func addToBulkWrite(bulk: OpaquePointer)
}
```

As I understand it, `internal` on the protocol will require the `addToBulkWrite()` method to be `internal` on implementing structs.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
kmahar: yes, since `opts` is initially an optional type, it remains that way.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
kmahar: top-level `private` declarations are accessible throughout the file they're defined in -- basically, top level `private` = `fileprivate`. 
all the public types conforming to the protocol are in this file, so they can access the protocol. but they need to explicitly mark their `asOpts` implementations as `fileprivate` to match the protocol. 

in contrast, making a protocol `internal` would allow it to be used in any of our source files.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -50,32 +56,61 @@ extension MongoCollection {
...
+        var opts = try options?.asOpts()
+
+        if let update = update {
+            // if no options were provided but we need to set update, create them
+            opts = opts ?? FindAndModifyOptions()
+            try opts!.setUpdate(update)
+        }
+
+        let reply = Document()
+        var error = bson_error_t()
```
jmikola: Is this using an initializer for the libmongoc struct, which is created by Swift? I assume this shows up in some generated file on macOS, which doesn't exist on Linux. Similar to when you were talking about Swift exporting libmongoc constants and such.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -147,4 +217,113 @@ public struct FindOneAndUpdateOptions: Encodable {
...
+        self._options = mongoc_find_and_modify_opts_new()
+    }
+
+    // swiftlint:disable:next cyclomatic_complexity
+    init(arrayFilters: [Document]? = nil, bypassDocumentValidation: Bool? = nil, collation: Document?,
+         maxTimeMS: Int64?, projection: Document?, remove: Bool? = nil, returnDocument: ReturnDocument? = nil,
+         sort: Document?, upsert: Bool? = nil, writeConcern: WriteConcern?) throws {
+        self._options = mongoc_find_and_modify_opts_new()
+
+        if let bypass = bypassDocumentValidation,
```
jmikola: Would it make sense to keep the constant named the same as the argument? I recall doing `if let tagSets = tagSets` for ReadPreference.

If this was just done to be more concise, no worries.

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -349,8 +377,7 @@ private class UpdateTest: CrudTest {
     override func execute(usingCollection coll: MongoCollection<Document>) throws {
         let filter: Document = try self.args.get("filter")
         let update: Document = try self.args.get("update")
-        let arrayFilters = self.args["arrayFilters"] as? [Document]
-        let options = UpdateOptions(arrayFilters: arrayFilters, collation: self.collation, upsert: self.upsert)
+        let options = UpdateOptions(arrayFilters: self.arrayFilters, collation: self.collation, upsert: self.upsert)
```
jmikola: This looks unrelated to `findAndModify`.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -296,6 +299,8 @@ final class MongoCollectionTests: XCTestCase {
         for doc in try coll1.find() {
             expect(doc).to(beAnInstanceOf(Basic.self))
         }
+
+        // add tests here using findAndModify commands
```
jmikola: Outstanding TODO item? Should there be a ticket for this if it's not being done in this PR?

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -320,4 +325,76 @@ final class MongoCollectionTests: XCTestCase {
...
+        expect(result1).to(equal(self.doc2))
+        expect(try self.coll.count()).to(equal(2))
+
+        // test using bypassDocumentValidation
+        let opts2 = FindOneAndUpdateOptions(bypassDocumentValidation: true)
+        let result2 = try self.coll.findOneAndUpdate(filter: ["cat": "dog"], update: ["$set": ["cat": "hi"] as Document], options: opts2)
+        expect(result2).to(equal(self.doc1))
+        expect(try self.coll.count()).to(equal(2))
+
+        // test using a write concern
```
jmikola: Interestingly, `{w:0}` has no effect with `findAndModify`.

```
> db.runCommand({findAndModify:"foo", update:{$set:{x:1}}, upsert:1, new:true, writeConcern: {w:1}})
{
	"lastErrorObject" : {
		"n" : 1,
		"updatedExisting" : false,
		"upserted" : ObjectId("5b4685dfdd9df12896b02f24")
	},
	"value" : {
		"_id" : ObjectId("5b4685dfdd9df12896b02f24"),
		"x" : 1
	},
	"ok" : 1
}
```

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -320,4 +325,76 @@ final class MongoCollectionTests: XCTestCase {
...
+    func testFindOneAndUpdate() throws {
+        // test using maxTimeMS
+        let opts1 = FindOneAndUpdateOptions(maxTimeMS: 100)
+        let result1 = try self.coll.findOneAndUpdate(filter: ["cat": "cat"], update: ["$set": ["cat": "blah"] as Document], options: opts1)
+        expect(result1).to(equal(self.doc2))
+        expect(try self.coll.count()).to(equal(2))
+
+        // test using bypassDocumentValidation
+        let opts2 = FindOneAndUpdateOptions(bypassDocumentValidation: true)
+        let result2 = try self.coll.findOneAndUpdate(filter: ["cat": "dog"], update: ["$set": ["cat": "hi"] as Document], options: opts2)
```
jmikola: If the collection doesn't actually have validation rules, this isn't testing much without APM. I suppose this is something I'll end up taking care of when improving the CRUD spec tests.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -320,4 +325,76 @@ final class MongoCollectionTests: XCTestCase {
...
+
+        // test using a write concern
+        let opts2 = FindOneAndDeleteOptions(writeConcern: try WriteConcern(w: .majority))
+        let result2 = try self.coll.findOneAndDelete([:], options: opts2)
+        expect(result2).to(equal(self.doc1))
+        expect(try self.coll.count()).to(equal(0))
+
+        // test invalid maxTimeMS throws error
+        let invalidOpts1 = FindOneAndDeleteOptions(maxTimeMS: 0)
+        let invalidOpts2 = FindOneAndDeleteOptions(maxTimeMS: -1)
```
jmikola: Would it make more sense to validate options when they're first constructed, or is that just be annoying for users since they'd have to `try` both this initialization and the actual executing method. I assume the convention is that this is a simple value container with no rules on its own.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -10,13 +10,17 @@ extension MongoCollection {
...
+     *   - `MongoError.invalidArgument` if any of the provided options are invalid
      *   - `MongoError.commandError` if there are any errors executing the command.
      *   - A `DecodingError` if the deleted document cannot be decoded to a `CollectionType` value
      */
     @discardableResult
-    public func findOneAndDelete(_ filter: Document, options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
-        throw MongoError.commandError(message: "Unimplemented command")
+    public func findOneAndDelete(_ filter: Document,
+                                 options: FindOneAndDeleteOptions? = nil) throws -> CollectionType? {
+        let opts = options ?? FindOneAndDeleteOptions()
```
kmahar: the new tests I added led me to discover a subtle bug I've fixed here. basically, for `findOneAndDelete`, we *always* need to send options because we have to set the `remove` flag to true. I added a test that uses no options and it led to a crash. turns out since you have to always set `remove` or `update`, libmongoc does not account for the possibility of us passing in `nil` for the options. so I think I will refactor the logic in `findAndModify` to reflect that better

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -147,4 +217,113 @@ public struct FindOneAndUpdateOptions: Encodable {
...
+
+        if let wc = writeConcern {
+            do {
+                extra["writeConcern"] = try BsonEncoder().encode(wc) 
+            } catch {
+                throw MongoError.invalidArgument(message: "Error encoding WriteConcern \(wc): \(error)")
+            }
+        }
+
+        if extra.keys.count > 0, !mongoc_find_and_modify_opts_append(self._options, extra.data) {
```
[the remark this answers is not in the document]
jmikola: > swift's source code seems to prefer && in this case.

I concur with using `&&` in this case.

PR #95, opened 2018-07-11
Replace Target() with .target in Package.swift example

jmikola:
`.target()` is consistent with current examples in https://swift.org/package-manager/. This is something I noticed while testing https://github.com/mongodb/mongo-swift-driver/pull/94 for [SWIFT-157](https://jira.mongodb.org/browse/SWIFT-157).

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,9 +65,7 @@ let package = Package(
         .package(url: "https://github.com/mongodb/mongo-swift-driver.git", from: "0.0.2"),
     ],
     targets: [
-        Target(
-            name: "MyPackage",
-            dependencies: ["MongoSwift"])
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,9 +65,7 @@ let package = Package(
         .package(url: "https://github.com/mongodb/mongo-swift-driver.git", from: "0.0.2"),
     ],
     targets: [
-        Target(
-            name: "MyPackage",
-            dependencies: ["MongoSwift"])
+        .target(name: "MyPackage", dependencies: ["MongoSwift"])
```
jmikola: Obviously, a large app may have many dependencies and use line breaks for declaring the `dependencies` array, but I don't see why we can't use one line for our concise example.

Willing to change if needed.

PR #96, opened 2018-07-12
SWIFT-149 MongoCollection.distinct should return a [BsonValue?]

kmahar:


Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -129,23 +126,8 @@ extension MongoCollection {
...
-        // mongoc_cursor_new_from_command_reply will bson_destroy the data we pass in,
-        // so copy it to avoid destroying twice (already done in Document deinit)
-        let fakeData = bson_copy(fakeReply.data)
-        guard let newCursor = mongoc_cursor_new_from_command_reply(client._client, fakeData, 0) else {
-            throw MongoError.invalidResponse()
-        }
-
-        return MongoCursor(fromCursor: newCursor, withClient: client)
+        // if no error, reply will always have a values array
+        return reply["values"] as! [BsonValue?]
```
mbroadst: why `as!` if the method already throws?

kmahar: it seemed like the only situation in where this wouldn't exist is if there is some change in how the server structures the reply document. otherwise if `mongoc_collection_read_command_with_opts` succeeds, we should always expect this field to be present, and be an array. but if you think we should be more defensive, I can throw an error if the cast fails

jmikola: I think strict errors (e.g. `as!`) works well for cases where we want to trust libmongoc's API, but anything involving a server response should be more defensive and fail with exceptions/errors instead of hard assertions.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -173,8 +173,9 @@ final class MongoCollectionTests: XCTestCase {
 
     func testDistinct() throws {
         let distinct = try coll.distinct(fieldName: "cat", filter: [:])
-        expect((distinct.next()?["values"] as? [String])!.sorted()).to(equal(["cat", "dog"]))
-        expect(distinct.next()).to(beNil())
+        expect((distinct as? [String])?.sorted()).to(equal(["cat", "dog"]))
+        // nonexistent field
+        expect(try self.coll.distinct(fieldName: "abc", filter: [:])).to(beEmpty())
```
jmikola: Would it make sense to demonstrate that `null` may be returned as one of several distinct values? IIRC, that was one of the reasons for opening [SWIFT-149](https://jira.mongodb.org/browse/SWIFT-149) in the first place.

PR #101, opened 2018-08-01
SWIFT-110 Hide bson_iter_t and refactor iterator code

kmahar:
- We already had a `DocumentIterator` type wrapping a `bson_iter_t` to support `Document` being a `Sequence`, so I have repurposed it to be a more general wrapper we can use throughout the BSON library anytime we need to iterate, including in the `BsonValue` protocol. 

I cleaned up some other things along the way, like converting a lot of precondition failures throughout the BSON library to errors now that the `BsonValue` method for reading from an iterator can throw. I think this will make it a lot easier when we are reasoning about document validation and where crashes can occur.

### DIFF
--- a/Sources/MongoSwift/BSON/BsonEncoder.swift
+++ b/Sources/MongoSwift/BSON/BsonEncoder.swift
@@ -527,10 +527,12 @@ private class MutableArray: BsonValue {
         try self.array.encode(to: storage, forKey: key)
     }
 
+    init() {}
+
     /// methods required by the BsonValue protocol that we don't actually need/use. MutableArray
     /// is just a BsonValue to simplify usage alongside true BsonValues within the encoder.
-    static func from(iter: inout bson_iter_t) -> BsonValue {
-        fatalError("MutableArray is not meant to be initialized from a bson_iter_t")
+    required init(from iter: DocumentIterator) {
+        fatalError("`MutableArray` is not meant to be initialized from a `DocumentIterator`")
     }
     func encode(to encoder: Encoder) throws {
         fatalError("`MutableArray` is not meant to be encoded with an `Encoder`")
@@ -575,10 +577,12 @@ private class MutableDictionary: BsonValue {
         try self.asDocument().encode(to: storage, forKey: key)
     }
 
+    init() {}
+
     /// methods required by the BsonValue protocol that we don't actually need/use. MutableDictionary
     /// is just a BsonValue to simplify usage alongside true BsonValues within the encoder.
-    static func from(iter: inout bson_iter_t) -> BsonValue {
-        fatalError("`MutableDictionary` is not meant to be initialized from a `bson_iter_t`")
+    required init(from iter: DocumentIterator) {
+        fatalError("`MutableDictionary` is not meant to be initialized from a `DocumentIterator`")
     }
     func encode(to encoder: Encoder) throws {
         fatalError("`MutableDictionary` is not meant to be encoded with an `Encoder`")
--- a/Sources/MongoSwift/BSON/BsonValue.swift
+++ b/Sources/MongoSwift/BSON/BsonValue.swift
@@ -91,7 +91,13 @@ extension Array: BsonValue {
             throw MongoError.bsonDecodeError(message: "Failed to create a bson_t from array data")
         }
 
-       self = Document(fromPointer: arrayData).values as! Array
+        let arrDoc = Document(fromPointer: arrayData)
+
+        guard let arr = arrDoc.values as? Array else {
+            preconditionFailure("Failed to cast values for document \(arrDoc) to array")
+        }
+
+       self = arr
     }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
--- a/Sources/MongoSwift/BSON/BsonValue.swift
+++ b/Sources/MongoSwift/BSON/BsonValue.swift
@@ -2,7 +2,7 @@ import Foundation
 import libbson
 
 /// The possible types of BSON values and their corresponding integer values.
-public enum BsonType: Int {
+public enum BsonType: UInt32 {
     /// An invalid type
     case invalid = 0,
     /// 64-bit binary floating point
@@ -45,37 +45,9 @@ public enum BsonType: Int {
     /// 128-bit decimal floating point
     decimal128,
     /// Special type which compares lower than all other possible BSON element values
-    minKey,
+    minKey = 0xff,
     /// Special type which compares higher than all other possible BSON element values
-    maxKey
-}
-
-internal let BsonTypeMap: [UInt32: BsonValue.Type] = [
-    0x01: Double.self,
-    0x02: String.self,
-    0x03: Document.self,
-    0x04: [BsonValue].self,
-    0x05: Binary.self,
-    0x07: ObjectId.self,
-    0x08: Bool.self,
-    0x09: Date.self,
-    0x0b: RegularExpression.self,
-    0x0c: DBPointer.self,
-    0x0d: CodeWithScope.self,
-    0x0e: Symbol.self,
-    0x0f: CodeWithScope.self,
-    0x10: Int.self,
-    0x11: Timestamp.self,
-    0x12: Int64.self,
-    0x13: Decimal128.self,
-    0xff: MinKey.self,
-    0x7f: MaxKey.self
-]
-
-internal func nextBsonValue(iter: inout bson_iter_t) -> BsonValue? {
-    let type = bson_iter_type(&iter)
-    guard let typeToReturn = BsonTypeMap[type.rawValue] else { return nil }
-    return typeToReturn.from(iter: &iter)
+    maxKey = 0x7f
 }
 
 /// A protocol all types representing BsonTypes must implement.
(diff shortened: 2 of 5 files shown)


Sources/MongoSwift/BSON/BsonEncoder.swift
```diff
@@ -522,10 +522,12 @@ private class MutableArray: BsonValue {
         try self.array.encode(to: storage, forKey: key)
     }
 
+    init() {}
```
jmikola: Is this required because `required init(from iter: DocumentIterator)` was added below and Swift would otherwise no longer create an empty initializer for us?

Sources/MongoSwift/BSON/BsonEncoder.swift
```diff
@@ -522,10 +522,12 @@ private class MutableArray: BsonValue {
...
+
     /// methods required by the BsonValue protocol that we don't actually need/use. MutableArray
     /// is just a BsonValue to simplify usage alongside true BsonValues within the encoder.
-    static func from(iter: inout bson_iter_t) -> BsonValue {
-        fatalError("MutableArray is not meant to be initialized from a bson_iter_t")
+    required init(from iter: DocumentIterator) {
+        fatalError("`MutableArray` is not meant to be initialized from a `DocumentIterator`")
```
jmikola: Does it make sense to format classes/types with backticks in error strings? (in contrast to doc blocks, which actually get rendered).

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -45,37 +45,9 @@ public enum BsonType: Int {
     /// 128-bit decimal floating point
     decimal128,
     /// Special type which compares lower than all other possible BSON element values
-    minKey,
+    minKey = 0xff,
```
jmikola: Since BsonTypeMap in its old form has been removed, would it make sense to explicitly define the enum values for all cases here? That would make it easier to casually glance at a case in this enum and immediately know what its code is, instead of needing to count forward from the initial case (i.e. `0x00`).

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -93,57 +65,36 @@ public protocol BsonValue {
...
         let arrayDoc = Document(fromPointer: arrayData)
 
-        var i = 0
-        var result = [BsonValue]()
-        while let v = arrayDoc[String(i)] {
-            result.append(v)
-            i += 1
-        }
-        return result
+        let arr: [BsonValue?] = (0..<arrayDoc.count).map { arrayDoc[String($0)] }
```
jmikola: While a well-formed BSON array will have indexes starting with "0", it's possible for BSON data to violate that but still be considered valid. See: https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/array.json

A robust/tolerant driver should iterate through the BSON array type and disregard possibly invalid keys.

As an aside, how does the Swift driver currently pass the aforementioned corpus tests? Would that bypass this bit of code?

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -277,51 +231,56 @@ extension Date: BsonValue {
...
 /// An internal struct to represent the deprecated DBPointer type. While DBPointers cannot
 /// be created, we may need to parse them into `Document`s, and this provides a place for that logic.
 internal struct DBPointer: BsonValue {
 
-    public var bsonType: BsonType { return .dbPointer }
+    var bsonType: BsonType { return .dbPointer }
 
-    public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        throw MongoError.bsonEncodeError(message: "DBPointers are deprecated; use a DBRef instead")
+    func encode(to storage: DocumentStorage, forKey key: String) throws {
```
jmikola: By removing the visibility here and on `var bsonType` above, do both of these now default to `internal` as well? If so, how would they be called?

Likewise a question for Symbol below.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -277,51 +231,56 @@ extension Date: BsonValue {
...
+                retrieveErrorMsg(type: "DBPointer ObjectId", key: iter.currentKey))
         }
         guard let collectionP = collectionPP.pointee else {
-            preconditionFailure(retrieveErrorMsg(type: "DBPointer collection name", key: String(cString: key)))
+            throw MongoError.bsonDecodeError(message:
+                retrieveErrorMsg(type: "DBPointer collection name", key: iter.currentKey))
         }
 
-        let dbRef: Document = [
+        return [
```
jmikola: Since the return type is a Document, does Swift automatically initialize one from this returned dictionary literal?

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -365,22 +327,26 @@ public struct Decimal128: BsonValue, Equatable, Codable {
...
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         if !bson_append_double(storage.pointer, key, Int32(key.count), self) {
             throw bsonEncodeError(value: self, forKey: key)
         }
     }
 
-    public static func from(iter: inout bson_iter_t) -> BsonValue {
-        return bson_iter_double(&iter)
+    public init(from iter: DocumentIterator) throws {
+        self = bson_iter_double(&iter.iter)
```
jmikola: Why isn't `self.init()` used here, as with other types? Likewise a question for Int and other primitive types below.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -483,32 +455,40 @@ public struct CodeWithScope: BsonValue, Equatable, Codable {
...
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         if !bson_append_minkey(storage.pointer, key, Int32(key.count)) {
             throw bsonEncodeError(value: self, forKey: key)
         }
     }
 
-    public static func from(iter: inout bson_iter_t) -> BsonValue { return MinKey() }
+    public init() {}
+
+    public init(from iter: DocumentIterator) throws { self.init() }
```
jmikola: There should be no reason for this to throw, since `init()` doesn't.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -483,32 +455,40 @@ public struct CodeWithScope: BsonValue, Equatable, Codable {
...
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         if !bson_append_maxkey(storage.pointer, key, Int32(key.count)) {
             throw bsonEncodeError(value: self, forKey: key)
         }
     }
 
-    public static func from(iter: inout bson_iter_t) -> BsonValue { return MaxKey() }
+    public init() {}
+
+    public init(from iter: DocumentIterator) throws { self.init() }
```
jmikola: There should be no reason for this to throw, since `init()` doesn't.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -553,11 +533,11 @@ public struct ObjectId: BsonValue, Equatable, CustomStringConvertible, Codable {
         }
     }
 
-    public static func from(iter: inout bson_iter_t) -> BsonValue {
-        guard let oid = bson_iter_oid(&iter) else {
-            preconditionFailure("Failed to retrieve ObjectID value")
+     public init(from iter: DocumentIterator) throws {
```
jmikola: Extra whitespace before `public` here.

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -674,53 +654,53 @@ struct RegularExpression: BsonValue, Equatable, Codable {
...
-
         return String(cString: strValue)
     }
 }
 
 /// A struct to represent the BSON Timestamp type.
 public struct Timestamp: BsonValue, Equatable, Codable {
-    public var bsonType: BsonType { return .timestamp }
 
+    public var bsonType: BsonType { return .timestamp }
```
jmikola: Should this have a newline after it for consistency with other classes?

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -0,0 +1,121 @@
...
+
+/// An iterator over the values in a `Document`. 
+public class DocumentIterator: IteratorProtocol {
+    /// the libbson iterator. it must be a `var` because we use it as
+    /// an inout argument
+    internal var iter: bson_iter_t
+
+    /// Initializes a new iterator over the contents of `doc`. `doc` must remain alive for 
+    /// the lifetime of the iterator. Returns `nil` if the key is not found, or if an iterator 
+    /// cannot be created over `doc` due to an error from e.g. corrupt data.
```
jmikola: What happens if `doc` does not remain alive for the lifetime of the iterator?

I imagine there's a risk we'd have dangling pointers and could crash trying to iterate further. Would it make sense to actually save a reference to it here and prevent it from being freed due to reference counting?

Similarly, that might avoid in-place updates where Document decides to duplicate its internal storage if there are multiple references to it.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /// Returns the current value. Assumes the iterator is in a valid position.
+    internal var currentValue: BsonValue? {
+        do {
+            switch self.currentType {
+            case .symbol:
+                return try Symbol.asString(from: self)
+            case .dbPointer:
+                return try DBPointer.asDocument(from: self)
```
jmikola: Ah, is this why the DBPointer and Symbol initializers can be internal? If so, aren't those initializers still throwing each time?

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Returns the current value's type. Assumes the iterator is in a valid position.
+    internal var currentType: BsonType {
+        return BsonType(rawValue: bson_iter_type(&iter).rawValue) ?? .invalid
+    }
+
+    /// Returns the keys from the iterator's current position to the end.
+    internal var keys: [String] {
+        var keys = [String]()
+        while self.advance() { keys.append(self.currentKey) }
```
jmikola: Is this accessor supposed to permanently advance the iterator, or should you revert to the previous position before returning? Likewise for `values` below.

Sources/MongoSwift/BSON/BsonValue.swift
jmikola: > so basically we only check the case for languages without a native (non-BSON data) representation since we're wrapping a `bson_t`.

Good point. I didn't consider that those tests were mainly for round-tripping data to a native language type and back to BSON. In that case, we would expect the round-tripping to correctly reset the array keys starting with "0".

> what about a skipped index? what if they keys are out of order?

I think the common rule is that when reading a BSON array into a native array type, the BSON keys should be entirely disregarded. And when writing a native type back to BSON we should start with "0" and go from there.

AFAIK, the server doesn't validate BSON array keys. When accessing BSON array indices in the query language, `array.1` simply looks up the "1" key in the BSON array, which is no different from a document apart from the type designation byte. I _would_ expect different behavior when using array operators, such as `$push`, as those do attempt to work with the BSON array as a proper list/vector.

----

Let's bring the question back to the LOC I'm commenting on. Is this BSON data coming from an untrusted source with potentially bad keys? If so, should we switch to using `append()` to build the Swift array instead of replying on `map()`?

Sources/MongoSwift/BSON/BsonValue.swift
jmikola: Ah, an `internal` class could never be used outside of the module itself, so I suppose it makes sense that its methods could never use a more public visibility (although you could certainly get more restrictive like `private` or `fileprivate`).

Sources/MongoSwift/BSON/Document+Sequence.swift
jmikola: If these are internal and only used on a disposable iterator, I don't think there's a problem. Perhaps a comment mentioning that they leave the iterator exhausted would be helpful, though.

I think resetting the iterator would entail stashing the current key and then later resetting the position back to the original key; although, documents can have duplicate keys so that's probably not fool-proof.

Sources/MongoSwift/BSON/BsonValue.swift
kmahar: it could be coming from a bad source, yes. I think we could just return `arrayDoc.values`, which will make an iterator and call `values` on that, which just goes through the doc sequentially, ignoring keys.

Sources/MongoSwift/BSON/BsonValue.swift
kmahar: yup 👍

PR #102, opened 2018-08-02
SWIFT-189 Add server version method

kmahar:
As part of the CRUD test updates we want to filter tests by current server version. This adds an extension within our tests to get the version string for a client, which we can compare to `minServerVersion` and `maxServerVersion` fields.


kmahar:
I will just address it there and close this PR

PR #104, opened 2018-08-02
SWIFT-162 Allow for null document IDs in inserted and upserted documents

kmahar:
https://jira.mongodb.org/browse/SWIFT-162

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -83,9 +83,8 @@ extension MongoCollection {
             self._collection, filter.data, replacementDoc.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-        let res = try BsonDecoder().decode(UpdateResult.self, from: reply)
-        return res
-        //return try BsonDecoder().decode(UpdateResult.self, from: reply)
+
+        return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
     /**
@@ -109,9 +108,8 @@ extension MongoCollection {
             self._collection, filter.data, update.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-        //return try BsonDecoder().decode(UpdateResult.self, from: reply)
-        let res = try BsonDecoder().decode(UpdateResult.self, from: reply)
-        return res
+
+        return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
     /**
@@ -135,9 +133,8 @@ extension MongoCollection {
             self._collection, filter.data, update.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-        //return try BsonDecoder().decode(UpdateResult.self, from: reply)
-        let res = try BsonDecoder().decode(UpdateResult.self, from: reply)
-        return res
+
+        return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
     /**
@@ -160,7 +157,7 @@ extension MongoCollection {
             self._collection, filter.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-        return DeleteResult(from: reply)
+        return try BsonDecoder().decode(DeleteResult.self, from: reply)
     }
 
     /**
@@ -183,7 +180,7 @@ extension MongoCollection {
             self._collection, filter.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-        return DeleteResult(from: reply)
+        return try BsonDecoder().decode(DeleteResult.self, from: reply)
     }
 }
 
@@ -308,17 +305,9 @@ public struct InsertManyResult {
 }
 
 /// The result of a `delete` command on a `MongoCollection`. 
-public struct DeleteResult {
+public struct DeleteResult: Decodable {
     /// The number of documents that were deleted.
     public let deletedCount: Int
-
-    /// Given a server response to a delete command, creates a corresponding
-    /// `DeleteResult`. If the `from` Document does not have a `deletedCount`
-    /// field, the initialization will fail.
-    internal init?(from: Document) {
-        guard let deletedCount = from["deletedCount"] as? Int else { return nil }
-        self.deletedCount = deletedCount
-    }
 }
 
 /// The result of an `update` operation a `MongoCollection`.
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -54,8 +54,8 @@ extension MongoCollection {
         let reply = Document()
         var error = bson_error_t()
         if !mongoc_collection_insert_many(
-             // TODO SWIFT-139: include writeErrors and writeConcernErrors from reply in the error
             self._collection, &docPointers, documents.count, opts?.data, reply.data, &error) {
+            // TODO SWIFT-139: include writeErrors and writeConcernErrors from reply in the error
             throw MongoError.commandError(message: toErrorString(error))
         }
 
@@ -114,8 +114,8 @@ extension MongoCollection {
         let reply = Document()
         var error = bson_error_t()
         if !mongoc_collection_update_one(
-            // TODO SWIFT-139: include writeErrors and writeConcernError from reply in the error
             self._collection, filter.data, update.data, opts?.data, reply.data, &error) {
+            // TODO SWIFT-139: include writeErrors and writeConcernError from reply in the error
             throw MongoError.commandError(message: toErrorString(error))
         }
         return try BsonDecoder().decode(UpdateResult.self, from: reply)
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -56,7 +56,6 @@ extension MongoCollection {
             self._collection, &docPointers, documents.count, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-
         return InsertManyResult(insertedIds: documents.map { $0["_id"] })
     }
 
@@ -83,7 +82,6 @@ extension MongoCollection {
             self._collection, filter.data, replacementDoc.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-
         return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
@@ -108,7 +106,6 @@ extension MongoCollection {
             self._collection, filter.data, update.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-
         return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
@@ -133,7 +130,6 @@ extension MongoCollection {
             self._collection, filter.data, update.data, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-
         return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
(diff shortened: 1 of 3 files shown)


Tests/Specs/crud/tests/write/updateMany-arrayFilters.json
```diff
@@ -97,7 +98,8 @@
       "outcome": {
         "result": {
           "matchedCount": 2,
-          "modifiedCount": 1
+          "modifiedCount": 1,
+          "upsertedCount": 0
```
kmahar: this line isn't actually here in the new file but I think @jmikola meant to add it along with adding it to all the other tests.
I resynced these because it was necessary to have the new field be present in order to decode to `UpdateResult`s in`CrudTests`

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -290,60 +290,32 @@ public struct DeleteOptions: Encodable {
...
-    /// Given an ordered array of insertedIds, creates a corresponding InsertManyResult.
-    internal init(fromArray arr: [BsonValue]) {
-        var inserted = [Int64: BsonValue]()
-        for (i, id) in arr.enumerated() {
-            let index = Int64(i)
-            inserted[index] = id
-        }
-        self.insertedIds = inserted
-    }
+    public let insertedIds: [BsonValue?]
```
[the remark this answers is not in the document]
mbroadst: My guess is because `insertMany` is generally implemented as a bulk operation, and bulk operations can have multiple operations within. The map tells you which of these operations is associated with the inserted id. I don't feel very strongly which we way we go here, but @jmikola may feel differently.

Tests/Specs/crud/tests/write/updateMany-arrayFilters.json
jmikola: I've since merged https://github.com/mongodb/specifications/pull/384 if you want to properly resync in a separate commit or PR.

When doing so, I find it's helpful to include the commit hash from the specs repo in the extended commit description.

Sources/MongoSwift/MongoCollection+Write.swift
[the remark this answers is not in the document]
jmikola: I vote for keeping this a map in accordance with the spec.

Consider a case where `insertMany` uses `ordered: false` and some documents in the middle of the requests fail. In that case, a map is necessary to know exactly which inserted IDs correspond to which documents in the original input.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -291,60 +291,32 @@ public struct DeleteOptions: Encodable {
...
-            return nil
-         }
-         self.matchedCount = matched
-         self.modifiedCount = modified
-         self.upsertedId = from["upsertedId"]
-    }
+    public let upsertedId: AnyBsonValue?
+
+    /// The number of documents that were upserted.
+    public let upsertedCount: Int
```
jmikola: Does libmongoc actually populate an `upsertedCount` field? I don't see it mentioned in http://mongoc.org/libmongoc/current/mongoc_collection_update_one.html#returns.

PHPC only uses libmongoc's bulk API (see: [WriteResult.c](https://github.com/mongodb/mongo-php-driver/blob/master/src/MongoDB/WriteResult.c)), so we have more fields to work with. In the event that `upsertedCount` isn't available, I think you'll need to infer it yourself based on whether the result document contains an `upsertedId` field. And `upsertedCount` in your result object is necessary for users to differentiate a `null` upserted ID from `upsertedId` not being set (i.e. no upsert occurred).

.travis.yml
```diff
@@ -25,7 +25,7 @@ install:
   - eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"
   
   # clone and build libmongoc
-  - git clone -b r1.10 https://github.com/mongodb/mongo-c-driver /tmp/libmongoc
+  - git clone -b r1.12 https://github.com/mongodb/mongo-c-driver /tmp/libmongoc
```
jmikola: Should this be in its own commit (related to bumping the dependency elsewhere)? It doesn't look like something that should end up in a single squashed commit from merging this PR.

kmahar: Sure, I will bump it in a separate commit to master.

kmahar: [done](https://github.com/mongodb/mongo-swift-driver/commit/57a799eef066a37371eba47874d14de6547cde69)

Tests/Specs/crud/tests/write/updateMany-arrayFilters.json
kmahar: ok, I resynced all the CRUD files (minus those for the new count API) in a separate commit [here](https://github.com/mongodb/mongo-swift-driver/commit/6e3adbdf80ce6bc26d1783d041b1cd0cbbda4c02).

Sources/MongoSwift/MongoCollection+Write.swift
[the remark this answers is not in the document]
kmahar: hmm, would we even get to that point? I think we would have already thrown an error because `mongoc_collection_insert_many` would return false. 
that said, I don't think we are correctly propagating writeErrors / writeConcernErrors in that case, and maybe we should also be sticking insertedIds / the `InsertManyResult` into the error?

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: while it is not mentioned in the docs, the reply I get back libmongoc looks like `{ "modifiedCount" : 1, "matchedCount" : 1, "upsertedCount" : 0 }`. 

Jesse's commit message [here](https://github.com/mongodb/mongo-c-driver/commit/84d7dda4232f102fec5b090b7a863df6139fa480) says it was added to mongoc_collection_update_one, update_many, and replace_one.

Sources/MongoSwift/MongoCollection+Write.swift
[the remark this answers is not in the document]
jmikola: Correct to your last point. Exceptions for bulk writes (`insertMany()` and `bulkWrite()`) should include a write result, so it'd make sense to create the proper `insertedIds` map to inject therein.

I'll point out that while the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) does say both of those methods throw a BulkWriteException, there is only one definition of that class in the spec and it includes a `writeResult` field as an optional BulkWriteResult. This is likely a failing of the spec, as it doesn't make sense for `insertMany` to throw something with the full result object from a `bulkWrite`. In that case, `writeResult` should be an InsertManyResult. Correcting this will likely require adding a new InsertManyException class to the spec. Mind creating a SPEC ticket for that with DCN: Maybe?

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: PR opened: https://github.com/mongodb/mongo-c-driver/pull/527

Sources/MongoSwift/MongoCollection+Write.swift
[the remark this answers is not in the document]
kmahar: I've created [SPEC-1147](https://jira.mongodb.org/browse/SPEC-1147), feel free to edit.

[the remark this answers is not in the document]
kmahar: also, I've gone back to using an `[Int: BsonValue?]` in `InsertManyResult`. I am also adding a note to throw a better error from here (and various other write methods) in SWIFT-139, once we've merged in your bulkWrite changes that include the structs like `WriteError` and `WriteConcernError` that we'll need to include in this error.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -53,10 +54,18 @@ extension MongoCollection {
         let reply = Document()
         var error = bson_error_t()
         if !mongoc_collection_insert_many(
+             // TODO SWIFT-139: include writeErrors and writeConcernErrors from reply in the error
```
jmikola: I think this comment should be moved down a line (above `throw`), as you did for `insert_one`.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -53,10 +54,18 @@ extension MongoCollection {
...
             self._collection, &docPointers, documents.count, opts?.data, reply.data, &error) {
             throw MongoError.commandError(message: toErrorString(error))
         }
-        return InsertManyResult(insertedIds: documents.map { $0["_id"] })
+
+        // if we succeeded, then we should store + return every `_id`
```
jmikola: Interesting. Since [`mongoc_collection_insert_many()`](http://mongoc.org/libmongoc/current/mongoc_collection_insert_many.html) doesn't report `insertedIds`, we might actually have to interpret the errors to determine how to (a) construct the map or (b) redact failed inserts from the map (assuming we construct it before executing the insert).

I know that PHPLIB reports inserted IDs in its InsertManyResult class. Looking at [the source](https://github.com/mongodb/mongo-php-library/blob/1.4.2/src/Operation/InsertMany.php#L142), it appears we don't report the map if an exception is thrown. In that case, we let the exception from [`executeBulkWrite()`](http://php.net/manual/en/mongodb-driver-manager.executebulkwrite.php) propagate. That exception will include a PHPC [WriteResult](http://php.net/manual/en/class.mongodb-driver-writeresult.php), which is derived from libmongoc's reply document and does not include a map of inserted IDs.

Given the complexity of building this map after encountering an error (mentioned in the first paragraph of this comment), I'm not sure it's worth the trouble. I might suggest poking around other drivers to see how they handle this before expending the effort to implement the TODO item above.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -103,6 +114,7 @@ extension MongoCollection {
         let reply = Document()
         var error = bson_error_t()
         if !mongoc_collection_update_one(
+            // TODO SWIFT-139: include writeErrors and writeConcernError from reply in the error
```
jmikola: Move this down a line.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -430,7 +430,7 @@ final class MongoCollectionTests: XCTestCase {
 
         let result2 = try self.coll.insertMany([["_id": nil], ["_id": 20]])
         expect(result2).toNot(beNil())
-        expect(result2?.insertedIds[0]).to(beNil())
+        expect(result2?.insertedIds[0]!).to(beNil())
```
jmikola: Why do you use `!` here but not on the line below? Should they be consistent?

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: yup, good catch.

Tests/MongoSwiftTests/MongoCollectionTests.swift
kmahar: This is an interesting thing I discovered yesterday that arises when your dictionary values are an optional type. Basically, a dictionary has an `associatedtype Value`, and the return type of the subscript getter is `Value?`. In the case of a dictionary where `Value` is already an optional type, that means the subscript getter returns a double-optional `??`. So because `insertedIds` is a `[String: BsonValue?]`, the return type of the subscript getter is double-optional, a `BsonValue??`. Essentially, if a key is present but its value is set to nil, you get back `Optional(nil)`. If a key is not present, you get back just plain `nil`. 

So in this line, I'm expecting an `Optional(nil)`, as the key should be present but set to nil.  But `isNil` is actually checking for a true `nil`, and `Optional(nil) != nil`. This is not is not really intuitive, because for all other Swift types, `Optional(x) == x` evaluates to `true`, as is the case in the following line where I compare an `Int?` to an `Int`. 

I alluded to this in [SWIFT-193](https://jira.mongodb.org/browse/SWIFT-193) yesterday as it could be a way to solve our issue of users being unable to differentiate between a key that isn't present and a key that's set to `nil`, though I think it has some other usability tradeoffs. here's an old but still pretty relevant [blog post](https://developer.apple.com/swift/blog/?id=12).

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: yes, building it does seem like it will be difficult, so maybe in the case of an error in `insertMany` or `bulkWrite` we should leave `insertedIds` out... either way, we can propagate the writeError / writeConcern error fields mentioned in the to-do.

PR #106, opened 2018-08-06
SWIFT-94 Filter tests by current server version

kmahar:
This builds on the work in SWIFT-189 to add a server version method to `MongoClient`. 

Basically, for the server version dependent tests we now get the version of the server and compare it to the min/max in the files. 
While I was in there I also did some code cleanup to get rid of unneeded try/catches, simplify file parsing by using a `BsonDecoder`, etc.

### DIFF
--- a/Tests/MongoSwiftTests/CommandMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/CommandMonitoringTests.swift
@@ -4,9 +4,6 @@ import XCTest
 
 let center = NotificationCenter.default
 
-// TODO: don't hardcode this
-let VERSION = "3.6"
-
 final class CommandMonitoringTests: XCTestCase {
     static var allTests: [(String, (CommandMonitoringTests) -> () throws -> Void)] {
         return [
@@ -20,6 +17,7 @@ final class CommandMonitoringTests: XCTestCase {
     }
 
     func testCommandMonitoring() throws {
+        let decoder = BsonDecoder()
         let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
         client.enableMonitoring(forEvents: .commandMonitoring)
 
@@ -34,7 +32,7 @@ final class CommandMonitoringTests: XCTestCase {
 
             let testFilePath = URL(fileURLWithPath: "\(cmPath)/\(filename)")
             let asDocument = try Document(fromJSONFile: testFilePath)
-            let testFile = try CMTestFile(fromDocument: asDocument)
+            let testFile = try decoder.decode(CMTestFile.self, from: asDocument)
 
             print("-----------------------")
             print("Executing tests for file \(name)...\n")
@@ -42,10 +40,8 @@ final class CommandMonitoringTests: XCTestCase {
             // execute the tests for this file
             for var test in testFile.tests {
 
-                // Check that the current version we're testing is within bounds for this cases
-                if let maxVersion = test.maxServerVersion, maxVersion < VERSION {
-                    continue
-                } else if let minVersion = test.minServerVersion, VERSION < minVersion {
+                if try !client.serverVersionIsInRange(test.minServerVersion, test.maxServerVersion) {
+                    print("Skipping test case \(test.description) for server version \(try client.serverVersion())")
                     continue
                 }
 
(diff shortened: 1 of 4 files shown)


Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -34,20 +32,15 @@ final class CommandMonitoringTests: XCTestCase {
...
             // execute the tests for this file
             for var test in testFile.tests {
 
-                // Check that the current version we're testing is within bounds for this cases
-                if let maxVersion = test.maxServerVersion, maxVersion < VERSION {
-                    continue
-                } else if let minVersion = test.minServerVersion, VERSION < minVersion {
-                    continue
-                }
+                if try !client.serverVersionIsInRange(test.minServerVersion, test.maxServerVersion) { continue }
```
mbroadst: prefer some sort of output indicating that a test was skipped rather than silently continuing.

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -18,6 +18,23 @@ extension XCTestCase {
     }
 }
 
+extension MongoClient {
+    internal func getServerVersion() throws -> String {
```
mbroadst: nit: prefer dropping `get`

Tests/MongoSwiftTests/CommandMonitoringTests.swift
jmikola: I think I already asked this, but XCTest has no notion of skipped tests, correct?

Is more custom output (thinking of the noisy CRUD tests) actually useful, given that we really have no way to meaningfully interpret it during CI? Perhaps this is something to put off until we can adopt a more full-featured test framework.

Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -104,71 +97,72 @@ final class CommandMonitoringTests: XCTestCase {
...
         // TODO SWIFT-31: use readPreferences for commands if provided
-        let filter = self.args["filter"] as? Document
-        switch self.operationName {
+        let filter: Document = self.op.args["filter"] as? Document ?? [:]
+
+        switch self.op.name {
 
         case "count":
-            do { _ = try collection.count(filter ?? [:]) } catch { }
+            _ = try? collection.count(filter)
```
jmikola: Is the `_` assignment necessary just so we can use `try?` and avoid a verbose `try`/`catch` block?

Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -182,38 +176,38 @@ private struct CMTest {
...
                                         returnKey: modifiers?["$returnKey"] as? Bool,
                                         showRecordId: modifiers?["$showDiskLoc"] as? Bool,
-                                        skip: self.args["skip"] as? Int64,
-                                        sort: self.args["sort"] as? Document)
+                                        skip: self.op.args["skip"] as? Int64,
+                                        sort: self.op.args["sort"] as? Document)
 
             // we have to iterate the cursor to make the command execute
-            do { for _ in try collection.find(filter ?? [:], options: options) {} } catch { }
+            for _ in try! collection.find(filter, options: options) {}
```
jmikola: Is full iteration necessary because the tests expect both an initial `find` and `getMore` command(s)? If not, would it make sense to only advance once to ensure that the initial `find` is executed and avoid anything beyond that?

Tests/MongoSwiftTests/CommandMonitoringTests.swift
kmahar: yeah, AFAIK there is no easy way to indicate a test should be skipped in XCTest. 

I agree that output is a good idea, though yes, it would be nice if it were more meaningful in CI. unfortunately testing from the command ljne in Swift is kind of garbage... when we explored other test frameworks like Quick a while back we ended up just coming back to XCTest, lol.

Tests/MongoSwiftTests/CommandMonitoringTests.swift
kmahar: yup exactly

Tests/MongoSwiftTests/CommandMonitoringTests.swift
kmahar: yup, for example [this](https://github.com/mongodb/specifications/blob/4f212acb5154a3c3d124525ad1f0f308efa93ef8/source/command-monitoring/tests/find.json#L170) test case expects a `getMore`.

kmahar:
Status on this: Matt was LGTM with a few changes which I've since made, waiting for review from Jeremy.

Tests/MongoSwiftTests/CommandMonitoringTests.swift
jmikola: What happens with this omitted? Is there a compile error saying that we're discarding the return value?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -18,6 +18,23 @@ extension XCTestCase {
     }
 }
 
+extension MongoClient {
+    internal func serverVersion() throws -> String {
+        let buildInfo = try self.db("admin").runCommand(["buildInfo": 1])
```
jmikola: Should this hard-code a primary read preference, just in case the client is using a different default read preference (e.g. `secondaryPreferred`)?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -18,6 +18,23 @@ extension XCTestCase {
...
+        guard let versionString = buildInfo["version"] as? String else {
+            throw TestError(message: "buildInfo reply missing version string: \(buildInfo)")
+        }
+        return versionString
+    }
+
+    internal func serverVersionIsInRange(_ min: String?, _ max: String?) throws -> Bool {
+        let version = try self.serverVersion()
+        if let min = min, min > version { return false }
+        if let max = max, max < version { return false }
```
jmikola: Is this robust enough for comparing version strings? I'm thinking that basic string comparison logic could be problematic. For instance, "3.6.4" should be considered smaller than "3.6.10".

It may be better to split the string on `.` to extract X.Y.Z and then compare them independently. Any non-numeric content after Z can likely be ignored (e.g. "beta", a hyphen).

Context: https://stackoverflow.com/questions/27932408/compare-two-version-strings-in-swift

PR #107, opened 2018-08-08
SWIFT-140 When UUID binary subtype is provided, validate that data is 16 bytes

kmahar:
https://jira.mongodb.org/browse/SWIFT-140

### DIFF
--- a/Sources/MongoSwift/BSON/BsonValue.swift
+++ b/Sources/MongoSwift/BSON/BsonValue.swift
@@ -140,28 +140,36 @@ public struct Binary: BsonValue, Equatable, Codable {
     }
 
     /// Initializes a `Binary` instance from a `Data` object and a `UInt8` subtype.
-    public init(data: Data, subtype: UInt8) {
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -140,28 +140,36 @@ public struct Binary: BsonValue, Equatable, Codable {
     }
 
     /// Initializes a `Binary` instance from a `Data` object and a `UInt8` subtype.
-    public init(data: Data, subtype: UInt8) {
+    /// Throws an error if the provided data is incompatible with the specified subtype.
+    public init(data: Data, subtype: UInt8) throws {
```
jmikola: Just want to confirm that this is only the code path for manually constructing a Binary (not decoding BSON into a Binary). If so, LGTM.

kmahar: yup, all the other `init`s get routed through here in the end.

PR #108, opened 2018-08-08
SWIFT-158 Add filter and mapValues methods to Document

kmahar:
This adds a `mapValues` method similar to [Dictionary's](https://developer.apple.com/documentation/swift/dictionary/2995348-mapvalues).

This also overrides the default `filter` implementation (which we got from conforming to the `Sequence` protocol).
Before the signature was 
```swift
public func filter(_ isIncluded: (Element) throws -> Bool) rethrows -> [Element]`
```
and now it's
```swift
public func filter(_ isIncluded: (Element) throws -> Bool) rethrows -> Document
```

We have default `map` and `reduce` implementations already from conforming to `Sequence`.

I also added `isEmpty` to go along with `count`, since it seems to be more idiomatic.

I'm going to dig further through the `Dictionary` docs and source code and potentially open more tickets to see if there are any other methods like this, either to add or override, that could improve usability/similarity to dictionaries and help alleviate append-only issues.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -20,10 +20,7 @@ public class DocumentStorage {
     }
 }
 
-/// A struct representing the BSON document type. `Document`s are append-only and do not support
-/// overwriting or removing keys in-place, or provide safeguards against duplicate keys. 
-/// To transform documents, use `Sequence` protocol methods along with `filter` and `mapValues`.
-/// To check for existing keys, use the `.keys` property.
+/// A struct representing the BSON document type.
 public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLiteral {
     /// the storage backing this document 
     internal var storage: DocumentStorage
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -20,8 +20,15 @@ public class DocumentStorage {
     }
 }
 
-/// A struct representing the BSON document type.
+/// A struct representing the BSON document type. `Document`s are append-only and do not support
+/// overwriting or removing keys in-place, or provide safeguards against duplicate keys. 
+/// To transform documents, use `Sequence` protocol methods along with `filter` and `mapValues`.
+/// To check for existing keys, use the `.keys` property.
 public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLiteral {
+
+    /// The element type of a document: a tuple containing an individual key-value pair.
+    public typealias Element = (key: String, value: BsonValue?)
+
     /// the storage backing this document 
     internal var storage: DocumentStorage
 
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -20,8 +20,15 @@ public class DocumentStorage {
...
 
-/// A struct representing the BSON document type.
+/// A struct representing the BSON document type. `Document`s are append-only and do not support
+/// overwriting or removing keys in-place, or provide safeguards against duplicate keys. 
+/// To transform documents, use `Sequence` protocol methods along with `filter` and `mapValues`.
+/// To check for existing keys, use the `.keys` property.
 public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLiteral {
+
+    /// The element type of a document: a tuple containing an individual key-value pair.
+    public typealias Element = (key: String, value: BsonValue?)
```
jmikola: Is an "Element" type alias a Swift convention?

kmahar: so the `Sequence` protocol in Swift has an `associatedtype Element` requirement ([docs](https://developer.apple.com/documentation/swift/sequence)). 
The compiler was actually inferring this already based on the type we return from `DocumentIterator.next()`, but inspired by the source code for Swift `Dictionary` etc., I think explicitly declaring it is nice for clarity, and also for giving us a shorthand way to refer to this specific tuple type.

kmahar:
@mbroadst can you take a look at this soon? I would like to use `mapValues` for SWIFT-196

kmahar:
now that I've started working on

kmahar:
oops, meant to just post a comment. I was going to say, now that I think about it, we should move these methods into Document+Sequence. `filter` is overriding the default impl provided by `Sequence`, and `mapValues` feels like it should go with that.
plus, as part of [SWIFT-195](https://jira.mongodb.org/browse/SWIFT-195) work, I will be adding a bunch of similar methods in Document+Sequence: prefix, suffix, dropFirst, dropLast, etc.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -20,7 +20,10 @@ public class DocumentStorage {
     }
 }
 
-/// A struct representing the BSON document type.
+/// A struct representing the BSON document type. `Document`s are append-only and do not support
+/// overwriting or removing keys in-place, or provide safeguards against duplicate keys. 
+/// To transform documents, use `Sequence` protocol methods along with `filter` and `mapValues`.
+/// To check for existing keys, use the `.keys` property.
```
jmikola: Would it make sense to use `/** */` here instead of a series of `///`?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -213,13 +218,11 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      *  let x: Int = try d.get("x")
      *  ```
      *
-     *  - Params:
+     *  - Parameters:
```
jmikola: Did you plan on making this change across the project, or are most other doc blocks already using "Parameters" over "Params"?

kmahar:
ok, this was approved by Jeremy and is pending review from Matt. The SWIFT-195 PR is directly on top of this code, so this should be reviewed first.

PR #109, opened 2018-08-13
SWIFT-89 Add failable initializer for Decimal128

kmahar:
This adds a new failable initializer for `Decimal128`, which validates the string by testing `bson_decimal128_t` initialization. 

It seemed to me that it should be an alternate initializer, differentiated by the argument label `ifValid`, so we don't force the performance penalty in unneeded cases.

I thought a failable initializer was more appropriate than throwing (as we do from some other initializers), reasons being that:
1) there is only one very opaque error we would throw from here ("invalid decimal128 string") 
2) this feels more consistent with other Swift methods for parsing types from strings, for example [Int](https://developer.apple.com/documentation/swift/int/2927504-init) and [Data](https://developer.apple.com/documentation/foundation/data/1780388-init).

### DIFF
--- a/Tests/MongoSwiftTests/BsonValueTests.swift
+++ b/Tests/MongoSwiftTests/BsonValueTests.swift
@@ -1,0 +1,30 @@
+import Foundation
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+final class BsonValueTests: XCTestCase {
+    static var allTests: [(String, (BsonValueTests) -> () throws -> Void)] {
+        return [
+            ("testInvalidDecimal128", testInvalidDecimal128),
+            ("testUUIDBytes", testUUIDBytes)
+        ]
+    }
+
+    func testInvalidDecimal128() throws {
+        expect(Decimal128(ifValid: "hi")).to(beNil())
+        expect(Decimal128(ifValid: "123.4.5")).to(beNil())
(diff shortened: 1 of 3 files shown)


kmahar:
Also, it felt like I was starting to put stuff that wasn't really related to documents at all in `DocumentTests` so I started splitting out general BSON library things into a new file

Tests/MongoSwiftTests/BsonValueTests.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+final class BsonValueTests: XCTestCase {
```
jmikola: Per my other comment, I think it'd ultimately make sense to split these by BSON type class when test coverage is expanded. I realize most of the successful inits for each type are probably covered in the Document tests.

kmahar:
with the exception of `Document`, none of the `BsonValue` types keep around a reference to a libbson type. we wanted to avoid keeping types like `bson_oid_t` and `bson_decimal128_t` around on the stack for potentially a long time. 

one thing we could do is round trip strings through libbson with `bson_decimal128_to_string` and `bson_decimal128_from_string` to normalize them, either when constructing them (though that adds the performance penalty I was trying to mitigate), or just at the time we compare them.

PR #110, opened 2018-08-14
WIP SWIFT-196: Check if a key exists already before setting it

kmahar:
I started on this but I ran into a bit of an issue trying to preserve copy-on-write behavior. I will add some inline comments to try to clarify, but the gist is:

Basically, I would like to use a `DocumentIterator` (initialized via the init that wraps `bson_iter_init_and_find`) to check if a key is already present in the `Document`, and if so, get its current type. This allows me to check whether it is a type that can be overwritten in place or a type that will require creating a new `bson_t` and copying over values. 

However, creating this iterator, which at least lives for the time I am trying to set the new value, means that the `Document`'s storage is no longer considered uniquely referenced and we will *always* end up making a copy before doing mutation, even if it's not really necessary, because the only other reference to the storage is the iterator that we're done with. 

I do think there is value in having the iterator keep its own reference to the storage, so that we don't need to caution users about modifying a doc while iterating through it, or destroying the doc too early, and so on. but in this case, it's also forcing us into some unneeded inefficiency.

some ideas for mitigating this:
- we could make `DocumentIterator.storage` optional and then add some additional `internal init` that doesn't set it, or provide a method to nil it out when done with the iterator. 
- we could create an alternative `internal` iterator class, that basically has behavior described in the previous bullet point
- we could use some of the convenience methods on `Document` that each create their own locally scoped iterators. this could lead to us creating more `bson_iter_t`s than necessary but it seems to alleviate the unique reference problem. 
- we could also just say screw it, if you're overwriting an existing value, we will always have to use a new`bson_t` and that is that. in which case I'm not even sure the overwriting optimizations I added would be worth it, because we'd always end up doing a `bson_copy` first anyway. 

Not sure how to best handle, let me know your thoughts.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -13,6 +13,10 @@ public class DocumentStorage {
         self.pointer = bson_copy(pointer)
     }
 
+    init(withoutCopying pointer: UnsafeMutablePointer<bson_t>) {
+        self.pointer = pointer
+    }
+
     deinit {
         guard let pointer = self.pointer else { return }
         bson_destroy(pointer)
@@ -31,6 +35,11 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
     /// Returns a `[String]` containing the keys in this `Document`.
     public var keys: [String] {
         return self.makeIterator().keys
+    }
+
+    /// Returns a `Boolean` indicating whether this `Document` contains the provided key.
+    public func hasKey(_ key: String) -> Bool {
+        return bson_has_field(self.data, key)
     }
 
     /// Returns a `[BsonValue?]` containing the values stored in this `Document`.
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -181,27 +186,52 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
-                if !bson_append_null(self.data, key, Int32(key.count)) {
-                    preconditionFailure("Failed to set the value for key \(key) to null")
+    private mutating func setValue(forKey key: String, to newValue: BsonValue?) throws {
+        // if the key already exists in the `Document`, we need to replace it
+
+        if let existingType = DocumentIterator(forDocument: self, advancedTo: key)?.currentType {
+            // if the new type is the same and it's a fixed length type, we can overwrite
+            if let ov = newValue as? Overwritable, ov.bsonType == existingType {
+                // we're about to mutate, so copy if any other docs reference our storage
+                self.copyStorageIfRequired()
```
kmahar: currently, this copy always happens, because the iterator created 2 lines up is still in scope.

kmahar:
lol, never mind to all of this, the reason it wasn't uniquely referenced was that I was trying to keep a reference around to *later* check if the references were still identical. but of course my saving it made it not identical... it seems like the iterator does get cleaned up in time as is so we're fine, will reopen PR once I finish writing tests

PR #111, opened 2018-08-15
SWIFT-196 Check if a key exists before setting it, and overwrite existing keys

kmahar:
Alright, round 2 now that I resolved my unique reference issues.

I moved all the logic for setting a value on a document to a private `setValue` method, please pay attention in particular to the logic there... I think it makes sense re: when we do or don't need to copy underlying storage and overwrite values vs. make new docs, but a sanity check is appreciated.

### DIFF
--- a/Sources/MongoSwift/BSON/BsonValue.swift
+++ b/Sources/MongoSwift/BSON/BsonValue.swift
@@ -322,6 +322,10 @@ public struct Decimal128: BsonValue, Equatable, Codable {
         }
     }
 
+    public static func == (lhs: Decimal128, rhs: Decimal128) -> Bool {
+        return lhs.data == rhs.data
+    }
+
     /// Returns the provided string as a `bson_decimal128_t`, or throws an error if initialization fails due an
     /// invalid string.
     internal static func encode(_ str: String) throws -> bson_decimal128_t {
@@ -332,12 +336,8 @@ public struct Decimal128: BsonValue, Equatable, Codable {
         return value
     }
 
-    public static func == (lhs: Decimal128, rhs: Decimal128) -> Bool {
-        return lhs.data == rhs.data
-    }
-
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-       var value = try Decimal128.encode(self.data)
+        var value = try Decimal128.encode(self.data)
         if !bson_append_decimal128(storage.pointer, key, Int32(key.count), &value) {
             throw bsonEncodeError(value: self, forKey: key)
         }
@@ -378,15 +378,19 @@ extension Double: BsonValue {
 /// The `Int` will be encoded as an Int32 if possible, or an Int64 if necessary.
 extension Int: BsonValue {
 
-    public var bsonType: BsonType { return .int32 }
+    public var bsonType: BsonType { return self.int32Value != nil ? .int32 : .int64 }
+
+    internal var int32Value: Int32? { return Int32(exactly: self) }
+    internal var int64Value: Int64? { return Int64(exactly: self) }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        if let int32 = Int32(exactly: self) {
+        if let int32 = self.int32Value {
             return try int32.encode(to: storage, forKey: key)
         }
-        if let int64 = Int64(exactly: self) {
+        if let int64 = self.int64Value {
             return try int64.encode(to: storage, forKey: key)
         }
+
         throw MongoError.bsonEncodeError(message: "`Int` value \(self) could not be encoded as `Int32` or `Int64`")
     }
 
--- a/Sources/MongoSwift/BSON/BsonValue.swift
+++ b/Sources/MongoSwift/BSON/BsonValue.swift
@@ -322,10 +322,6 @@ public struct Decimal128: BsonValue, Equatable, Codable {
         }
     }
 
-    public static func == (lhs: Decimal128, rhs: Decimal128) -> Bool {
-        return lhs.data == rhs.data
-    }
-
     /// Returns the provided string as a `bson_decimal128_t`, or throws an error if initialization fails due an
     /// invalid string.
     internal static func encode(_ str: String) throws -> bson_decimal128_t {
@@ -334,6 +330,10 @@ public struct Decimal128: BsonValue, Equatable, Codable {
             throw MongoError.bsonEncodeError(message: "Invalid Decimal128 string \(str)")
         }
         return value
+    }
+
+    public static func == (lhs: Decimal128, rhs: Decimal128) -> Bool {
+        return lhs.data == rhs.data
     }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -223,7 +223,7 @@ public class DocumentIterator: IteratorProtocol {
 
     /// Returns the current value's type. Assumes the iterator is in a valid position.
     internal var currentType: BsonType {
-        return BsonType(rawValue: bson_iter_type(&iter).rawValue) ?? .invalid
+        return BsonType(rawValue: bson_iter_type(&self.iter).rawValue) ?? .invalid
     }
 
     /// Returns the keys from the iterator's current position to the end. The iterator
@@ -273,6 +273,12 @@ public class DocumentIterator: IteratorProtocol {
         return self.advance() ? (self.currentKey, self.currentValue) : nil
     }
 
+    internal func overwriteCurrentValue(with newValue: Overwritable) throws {
+        precondition(newValue.bsonType == self.currentType,
+                     "Expected \(newValue) to have BSON type \(self.currentType), but has type \(newValue.bsonType)")
+        try newValue.writeToCurrentPosition(of: self)
+    }
+
     private static let BsonTypeMap: [BsonType: BsonValue.Type] = [
         .double: Double.self,
         .string: String.self,
@@ -294,4 +300,45 @@ public class DocumentIterator: IteratorProtocol {
         .minKey: MinKey.self,
         .maxKey: MaxKey.self
     ]
+}
+
+/// A protocol indicating that a type can be overwritten in-place on a `bson_t`.
+internal protocol Overwritable: BsonValue {
+    /// Overwrites the value at the current position of the iterator with self.
+    func writeToCurrentPosition(of iter: DocumentIterator) throws
+}
+
+extension Bool: Overwritable {
+    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_bool(&iter.iter, self) }
+}
+
+extension Int: Overwritable {
+    func writeToCurrentPosition(of iter: DocumentIterator) throws {
+        if let int32 = self.int32Value {
+            return int32.writeToCurrentPosition(of: iter)
+        } else if let int64 = self.int64Value {
+            return int64.writeToCurrentPosition(of: iter)
+        }
+
+        throw MongoError.bsonEncodeError(message: "`Int` value \(self) could not be encoded as `Int32` or `Int64`")
+    }
+}
+
+extension Int32: Overwritable {
+    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_int32(&iter.iter, self) }
+}
+
+extension Int64: Overwritable {
+    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_int64(&iter.iter, self) }
+}
+
+extension Double: Overwritable {
+    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_double(&iter.iter, self) }
+}
+
+extension Decimal128: Overwritable {
+    func writeToCurrentPosition(of iter: DocumentIterator) throws {
+        var encoded = try Decimal128.encode(self.data)
+        bson_iter_overwrite_decimal128(&iter.iter, &encoded)
+    }
 }
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -256,6 +256,7 @@ public class DocumentIterator: IteratorProtocol {
 
         var output = Document()
 
+        // TODO SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
         for _ in startIndex..<endIndex {
             if let next = iter.next() {
                 output[next.key] = next.value
@@ -300,45 +301,4 @@ public class DocumentIterator: IteratorProtocol {
         .minKey: MinKey.self,
         .maxKey: MaxKey.self
     ]
-}
-
-/// A protocol indicating that a type can be overwritten in-place on a `bson_t`.
-internal protocol Overwritable: BsonValue {
-    /// Overwrites the value at the current position of the iterator with self.
-    func writeToCurrentPosition(of iter: DocumentIterator) throws
-}
-
-extension Bool: Overwritable {
-    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_bool(&iter.iter, self) }
-}
-
-extension Int: Overwritable {
-    func writeToCurrentPosition(of iter: DocumentIterator) throws {
-        if let int32 = self.int32Value {
-            return int32.writeToCurrentPosition(of: iter)
-        } else if let int64 = self.int64Value {
-            return int64.writeToCurrentPosition(of: iter)
-        }
-
-        throw MongoError.bsonEncodeError(message: "`Int` value \(self) could not be encoded as `Int32` or `Int64`")
-    }
-}
-
-extension Int32: Overwritable {
-    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_int32(&iter.iter, self) }
-}
-
-extension Int64: Overwritable {
-    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_int64(&iter.iter, self) }
-}
-
-extension Double: Overwritable {
-    func writeToCurrentPosition(of iter: DocumentIterator) { bson_iter_overwrite_double(&iter.iter, self) }
-}
-
-extension Decimal128: Overwritable {
-    func writeToCurrentPosition(of iter: DocumentIterator) throws {
-        var encoded = try Decimal128.encode(self.data)
-        bson_iter_overwrite_decimal128(&iter.iter, &encoded)
-    }
 }
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -33,6 +33,11 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         return self.makeIterator().keys
     }
 
+    /// Returns a `Boolean` indicating whether this `Document` contains the provided key.
+    public func hasKey(_ key: String) -> Bool {
+        return bson_has_field(self.data, key)
+    }
+
     /// Returns a `[BsonValue?]` containing the values stored in this `Document`.
     public var values: [BsonValue?] {
         return self.makeIterator().values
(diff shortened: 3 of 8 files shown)


Package.resolved
```diff
@@ -6,26 +6,26 @@
         "repositoryURL": "https://github.com/Quick/Nimble.git",
         "state": {
           "branch": null,
-          "revision": "8023e3980d91b470ad073d6da843b73f2eeb1844",
-          "version": "7.1.2"
+          "revision": "9c1379fdcd58c4f2278aea5e029394ba9a2b8f07",
+          "version": "7.1.3"
```
jmikola: Should these `Package.resolved` changes be in their own commit?

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -357,16 +357,20 @@ extension Double: BsonValue {
...
+        if let int32 = self.int32Value {
             return try int32.encode(to: storage, forKey: key)
         }
-        if let int64 = Int64(exactly: self) {
+        if let int64 = self.int64Value {
             return try int64.encode(to: storage, forKey: key)
         }
-        throw MongoError.bsonEncodeError(message: "`Int` value \(self) could not be encoded as `Int32` or `Int64`")
+
+        preconditionFailure("`Int` value \(self) could not be encoded as `Int32` or `Int64`")
```
jmikola: Why not continue to throw for this case as well?

Sources/MongoSwift/BSON/BsonValue.swift
```diff
@@ -357,16 +357,20 @@ extension Double: BsonValue {
 /// The `Int` will be encoded as an Int32 if possible, or an Int64 if necessary.
 extension Int: BsonValue {
 
-    public var bsonType: BsonType { return .int32 }
+    public var bsonType: BsonType { return self.int32Value != nil ? .int32 : .int64 }
```
jmikola: I originally commented that these Int changes did not seem relevant to "check for existing key and overwrite values", but I realize they are appropriate given the Overwritable protocol you introduce later. :+1:

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -181,26 +186,44 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
-                    preconditionFailure("Failed to set the value for key \(key) to null")
+    private mutating func setValue(forKey key: String, to newValue: BsonValue?) throws {
+        // if the key already exists in the `Document`, we need to replace it
+        if let existingType = DocumentIterator(forDocument: self, advancedTo: key)?.currentType {
+            // if the new type is the same and it's a fixed length type, we can overwrite
+            if let ov = newValue as? Overwritable, ov.bsonType == existingType {
+                self.copyStorageIfRequired()
+                // make a new iterator referencing our new storage. we already know the key is present
+                /// so initialization will succeed and ! is safe.
+                DocumentIterator(forDocument: self, advancedTo: key)!.overwriteCurrentValue(with: ov)
```
jmikola: In the event that a duplicate key exists in the same BSON document, are we guaranteed that advancing here before overwriting will reach the same key (likely the first occurrence) as the `existingType` declaration a few lines up? I believe so, but just want to confirm that is a not a problematic edge case.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -181,26 +186,44 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
+
+            // otherwise, we just create a new document and replace this key
+            } else {
+                var newSelf = Document()
+                self.forEach {
+                    newSelf[$0.key] = $0.key == key ? newValue : $0.value
```
jmikola: This looks like it might have trouble with duplicate keys. Would it make sense to add some logic to guarantee that we only replace the first occurrence to be consistent with the Overwritable case above?

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,101 @@ final class DocumentTests: XCTestCase {
...
+            // overwrite int32 with int32
+            doc["int32"] = Int32(15)
+            expect(doc["int32"] as? Int).to(equal(15))
+
+            // overwrite int32 with int
+            doc["int32"] = 20
+            expect(doc["int32"] as? Int).to(equal(20))
+
+            // overwrite int64 with int64
+            let bigInt64 = Int64(Int32.max) + 1
```
jmikola: Do we need to skip some of these "big int" tests when running on 32-bit platforms? I seem to recall another test doing that (perhaps with macros) when working with large integers.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,101 @@ final class DocumentTests: XCTestCase {
...
+        // test replacing some of the non-Overwritable types with values of their own types
+        do {
+            var doc = nonOverwritables
+
+            doc["string"] = "hi"
+            doc["nil"] = nil
+            let newDoc: Document = ["y": 1]
+            doc["doc"] = newDoc
+            doc["arr"] = [3, 4]
+            let oid = ObjectId()
```
jmikola: Interesting that ObjectId is not considered an Overwritable type, considering it's fixed length. I suppose that's something to address in libbson, though.

I suppose that could also apply to Decimal128.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,101 @@ final class DocumentTests: XCTestCase {
...
+            // final values
+            expect(doc).to(equal(["double": 3.0, "int32": 20, "int64": bigInt, "bool": true]))
+        }
+
+        // test replacing some of the non-Overwritable types with values of their own types
+        do {
+            var doc = nonOverwritables
+
+            doc["string"] = "hi"
+            doc["nil"] = nil
```
jmikola: Does it make sense to optimize for the `nil` case? BSON types with no custom data (e.g. nil, undefined, MinKey, MaxKey) should be overwritable as a no-op. That would allow us to avoid creating a new document and re-copying all other keys.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,101 @@ final class DocumentTests: XCTestCase {
...
+            doc["int64"] = bigInt
+
+            doc["bool"] = true
+            doc["double"] = 3.0
+
+            // final values
+            expect(doc).to(equal(["double": 3.0, "int32": 20, "int64": bigInt, "bool": true]))
+        }
+
+        // test replacing some of the non-Overwritable types with values of their own types
```
jmikola: This comment smells like we should have a new test method. Can we use a data provider to create the initial documents (assuming they need to be the same for each case)?

Shorter test methods would be more maintainable and we could also give them more verbose names.

Package.resolved
kmahar: done: 115294ca49121811267edd0b33cb3b691810ce8c

Tests/MongoSwiftTests/DocumentTests.swift
kmahar: good point. I've moved your code from that test to be in an `XCTestCase` extension so we can access `is32Bit` from any test. I have refactored this test so that we use a different version of `overwritables` not containing an `Int64` on 32-bit platforms.

however, there are most definitely other cases throughout our older tests where we are not performing this check, so I've opened [SWIFT-197](https://jira.mongodb.org/browse/SWIFT-197) so we remember to do an audit in the future.

Sources/MongoSwift/BSON/Document.swift
kmahar: sure, done. I think that as part of document validation, we should include adding our own validation against duplicate keys - libbson validation does not check for this 😱 - and then we can revisit what we are doing here

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -181,26 +186,57 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
+            // if the new type is the same and it's a fixed length type, we can overwrite
+            if let ov = newValue as? Overwritable, ov.bsonType == existingType {
+                self.copyStorageIfRequired()
+                // make a new iterator referencing our new storage. we already know the key is present
+                /// so initialization will succeed and ! is safe.
+                try DocumentIterator(forDocument: self, advancedTo: key)!.overwriteCurrentValue(with: ov)
+
+            // otherwise, we just create a new document and replace this key
+            } else {
+                var newSelf = Document()
```
kmahar: once `mapValues` is merged in we can use that here instead

kmahar:
another question here is whether this behavior should be a setting users can turn on and off, and if so, when would be the appropriate time for them to specify that?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -76,10 +76,15 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      *
      * - Returns: a new `Document`
      */
-    public init(dictionaryLiteral doc: (String, BsonValue?)...) {
+    public init(dictionaryLiteral keyValuePairs: (String, BsonValue?)...) {
+        // make sure all keys are unique
+        if Set(keyValuePairs.map { $0.0 }).count != keyValuePairs.count {
```
kmahar: this is the same behavior as `Dictionary`:
```
9> let d = ["a": 1, "a": 2]
Fatal error: Dictionary literal contains duplicate keys
```

Sources/MongoSwift/BSON/BsonValue.swift
jmikola: I often err on the side of defensive programming, especially in a library/driver. I'll defer to your judgement either way.

Sources/MongoSwift/BSON/Document.swift
jmikola: If we're mirroring Dictionary here, I don't think we need to make this behavior something users can turn on/off, per your earlier [comment](https://github.com/mongodb/mongo-swift-driver/pull/111#issuecomment-414538230):

> another question here is whether this behavior should be a setting users can turn on and off, and if so, when would be the appropriate time for them to specify that?

I can't think of a valid use case for creating BSON documents with duplicate keys, and most other drivers that use native types for document representations (thinking Python, PHP, Node) wouldn't support this at all.

jmikola:
Per your earlier [comment](https://github.com/mongodb/mongo-swift-driver/pull/111#issuecomment-414538230):

> another question here is whether this behavior should be a setting users can turn on and off, and if so, when would be the appropriate time for them to specify that?

Are you asking about allowing users to turn off behavior for overwriting the Overwritable types (e.g. integers)? Or allowing them to turn off restrictions from creating duplicate keys?

I can't think of a valid use case for creating BSON documents with duplicate keys, and most other drivers that use native types for document representations (thinking Python, PHP, Node) wouldn't support that at all.

In any event, I think you can err on the side of being conservative for now. Always allowing the fixed-size types to be overwritten makes sense to me (libbson supports it), as does disallowing duplicates key creation so long as we don't have a way to overwrite variable-length types.

kmahar:
I meant basically a way to go back to the state before these changes basically, we wouldn't do any check for duplicate keys and would just append, with the intent of improving performance in cases where you know keys won't be duplicate.

I'm fine with leaving it out of the public API, but this does make me realize: there are a few cases internally where we can guarantee keys are unique and we should bypass all this. when we are [initializing a document from an array](https://github.com/mongodb/mongo-swift-driver/blob/9d413fba270c80d185e43aed9e27ea0cbcba6a6e/Sources/MongoSwift/BSON/Document.swift#L107), we assign the keys, and we can guarantee they're all going to be unique. similarly, the [dictionary literal init](https://github.com/mongodb/mongo-swift-driver/pull/111/files#diff-fa227eb40bcf532cab60f2c069a062caR81) now asserts the keys are all unique before creating the document. so I'm going to add a flag to `setValue` (which is `private`) that we can use in those cases.

kmahar:
waiting for lgtm from Jeremy and for Matt review. libmongoc 1.13 is adding new methods we'd want to use in some of this work - see [SWIFT-200](https://jira.mongodb.org/browse/SWIFT-200) - so depending on timing, will maybe incorporate those within this PR or do it in a future PR.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,131 @@ final class DocumentTests: XCTestCase {
...
+        doc["double"] = 3.0
+        doc["decimal"] = Decimal128("100")
+
+        if XCTestCase.is32Bit {
+            expect(doc).to(equal(["double": 3.0, "int32": 20, "bool": true, "decimal": Decimal128("100")]))
+            return
+        }
+
+        // overwrite int64 with int64
+        let bigInt64 = Int64(Int32.max) + 1
```
jmikola: Is there actually a problem working with `Int64` objects on a 32-bit platform? IIRC, the skip in `testIntEncodesAsInt32OrInt64()` was only used because we were doing arithmetic.

In this case, I wonder if you could just create `Int64` objects from `Int64.max` and overwrite it with `Int64.min`.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,131 @@ final class DocumentTests: XCTestCase {
...
+        doc["arr"] = [3, 4]
+        let oid = ObjectId()
+        doc["oid"] = oid
+
+        expect(doc).to(equal(["string": "hi", "nil": nil, "doc": newDoc, "arr": [3, 4] as [Int], "oid": oid]))
+    }
+
+    // test replacing both overwritable and nonoverwritable values with values of different types
+    func testReplaceValueWithNewType() throws {
+
```
jmikola: Extra newline.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,131 @@ final class DocumentTests: XCTestCase {
...
     }
+
+    // exclude Int64 value on 32-bit platforms
+    static let overwritables: Document = XCTestCase.is32Bit ?
+                                        ["double": 2.5, "int32": Int32(32), "bool": false, "decimal": Decimal128("1.2E+10")] :
+                                        ["double": 2.5, "int32": Int32(32), "int64": Int64(64), "bool": false, "decimal": Decimal128("1.2E+10")]
+
+    static let nonOverwritables: Document = ["string": "hello", "nil": nil, "doc": ["x": 1] as Document, "arr": [1, 2] as [Int], "oid": ObjectId()]
+
+    // test replacing `Overwritable` types with values of their own type
```
jmikola: If I understand correctly, this test ovewrites values but doesn't copy the underlying storage, since libbson lets us overwrite them in place. But we also have tests for other overwriting behavior:

 * test replacing some of the non-Overwritable types with values of their own types
 * test replacing both overwritable and nonoverwritable values with values of different types

I wonder if _this_ test should actually assert that we don't duplicate storage. If so, perhaps that other tests can assert that storage _is_ duplicated (where appropriate).

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -372,4 +373,131 @@ final class DocumentTests: XCTestCase {
...
+        doc2["string"] = nil
+        doc2["nil"] = nil
+        doc2["doc"] = nil
+        doc2["arr"] = nil
+        doc2["oid"] = nil
+
+        expect(doc2).to(equal(["string": nil, "nil": nil, "doc": nil, "arr": nil, "oid": nil]))
+    }
+
+    // Test types where replacing them with an instance of their own type is a no-op
```
jmikola: If this is comparable to replacing Overwritable types, perhaps it can also assert that storage is not duplicated.

kmahar: 👍 done

Tests/MongoSwiftTests/DocumentTests.swift
kmahar: ah, yeah you're right. in this case, I will switch to overwriting with Int64.min. 

However, for the next part, where I want to try replacing an `Int64` with an `Int`, I need an `Int` whose `bsonType` will be `Int64` (which will only be the case if `Int32(exactly:)` fails). I don't think an `Int` larger than `Int32.max` can exist on a 32 bit platform, right? but we can at least move the early return a bit later.

kmahar:
C driver 1.13 has now been released, so I just briefly tried to start incorporating that in this PR (see [SWIFT-200](https://jira.mongodb.org/browse/CDRIVER-2811)). however, bumping our dependency to 1.13 is going to require at least one more fix - I am getting a crash related to SDAM which I assume is somehow related to [CDRIVER-2811](https://jira.mongodb.org/browse/CDRIVER-2811). 

I'll try to investigate that a bit more today and figure out what we need to change, and see whether bumping to 1.13 before this PR is merged in is reasonable.

Tests/MongoSwiftTests/DocumentTests.swift
jmikola: > I don't think an Int larger than Int32.max can exist on a 32 bit platform, right?

Assuming `Int` is platform dependent, I'd agree.

> However, for the next part, where I want to try replacing an Int64 with an Int

Since libbson requires integers to be replaced with the same size, would it make sense that we should allow an Int64 BSON type to be replaced with any integer? If it's in the range of 32-bit values (e.g. result of decrementing a value in the 64-bit range), perhaps we should cast it up to a 64-bit value to still take advantage of in-place replacement in libbson. Obviously, the reverse case (`Int` in 32-bit range growing to 64-bit range) would require re-encoding and could not utilize in-place replacement.

I'm not sure how the server handles this (if it still does). With MMAPv1, there were optimizations for in-place replacements, but I don't think those apply to WT, since documents are always re-written.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -141,6 +141,12 @@ public class DocumentIterator: IteratorProtocol {
         return nil
     }
 
+    internal func overwriteCurrentValue(with newValue: Overwritable) throws {
+        precondition(newValue.bsonType == self.currentType,
+                     "Expected \(newValue) to have BSON type \(self.currentType), but has type \(newValue.bsonType)")
```
jmikola: Correct me if I'm wrong, but this is just a sanity check to ensure one Overwritable (calling this method) isn't trying to overwrite itself with a different Overwritable type. Is that right?

And this ties in with the logic before this method is called in `setValue()`:

```
if let ov = newValue as? Overwritable, ov.bsonType == existingType
```

kmahar: yes, exactly, just a sanity check. I don't expect we will ever call this method from anywhere besides the current call site.

Tests/MongoSwiftTests/DocumentTests.swift
kmahar: [it is](https://developer.apple.com/documentation/swift/int) - 
> On 32-bit platforms, Int is the same size as Int32, and on 64-bit platforms, Int is the same size as Int64.

I was thinking about that too. my only concern was that when reading the value out of the document, the user would then get an `Int64` when they expect an `Int32`. I can't think of any situations where that would be really catastrophic, but it may be unexpected. what do you think?

kmahar:
nvm, crash was just because I didn't build libmongoc with automatic init/cleanup enabled, which led to a segfault when libmongoc expected the driver name to be set when it constructed the first handshake. making separate PR to bump libmongoc to 1.13

Tests/MongoSwiftTests/DocumentTests.swift
jmikola: Sorry for losing track of this conversation. I don't follow the concern you're raising.

IIRC, we were discussing allowing an Int64 to be overwritten by _any_ integer value. If an Int, which happens to be an Int32, or an explicit Int32 is passed, we can cast it up and widen it to an Int64 in order to do an in-place update in libbson.

I'm not sure how this would adversely affect a user. Even if they did want to assert that they had overwritten a large-than-32-bit value with a value within the 32-bit range, Swift should still allow them to downcast safely to an Int32 when accessing the field (I assume using `as Int32`), no?

kmahar: unfortunately no, consider the following snippet:
```swift
let doc: Document = ["a": Int64(100)]
print(doc["a"] as? Int32) // nil
print(doc["a"] as? Int64) // 100
```
the subscript operator has no knowledge of what type we're requesting, so it just returns a value of whatever Swift type corresponds to BSON type specified in the `bson_t`. 

we aren't allowed to cast `Int64` to `Int32` using `as`, so a user would have to do something like
```swift
guard let int64Value = doc["a"] as? Int64 else { /// ... }
let int32Value = Int32(int64Value)
```
so if the user needs to be able to extract the field specifically as an `Int32` this would be problematic.

jmikola: And it looks like we also can't do `doc["a"] as? Int` to return the Int64 as an Int, even if we're on a 64-bit platform. `DocumentIterator.BsonTypeMap` says that all BSON int32 and int64 types are converted to an `Int` and `Int64`, respectively.

I wonder if the existing behavior is problematic as well. Consider:

```swift
let a = Int(Int64.max)
let doc: Document = ["a": a]
print(a) // 9223372036854775807
print(doc["a"] as? Int) // nil
print(doc["a"] as? Int64) // Optional(9223372036854775807)
```

If users are accessing a BSON integer field, they must anticipate its range before they can cast it to a native type. And they can't rely on always casting to `Int64`, as a BSON int32 type will yield nil if cast to `Int64`.

To bring this back to the current PR, I think it's sensible to only allow overwriting (really just an optimization) when the exact BSON type is used. Separate from that, it seems like there's room for improvement with integer access.

jmikola: This can be addressed by [SWIFT-221](https://jira.mongodb.org/browse/SWIFT-221). Resolving.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -295,3 +301,44 @@ public class DocumentIterator: IteratorProtocol {
         .maxKey: MaxKey.self
     ]
 }
+
+/// A protocol indicating that a type can be overwritten in-place on a `bson_t`.
+internal protocol Overwritable: BsonValue {
```
mbroadst: do these belong in a different file, e.g. `Document+Overwritable.swift`?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -177,26 +195,58 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
+            // if the new type is the same and it's a fixed length type, we can overwrite
+            if let ov = newValue as? Overwritable, ov.bsonType == existingType {
+                self.copyStorageIfRequired()
+                // make a new iterator referencing our new storage. we already know the key is present
+                /// so initialization will succeed and ! is safe.
+                try DocumentIterator(forDocument: self, advancedTo: key)!.overwriteCurrentValue(with: ov)
+
+            // otherwise, we just create a new document and replace this key
+            } else {
+                var newSelf = Document()
```
mbroadst: please make a note that this should eventually use `bson_copy_to_excluding` for performance reasons

PR #112, opened 2018-08-17
SWIFT-195 Make Document SubSequence type a Document

kmahar:
I use map/filter in these methods so this is on top of [SWIFT-158](https://github.com/mongodb/mongo-swift-driver/pull/108).

there are a number of methods that come along with the sequence protocol such as `dropFirst`, `dropLast`, `prefix`, etc. that return subsequences. By default, the type of these subsequences is `[whatever your iterator returns]`, meaning we would return a`[(String, BsonValue?)]`.

Swift's `Dictionary` provides custom implementations of the `SubSequence` related methods, so that you get a new `Dictionary` back rather than an `[Dictionary.Element]`. This does the same for `Document`s.

### DIFF
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -13,6 +13,12 @@ extension Document: Sequence {
     /// The element type of a document: a tuple containing an individual key-value pair.
     public typealias KeyValuePair = (key: String, value: BsonValue?)
 
+    /// Returns the number of (key, value) pairs stored at the top level of this `Document`.
+    public var count: Int { return Int(bson_count_keys(self.data)) }
+
+    /// Returns a `Bool` indicating whether the document is empty.
+    public var isEmpty: Bool { return self.makeIterator().advance() }
+
     /// Returns a `DocumentIterator` over the values in this `Document`. 
     public func makeIterator() -> DocumentIterator {
         guard let iter = DocumentIterator(forDocument: self) else {
@@ -28,7 +34,9 @@ extension Document: Sequence {
      * - Parameters:
      *   - isIncluded: A closure that takes a key-value pair as its argument and returns a `Bool` indicating whether 
      *                 the pair should be included in the returned document.
+     *
      * - Returns: A document of the key-value pairs that `isIncluded` allows.
+     *
      * - Throws: An error if `isIncluded` throws an error.
      */
     public func filter(_ isIncluded: (KeyValuePair) throws -> Bool) rethrows -> Document {
@@ -47,7 +55,9 @@ extension Document: Sequence {
      *   - transform: A closure that transforms a `BsonValue?`. `transform` accepts each value of the
      *                document as its parameter and returns a transformed `BsonValue?` of the same or 
      *                of a different type.
+     *
      * - Returns: A document containing the keys and transformed values of this document.
+     *
      * - Throws: An error if `transform` throws an error.
      */
     public func mapValues(_ transform: (BsonValue?) throws -> BsonValue?) rethrows -> Document {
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -10,13 +10,156 @@ import libmongoc
...
+        for (k, v) in self {
+            output[k] = try transform(v)
+        }
+        return output
+    }
+
+    public func dropFirst(_ n: Int) -> Document {
+        switch n {
+        case ..<0:
+            preconditionFailure("Can't drop a negative number of elements from a document")
```
kmahar: while this seems not ideal, we're unfortunately prevented from throwing here because it's a protocol method. `Dictionary` fails similarly for negative n.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -10,13 +10,156 @@ import libmongoc
...
+        default:
+            let start = self.count - maxLength
+            // short circuit if there are fewer elements in the doc than requested
+            return start <= 0 ? self : DocumentIterator.subsequence(of: self, startIndex: start)
+        }
+    }
+
+    public func split(maxSplits: Int = Int.max, omittingEmptySubsequences: Bool = true,
+                      whereSeparator isSeparator: (KeyValuePair) throws -> Bool) rethrows -> [Document] {
+        // rather than implementing the complex logic necessary for split, convert to an array and call split on that
```
kmahar: for several methods besides `split` added in this PR, I wrote custom implementations relying on a new static method `DocumentIterator.subsequence`. my reasoning was that several of those methods allow us to get away with only reading out some portion of the `Document` into Swift types --  we often get to skip over values at the front of the document or terminate early. but calling `Array(self)` and then doing the operations on that array would always requires us to read every single value out to the array, which isn't very efficient. 
On the other hand, `split` is always going to require reading every single element anyway, plus the source code for the array version is pretty complicated (see [here](https://github.com/apple/swift/blob/a17dbfbf29aeed3f80b484a5a66c9ea5b48af3d3/stdlib/public/core/Collection.swift#L1652)), so that was another compelling reason to just do this rather than trying to duplicate.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -93,8 +236,34 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    // uses an iterator to copy (key, value) pairs of the provided document from range [startIndex, endIndex) into a new
+    // document. starts at the startIndex-th pair and ends at the end of the document or the (endIndex-1)th index, 
+    // whichever comes first. 
+    internal static func subsequence(of doc: Document, startIndex: Int = 0, endIndex: Int = Int.max) -> Document {
+        precondition(endIndex >= startIndex, "endIndex must be >= startIndex")
+
+        guard let iter = DocumentIterator(forDocument: doc) else { return [:] }
```
kmahar: not really sure what to do in this case. none of the methods calling subsequence can throw. maybe just preconditionFailure? idk

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -158,26 +157,6 @@ final class DocumentTests: XCTestCase {
        expect(doc2["2"]).to(beNil())
     }
 
-    func testIterator() {
```
kmahar: since this test relates to sequence protocol, moved to new test class

Sources/MongoSwift/BSON/Document+Sequence.swift
jmikola: I suppose consistency with Swift is the best we can do. I assume that means precondition failures?

kmahar: well, in general we would do that when we can't error, but this reminds me of the cases in `.extendedJSON` and `.canonicalExtendedJSON` where we opted out of a `preconditionFailure` and instead returned an empty string if parsing the data failed, because someone could call the method on unvalidated/corrupt BSON and have a crash.

similarly, here we could experience
```swift
let doc: Document = Document(fromBSON: corrupt_bson_data_here)
print(doc.suffix(maxLength: 1)) // precondition failure
```

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -20,7 +20,10 @@ public class DocumentStorage {
     }
 }
 
-/// A struct representing the BSON document type.
+/// A struct representing the BSON document type. `Document`s are append-only and do not support
+/// overwriting or removing keys in-place, or provide safeguards against duplicate keys. 
+/// To transform documents, use `Sequence` protocol methods along with `filter` and `mapValues`.
+/// To check for existing keys, use the `.keys` property.
```
jmikola: Use case for `/** */`?

Tests/MongoSwiftTests/Document+SequenceTests.swift
```diff
@@ -0,0 +1,192 @@
...
+    func isInt(_ pair: Document.KeyValuePair) -> Bool { return pair.value is Int }
+    func isNotNil(_ pair: Document.KeyValuePair) -> Bool { return pair.value != nil }
+    func is10(_ pair: Document.KeyValuePair) -> Bool {
+        if let int = pair.value as? Int { return int == 10 } else { return false }
+    }
+    func isNot10(_ pair: Document.KeyValuePair) -> Bool { return !is10(pair) }
+
+    func testDropFirst() throws {
+        expect(self.emptyDoc.dropFirst(0)).to(equal([:]))
+        expect(self.emptyDoc.dropFirst(1)).to(equal([:]))
```
jmikola: I assume this behavior (dropping from an empty sequence returns the empty sequence) is based on Dictionary's `dropFirst()`?

Tests/MongoSwiftTests/Document+SequenceTests.swift
```diff
@@ -0,0 +1,192 @@
...
+
+        expect(self.smallDoc.dropFirst(0)).to(equal(smallDoc))
+        expect(self.smallDoc.dropFirst()).to(equal([:]))
+        expect(self.smallDoc.dropFirst(2)).to(equal([:]))
+
+        expect(self.doc.dropFirst(0)).to(equal(doc))
+        expect(self.doc.dropFirst()).to(equal(["b": "hi", "c": [1, 2] as [Int], "d": false, "e": nil, "f": MinKey(), "g": 10]))
+        expect(self.doc.dropFirst(4)).to(equal(["e": nil, "f": MinKey(), "g": 10]))
+        expect(self.doc.dropFirst(7)).to(equal([:]))
```
jmikola: To test the boundary case, would it be better to use `3` here and assert an empty sequence?

Tests/MongoSwiftTests/Document+SequenceTests.swift
```diff
@@ -0,0 +1,192 @@
...
+        expect(self.emptyDoc.dropLast(1)).to(equal([:]))
+
+        expect(self.smallDoc.dropLast(0)).to(equal(smallDoc))
+        expect(self.smallDoc.dropLast()).to(equal([:]))
+        expect(self.smallDoc.dropLast(2)).to(equal([:]))
+
+        expect(self.doc.dropLast(0)).to(equal(doc))
+        expect(self.doc.dropLast()).to(equal(["a": 1, "b": "hi", "c": [1, 2] as [Int], "d": false, "e": nil, "f": MinKey()]))
+        expect(self.doc.dropLast(4)).to(equal(["a": 1, "b": "hi", "c": [1, 2] as [Int]]))
+        expect(self.doc.dropLast(7)).to(equal([:]))
```
jmikola: Likewise for testing the boundary case here as well, instead of using `7`.

kmahar:
approved by Jeremy, pending Matt review. SWIFT-158 should be reviewed first as this is on top of it.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -93,8 +236,34 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    // uses an iterator to copy (key, value) pairs of the provided document from range [startIndex, endIndex) into a new
+    // document. starts at the startIndex-th pair and ends at the end of the document or the (endIndex-1)th index, 
+    // whichever comes first. 
+    internal static func subsequence(of doc: Document, startIndex: Int = 0, endIndex: Int = Int.max) -> Document {
```
mbroadst: shouldn't we be using [bson_copy_to_excluding_noinit](http://mongoc.org/libbson/current/bson_copy_to_excluding_noinit.html) here instead of doing this all in Swift?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -38,10 +38,12 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         return self.makeIterator().values
     }
 
-    /// Returns the number of (key, value) pairs stored at the top level
-    /// of this `Document`.
+    /// Returns the number of (key, value) pairs stored at the top level of this `Document`.
     public var count: Int { return Int(bson_count_keys(self.data)) }
 
+    /// Returns a `Bool` indicating whether the document is empty.
+    public var isEmpty: Bool { return self.count == 0 }
```
mbroadst: is this required of the interface?

Tests/MongoSwiftTests/Document+SequenceTests.swift
```diff
@@ -0,0 +1,192 @@
...
+            "int32": Int32(5),
+            "double": Double(15),
+            "decimal128": Decimal128("1.2E+10"),
+            "minkey": MinKey(),
+            "maxkey": MaxKey(),
+            "date": Date(timeIntervalSince1970: 5000),
+            "timestamp": Timestamp(timestamp: 5, inc: 10)
+        ]
+
+        for (_, _) in doc { }
```
mbroadst: there's no actual assertions in this test 🤔

Sources/MongoSwift/BSON/Document+Sequence.swift
mbroadst: That's a bummer, maybe we should open a ticket in the C driver for this. The way we are doing this is pretty expensive

kmahar: I've opened [CDRIVER-2829](https://jira.mongodb.org/browse/CDRIVER-2829) for discussion.

PR #114, opened 2018-09-06
Correct linting violations and update travis XCode version

kmahar:
this corrects the violations I was missing before w/ old version of swiftlint.

it seems that travis hasn't upgraded to including the newest swiftlint yet, so I guess we can just catch these rule violations locally for now...

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -15,7 +15,6 @@ enabled_rules:
   - modifier_order
   - missing_docs
   - implicit_return
-  - function_default_parameter_at_end
   - fatal_error_message
 
 excluded:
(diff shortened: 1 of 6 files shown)


.swiftlint.yml
```diff
@@ -15,7 +15,6 @@ enabled_rules:
   - modifier_order
   - missing_docs
   - implicit_return
-  - function_default_parameter_at_end
```
kmahar: I removed this rule because pretty much every violation it found was justified imo - i.e. if a method has like 20 parameters I prefer to have them alphabetically ordered rather than order by default or not.

PR #116, opened 2018-09-12
SWIFT-210: Add very basic Kitura and Vapor examples

mbroadst:
These are the most basic examples you can have for using these services with MongoDB. Use the `loadExampleData.sh` script to insert the expected data, and then each of the examples run a server on `localhost:8080` with a single route `/kittens` provided

### DIFF
--- a/Examples/Kitura/Package.resolved
+++ b/Examples/Kitura/Package.resolved
@@ -1,0 +1,124 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "Socket",
+        "repositoryURL": "https://github.com/IBM-Swift/BlueSocket.git",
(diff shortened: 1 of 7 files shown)


kmahar:
also I got both running locally myself, only weirdness was having to switch to 4.1 for vapor example

mbroadst:
@kmahar dang, and I was looking at a picture of them when I wrote that too

jmikola:
Created SWIFT-210](https://jira.mongodb.org/browse/SWIFT-210) to track this issue, so please use that in the final merge/commit.

Also made [SWIFT-211](https://jira.mongodb.org/browse/SWIFT-211) to track adding a readme to these examples.

PR #118, opened 2018-09-17
SWIFT-208: CRUD methods return nil for unacknowledged write concern

jmikola:
https://jira.mongodb.org/browse/SWIFT-208

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -26,6 +26,11 @@ extension MongoCollection {
             // TODO SWIFT-139: include writeErrors and writeConcernErrors from reply in the error
             throw MongoError.commandError(message: toErrorString(error))
         }
+
+        guard isAcknowledged(options?.writeConcern) else {
+            return nil
+        }
+
         return InsertOneResult(insertedId: document["_id"])
     }
 
@@ -57,6 +62,10 @@ extension MongoCollection {
             self._collection, &docPointers, documents.count, opts?.data, reply.data, &error) {
             // TODO SWIFT-139: include writeErrors and writeConcernErrors from reply in the error
             throw MongoError.commandError(message: toErrorString(error))
+        }
+
+        guard isAcknowledged(options?.writeConcern) else {
+            return nil
         }
 
         // if we succeeded, then we should store + return every `_id`
@@ -93,6 +102,10 @@ extension MongoCollection {
             throw MongoError.commandError(message: toErrorString(error))
         }
 
+        guard isAcknowledged(options?.writeConcern) else {
+            return nil
+        }
+
         return try BsonDecoder().decode(UpdateResult.self, from: reply)
     }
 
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -191,8 +219,33 @@ extension MongoCollection {
...
+     *
+     * - Parameters:
+     *   - writeConcern: `WriteConcern` from the operation's options struct
+     *
+     * - Returns: Whether the operation will use an acknowledged write concern
+     */
+    fileprivate func isAcknowledged(_ writeConcern: WriteConcern?) -> Bool {
+        /* If the collection's write concern is also `nil` it is the default. We
+         * can safely assume it is acknowledged, sincethe server requires that
+         * getLastErrorDefaults is acknowledged by at least one member. */
```
jmikola: FYI: I tested this locally using MongoDB 4.0 where it would not allow my to reconfigure the replica set with an unacknowledged write concern (see: [`settings.getLastErrorDefaults`](https://docs.mongodb.com/manual/reference/replica-configuration/#rsconf.settings.getLastErrorDefaults)). Therefore, I think we can safely assume that all default write concerns are acknowledged.

I didn't test earlier versions, but I doubt that restriction is new to 4.0.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -26,6 +26,11 @@ extension MongoCollection {
             // TODO SWIFT-139: include writeErrors and writeConcernErrors from reply in the error
             throw MongoError.commandError(message: toErrorString(error))
         }
+
+        guard isAcknowledged(options?.writeConcern) else {
```
kmahar: I ws going to say "nit: I would probably one-line all these", but I realize we have no stylistic standard for when to use one line vs. two. I usually just kind of do it based on what feels right and how long the line is... any thoughts on how to establish a consistent pattern on that throughout the codebase?

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: sounds like a reasonable assumption to me.

Sources/MongoSwift/MongoCollection+Write.swift
jmikola: I'm not a fan of using one-liners for conditionals. You could make the argument that `guard` is special since we can trust that it will exit the current scope one way or another (`return` or `throw`).

That said, I think they're OK for short lambas (e.g. something that just returns a value and is used with `map`).

I suppose this is something that can be detected with SwiftLint. Does that actually support fixing the code, or can it only alert us when code passes certain checks?

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -191,8 +219,33 @@ extension MongoCollection {
...
+     * the collection's write concern will be considered.
+     *
+     * - Parameters:
+     *   - writeConcern: `WriteConcern` from the operation's options struct
+     *
+     * - Returns: Whether the operation will use an acknowledged write concern
+     */
+    fileprivate func isAcknowledged(_ writeConcern: WriteConcern?) -> Bool {
+        /* If the collection's write concern is also `nil` it is the default. We
+         * can safely assume it is acknowledged, sincethe server requires that
```
jmikola: "since the"

Sources/MongoSwift/MongoCollection+Write.swift
jmikola: https://jira.mongodb.org/browse/SWIFT-217

PR #120, opened 2018-09-24
Use libmongoc 1.13

kmahar:
This will enable us to take advantage of newly added methods for overwriting fixed-length types in place ([SWIFT-200](https://jira.mongodb.org/browse/SWIFT-200)).

Since we have no way of requiring a particular libmongoc version currently, Travis is the only thing we need to change.

I have also updated the README to instruct users to install 1.13. while our latest release 0.0.3 doesn't actually require it, I fear we will forget to update this minimum version whenever we release 0.0.4, so better to just do it now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ install:
   - eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"
   
   # clone and build libmongoc
-  - git clone -b r1.12 https://github.com/mongodb/mongo-c-driver /tmp/libmongoc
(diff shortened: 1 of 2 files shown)


jmikola:
> Since we have no way of requiring a particular libmongoc version currently

Is this a limitation of Swift itself, or perhaps just something we'd need to do in [swift-mongoc](https://github.com/mongodb/swift-mongoc)?

kmahar:
we have a ticket ([SWIFT-130](https://jira.mongodb.org/browse/SWIFT-130)) for figuring out how to do that but haven't gotten to it yet. I've also opened [SWIFT-220](https://jira.mongodb.org/browse/SWIFT-220) for updating the includes for 1.13

PR #122, opened 2018-10-08
Refactor insertMany and test CRUD continue-on-error behavior

jmikola:
This refactors `insertMany()` a bit to be consistent with `bulkWrite()`. We introduce a `MongoError.insertManyError`, which contains an optional `InsertManyResult`.

This also syncs write tests from https://github.com/mongodb/specifications/pull/399, which test for continue-on-error behavior. Additionally, bulkWrite collation tests have been pulled in (and pass). CRUD spec read tests have not been synced due to new count helpers not being implemented (see: [SWIFT-227](https://jira.mongodb.org/browse/SWIFT-227)).

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -35,46 +35,53 @@ extension MongoCollection {
     }
 
     /**
-     * Encodes the provided values to BSON and inserts them. If any values are missing identifiers,
-     * the driver will generate them.
+     * Encodes the provided values to BSON and inserts them. If any values are
+     * missing identifiers, the driver will generate them.
      *
      * - Parameters:
-     *   - documents: The `CollectionType` values to insert
-     *   - options: Optional `InsertManyOptions` to use when executing the command
+     *   - values: The `CollectionType` values to insert
+     *   - options: optional `InsertManyOptions` to use while executing the operation
      *
-     * - Returns: The optional result of attempting to performing the insert. If the write concern
-     *            is unacknowledged, nil is returned
+     * - Returns: an `InsertManyResult`, or `nil` if the write concern is unacknowledged.
+     *
+     * - Throws:
+     *   - `MongoError.invalidArgument` if `values` is empty
+     *   - `MongoError.insertManyError` if any error occurs while performing the writes
      */
     @discardableResult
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -35,46 +35,53 @@ extension MongoCollection {
...
 
+        let encoder = BsonEncoder()
+        let opts = try encoder.encode(options)
         let documents = try values.map { try encoder.encode($0) }
-        for doc in documents where !doc.keys.contains("_id") {
-            try ObjectId().encode(to: doc.storage, forKey: "_id")
+        var insertedIds: [Int: BsonValue?] = [:]
+
+        try documents.enumerated().forEach { (index, document) in
+            if !document.keys.contains("_id") {
```
kmahar: now that [`Document.hasKey`](https://github.com/mongodb/mongo-swift-driver/blob/185829de1358094b25838781c126db000cad8a7c/Sources/MongoSwift/BSON/Document.swift#L37) exists, we should use that in these cases to avoid reading all the keys out into Swift.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -35,46 +35,53 @@ extension MongoCollection {
...
+        let success = mongoc_collection_insert_many(self._collection, &docPointers, values.count, opts?.data, reply.data, &error)
+        let result = try InsertManyResult(reply: reply, insertedIds: insertedIds)
+        let isAcknowledged = self.isAcknowledged(options?.writeConcern)
 
-        // if we succeeded, then we should store + return every `_id`
-        var insertedIds = [Int: BsonValue?]()
-        for (index, doc) in documents.enumerated() {
-            insertedIds[index] = doc["_id"]
+        guard success else {
+            throw MongoError.insertManyError(code: error.code, message: toErrorString(error),
```
kmahar: nit: message on new line

jmikola: I thought this was kosher because we're just abiding a 120-character wrap and not necessarily placing each argument on its own line. We were doing the same in https://github.com/mongodb/mongo-swift-driver/blob/185829de1358094b25838781c126db000cad8a7c/Sources/MongoSwift/MongoCollection%2BBulkWrite.swift#L281.

Should I change both?

PR #123, opened 2018-10-19
SWIFT-203: Rename BsonEncoder/BsonDecoder to BSONEncoder/BSONDecoder

Utagai:
[SWIFT-203](https://jira.mongodb.org/browse/SWIFT-203?jql=key%20%3D%20SWIFT-203)

This PR does a bit more than what the ticket explicitly mentions. This ticket renames mentions of `Bson` to `BSON`, a superset of the changes expected in the ticket. This was done after I spoke with @kmahar in order to be consistent across the codebase.


Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -185,21 +185,21 @@ internal struct _BsonEncodingStorage {
...
 
-    fileprivate mutating func popContainer() -> BsonValue? {
+    fileprivate mutating func popContainer() -> BSONValue? {
         precondition(self.containers.count > 0, "Empty container stack.")
         return self.containers.popLast()!
     }
 }
 
-/// _BsonReferencingEncoder is a special subclass of _BsonEncoder which has its own storage, but references the 
+/// _BSONReferencingEncoder is a special subclass of _BSONEncoder which has its own storage, but references the
```
jmikola: Unrelated to the change, but we can probably add backticks around these class references for consistency:

```suggestion
/// `_BSONReferencingEncoder` is a special subclass of `_BSONEncoder` which has its own storage, but references the 
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -260,29 +260,29 @@ private class _BsonReferencingEncoder: _BsonEncoder {
 
 }
 
-/// Extend _BsonEncoder to add methods for "boxing" values.
-extension _BsonEncoder {
+/// Extend _BSONEncoder to add methods for "boxing" values.
```
jmikola:
```suggestion
/// Extend `_BSONEncoder` to add methods for "boxing" values.
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -152,19 +152,19 @@ internal class _BsonEncoder: Encoder {
...
     }
 }
 
-internal struct _BsonEncodingStorage {
+internal struct _BSONEncodingStorage {
 
     /// The container stack.
-    /// Elements may be any BsonValue type.
-    internal var containers: [BsonValue?] = []
+    /// Elements may be any BSONValue type.
```
jmikola:
```suggestion
    /// Elements may be any `BSONValue` type.
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -260,29 +260,29 @@ private class _BsonReferencingEncoder: _BsonEncoder {
...
         return number
     }
 
-    fileprivate func box<T: Encodable>(_ value: T) throws -> BsonValue? {
-        // if it's already a BsonValue, just return it, unless if it is an 
-        // array. technically [Any] is a BsonValue, but we can only use this
-        // short-circuiting if all the elements are actually BsonValues.
-        if let bsonValue = value as? BsonValue, !(bsonValue is [Any]) {
+    fileprivate func box<T: Encodable>(_ value: T) throws -> BSONValue? {
+        // if it's already a BSONValue, just return it, unless if it is an
```
jmikola:
```suggestion
        // if it's already a `BSONValue`, just return it, unless if it is an 
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -260,29 +260,29 @@ private class _BsonReferencingEncoder: _BsonEncoder {
...
     }
 
-    fileprivate func box<T: Encodable>(_ value: T) throws -> BsonValue? {
-        // if it's already a BsonValue, just return it, unless if it is an 
-        // array. technically [Any] is a BsonValue, but we can only use this
-        // short-circuiting if all the elements are actually BsonValues.
-        if let bsonValue = value as? BsonValue, !(bsonValue is [Any]) {
+    fileprivate func box<T: Encodable>(_ value: T) throws -> BSONValue? {
+        // if it's already a BSONValue, just return it, unless if it is an 
+        // array. technically [Any] is a BSONValue, but we can only use this
```
jmikola:
```suggestion
        // array. technically `[Any]` is a `BSONValue`, but we can only use this
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -260,29 +260,29 @@ private class _BsonReferencingEncoder: _BsonEncoder {
...
+        // short-circuiting if all the elements are actually BSONValues.
```
jmikola:
```suggestion
        // short-circuiting if all the elements are actually `BSONValue`s.
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -529,8 +529,8 @@ private class MutableArray: BsonValue {
 
     init() {}
 
-    /// methods required by the BsonValue protocol that we don't actually need/use. MutableArray
-    /// is just a BsonValue to simplify usage alongside true BsonValues within the encoder.
+    /// methods required by the BSONValue protocol that we don't actually need/use. MutableArray
```
jmikola:
```suggestion
    /// methods required by the `BSONValue` protocol that we don't actually need/use. `MutableArray`
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -529,8 +529,8 @@ private class MutableArray: BsonValue {
...
+    /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
```
jmikola:
```suggestion
    /// is just a `BSONValue` to simplify usage alongside true `BSONValue`s within the encoder.
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -579,8 +579,8 @@ private class MutableDictionary: BsonValue {
 
     init() {}
 
-    /// methods required by the BsonValue protocol that we don't actually need/use. MutableDictionary
-    /// is just a BsonValue to simplify usage alongside true BsonValues within the encoder.
+    /// methods required by the BSONValue protocol that we don't actually need/use. MutableDictionary
```
jmikola:
```suggestion
    /// methods required by the `BSONValue` protocol that we don't actually need/use. `MutableDictionary`
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -579,8 +579,8 @@ private class MutableDictionary: BsonValue {
...
+    /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
```
jmikola:
```suggestion
    /// is just a `BSONValue` to simplify usage alongside true `BSONValue`s within the encoder.
```

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -50,13 +50,13 @@ public enum BsonType: UInt32 {
     maxKey = 0x7f
 }
 
-/// A protocol all types representing BsonTypes must implement.
-public protocol BsonValue {
-    /// The `BsonType` of this value.
-    var bsonType: BsonType { get }
+/// A protocol all types representing BSONTypes must implement.
```
jmikola:
```suggestion
/// A protocol all types representing `BSONType`s must implement.
```

Sources/MongoSwift/BSON/CodableNumber.swift
```diff
@@ -34,10 +34,10 @@ extension CodableNumber {
         return nil
     }
 
-    /// By default, just try casting the number to a `BsonValue`. Types
-    /// where that will not work provide their own `asBsonValue` impl. 
-    var bsonValue: BsonValue? {
-        return self as? BsonValue
+    /// By default, just try casting the number to a `BSONValue`. Types
+    /// where that will not work provide their own `asBSONValue` impl.
```
jmikola: @kmahar: I realize this last sentence predates @Utagai's change, but is "`asBSONValue` impl" technically correct? I read this assuming it was a protocol, but it appears to be referring to a type being extended with a `var bsonValue: BsonValue?` computed property. Would it make more sense to refer to "their own `bsonValue` computed property"?

Tests/MongoSwiftBenchmarks/BsonBenchmarks.swift
```diff
@@ -3,16 +3,16 @@ import Foundation
 import XCTest
 
 let basePath = XCTestCase.getSpecsPath()
-let flatBsonFile = URL(fileURLWithPath: basePath + "/flat_bson.json")
+let flatBSONFile = URL(fileURLWithPath: basePath + "/flat_bson.json")
```
jmikola: Should these remain lowercase, since they're essentially variable/property names?

Or did we decide that "bson" would only be lowercase if it was the first segment of a variable/property name? I think we may have decided that when I brought up the `fromBSON` argument name remaining as-is.

Tests/MongoSwiftTests/Document+SequenceTests.swift
```diff
@@ -87,10 +87,10 @@ final class Document_SequenceTests: XCTestCase {
                             "decimal128", "minkey", "maxkey", "date", "timestamp"]
         for (k, v) in doc {
             expect(k).to(equal(expectedKeys.removeFirst()))
-            // we can't compare BsonValues for equality, nor can we cast v
+            // we can't compare BSONValues for equality, nor can we cast v
```
jmikola:
```suggestion
            // we can't compare `BSONValue`s for equality, nor can we cast v
```

Tests/MongoSwiftTests/Document+SequenceTests.swift
```diff
@@ -87,10 +87,10 @@ final class Document_SequenceTests: XCTestCase {
...
             // to a dynamically determined equatable type, so just verify
-            // it's a BsonValue anyway 
-            expect(v).to(beAKindOf(BsonValue.self))
+            // it's a BSONValue anyway
```
jmikola:
```suggestion
            // it's a `BSONValue` anyway 
```

Tests/MongoSwiftBenchmarks/BsonBenchmarks.swift
Utagai: I think the correct thing to do here is to keep this capitalized. The idea was to keep with camelcase, stay consistent and do so as simply as possible:

1. When 'bson' needs to be lowercased, it is 'bson'.
2. When 'bson' needs to be uppercased, it is 'BSON'.

Those variables can be immediately recognized as variable/property names because their very first word (`flat`) is lowercased. If the `flat` was not there, it would be `bsonFile`. Since the `flat` is there, we have `flatBSONFile`.

Let me know what you think about this.

Sources/MongoSwift/BSON/CodableNumber.swift
kmahar: yes, I think I initially had named it `asBsonValue` when I wrote the comment, and later changed the name. suggesting fix to clarify.
```suggestion
    /// where that will not work provide their own implementation of the `bsonValue` computed property. 
```
@Utagai, you might want to fix the line break position in this comment, since this suggestion is a lot longer than the line before. but I can't figure out how to suggest for both lines at once.

Tests/MongoSwiftBenchmarks/BsonBenchmarks.swift
kmahar: I think what May has here is good. basically, all the letters in BSON should always be a single case. 

this is consistent with Apple's code - for example, note [this method](https://github.com/apple/swift-corelibs-foundation/blob/3995b2cde03b7d12e3e8f8f783b6e43e6322d60a/Foundation/JSONSerialization.swift#L311) is `serializeJSON`, not `serializeJson`.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -280,11 +280,11 @@ public class DocumentIterator: IteratorProtocol {
         try newValue.writeToCurrentPosition(of: self)
     }
 
-    private static let BsonTypeMap: [BsonType: BsonValue.Type] = [
+    private static let BSONTypeMap: [BSONType: BSONValue.Type] = [
```
kmahar: this should probably be lowercase (my bad for the original naming.)
```suggestion
    private static let bsonTypeMap: [BSONType: BSONValue.Type] = [
```

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -206,24 +206,24 @@ public class DocumentIterator: IteratorProtocol {
...
+    internal var currentValue: BSONValue? {
         do {
             switch self.currentType {
             case .symbol:
                 return try Symbol.asString(from: self)
             case .dbPointer:
                 return try DBPointer.asDocument(from: self)
             default:
-                return try DocumentIterator.BsonTypeMap[currentType]?.init(from: self)
+                return try DocumentIterator.BSONTypeMap[currentType]?.init(from: self)
```
kmahar: to be consistent w/ my suggestion below
```suggestion
                return try DocumentIterator.bsonTypeMap[currentType]?.init(from: self)
```

Sources/MongoSwift/BSON/CodableNumber.swift
Utagai: Done.

Utagai:
@kmahar I had made sure to run the benchmarks completely before I submitted the initial PR.

I just ran `make benchmark` really quickly again to be on the safe side; I've seen the started output and a few of the benchmarks passed, so no complaints from the benchmarks it seems.

PR #124, opened 2018-10-22
Add Neovim/Vim related info to Development.md

Utagai:
Personally, I use Neovim, but the added info should work for Vim as well (actually, I think both of the mentioned things were made for Vim).

Finding a spot for tossing in the information about getting proper syntax and indenting for Swift in Vim was a bit of a challenge. I put it under 'Things to Install' since it is probably something you _need_ to install if you are indeed using Vim or Neovim, but other suggestions are welcome.

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -17,6 +17,10 @@
 * [swiftlint](https://github.com/realm/SwiftLint#using-homebrew): the Swift linter we use. 
 * [libmongoc](http://mongoc.org/libmongoc/current/api.html): the MongoDB C driver, which this library wraps. See the installation instructions provided in our [README](README.md#first-install-the-mongodb-c-driver) or on the [libmongoc docs](http://mongoc.org/libmongoc/current/installing.html). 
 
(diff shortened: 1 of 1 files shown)


Development.md
```diff
@@ -16,6 +16,9 @@
 * [jazzy](https://github.com/realm/jazzy#installation): the tool we use to generate documentation.
 * [swiftlint](https://github.com/realm/SwiftLint#using-homebrew): the Swift linter we use. 
 * [libmongoc](http://mongoc.org/libmongoc/current/api.html): the MongoDB C driver, which this library wraps. See the installation instructions provided in our [README](README.md#first-install-the-mongodb-c-driver) or on the [libmongoc docs](http://mongoc.org/libmongoc/current/installing.html). 
+### If you are using (Vim/Neovim)
```
jmikola: Extra newline between sections.

```suggestion

### If you are using (Vim/Neovim)
```

PR #125, opened 2018-10-22
SWIFT-201: Refactor "if failure" conditions with guard statements

Utagai:
[SWIFT-201](https://jira.mongodb.org/browse/SWIFT-201)

This PR finds places in the entire codebase where we should probably be using `guard` instead of an `if` statement.

Generally speaking, it seems like `guard` should be used to catch or, _guard_ against exceptional violations of (an) assumption(s) or invariant(s) in the rest of the current scope. Usually, this is involved with condition checks at the beginning of functions or protecting oneself from exceptional failures, e.g., calls to `mongoc_*` functions. Therefore, there may be a few cases where a `guard` is added for something that is not an explicit failure.

I've used the above idea to determine when to use `guard`. Since I essentially looked through the entire codebase, it is possible that I have missed some cases. If you think I've missed anything, just suggest it :).

### DIFF
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -17,7 +17,7 @@ extension MongoCollection {
      */
     @discardableResult
     public func bulkWrite(_ requests: [WriteModel], options: BulkWriteOptions? = nil) throws -> BulkWriteResult? {
-        if requests.isEmpty {
+        guard !requests.isEmpty else {
             throw MongoError.invalidArgument(message: "requests cannot be empty")
         }
 
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -84,8 +84,8 @@ extension MongoCollection {
         let reply = Document()
         var error = bson_error_t()
 
-        if !mongoc_collection_find_and_modify_with_opts(self._collection, filter.data,
-                                                        opts._options, reply.data, &error) {
+        guard mongoc_collection_find_and_modify_with_opts(self._collection, filter.data,
+                                                          opts._options, reply.data, &error) else {
             // TODO SWIFT-144: replace with more descriptive error type(s)
             throw MongoError.commandError(message: toErrorString(error))
         }
@@ -317,7 +317,7 @@ private class FindAndModifyOptions {
     /// Sets the `update` value on a `mongoc_find_and_modify_opts_t`. We need to have this separate from the 
     /// initializer because its value comes from the API methods rather than their options types.
     fileprivate func setUpdate(_ update: Document) throws {
-        if !mongoc_find_and_modify_opts_set_update(self._options, update.data) {
+        guard mongoc_find_and_modify_opts_set_update(self._options, update.data) else {
             throw MongoError.invalidArgument(message: "Error setting update to \(update)")
         }
     }
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -313,8 +313,7 @@ private class FindAndModifyOptions {
             }
         }
 
-        let appendSuccess = extra.keys.count <= 0 || mongoc_find_and_modify_opts_append(self._options, extra.data)
-        guard appendSuccess else {
+        guard extra.isEmpty || mongoc_find_and_modify_opts_append(self._options, extra.data) else {
             throw MongoError.invalidArgument(message: "Error appending extra fields \(extra)")
         }
     }
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -262,8 +262,10 @@ private class FindAndModifyOptions {
             throw MongoError.invalidArgument(message: "Error setting bypassDocumentValidation to \(bypass)")
         }
 
-        if let fields = projection, !mongoc_find_and_modify_opts_set_fields(self._options, fields.data) {
-            throw MongoError.invalidArgument(message: "Error setting fields to \(fields)")
+        if let fields = projection {
+            guard mongoc_find_and_modify_opts_set_fields(self._options, fields.data) else {
+                throw MongoError.invalidArgument(message: "Error setting fields to \(fields)")
+            }
         }
 
         // build a mongoc_find_and_modify_flags_t
@@ -282,8 +284,10 @@ private class FindAndModifyOptions {
                 "Error setting flags to \(flags); remove=\(remStr), upsert=\(upsStr), returnDocument=\(retStr)")
         }
 
-        if let sort = sort, !mongoc_find_and_modify_opts_set_sort(self._options, sort.data) {
-            throw MongoError.invalidArgument(message: "Error setting sort to \(sort)")
+        if let sort = sort {
+            guard mongoc_find_and_modify_opts_set_sort(self._options, sort.data) else {
+                throw MongoError.invalidArgument(message: "Error setting sort to \(sort)")
+            }
         }
 
         // build an "extra" document of fields without their own setters
@@ -309,7 +313,8 @@ private class FindAndModifyOptions {
             }
         }
 
-        if extra.keys.count > 0 && !mongoc_find_and_modify_opts_append(self._options, extra.data) {
+        let appendSuccess = extra.keys.count <= 0 || mongoc_find_and_modify_opts_append(self._options, extra.data)
+        guard appendSuccess else {
             throw MongoError.invalidArgument(message: "Error appending extra fields \(extra)")
         }
     }
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -212,7 +212,7 @@ extension MongoCollection {
         let opts = try encoder.encode(options)
         var error = bson_error_t()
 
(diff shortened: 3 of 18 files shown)


Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -309,15 +313,16 @@ private class FindAndModifyOptions {
             }
         }
 
-        if extra.keys.count > 0 && !mongoc_find_and_modify_opts_append(self._options, extra.data) {
+        let appendSuccess = extra.keys.count <= 0 || mongoc_find_and_modify_opts_append(self._options, extra.data)
```
kmahar: so, a few things for this line, most of which are artifacts and not related to your change. but I'll do my best to explain why we should fix them because it's a good lesson in Swift and Swift-C performance.

so in general, it's preferred to use `isEmpty` on [`Collection`](https://developer.apple.com/documentation/swift/collection) types in Swift (which includes `Array`) rather than comparing `.count` to 0. in short, the reason for this is that you can usually check in O(1) if *something* is there, but it takes O(n) to count *how many* things are there. 

ok, so that means we should use `extra.keys.isEmpty`. BUT, that's actually still O(n).... reason being that [`Document.keys`](https://github.com/mongodb/mongo-swift-driver/blob/e98bdf5806458caf911db56aa5e7aed67d41b5bd/Sources/MongoSwift/BSON/Document.swift#L32) is a computed property and it's created on the fly when we ask for it, by iterating over the document and converting all the C string keys to Swift strings. 

so, the most optimal thing here is to use the recently added  [`Document.isEmpty`](https://github.com/mongodb/mongo-swift-driver/blob/e98bdf5806458caf911db56aa5e7aed67d41b5bd/Sources/MongoSwift/BSON/Document%2BSequence.swift#L20) property. this works by just asking libbson directly if there is at least one element in the underlying `bson_t`.

hopefully that all makes sense.... if not let me know.

ok, finally... prefer just line-breaking to deal with long lines rather than introducing new variables that will only be used once.

so basically, I can't encompass this in a suggestion, but let's make this line something like
```swift
guard extra.isEmpty || mongoc_find_and_modify_opts_append(self._options, extra.data) else { // ... 
```

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -17,7 +17,7 @@ extension MongoCollection {
      */
     @discardableResult
     public func bulkWrite(_ requests: [WriteModel], options: BulkWriteOptions? = nil) throws -> BulkWriteResult? {
-        if requests.isEmpty {
+        guard !requests.isEmpty else {
```
jmikola: `guard !...` stood out in contrast to most of the other changes in this PR -- but I think it's still preferable to something like `guard requests.count > 0`.

Side observation while I was looking into this: Both [`Array.isEmpty`]((https://developer.apple.com/documentation/swift/array/1688398-isempty)) and `.count` are constant-time. [Collection.count](https://developer.apple.com/documentation/swift/array/2949997-count) is either O(1) or O(n) depending on whether the implementation supports random access, which Array does.

Anywho, this is fine as-is. Just wanted to record my thought process :)

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -262,8 +262,10 @@ private class FindAndModifyOptions {
             throw MongoError.invalidArgument(message: "Error setting bypassDocumentValidation to \(bypass)")
         }
 
-        if let fields = projection, !mongoc_find_and_modify_opts_set_fields(self._options, fields.data) {
-            throw MongoError.invalidArgument(message: "Error setting fields to \(fields)")
+        if let fields = projection {
+            guard mongoc_find_and_modify_opts_set_fields(self._options, fields.data) else {
```
jmikola: I think this is much clearer than mixing assignments and a function call in a comma-delimited `if` statement. :+1:

Sources/MongoSwift/MongoCollection+FindAndModify.swift
jmikola: AFAIK, `Document.keys` returns `[String]`. In this case, would it make more sense to use `isEmpty` rather than `<= 0`? I imagine `<=` was probably used to be defensive, but I think we're only checking whether or not there are any keys at all.

This might also be clearer if written like the blocks above:

```
if !extra.keys.isEmpty {
    guard mongoc_find_and_modify_opts_append(...) else {
        throw ...
    }
}
```

Alternatively, we can craft a single guard statement:

```
guard extra.keys.isEmpty || mongoc_find_and_modify_opts_append(...) else {
   throw ...
}
```

In this case, `mongoc_find_and_modify_opts_append()` would never be called if the keys were empty due to short-circuiting.

I'm happy to defer to @kmahar as to which is more Swifty. The first is a bit more verbose but is probably the most readable.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -284,7 +284,8 @@ extension MongoCollection {
         let opts = try BSONEncoder().encode(options)
         let reply = Document()
         var error = bson_error_t()
-        if !mongoc_collection_write_command_with_opts(self._collection, command.data, opts?.data, reply.data, &error) {
+        guard mongoc_collection_write_command_with_opts(self._collection, command.data, opts?.data, reply.data, &error)
+        else {
```
jmikola: Was `else` put on a new line by the linter? If it's being wrapped I might have expected some indentation here, although I honestly haven't encountered this case before so I'm not sure how linting handles it.

Curious if the syntax from `MongoCollection+Read.swift` may be the way to go. For example:

```
guard mongoc_collection_read_command_with_opts(
    self._collection, command.data, rp, opts?.data, reply.data, &error) else {
```

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -26,7 +26,7 @@ final class MongoClientTests: XCTestCase {
         }
 
         let client_t = mongoc_client_new_from_uri(uri)
-        if client_t == nil {
+        guard client_t != nil else {
```
jmikola: I wonder if we can combine this with the assignment above. I believe `guard let client_t = mongoc_client_new_from_uri(uri)` should only fail if `nil` is returned.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: there's been [some discussion](https://forums.swift.org/search?q=isnotempty) of adding `.isNotEmpty` in the past... but verdict seems to be it's not gonna happen

Sources/MongoSwift/MongoCollection+FindAndModify.swift
kmahar: lol oops didn't see your comment came in as I was writing mine :) 

I'm fine with either suggestion of yours, so long as we use `isEmpty` directly on the doc rather than on the keys array, for the reasons explained in my above comment.

kmahar: I've opened [SWIFT-229](https://jira.mongodb.org/browse/SWIFT-229) about looking for these cases throughout the whole codebase

Sources/MongoSwift/MongoCollection+Indexes.swift
kmahar: this is something else I want a linter rule for... I am rather inconsistent and sometimes break after the open parens and sometimes break mid argument list. but I think the former is probably better.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
Utagai: It is always odd to me to see things like negations of library functions like `isEmpty -> isNonEmpty` being rejected. The usual arguments of trying to only add functions when absolutely necessary just doesn't convince me. There _is_ a cognitive burden to having to do `!isEmpty()` because you have to catch the `!`, which I'm sure we have all at some point missed. `isNonEmpty()` just reads so much better because its just straight up natural language.

In either case, @jmikola, feel free to resolve this since I think we want to follow the insight Kaitlin got from Jeff. :).

Sources/MongoSwift/MongoCollection+Indexes.swift
Utagai: Yeah, this was something that the linter did not care about. My guess is that this is because it is seeing it the same way it sees `if` and `else`:
```
if { ... }
else { ... }
```
Where `if` and `else` is aligned. But I like the syntax you've shown from `MongoCollection+Read.swift` better.

Tests/MongoSwiftTests/MongoClientTests.swift
Utagai: We can, and we did :).

Sources/MongoSwift/MongoCollection+FindAndModify.swift
Utagai: @jmikola Yup, this is waiting on PR#126 from Kaitlin. So maybe feel free to resolve this thread and we can use Kaitlin's comment below as the thread for this line of the PR?

kmahar: #126 merged

Utagai: What you said makes perfect sense. TIL!!

The addition of `appendSuccess` wasn't made for the purpose of shortening the line length though. I did that because generally I tend to assign longer boolean expressions to variables so it becomes more readable. i.e. `appendSuccess` is more immediately obvious in meaning than `extra.keys.count <= 0 || mongoc_find_and_modify_opts_append(self._options, extra.data)`. However, it seems like this is not good Swift, so I've reverted that.

I've also included the suggestion to switch over to `extra.isEmpty`.

PR #127, opened 2018-10-24
SWIFT-232: rename `libbson`/`libmongoc` references

mbroadst:
https://jira.mongodb.org/browse/SWIFT-232

The platforms team renamed these frameworks during their attempts to frameworkify them for mobile builds. It's much easier for us to use their new names than to make the change on the other end, so
here's the rename!

**NOTE**: sorry, I updated documentation hoping to catch any references to libmongoc/libbson and it looks like this hadn't been done since at least @jmikola's bulk changes - so you're getting those here too :)


Package.swift
```diff
@@ -6,12 +6,12 @@ let package = Package(
         .library(name: "MongoSwift", targets: ["MongoSwift"])
     ],
     dependencies: [
-        .package(url: "https://github.com/mongodb/swift-bson", from: "1.0.0"),
-        .package(url: "https://github.com/mongodb/swift-mongoc", from: "1.0.0"),
+        .package(url: "https://github.com/mongodb/swift-bson", .branch("SWIFT-232/update-exported-name")),
```
mbroadst: these will change to proper tags once the PRs for those changes over there are accepted, merged, and tagged

Makefile
```diff
@@ -39,4 +39,4 @@ clean:
 documentation:
 	make project
 	@gem list jazzy -i > /dev/null || gem install jazzy || { echo "ERROR: Failed to locate or install the ruby gem jazzy; please install yourself with 'gem install jazzy' (you may need to use sudo)"; exit 1; }
-	jazzy --module MongoSwift --module-version 0.0.2 --root-url https://mongodb.github.io/mongo-swift-driver/ --documentation Development.md
+	jazzy --module MongoSwift --module-version 0.0.3 --root-url https://mongodb.github.io/mongo-swift-driver/ --documentation Development.md
```
jmikola: Since 0.0.3 is already tagged and we'll need to create a new tag following this change, shouldn't this be 0.0.4?

PR #130, opened 2018-10-25
SWIFT-226: Provide a way to compare equality of BSONValues

Utagai:
[SWIFT-226](https://jira.mongodb.org/browse/SWIFT-226)

This PR adds a convenience function, `bsonEquals` that lets the caller compare two arbitrary `BSONValue`s. This does _not_ add `Equatable` conformance to `BSONValue`, because this would cause us to provide an `Equatable` conformance to all arrays. I, or @kmahar can provide further detail on this if necessary.

This PR also adds many test cases for checking the code paths through `bsonEquals`, but does not check for `preconditionFailure`s, since these cause the program, including the tests, to halt. There is apparently [a way](https://www.cocoawithlove.com/blog/2016/02/02/partial-functions-part-two-catching-precondition-failures.html) to get around this, but it is apparently very hacky.

I've included some comments on my own patch here, that I'd like to draw attention to and get feedback for in particular, as I'm not experience enough in Swift to know if they are good practices.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -812,7 +812,10 @@ func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     case (let l as [BSONValue?], let r as [BSONValue?]):
         return try zip(l, r).reduce(true, {prev, next in try bsonEquals(lhs: next.0, rhs: next.1) && prev})
     case (_ as [Any], _ as [Any]):
-        preconditionFailure("arrays not of type [BSONValue?] should not be compared with bsonEqual()")
+        enum BSONEqualsError: Error { case InvalidArrayArgument(String) }
+        throw BSONEqualsError.InvalidArrayArgument(
+            "arrays not of type [BSONValue?] should not be compared with bsonEqual()"
+        )
     default:
         // For BSONTypes that do not have actual classes/structs associated
         // with them, the behavior is known regardless.
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -795,7 +795,7 @@ public struct Timestamp: BSONValue, Equatable, Codable {
  *   - lhs: The left-hand-side BSONValue to compare.
  *   - rhs: The right-hand-side BSONValue to compare.
  *
- * - Returns: True if lhs is equal to rhs, false otherwise.
+ * - Returns: `true` if `lhs` is equal to `rhs`, `false` otherwise.
  */
 func bsonEquals(_ lhs: BSONValue, _ rhs: BSONValue) -> Bool {
     validateBSONTypes(lhs, rhs)
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -781,12 +781,16 @@ public struct Timestamp: BSONValue, Equatable, Codable {
 
 enum BSONEqualsError: Error { case InvalidArrayArgument(String) }
 
+
 // See https://github.com/realm/SwiftLint/issues/461
 // swiftlint:disable cyclomatic_complexity
-// A helper function to test equality between two BSONValues. This function tests for exact BSON equality.
-// This means that, for example, differing types with equivalent value are not equivalent.
-// e.g.
-//  4.0 (Double) != 4 (Int)
+/**
+ *  A helper function to test equality between two BSONValues. This function tests for exact BSON equality.
+ *  This means that differing types with equivalent value are not equivalent.
+ *
+ *  e.g.
+ *      4.0 (Double) != 4 (Int)
+ */
 func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     validateBSONTypes(lhs: lhs, rhs: rhs)
 
@@ -817,8 +821,8 @@ func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     }
 }
 
-// A helper function to test equality between two BSONValue?s. See bsonEquals for BSONValues (non-optional) for more
-// information.
+/// A helper function to test equality between two BSONValue?s. See bsonEquals for BSONValues (non-optional) for more
+/// information.
 func bsonEquals(lhs: BSONValue?, rhs: BSONValue?) throws -> Bool {
     guard let left = lhs, let right = rhs else {
         return lhs == nil && rhs == nil
@@ -827,8 +831,8 @@ func bsonEquals(lhs: BSONValue?, rhs: BSONValue?) throws -> Bool {
     return try bsonEquals(lhs: left, rhs: right)
 }
 
-// A function for catching invalid BSONTypes that should not ever arise, and triggering a preconditionFailure when it
-// finds such types.
+/// A function for catching invalid BSONTypes that should not ever arise, and triggering a preconditionFailure when it
+/// finds such types.
 internal func validateBSONTypes(lhs: BSONValue, rhs: BSONValue) {
     let invalidTypes: [BSONType] = [.symbol, .dbPointer, .invalid, .undefined, .null]
     guard !invalidTypes.contains(lhs.bsonType) else {
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -797,8 +797,8 @@ public struct Timestamp: BSONValue, Equatable, Codable {
  *
  * - Returns: True if lhs is equal to rhs, false otherwise.
  */
-func bsonEquals(lhs: BSONValue, rhs: BSONValue) -> Bool {
-    validateBSONTypes(lhs: lhs, rhs: rhs)
+func bsonEquals(_ lhs: BSONValue, _ rhs: BSONValue) -> Bool {
+    validateBSONTypes(lhs, rhs)
 
     switch (lhs, rhs) {
     case (let l as Int, let r as Int): return l == r
@@ -818,7 +818,7 @@ func bsonEquals(lhs: BSONValue, rhs: BSONValue) -> Bool {
     case (let l as Binary, let r as Binary): return l == r
     case (let l as Document, let r as Document): return l == r
     case (let l as [BSONValue?], let r as [BSONValue?]): // TODO: SWIFT-242
-        return zip(l, r).reduce(true, {prev, next in bsonEquals(lhs: next.0, rhs: next.1) && prev})
+        return zip(l, r).reduce(true, {prev, next in bsonEquals(next.0, next.1) && prev})
     case (_ as [Any], _ as [Any]): return false
     default: return false
     }
@@ -834,17 +834,17 @@ func bsonEquals(lhs: BSONValue, rhs: BSONValue) -> Bool {
  *
  * - Returns: True if lhs is equal to rhs, false otherwise.
  */
-public func bsonEquals(lhs: BSONValue?, rhs: BSONValue?) -> Bool {
+public func bsonEquals(_ lhs: BSONValue?, _ rhs: BSONValue?) -> Bool {
     guard let left = lhs, let right = rhs else {
         return lhs == nil && rhs == nil
     }
 
-    return bsonEquals(lhs: left, rhs: right)
+    return bsonEquals(left, right)
 }
 
 /// A function for catching invalid BSONTypes that should not ever arise, and triggering a preconditionFailure when it
 /// finds such types.
-internal func validateBSONTypes(lhs: BSONValue, rhs: BSONValue) {
+fileprivate func validateBSONTypes(_ lhs: BSONValue, _ rhs: BSONValue) {
     let invalidTypes: [BSONType] = [.symbol, .dbPointer, .invalid, .undefined, .null]
     guard !invalidTypes.contains(lhs.bsonType) else {
         preconditionFailure("\(lhs.bsonType) should not be used")
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -779,17 +779,17 @@ public struct Timestamp: BSONValue, Equatable, Codable {
     }
 }
 
-func bsonEqual(lhs: BSONValue?, rhs: BSONValue?) throws -> Bool {
+func bsonEquals(lhs: BSONValue?, rhs: BSONValue?) throws -> Bool {
     guard let left = lhs, let right = rhs else {
         return lhs == nil && rhs == nil
     }
 
-    return try bsonEqual(lhs: left, rhs: right)
+    return try bsonEquals(lhs: left, rhs: right)
 }
 
 // See https://github.com/realm/SwiftLint/issues/461
 // swiftlint:disable cyclomatic_complexity
-func bsonEqual(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
+func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     validateBSONTypes(lhs: lhs, rhs: rhs)
 
     switch (lhs, rhs) {
@@ -810,7 +810,7 @@ func bsonEqual(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     case (let l as Binary, let r as Binary): return l == r
     case (let l as Document, let r as Document): return l == r
     case (let l as [BSONValue?], let r as [BSONValue?]):
-        return try zip(l, r).reduce(true, {prev, next in try bsonEqual(lhs: next.0, rhs: next.1) && prev})
+        return try zip(l, r).reduce(true, {prev, next in try bsonEquals(lhs: next.0, rhs: next.1) && prev})
     case (_ as [Any], _ as [Any]):
         preconditionFailure("arrays not of type [BSONValue?] should not be compared with bsonEqual()")
     default:
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -790,6 +790,12 @@ enum BSONEqualsError: Error { case InvalidArrayArgument(String) }
  *
  *  e.g.
  *      4.0 (Double) != 4 (Int)
+ *
+ *  * - Parameters:
+ *   - lhs: The left-hand-side BSONValue to compare.
+ *   - rhs: The right-hand-side BSONValue to compare.
+ *
+ * - Returns: True if lhs is equal to rhs, false otherwise.
  */
 func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     validateBSONTypes(lhs: lhs, rhs: rhs)
@@ -821,8 +827,16 @@ func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
     }
 }
 
-/// A helper function to test equality between two BSONValue?s. See bsonEquals for BSONValues (non-optional) for more
-/// information.
+/**
+ *  A helper function to test equality between two BSONValue?s. See bsonEquals for BSONValues (non-optional) for more
+ *  information.
+ *
+ *  * - Parameters:
+ *   - lhs: The left-hand-side BSONValue? to compare.
+ *   - rhs: The right-hand-side BSONValue? to compare.
+ *
+ * - Returns: True if lhs is equal to rhs, false otherwise.
+ */
 public func bsonEquals(lhs: BSONValue?, rhs: BSONValue?) throws -> Bool {
     guard let left = lhs, let right = rhs else {
         return lhs == nil && rhs == nil
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -844,7 +844,7 @@ public func bsonEquals(_ lhs: BSONValue?, _ rhs: BSONValue?) -> Bool {
 
 /// A function for catching invalid BSONTypes that should not ever arise, and triggering a preconditionFailure when it
 /// finds such types.
-fileprivate func validateBSONTypes(_ lhs: BSONValue, _ rhs: BSONValue) {
+private func validateBSONTypes(_ lhs: BSONValue, _ rhs: BSONValue) {
     let invalidTypes: [BSONType] = [.symbol, .dbPointer, .invalid, .undefined, .null]
     guard !invalidTypes.contains(lhs.bsonType) else {
         preconditionFailure("\(lhs.bsonType) should not be used")
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -788,8 +788,8 @@ public struct Timestamp: BSONValue, Equatable, Codable {
  *  e.g.
  *      4.0 (Double) != 4 (Int)
  *
- *  NOTE: This function will always return `false` if it is used with two arrays that are not of the type `[BSONValue]` or
- *  `[BSONValue?]`, because any arrays that are not of these types are not valid `BSONValue`'s.
+ *  NOTE: This function will always return `false` if it is used with two arrays that are not of the type `[BSONValue]`
+ *  or `[BSONValue?]`, because any arrays that are not of these types are not valid `BSONValue`'s.
  *
  *  * - Parameters:
  *   - lhs: The left-hand-side `BSONValue` to compare.
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -816,20 +816,12 @@ func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
         throw BSONEqualsError.InvalidArrayArgument(
             "arrays not of type [BSONValue?] should not be compared with bsonEqual()"
         )
-    default:
-        // For BSONTypes that do not have actual classes/structs associated
-        // with them, the behavior is known regardless.
-        switch (lhs.bsonType, rhs.bsonType) {
-        case (.invalid, .invalid): return true
-        case (.undefined, .undefined): return true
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,66 @@ public struct Timestamp: BSONValue, Equatable, Codable {
     }
 }
 
+enum BSONEqualsError: Error { case InvalidArrayArgument(String) }
```
Utagai: The question of whether to throw an error for arrays or trigger a `preconditionFailure` is something I had to wrestle with. I settled on throwing an error, because it seemed excessive to halt the user's application if they called this function unaware of the fact that, even though `Array` does conform to `BSONValue`, our function accepting two `BSONValue`'s does not accept all kinds of Arrays. This is not really the same as feeding in a `bsonType` that should never really arise, such as `.undefined`, so an error seemed much more tame.

The downside is that we add a lone, random error case that is used and thrown only once in the entire codebase, and, now the user must wrap their calls to `bsonEquals()` with a `try`.

Feedback is welcome. :)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,66 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+        throw BSONEqualsError.InvalidArrayArgument(
+            "arrays not of type [BSONValue?] should not be compared with bsonEqual()"
+        )
+    default: return false
+    }
+}
+
+// A helper function to test equality between two BSONValue?s. See bsonEquals for BSONValues (non-optional) for more
+// information.
+func bsonEquals(lhs: BSONValue?, rhs: BSONValue?) throws -> Bool {
```
Utagai: This auxiliary method overloads the `bsonEquals()` implementation above. This seemed like a good idea for users that have optional `BSONValues` that they wished to compare, since those distinctly different types in Swift. This also makes the `[BSONValue?]` case much cleaner.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,66 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+    case (let l as Timestamp, let r as Timestamp): return l == r
+    case (let l as Date, let r as Date): return l == r
+    case (_ as MinKey, _ as MinKey): return true
+    case (_ as MaxKey, _ as MaxKey): return true
+    case (let l as ObjectId, let r as ObjectId): return l == r
+    case (let l as CodeWithScope, let r as CodeWithScope): return l == r
+    case (let l as Binary, let r as Binary): return l == r
+    case (let l as Document, let r as Document): return l == r
+    case (let l as [BSONValue?], let r as [BSONValue?]):
+        return try zip(l, r).reduce(true, {prev, next in try bsonEquals(lhs: next.0, rhs: next.1) && prev})
```
Utagai: This is a nice functional implementation for the `[BSONValue?]` case, but perhaps making this iterative over the result of `zip(l, r)` is the more clean/readable/Swift-y approach.

I'm not sure what the Swift community's feelings are towards functional implementations like these. It's pretty easily readable for me, but I'm partial to functional code. :)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,70 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+    case (_ as [Any], _ as [Any]):
+        throw BSONEqualsError.InvalidArrayArgument(
+            "arrays not of type [BSONValue?] should not be compared with bsonEqual()"
+        )
```
jmikola: Is there a particular reason to raise an error instead of rely on the default case?

Alternatively, is this something `validateBSONTypes()` should catch? I would have thought a `[BSONValue?]` _is_ a valid BSON type, while an arbitrary `[Any]` is not.

Sources/MongoSwift/BSON/BSONValue.swift
jmikola: Would we also need `bsonEquals()` to an optional `lhs` and non-optional `rhs`, and vice versa?

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -27,4 +28,89 @@ final class BSONValueTests: XCTestCase {
...
+            } else {
+                return PredicateResult(
+                    status: .fail,
+                    message: msg.appendedBeNilHint()
+                )
+            }
+        }
+    }
+
+    func checkTrueAndFalse(val: BSONValue, alternate: BSONValue) {
```
jmikola: I'm not sure if we're strict about this across our test suite, but should we stick to `fileprivate` or some other limited visibility here? Likely also for `bsonEqual()` above, too.

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -27,4 +28,89 @@ final class BSONValueTests: XCTestCase {
...
+        // Timestamp
+        checkTrueAndFalse(val: Timestamp(timestamp: 1, inc: 2), alternate: Timestamp(timestamp: 5, inc: 10))
+        // Date
+        checkTrueAndFalse(
+            val: Date(timeIntervalSinceReferenceDate: 5000),
+            alternate: Date(timeIntervalSinceReferenceDate: 5001)
+        )
+        // MinKey & MaxKey
+        expect(MinKey()).to(bsonEqual(expectedValue: MinKey()))
+        expect(MaxKey()).to(bsonEqual(expectedValue: MaxKey()))
```
jmikola: Any reason not to use `checkTrueAndFalse()` here? I think we're skipping the `false` case for both types as-is.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Since `bson_compare` seems like it is calling a function from `mongoc`, it is totally reasonable to imagine that we'd see a performance increase from using it. However, instantiating a new document using the array may be expensive enough to not make a difference. If we can use `bson_compare` directly however, it may be a better move, but it seems like the best we can do currently is use `encode` to append it onto a `Document` and use `Document`'s `==` operator.

In other words, it seems like this boils down to whether the initialization of a `Document` for holding an array for comparison, and its storage, is fast enough to still get gains from calling a `mongoc` function.

We could try doing some low quality benchmarking to see which one wins.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Basically, the issue here is what I mentioned about making the single `BSONEqualsError` with `InvalidArrayArgument`. The other option is to let such an array trigger a precondition failure. However, the ultimate judgment I made is that this is too harsh of a reaction, _especially_ because we have extended arrays to conform to `BSONValue`.

The difference that I noticed between the `[Any]` and other invalid types is principally this:
You should never be able to reach those preconditionFailures unless something really weird has happened or you've done some hackery. `[Any]` however, is something you could totally have passed into the function, even as a mistake. Throwing an error at least doesn't immediately halt your application.

But of course, this wasn't something I was ever very decisive about. I ended up choosing this approach, but I saw merits to both sides. I agree that arbitrary `[Any]` shouldn't be a valid `BSONValue`, and also, considering it as a `preconditionFailure()` case would mean we don't have to `throws` for `bsonEquals()` which is a great plus for the driver's ergonomics.

Let me know what you guys think on this, I'm very on-the-fence as it is.

Tests/MongoSwiftTests/BSONValueTests.swift
Utagai: `MinKey` and `MaxKey` are both... interesting types. The `MinKey` type only ever has a single value, so the equality operation on it is essentially vacuously true. This of course also applies to `MaxKey`.

So, testing something like `MaxKey =?= MinKey` or vice-versa, is really just testing the differing types case, which is tested in `BSONValueTests.swift` in the last few lines:
```
// Different types
 expect(4).toNot(bsonEqual(expectedValue: "swift"))
```

Tests/MongoSwiftTests/BSONValueTests.swift
Utagai: I think this is a good idea. Reducing the visibility of functions like these that are only ever used here should only help us out. I believe I just genuinely missed these when going over the added functions and adding visibilities, so thanks for this catch.

This is done in 69c9dea.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Ah, I didn't see @kmahar's response when I posted my response, so there may be some repetition, but seems like this isn't obviously better or worse.

Should we try some simple benchmarking?

jmikola: This would afford you an opportunity to play with the benchmarking facilities in XCTest, so it sounds worthwhile. From a maintainability standpoint, I think the code that wraps the array operands in small documents and uses `==` is more straightforward. Famous last words, but I don't think this will be a particularly hot code path in the driver.

That said, isn't Array already Equatable as of Swift 4.1 (see: [Conditional Conformance](https://swift.org/blog/conditional-conformance/)) if all of its members conform? If so, can we just rely on `==`? Or does this take us back to [SWIFT-226](https://jira.mongodb.org/browse/SWIFT-226), where we can't rely on BSONValue being Equatable since it's not a concrete type?

Sources/MongoSwift/BSON/BSONValue.swift
jmikola: In that case, would it make sense to do everything under a single `bsonEquals(lhs: BSONValue?, rhs: BSONValue?)` method?

Utagai: It should be fine to do it that way too. I just had them separated because it separated the logic for handling optional arguments away from the actual logic of handling the equality checking of `BSONValue`.

Do you think we should just merge these two functions?

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Yes, in 4.1+, arrays in Swift have conditional conformance. However, conditional conformance only works if the members conform to `Equatable` as you've said. The members in this case are `BSONValue`, which _do not_ conform to `Equatable`.

This may make you wonder why we don't just make `BSONValue` conform to this. However, if we did this, then, since arrays are extended to conform to `BSONValue`, doing `arr1 == arr2` would call the implementation for `==` of `BSONValue`, and _not_ trigger conditional conformance. Conditional conformance only kicks in if the container type itself does not conform to `Equatable`.

Sources/MongoSwift/BSON/BSONValue.swift
jmikola: > `[Any]` however, is something you could totally have passed into the function, even as a mistake.

My first question was why we need to raise any error at all instead of allowing the default case to return `false`. It sounds like the goal is to inform the user that they've made a mistake rather than quietly report inequality for two `[Any]` values that might otherwise be considered equal via their `==` operator. I'm also on the fence if handling that edge case is worth making `bsonEquals()` throw. If @kmahar doesn't object, I'd probably vote for removing this case, the `throws`, and simply letting the function return `false`.

If not: I know the earlier conversation about the error is since resolved, but giving it some more thought I think we'd do well to stick to `MongoError.invalidArgument()` here for the time being. That would give us one less thing to worry about when revisiting the error hierarchy down the line in [SWIFT-145](https://jira.mongodb.org/browse/SWIFT-145).

kmahar: yeah, thinking about it more, I think that making this throw and forcing users to `try` for the array case may be a little excessive, given that *most* of the time, users won't be passing in arrays. Also, down the road, we'll be able to drop versions of Swift that don't support conditional conformance, and make it so that arrays are only BSONValue when their elements are also BSONValues. In which case this won't need to throw anymore at all. 

I also could be wrong, but my intuition is that most of the time when people are working with an `[BSONValue?]`/`[BSONValue]`, it's because they looked up some array value from a document. Which means that in most cases, this won't be a problem anyway. 

However, since users could plausibly make this error, preconditionFailure does seem harsh. 

So in short, I support @jmikola's point about just returning false here as a happy medium. If you'd like, we could add a note in a docstring about the behavior with arrays.

Utagai: @jmikola Ah, got it, my bad. Actually, an earlier version of this function just returned `false` in this case. I gravitated away from `false` because it implies that we actually have some idea about the equality of the two arguments, when in reality we are just always returning false. This seems like a bad idea, especially since there really is no way to realize this is wrong until users get some confusing behavior from the function.

I think if we want to ditch the throws, the better idea is actually using a `preconditionFailure()`, since at least we can give a useful message about what went wrong and even if its a bit of an extreme reaction, it is solvable for the user and is not hiding anything behind a seemingly innocuous return value.

The idea of using a `MongoError.invalidArgument()` occurred to me as well, but it seemed like `MongoError`'s were about a different thing. But I can totally see how that really helps us out with SWIFT-125.

(!!!) An idea I just had is: can we make it so that we only apply the extension to Array if the elements of the Array conform to `BSONValue`? Does something like this do what we want and seem promising to anyone?:

```
extension Array: BSONValue where Element: BSONValue {
    ...
}
```

This would actually solve our issues by:
1. No longer weirdly calling `[Any]` a `BSONValue` when it probably really isn't one.
2. No longer have to deal with this case.

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -27,4 +28,89 @@ final class BSONValueTests: XCTestCase {
         expect(try Binary(data: sixteenBytes, subtype: .uuidDeprecated)).toNot(throwError())
         expect(try Binary(data: sixteenBytes, subtype: .uuid)).toNot(throwError())
     }
+
+    fileprivate func bsonEqual(expectedValue: BSONValue?) -> Predicate<BSONValue> {
```
kmahar: 1) AFAIK, no existing predicates require parameter labels. let's follow that style and add `_ `  before `expectedValue`, and just have the syntax be `expect(x).to(bsonEqual(y))`.

2) let's move this into `TestUtils.swift`. It will be useful throughout all of our tests. 

at some point we should update tests to make them less verbose and use this. You'll notice a lot of places where I do something like
```swift
expect(doc["a"] as? Int).to(equal(5))`
```

but now we can do
```swift
expect(doc["a"]).to(bsonEqual(5))
```

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,70 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+// See https://github.com/realm/SwiftLint/issues/461
+// swiftlint:disable cyclomatic_complexity
+/**
+ *  A helper function to test equality between two BSONValues. This function tests for exact BSON equality.
+ *  This means that differing types with equivalent value are not equivalent.
+ *
+ *  e.g.
+ *      4.0 (Double) != 4 (Int)
+ */
+func bsonEquals(lhs: BSONValue, rhs: BSONValue) throws -> Bool {
```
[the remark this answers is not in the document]
Utagai: TIL; done in f6f12c3.

This is a little weird because it was still accessible in the tests, which made me think that this was `public` by default. However, I see now that this is due to the `@testable import`!

Awesome new tidbit I learned about Swift.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: aha, overlapping comments. yeah, so eventually we'll use your idea there, but we can't atm until we drop swift 4.0. and swift 4.1 conditional conformance isn't quite complete either; there's some issue around dynamically checking for conformance that I can't quite remember right now... could be a workaround, but I didn't try too hard as we are still supporting 4.0.

so basically, until we drop <4.2 or maaaaybe <4.1 we're kinda stuck with the current state of affairs. 

but yes, definitely a promising idea :)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,70 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+
+
+// See https://github.com/realm/SwiftLint/issues/461
+// swiftlint:disable cyclomatic_complexity
+/**
+ *  A helper function to test equality between two BSONValues. This function tests for exact BSON equality.
+ *  This means that differing types with equivalent value are not equivalent.
+ *
+ *  e.g.
+ *      4.0 (Double) != 4 (Int)
```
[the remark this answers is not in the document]
Utagai: Done in a12e1e1.

I left the `internal` helper function alone, but let me know if we should add explicit parameter and return values for that function as well.

Tests/MongoSwiftTests/BSONValueTests.swift
Utagai: Great, done in e81efdb.

I'll make a ticket for updating the other parts of the tests using `expect()` and edit this comment to include a link.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Aw :(. Hopefully we can move to this in the future then.

I'll make it return false for that case and get rid of the `throws`. I'll also update the documentation to include a note about this behavior.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,83 @@ public struct Timestamp: BSONValue, Equatable, Codable {
     }
 }
 
+enum BSONEqualsError: Error { case InvalidArrayArgument(String) }
```
jmikola: Since `bsonEquals()` no longer throws, this can be deleted.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,83 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+    case (let l as Timestamp, let r as Timestamp): return l == r
+    case (let l as Date, let r as Date): return l == r
+    case (_ as MinKey, _ as MinKey): return true
+    case (_ as MaxKey, _ as MaxKey): return true
+    case (let l as ObjectId, let r as ObjectId): return l == r
+    case (let l as CodeWithScope, let r as CodeWithScope): return l == r
+    case (let l as Binary, let r as Binary): return l == r
+    case (let l as Document, let r as Document): return l == r
+    case (let l as [BSONValue?], let r as [BSONValue?]):
+        return zip(l, r).reduce(true, {prev, next in bsonEquals(lhs: next.0, rhs: next.1) && prev})
```
jmikola: Fine to leave this as-is, but I might suggest adding `// TODO: Check if wrapping arrays and using Document `==` is more performant` or creating a ticket to follow-up later.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,83 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+ *  NOTE: This function will always return false if it is used with two arrays that are not of the type [BSONValue] or
+ *  [BSONValue?], because any arrays that are not of these types are not valid `BSONValue`'s.
+ *
+ *  * - Parameters:
+ *   - lhs: The left-hand-side BSONValue to compare.
+ *   - rhs: The right-hand-side BSONValue to compare.
+ *
+ * - Returns: True if lhs is equal to rhs, false otherwise.
+ */
+func bsonEquals(lhs: BSONValue, rhs: BSONValue) -> Bool {
```
jmikola: Do these actually need to use named parameters when calling `bsonEquals()`? I think we can prefix these with `_` and allow users to rely on their position without having to specify `lhs` and `rhs`.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,83 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+    guard let left = lhs, let right = rhs else {
+        return lhs == nil && rhs == nil
+    }
+
+    return bsonEquals(lhs: left, rhs: right)
+}
+
+/// A function for catching invalid BSONTypes that should not ever arise, and triggering a preconditionFailure when it
+/// finds such types.
+internal func validateBSONTypes(lhs: BSONValue, rhs: BSONValue) {
```
jmikola: Likewise here, perhaps we can ditch the argument labels from the caller's perspective.

jmikola: On a separate note, if this is only used within this file we can use `fileprivate` visibility.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Done in [SWIFT](https://jira.mongodb.org/browse/SWIFT-242)

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Done as well!

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Sure, done.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,81 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+    default: return false
+    }
+}
+
+/**
+ *  A helper function to test equality between two BSONValue?s. See bsonEquals for BSONValues (non-optional) for more
+ *  information.
+ *
+ *  * - Parameters:
+ *   - lhs: The left-hand-side BSONValue? to compare.
```
Utagai: Now that these parameters have had their labels removed, does the documentation have to change somehow?

jmikola: I think this is fine. Looking at Collection's `insertMany()` method, its first argument uses `_` but the docs still show the argument name.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,81 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
+    case (let l as Timestamp, let r as Timestamp): return l == r
+    case (let l as Date, let r as Date): return l == r
+    case (_ as MinKey, _ as MinKey): return true
+    case (_ as MaxKey, _ as MaxKey): return true
+    case (let l as ObjectId, let r as ObjectId): return l == r
+    case (let l as CodeWithScope, let r as CodeWithScope): return l == r
+    case (let l as Binary, let r as Binary): return l == r
+    case (let l as Document, let r as Document): return l == r
+    case (let l as [BSONValue?], let r as [BSONValue?]): // TODO: SWIFT-242
+        return zip(l, r).reduce(true, {prev, next in bsonEquals(next.0, next.1) && prev})
```
kmahar: I was going to suggest  we use `allSatisfy` rather than `reduce` as I think it's a bit clearer, but turns out that was only added in 4.2. so we *could* do an if/else on the version to have something like this, but I don't feel too strongly, esp. as we may change this in SWIFT-242 anyway:
```swift
return zip(l,r).allSatisfy { bsonEquals($0.0, $0.1) }
```

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -206,3 +206,21 @@ internal func rearrangeDoc(_ input: Document, toLookLike standard: Document) ->
...
     return output
 }
+
+/// A Nimble matcher for testing BSONValue equality.
+internal func bsonEqual(_ expectedValue: BSONValue?) -> Predicate<BSONValue> {
+    return Predicate { (actualExpression: Expression<BSONValue>) throws -> PredicateResult in
+        let msg = ExpectationMessage.expectedActualValueTo("equal <\(String(describing: expectedValue)))>")
+        if let actualValue = try actualExpression.evaluate() {
+            return PredicateResult(
+                bool: bsonEquals(actualValue, expectedValue!),
```
kmahar: won't this cause a crash if the expected value we pass in is nil? let's follow the logic from the original `equals` matcher [here](https://github.com/Quick/Nimble/blob/c68f5fd88997ab93a53cab07434349621882c817/Sources/Nimble/Matchers/Equal.swift#L7)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -779,6 +779,81 @@ public struct Timestamp: BSONValue, Equatable, Codable {
     }
 }
 
+// See https://github.com/realm/SwiftLint/issues/461
+// swiftlint:disable cyclomatic_complexity
+/**
+ *  A helper function to test equality between two BSONValues. This function tests for exact BSON equality.
```
```suggestion
 *  A helper function to test equality between two `BSONValue`s. This function tests for exact BSON equality.
```

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Actually, seems like we should use `private` over `fileprivate` here, according to the linter. So I've added an extra commit that fixes this linter check issue.

kmahar:
just docstring nits and a question

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: This is actually why I didn't use `allSatisfy()`.

I think it should be fine to keep it as is since SWIFT-242 may change it, and if it doesn't, it shouldn't be too unclear what we are doing here (since boolean squashing like this is a common functional idiom with `reduce`).

What do you think?

kmahar: sgtm 👍

Tests/MongoSwiftTests/TestUtils.swift
Utagai: Ah, yes, it would because of `!`. I was following the Nimble example in their README, but I now see that they include small note after it about how their example does not handle `nil`!!!

I've adapted the original `equals` matcher and tested to see that we are not crashing on any of the permutations of `nil`:
1. `nil`, `non-nil` 
1. `non-nil`, `nil` 
1. `nil`, `nil`

kmahar: 💯

PR #131, opened 2018-10-25
SWIFT-236: Fix linter issues

Utagai:
[SWIFT-236](https://jira.mongodb.org/browse/SWIFT-236)

This PR just fixes some linter issues we were not catching because of a slight issue with our Travis configuration that was ignoring warnings. That issue will be fixed as [SWIFT-237](https://jira.mongodb.org/browse/SWIFT-237).

This PR also renames `enabled_rules` to `opt_in_rules`, since `renamed_rules` will be removed, so we might as well do this now and not deal with issues later.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -205,7 +205,8 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
     }
 
-    /// Sets key to newValue. if checkForKey=false, the key/value pair will be appended without checking for the key's presence first.
+    /// Sets key to newValue. if checkForKey=false, the key/value pair will be appended without checking for the key's
+    // presence first.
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -205,7 +205,8 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
     }
 
-    /// Sets key to newValue. if checkForKey=false, the key/value pair will be appended without checking for the key's presence first.
+    /// Sets key to newValue. if checkForKey=false, the key/value pair will be appended without checking for the key's
+    // presence first.
```
kmahar:
```suggestion
    /// presence first.
```
although this is internal, we try to follow the same docstring policies as public types anyway.

Utagai: Right, I recall this from the `Development.md` but forgot about this. Done.

PR #132, opened 2018-10-25
SWIFT-237: Fix .travis.yml to fail whenever swiftlint has errors OR warnings

Utagai:
[SWIFT-237](https://jira.mongodb.org/browse/SWIFT-237)

Before, our Travis configuration for running `swiftlint` would only fail the build if we had 'serious' violations. This means that if we had warnings that `swiftlint` deemed not 'serious', then we would have a successful exit code of 0, and Travis would not fail the build. This PR adds the `--strict` command line option for `swiftlint` to the Travis configuration, making it so that warnings are treated as errors and therefore give a non-zero exit code. We should not be able to catch linter issues introduced as part of a PR.

This PR is a little weird. Technically, the fix for the things causing the linter to fail is part of #131, so this is somewhat blocked on that, and furthermore, we needed this branch to fail so that we can confirm that Travis is correctly failing.

Once #131 is finished, we can rebase this and verify that it also passes when all linter issues are resolved. These cases were tested locally, but it would be good to see that it works on Travis as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ install:
 before_script:
   - mkdir ${PWD}/mongodb-${MONGODB_VERSION}/data
   - ${PWD}/mongodb-${MONGODB_VERSION}/bin/mongod --dbpath ${PWD}/mongodb-${MONGODB_VERSION}/data --logpath ${PWD}/mongodb-${MONGODB_VERSION}/mongodb.log --enableMajorityReadConcern --fork
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then swiftlint; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then swiftlint --strict; fi
 
 script:
   - swift build -v


kmahar:
nice! so confirmed failure in case of warnings. one thing to note in the travis output is that there are a couple of errors related to the configuration because of unsupported rules. I think this is because both of the listed rules (redundant_type_annotation and missing_docs) were added rather recently and travis has not caught up their swiftlint version. 
once this is rebased on your other PR, we should make sure the lack of those rules doesn't cause a problem with strict.... and maybe ideally do a bit of investigation to see if we can get travis to use a new swiftlint version.

jmikola:
#131 LGTM, so this can wait on a rebase to confirm @kmahar's question above re: not failing due to lack of rules in Travis' version of `swiftlint`.

Utagai:
@kmahar Is it not possible to just run `brew` and install `swiftlint` using `brew` as part of our Travis configuration in order to ensure we get a more recent version of `swiftlint`?

I looked around to see if maybe this was the 'wrong' way or impossible, but [this link](https://alexplescan.com/posts/2016/03/03/setting-up-swiftlint-on-travis-ci/) here seems to suggest this as one of the ideas!

kmahar:
@Utagai  sounds definitely possible, want to give it a shot?

Utagai:
@kmahar Why not? Should we include it as part of this branch?

jmikola:
> Should we include it as part of this branch?

I don't see the harm of it. Just make sure you do it only for the osx environment :)

If there's any trouble, we can merge this as-is (since we're already green) and punt that to a separate issue.

Utagai:
@jmikola Good points.

I'll merge this PR separately (once you guys LGTM) and just make a ticket for using `brew` to get `swiftlint` on Travis for now. It is technically separate from the purpose of this ticket, and we already have things green here.

I am pretty sure I know how to do this (including the check for OS), so it shouldn't really be much of a difference in either case, but it would be nice to keep things clean. :)

Utagai:
Wow you guys are fast.

PR #134, opened 2018-10-29
SWIFT-239 use BSONType instead of libbson equivalent

kmahar:
Due to reasons we've yet to exactly determine, the driver will not build in Swift 4.2 in some cases due to an error about BSON_TYPE_CODE not being defined. This switches to using our own BSONType enum (which we should've been doing in the first place) and fixes the compilation errors in 4.2.


kmahar:
That's my suspicion / hope :)

mbroadst:
Is this the only place you are still using the `BSON_TYPE` enums?

kmahar:
yep, no other BSON_TYPE_X usages

PR #135, opened 2018-10-30
SWIFT-231: Don't hard-code in port number for tests

Utagai:
[SWIFT-231](https://jira.mongodb.org/browse/SWIFT-231)

This PR adds an ability for us to take a MDB connection string from the environment, defined via `$MDB_CONN_STR`, and use it to determine what `mongod` server to connect to for our tests. This is implemented with some convenience initializers, which lets us implement this ticket in a relatively small diff. Alternative approaches are welcome if this is not 'Swift-y'.

### DIFF
--- a/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
@@ -58,7 +58,7 @@ final class SDAMTests: XCTestCase {
         center.removeObserver(observer)
 
         var error = bson_error_t()
-        guard let uri = mongoc_uri_new_with_error(XCTestCase.getConnStr(), &error) else {
+        guard let uri = mongoc_uri_new_with_error(XCTestCase.connStr, &error) else {
             throw MongoError.invalidUri(message: toErrorString(error))
         }
 
--- a/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
@@ -104,6 +104,7 @@ final class SDAMTests: XCTestCase {
         expect(event4.topologyId).to(equal(event3.topologyId))
         let prevTopology = event4.previousDescription
         expect(prevTopology.type).to(equal(TopologyType.single))
+        expect(prevTopology.servers).to(beEmpty())
 
         let newTopology = event4.newDescription
         expect(newTopology.type).to(equal(TopologyType.single))
--- a/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
@@ -59,7 +59,8 @@ final class SDAMTests: XCTestCase {
 
         var error = bson_error_t()
         guard let uri = mongoc_uri_new_with_error(XCTestCase.connStr, &error) else {
-            throw MongoError.invalidUri(message: toErrorString(error))
+            XCTFail(toErrorString(error))
+            return
         }
 
         guard let hostlist = mongoc_uri_get_hosts(uri) else {
--- a/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/SDAMMonitoringTests.swift
@@ -1,5 +1,6 @@
 import Foundation
 @testable import MongoSwift
+import mongoc
 import Nimble
 import XCTest
 
@@ -24,8 +25,8 @@ final class SDAMTests: XCTestCase {
         expect(desc.type).to(equal(ServerType.unknown))
     }
 
-    func checkDefaultHostPort(_ desc: ServerDescription) {
-        expect(desc.connectionId).to(equal(ConnectionId("localhost:27017")))
+    func checkDefaultHostPort(_ desc: ServerDescription, _ hostlist: UnsafePointer<mongoc_host_list_t>) {
+        expect(desc.connectionId).to(equal(ConnectionId(hostlist)))
     }
 
     // Basic test based on the "standalone" spec test for SDAM monitoring:
@@ -56,6 +57,15 @@ final class SDAMTests: XCTestCase {
 
         center.removeObserver(observer)
 
+        var error = bson_error_t()
+        guard let uri = mongoc_uri_new_with_error(getConnStr(), &error) else {
+            throw MongoError.invalidUri(message: toErrorString(error))
+        }
+
+        guard let hostlist = mongoc_uri_get_hosts(uri) else {
+            throw MongoError.invalidResponse()
+        }
+
         // check event count and that events are of the expected types
         expect(receivedEvents.count).to(equal(5))
         expect(receivedEvents[0]).to(beAnInstanceOf(TopologyOpeningEvent.self))
@@ -72,25 +82,21 @@ final class SDAMTests: XCTestCase {
         expect(event1.topologyId).to(equal(event0.topologyId))
         expect(event1.previousDescription.type).to(equal(TopologyType.unknown))
         expect(event1.newDescription.type).to(equal(TopologyType.single))
-        let server0 = event1.newDescription.servers[0]
-
-        checkDefaultHostPort(server0)
-        expect(server0.type).to(equal(ServerType.unknown))
-        checkEmptyLists(server0)
+        expect(event1.newDescription.servers).to(beEmpty())
 
         let event2 = receivedEvents[2] as! ServerOpeningEvent
         expect(event2.topologyId).to(equal(event1.topologyId))
-        expect(event2.connectionId).to(equal(ConnectionId("localhost:27017")))
+        expect(event2.connectionId).to(equal(ConnectionId(hostlist)))
 
         let event3 = receivedEvents[3] as! ServerDescriptionChangedEvent
         expect(event3.topologyId).to(equal(event2.topologyId))
         let prevServer = event3.previousDescription
-        checkDefaultHostPort(prevServer)
+        checkDefaultHostPort(prevServer, hostlist)
         checkEmptyLists(prevServer)
         checkUnknownServerType(prevServer)
 
         let newServer = event3.newDescription
-        checkDefaultHostPort(newServer)
+        checkDefaultHostPort(newServer, hostlist)
         checkEmptyLists(newServer)
         expect(newServer.type).to(equal(ServerType.standalone))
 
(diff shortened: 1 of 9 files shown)


Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -113,6 +113,22 @@ extension MongoClient {
...
+    internal convenience init(options: ClientOptions? = nil) throws {
+        try self.init(connectionString: getConnStr(), options: options)
+    }
+
+    internal convenience init() throws {
+        try self.init(connectionString: getConnStr())
+    }
+}
+
+func getConnStr() -> String {
```
kmahar: Let's 
1) Move this into the `XCTestCase` extension above, next to `getSpecsPath`. Eventually, we're going to [create a base test class](https://jira.mongodb.org/browse/SWIFT-207), which will inherit from `XCTestCase` and be a natural place for methods such as this and `getSpecsPath`, and `is32Bit` to live. In the meantime, let's keep them all together, and can you please add a TODO above the extension noting its contents will be moved into the base test class in SWIFT-207?

2) Rewrite this as a computed property called `connStr`. Generally a computed property named `x` is more Swift-y than a `getX` method... which I did not realize back when I wrote that method. 🙂 If you feel like it, you could also rename `getSpecsPath` to `specsPath` and fix that wherever it's used. Or we can just do it later as part of SWIFT-207. :)

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -113,6 +113,22 @@ extension MongoClient {
 
         return true
     }
+
+    internal convenience init(options: ClientOptions? = nil) throws {
+        try self.init(connectionString: getConnStr(), options: options)
+    }
+
+    internal convenience init() throws {
```
kmahar: Do we need both of these? I'd think the first method with a default value for options would cover both. 

Also, have you confirmed somehow that the inits here take priority over the ones initially defined in the `MongoClient` class?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -113,6 +113,22 @@ extension MongoClient {
...
+        try self.init(connectionString: getConnStr())
+    }
+}
+
+func getConnStr() -> String {
+    if let connStr = ProcessInfo.processInfo.environment["MDB_CONN_STR"] {
```
jmikola: libmongoc uses [`MONGOC_TEST_URI`](https://github.com/mongodb/mongo-c-driver/blob/master/CONTRIBUTING.md#testing), alongside many other similarly named options. PHP uses [`MONGODB_URI`](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#local-mongo-orchestration-and-travis-ci).

I vote for renaming this to `MONGODB_URI`.

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -113,6 +113,22 @@ extension MongoClient {
...
+        return connStr
+    } else {
+        return "mongodb://localhost:27017"
```
jmikola: libmongoc defaults to [`mongodb://127.0.0.1/`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-client.c#L890) in its client constructor. I realize that Swift uses [`mongodb://localhost:27017`](https://github.com/mongodb/mongo-swift-driver/blob/v0.0.5/Sources/MongoSwift/MongoClient.swift#L124), but I would suggest sticking to libmongoc's default. Perhaps that belongs in a new ticket.

Despite the client constructor default, libmongoc does say its test suite defaults to `localhost` in [`CONTRIBUTING.md`](https://github.com/mongodb/mongo-c-driver/blob/master/CONTRIBUTING.md#testing) if the `MONGOC_TEST_HOST` variable is not defined. PHP's test suite always defaults to `mongodb://127.0.0.1/` (see: [here](https://github.com/mongodb/mongo-php-driver/blob/1.5.3/tests/utils/basic.inc#L5)).

I'm a proponent of using the IPv4 address over localhost as some platforms will resolve `localhost` as an IPv6 address, and some older versions of MongoDB may not listen on IPv6 by default. For that reason, the IPv4 address is more portable IMO. In both cases, the port isn't necessary in the default URI as libmongoc already supplies `27017` as its default port if one isn't specified in the URI.

Tests/MongoSwiftTests/TestUtils.swift
Utagai: You're right, I've removed the second `init()`.

Yes, I've confirmed this. I did this by examining the original `init()` signatures:
```
init(connectionString: String = "mongodb://localhost:27017", options: ClientOptions? = nil)
init(fromPointer: OpaquePointer)
```
The second of these can't be used because the signature is different. The first, since it has default values for connectionString, may end up being called. However, convenience initializers are called first. This is testable by trying to run the tests under a non-default connection string and non-default `mongod` port. If the original initializer is called, then the tests should fail to connect to a `mongod`.

Tests/MongoSwiftTests/TestUtils.swift
Utagai: Yeah, I think this is a good justification for changing the name. I don't really know much about what the other drivers are doing but having consistency across them is better than having a new variable name. I've changed this to agree with what PHP is using.

Tests/MongoSwiftTests/TestUtils.swift
Utagai: TIL! Done.

Tests/MongoSwiftTests/TestUtils.swift
Utagai: So I've added the `TODO` comment, but the other ideas related to moving the function into `XCTestCase` would mean that the convenience initializers no longer have the ability to get the connection string.

We can of course fix this, by either changing the signature to take an XCTestCase (seems pretty weird) or change all instances where we initialize a `MongoClient` to explicitly specify a connection string by using the computed property.

Honestly, I don't think this is a _bad_ thing, but I just figured I'd hold off on this in case this changes what we want to do with `getConnStr()`, since this change probably has ramifications outside of `TestUtils.swift`.

Utagai: Sounds brilliant. Done (both for `getConnStr()` -> `connStr` and `getSpecsPath()` -> `specsPath`).

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -3,18 +3,33 @@ import MongoSwift
...
-        // if we can access the "/Tests" directory, assume we're running from command line
-        if FileManager.default.fileExists(atPath: "./Tests") { return "./Tests/Specs" }
-        // otherwise we're in Xcode, get the bundle's resource path
-        guard let path = Bundle(for: type(of: self)).resourcePath else {
-            XCTFail("Missing resource path")
-            return ""
-        }
-        return path
+    static var specsPath: String {
+        get {
```
kmahar: nit: prefer omitting the `get { ... }` block when there's no `set`.

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -3,18 +3,33 @@ import MongoSwift
...
+                return ""
+            }
+            return path
+        }
+    }
+
+    /// Gets the connection string for the database being used for testing from the environment variable, $MONGODB_URI.
+    /// If the environment variable does not exist, this will use a default of "mongodb://127.0.0.1/".
+    static var connStr: String {
+        get {
```
kmahar: same nit as above.

Tests/MongoSwiftTests/TestUtils.swift
Utagai: Done.

Tests/MongoSwiftTests/TestUtils.swift
Utagai: Done.

Tests/MongoSwiftTests/SDAMMonitoringTests.swift
```diff
@@ -56,6 +57,15 @@ final class SDAMTests: XCTestCase {
 
         center.removeObserver(observer)
 
+        var error = bson_error_t()
+        guard let uri = mongoc_uri_new_with_error(XCTestCase.connStr, &error) else {
+            throw MongoError.invalidUri(message: toErrorString(error))
+        }
+
+        guard let hostlist = mongoc_uri_get_hosts(uri) else {
+            throw MongoError.invalidResponse()
```
jmikola: Should we use a more descriptive error here, or is it clear from this enum case why the test might fail? I'm just contrasting this with the `invalidUri` error above, which includes a descriptive message.

Granted, I'm also not familiar with these tests.

Utagai: The above case has the fortune of having an error returned from the `mongoc` function, which lets you actually say something meaningful about why something went wrong. In this case however, we don't really know why `mongoc` did not give us something back, so the best we can do is say something along the lines of "Getting the hostlist did not work".

Also, `invalidResponse()` seemed most appropriate, and unfortunately does not take a `message` argument.  :( Otherwise, I'd have put in at least some kind of message about the `hostlist` not working out.

kmahar: didn't think of this before, but we should probably just call `XCTFail` here with a descriptive message rather than throwing.

Utagai: Great, done.

I've kept the [other throw](https://github.com/mongodb/mongo-swift-driver/pull/135/files/ba5716babeccbc766f6a0be78f2033f35deeb336#diff-bf6491c15a31640b2ab0c3bbf0fb4494R62), since that has a descriptive message and that seems to satisfy the principal goal of this discussion.

But since I'm not sure if throwing in general is something we want to avoid in Swift tests, I'll refrain from merging this until we're sure that we're okay with throwing a descriptive error for the above case. Let me know if we should use `XCTFail()` again instead.

kmahar: I think the actual error is fine, but also support changing to XCTFail if @jmikola cares.

Tests/MongoSwiftTests/SDAMMonitoringTests.swift
```diff
@@ -56,6 +57,16 @@ final class SDAMTests: XCTestCase {
...
         center.removeObserver(observer)
 
+        var error = bson_error_t()
+        guard let uri = mongoc_uri_new_with_error(XCTestCase.connStr, &error) else {
+            throw MongoError.invalidUri(message: toErrorString(error))
+        }
+
+        guard let hostlist = mongoc_uri_get_hosts(uri) else {
+            XCTFail("Could not get hostlists for uri: \(uri)")
+            return
```
jmikola: Do I assume correctly that the `return` is necessary due to `continueOnError` (per [this Stack Overflow thread](https://stackoverflow.com/q/36284251/162228)).

Utagai: `SDAMMonitoringTests.swift` has [`self.continueAfterFailure = false`](https://github.com/mongodb/mongo-swift-driver/pull/135/files/a21d4f98514b2568ef59a5cc3ab0ca1d6a85c347#diff-bf6491c15a31640b2ab0c3bbf0fb4494L14). I added in a `return` because we _always_ want to skip the rest of this test if the `guard` fails, but we may change the `continueAfterFailure` setting.

So in the current code, it isn't really _necessary_. This is just paying a single LoC in exchange for being protected against future changes.

Tests/MongoSwiftTests/SDAMMonitoringTests.swift
Utagai: Sounds like a plan! @jmikola Let us know what you think.

Utagai: Done. Now they both trigger `XCTFail()`. I'll merge once Travis is green.

PR #136, opened 2018-10-31
SWIFT-198: Use setValue where possible rather than subscript setter

Utagai:
[SWIFT-198](https://jira.mongodb.org/browse/SWIFT-198)

This PR replaces instances where we use subscript-setting in a throwing function with `setValue()`. This should help avoid `preconditionFailure`'s and instead propagate a non-fatal error for the end-user.

There were some instances of functions that were declared as `rethrows`. From my understanding, these functions only throw if their argument which is a function/closure throws itself, so we cannot use `setValue()` without marking those function as `throws`. This was enough of a change that I figured it shouldn't be included in this ticket. Plus, these were implemented in conformance to `Sequence`, and I believe marking them as `throws` would break the conformance.

As usual with these kinds of PRs, let me know if you think I have missed any instances.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -103,7 +103,7 @@ extension Array: BSONValue {
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         // An array is just a document with keys '0', '1', etc. corresponding to indexes
         var arr = Document()
-        for (i, v) in self.enumerated() { arr[String(i)] = v as? BSONValue }
+        for (i, v) in self.enumerated() { try arr.setValue(forKey: String(i), to: v as? BSONValue) }
         guard bson_append_array(storage.pointer, key, Int32(key.count), arr.data) else {
             throw bsonEncodeError(value: self, forKey: key)
         }
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -103,7 +103,12 @@ extension Array: BSONValue {
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         // An array is just a document with keys '0', '1', etc. corresponding to indexes
         var arr = Document()
-        for (i, v) in self.enumerated() { try arr.setValue(forKey: String(i), to: v as? BSONValue) }
+        for (i, v) in self.enumerated() {
+            if self as? [BSONValue?] == nil {
+                throw MongoError.invalidArgument(message: "Cannot encode a non-BSONValue array element: \(v)")
+            }
+            try arr.setValue(forKey: String(i), to: v as? BSONValue)
+        }
         guard bson_append_array(storage.pointer, key, Int32(key.count), arr.data) else {
             throw bsonEncodeError(value: self, forKey: key)
         }
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -102,11 +102,11 @@ extension Array: BSONValue {
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         // An array is just a document with keys '0', '1', etc. corresponding to indexes
+        if self as? [BSONValue?] == nil {
+            throw MongoError.invalidArgument(message: "Cannot encode a non-BSONValue array element")
+        }
         var arr = Document()
         for (i, v) in self.enumerated() {
-            if self as? [BSONValue?] == nil {
-                throw MongoError.invalidArgument(message: "Cannot encode a non-BSONValue array element: \(v)")
-            }
             try arr.setValue(forKey: String(i), to: v as? BSONValue)
         }
         guard bson_append_array(storage.pointer, key, Int32(key.count), arr.data) else {
--- a/Sources/MongoSwift/BSON/Document+Codable.swift
+++ b/Sources/MongoSwift/BSON/Document+Codable.swift
@@ -46,9 +46,10 @@ extension Document: Codable {
         for key in container.allKeys {
             let k = key.stringValue
             if try container.decodeNil(forKey: key) {
-                output[k] = nil
+                try output.setValue(forKey: k, to: nil)
             } else {
-                output[k] = try container.decode(AnyBSONValue.self, forKey: key).value
+                let val = try container.decode(AnyBSONValue.self, forKey: key).value
+                try output.setValue(forKey: k, to: val)
             }
         }
         self = output
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -293,7 +293,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      * For example:
      *      let doc1: Document = ["a": 1]
      *      var doc2 = doc1
-     *      doc2["b"] = 2
+     *      doc2.setValue(forKey: "b", to: 2)
      *
      * Therefore, this function should be called just before we are about to modify a document - either by
      * setting a value or merging in another doc.
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -270,7 +270,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      *  - Throws: A `MongoError.typeError` if the value cannot be cast to type `T` or is not in the `Document`
      *
      */
-    public func get<T: BSONValue>(_ key: String) throws -> T {
+    internal func get<T: BSONValue>(_ key: String) throws -> T {
         guard let value = self[key] as? T else {
             throw MongoError.typeError(message: "Could not cast value for key \(key) to type \(T.self)")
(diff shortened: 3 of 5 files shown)


Sources/MongoSwift/BSON/Document+Codable.swift
```diff
@@ -46,9 +46,10 @@ extension Document: Codable {
         for key in container.allKeys {
             let k = key.stringValue
             if try container.decodeNil(forKey: key) {
-                output[k] = nil
+                try output.setValue(forKey: k, to: nil)
             } else {
-                output[k] = try container.decode(AnyBSONValue.self, forKey: key).value
+                let val = try container.decode(AnyBSONValue.self, forKey: key).value
```
Utagai: This line was broken up into two because the idea of having a nested `try` inside of an outer `try` seemed less readable and less explicit about what points in our code can actually throw.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -103,7 +103,7 @@ extension Array: BSONValue {
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         // An array is just a document with keys '0', '1', etc. corresponding to indexes
         var arr = Document()
-        for (i, v) in self.enumerated() { arr[String(i)] = v as? BSONValue }
+        for (i, v) in self.enumerated() { try arr.setValue(forKey: String(i), to: v as? BSONValue) }
```
kmahar: while you're here: note that this method will silently fail and encode a `nil` when the Array contains a non-`BsonValue?`, as the `as? BSONValue` cast just returns nil.
so if I had an array like `[1, "hi", SomeNonBSONValueType(), 3]`, it ends up getting encoded as `[1, "hi", nil, 3]`. 

Since this method throws now anyway (it didn't use to), how about throwing an error here when v is neither `nil` nor castable to a `BSONValue`? 

the downside of that is that code would now crash if you tried to put such an array in a document via th subscript setter. I'm not sure whether this silent unexpected behavior is preferable to a crash... maybe @jmikola will have thoughts as well.

Utagai: How is this case possible if arrays in Swift are homogenous? The only way this can happen from what I can tell if we are playing with an array of type `[Any]`. However, with [SWIFT-215](https://jira.mongodb.org/browse/SWIFT-215), this should not happen right?

In either case, since we may not know when SWIFT-215 will be done, it may still be better to do something explicit here instead of silently fail. My own opinion almost always about this sort of stuff is being explicit and annoying the user away from doing dumb or bad things is better than being quiet.

kmahar: yup, that's correct, it could only be an `[Any]` or `[Non-BSONValue]`, and indeed, SWIFT-215 will solve that once we drop 4.0.
I'm good with throwing an error here in those cases. we should probably also add an explicit note to the docstring for the Array extension noting that it's currently on the user to ensure their arrays only contain `BSONValue`s.

Utagai: Sounds good to me as well. Since this seems to be the only thing, I'll wait to see if @jmikola has any thoughts on this as well.

jmikola: >  how about throwing an error here when v is neither nil nor castable to a BSONValue?

Would this be done by checking if `v` is first nil and, if not, catching if the result of `v as? BSONValue` yields nil?

I don't quite follow how this would result in a crash. Is the subscript setter using `encode()` and raising a precondition failure if an exception is thrown?

Utagai: @jmikola I don't think the issue is related to the new code crashing. The issue is that it will silently give `nil` for non-`BSONValue`'s. What we are proposing however is to throw something if `v` is not a `BSONValue`.

If you're asking about how the old code would have crashed, then yes, the current subscript setter will raise a `preconditionFailure()` whenever `setValue()` throws. `setValue()` eventually calls `encode()` (if it is a new key, which occurs every time in this case since we are starting with a new `Document`).

> Would this be done by checking if v is first nil and, if not, catching if the result of v as? BSONValue yields nil?

And yes, I think that is how it would be done.

Let me know if I misinterpreted you.

kmahar: yes, it would lead to a crash because the subscript setter calls through to this function.

kmahar:
yeah, per our previous conversations about having too many ways to do stuff with documents, I intended for this method to be internal. on that note, we might want to consider making that `get`method internal. 
if we *do* keep it public, I now know it's more idiomatic to have it take in the requested type as a parameter rather than having it be given by declaring the type of the requested variable. i.e, now you have to do `let x: Int = try doc.get("x")`, but more idiomatic is something like `let x = try doc.get("x", as: Int.self)`.

alternatively, we could consider having public "safe" getters and setters for when you're using unvalidated BSON data. this somewhat comes back to that document validation epic we have somewhere....

kmahar:
the signature would be as follows:
```swift
func get<T: BsonValue>(_ key: String, as: T.self) throws -> T
```

some existing examples in our codebase :
```swift
// BSONDecoder.decode
 public func decode<T: Decodable>(_ type: T.Type, from document: Document) throws -> T

// MongoDatabase.collection
public func collection<T: Codable>(_ name: String, withType: T.Type,
                                       options: CollectionOptions? = nil) throws -> MongoCollection<T>
```

Sources/MongoSwift/BSON/BSONValue.swift
jmikola: If the current code would raise a precondition failure when encoding failed, I'm all for keeping that here. My instincts tell me it's worse to silently encode a nil and hide errors from the user.

I don't have the bandwidth to think about it now, but we'd probably do well to revisit all of this and audit the driver to places where users might experience precondition failures due to runtime errors. Then we can properly document all of those cases and chalk them up to necessary evils due to language limitations (i.e. not being able to throw from certain contexts).

kmahar:
side note, wtf? 
screen shot 2018-11-02 at 5 30 08 pm

jmikola:
> side note, wtf?
> screen shot 2018-11-02 at 5 30 08 pm

I wonder if you're experiencing the same curious macOS bug as the OP in https://github.com/briannesbitt/Carbon/issues/1490 :smiley:

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: yeah, it was still possible if something else went wrong in encoding; this is just adding another possible error that could result in a crash. agreed on the audit

kmahar:
that comment was from Friday actually, long before the time changed happened. I think @jmikola's suggestion is that it's a new bug on MacOS with times in general, and it was cropping up in different libraries that use system time. 

anyway, I think I've figured out the issue. my clock is a bit behind. see the times in this screenshot: according to time.is, my clock is 39.2 seconds behind. So I think GitHub is sending back a timestamp of the actual time it was posted, and then moment.js etc. is used to render the difference between my time and the timestamp. 

screen shot 2018-11-04 at 12 36 11 pm

I also noticed the other day that I got a "10 minutes before" calendar notification earlier on my phone than on my computer. 

I have turned on and off the "set time automatically" function to try to get it to reset, but it doesn't work. it's behind the time on my iPhone, which is *also* set automatically by Apple... there's a 10.14.1 update I haven't done yet. release notes don't mention a bug fix anywhere but I'll try it out anyway and do a restart. so, perhaps the user having the carbon bug on that thread is also having this bug. but no relation to Carbon.

Utagai:
Was fun reading these comments of odd time behavior :).

Lots of stuff was mentioned about `get`. I like the idea of reducing the number of ways to do the same thing if it is all functionally equivalent. Should we perhaps make it internal as part of this ticket? (Or, if we don't want to make it internal, make it more idiomatic by taking in the type?). The mention about safe getters/setters however seems like it would part of a different epic.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: let's definitely not revert, we want to throw from here and have the precondition failure happen at the top-level subscript `set` call rather than in this recursive call (since subscript `set` calls this method...) we were silent encoding nil with the old way too. let's also add an error in the non-nil case. 
as an interesting side note, technically we could support anything Codable being in an array, not just anything that's a BSONValue. because Codable things can be turned into Documents via BSONEncoder, and are therefore kinda BSONValues themselves... 🤯

kmahar:
ya, I think I'm in favor of making it internal.

something between the update and the restart that the update forced fixed my clock, btw. should have tried just restart first to figure out where it was coming from... oh well

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: @kmahar I've added this detail about `Codable` to [SWIFT-215](https://jira.mongodb.org/browse/SWIFT-215).

Also, we now throw in the case that we get a non-`BSONValue`.

jmikola:
@kylekatarnls: Sorry to distract you. I just happened to be reading that Carbon issue while looking at this PR and found it to be an amusing coincidence. I wasn't suggesting that the issues were related in any way -- in hindsight, I should have saved that comment for a private Slack message.

PR #138, opened 2018-11-01
SWIFT-200: Use libbson overwrite methods for ObjectId, Timestamp, Datetime 

Utagai:
[SWIFT-200](https://jira.mongodb.org/browse/SWIFT-200)

This PR has a fairly self-explanatory title. It extends `ObjectId`, `Timestamp` and `Datetime` with overwrite methods so that they can conform to `Overwritable`. Alongside this, it updates the tests in `DocumentTests.swift` to include the new conformances.

This PR also includes some fixes to linter issues in `DocumentTests.swift`, as well as a rename of some variables in the related test cases, which should improve the readability of the test cases.

Is there a way to test this stuff on 32-bit systems? It seems like our Travis configuration does not run on 32-bit systems.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -315,7 +315,7 @@ public struct Decimal128: BSONValue, Equatable, Codable {
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst
     public init?(ifValid data: String) {
         do {
-            _ = try Decimal128.encode(data)
+            _ = try Decimal128.toLibBSONType(data)
             self.init(data)
         } catch {
             return nil
@@ -324,7 +324,7 @@ public struct Decimal128: BSONValue, Equatable, Codable {
 
     /// Returns the provided string as a `bson_decimal128_t`, or throws an error if initialization fails due an
     /// invalid string.
-    internal static func encode(_ str: String) throws -> bson_decimal128_t {
+    internal static func toLibBSONType(_ str: String) throws -> bson_decimal128_t {
         var value = bson_decimal128_t()
         guard bson_decimal128_from_string(str, &value) else {
             throw MongoError.bsonEncodeError(message: "Invalid Decimal128 string \(str)")
@@ -337,7 +337,7 @@ public struct Decimal128: BSONValue, Equatable, Codable {
     }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        var value = try Decimal128.encode(self.data)
+        var value = try Decimal128.toLibBSONType(self.data)
         guard bson_append_decimal128(storage.pointer, key, Int32(key.count), &value) else {
             throw bsonEncodeError(value: self, forKey: key)
         }
@@ -565,7 +565,7 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     }
 
     /// Returns the provided string as a `bson_oid_t`.
-    internal static func encode(_ str: String) -> bson_oid_t {
+    internal static func toLibBSONType(_ str: String) -> bson_oid_t {
         var value = bson_oid_t()
         bson_oid_init_from_string(&value, str)
         return value
@@ -573,7 +573,7 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         // create a new bson_oid_t with self.oid
-        var oid = ObjectId.encode(self.oid)
+        var oid = ObjectId.toLibBSONType(self.oid)
         // encode the bson_oid_t to the bson_t
         guard bson_append_oid(storage.pointer, key, Int32(key.count), &oid) else {
             throw bsonEncodeError(value: self, forKey: key)
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -559,11 +559,10 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     /// ObjectId.
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init?(ifValid oid: String) {
-        do {
-            _ = try ObjectId.toLibBSONType(oid)
-            self.init(fromString: oid)
-        } catch {
+        if !bson_oid_is_valid(oid, oid.count) {
             return nil
+        } else {
+            self.init(fromString: oid)
         }
     }
 
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -564,10 +564,16 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
         }
     }
 
+    /// Returns the provided string as a `bson_oid_t`.
+    internal static func encode(_ str: String) -> bson_oid_t {
+        var value = bson_oid_t()
+        bson_oid_init_from_string(&value, str)
+        return value
+    }
+
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         // create a new bson_oid_t with self.oid
-        var oid = bson_oid_t()
-        bson_oid_init_from_string(&oid, self.oid)
+        var oid = ObjectId.encode(self.oid)
         // encode the bson_oid_t to the bson_t
         guard bson_append_oid(storage.pointer, key, Int32(key.count), &oid) else {
             throw bsonEncodeError(value: self, forKey: key)
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -549,9 +549,22 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
         self.init(fromPointer: &oid_t)
     }
 
-    /// Initializes an `ObjectId` from the provided `String`.
+    /// Initializes an `ObjectId` from the provided `String`. Assumes that the given string is a valid ObjectId.
+    /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(fromString oid: String) {
         self.oid = oid
+    }
+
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -564,10 +564,16 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
         }
     }
 
+    /// Returns the provided string as a `bson_oid_t`.
+    internal static func encode(_ str: String) -> bson_oid_t {
```
jmikola: I'm not sure it makes sense to introduce a new static function instead of sticking with 2 LOC in the two places this would be called:

```
var oid = bson_oid_t()
bson_oid_init_from_string(&oid, self.oid)
```

Part of my objection is also with the name itself implying that it's an overload, until you realize it's very different than the `encode()` protocol method.

Sources/MongoSwift/BSON/Overwritable.swift
```diff
@@ -1,4 +1,5 @@
 import bson
+import Foundation
```
jmikola: Is this required due to the changes below, or was it just something that was missing?

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -354,9 +354,23 @@ final class DocumentTests: XCTestCase {
...
-    static let nonOverwritables: Document = ["string": "hello", "nil": nil, "doc": ["x": 1] as Document, "arr": [1, 2] as [Int], "oid": ObjectId()]
+    static let overwritables: Document = [
+        "double": 2.5,
+        "int32": Int32(32),
+        "int64": Int64.max,
+        "bool": false,
+        "decimal": Decimal128("1.2E+10"),
+        "oid": ObjectId(),
+        "timestamp": Timestamp(timestamp: 1, inc: 2),
+        "datetime": Date(msSinceEpoch: 1000)
```
jmikola: Just confirming that "oid" was moved from `nonOverwritables`, and "timestamp" and "datetime" were added (previously not tested at all).

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -389,17 +403,45 @@ final class DocumentTests: XCTestCase {
...
+        doc["timestamp"] = Timestamp(timestamp: 5, inc: 10)
+        expect(doc.data).to(equal(pointer))
+
+        doc["datetime"] = Date(msSinceEpoch: 2000)
+        expect(doc.data).to(equal(pointer))
+
         // return early as we will to use an Int requiring > 32 bits after this 
         if XCTestCase.is32Bit {
-            expect(doc).to(equal(["double": 3.0, "int32": 20, "int64": Int64.min, "bool": true, "decimal": Decimal128("100")]))
+            expect(doc).to(equal([
```
jmikola: This comment is intended for comment two lines up. This doesn't actually return early. Should the comment be updated?

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -423,67 +464,86 @@ final class DocumentTests: XCTestCase {
...
+        ]
 
-        newPairs1.forEach { (k, v) in
-            doc1[k] = v
-            expect(doc1.data).toNot(equal(pointer1))
-            pointer1 = doc1.data
+        overwritablePairs.forEach { (k, v) in
+            overwritableDoc[k] = v
+            expect(overwritableDoc.data).toNot(equal(overwritablePointer))
+            overwritablePointer = overwritableDoc.data
```
jmikola: :+1: for more descriptive variables names.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: This is something I spoke to Kaitlin about in [#swift](https://mongodb.slack.com/archives/C89CNDNF6/p1541095355137600).

I suppose I'll tag Kaitlin here to see if we want to actually not do this after all @kmahar.

Sources/MongoSwift/BSON/Overwritable.swift
Utagai: Yes, it is required due to the changes below. It is because we are referencing the `Date` type, which requires an `Foundation`.

Tests/MongoSwiftTests/DocumentTests.swift
Utagai: Yup that's exactly right. Previously, we were not testing `"timestamp"` or `"datetime"`, in `nonOverwritables` or `overwritables`, so when I noticed that I made sure to include them now.

Tests/MongoSwiftTests/DocumentTests.swift
Utagai: Good catch. However, I think the right thing to do is keep the comment but correct the code to match the comment. The right thing to do here definitely seems to be to return early if we are on a 32 bit system. I will try to see if I can use the Github suggestion thing to chuck in a `return`, and if you think that sounds good, you can approve it?

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I think we should keep this abstraction to be consistent with `Decimal128`, which is the only other type where we have to do some kind of conversion b.w our storage format and libbson's. (in that case it's 4 lines that are called 3x, so we should definitely keep that.)

that said, I would support renaming these methods, to something more descriptive that clarifies what we're doing. `convertToLibbsonType()` or something? idk.

Tests/MongoSwiftTests/DocumentTests.swift
kmahar: that's on me 😛

Sources/MongoSwift/BSON/BSONValue.swift
jmikola: I'm not sure about Decimal128's `encode()` method, but given that this is always called with an ObjectID instance in mind, I'd propose we remove `static` and make this an actual internal method, such as `toInternalType()` with no arguments.

I would suggest a computed property, but Decimal128 initialization does throw. And we can't assume its internal data is valid because there is an second `init()` method that doesn't validate and trusts the user provided something valid.

Tests/MongoSwiftTests/DocumentTests.swift
Utagai: From seeing the logic, it seems like it should work fine on a 32-bit system too. I've made it so that we always run the check and then return if the system is 32 bit.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: I left the office before I could really respond to Jeremy's points above, but I gave him the very solid thumbs up. I think that's a good way to meet in the middle here, and the proposed changes of making the method internal sounds good.

I am not completely sure about the removal of the arguments however, because it seems like this method is at some point being called in a `static` manner, for example, in the [`Decimal128`'s `init?()`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/BSON/BSONValue.swift#L318).

However, from what I can tell that is the _only_ time we use the method in a `static` fashion, both for `Decimal128` and `ObjectId`.

kmahar: ya, the Decimal128 one is static so it can be used from within an init to validate data before actually going ahead with creating the object.
all of this making me look at libbson's [bson_oid_t](http://mongoc.org/libbson/current/bson_oid_t.html) and I'm a little confused now - should we be calling `bson_oid_is_valid` ever? does `bson_oid_init_from_string` ever fail / how can we tell?

Utagai: @kmahar had the same kind of questions about `bson_oid_init_from_string()` failing, but did not notice a `bson_oid_is_valid()`.

It seems like if such a function for checking validity exists, we are supposed to using it. I'll see if perhaps Kevin has an idea on this and get back to this thread.

Utagai: Did a rename of `encode()` to `toLibBSONType()`.

Utagai: @jmikola Answered this in #mongoc:
> PHP does use is_valid: https://github.com/mongodb/mongo-php-driver/blob/master/src/BSON/ObjectId.c#L49

And:
> checks the string buffer length and validates its characters: https://github.com/mongodb/mongo-c-driver/blob/master/src/libbson/src/bson/bson-oid.c#L249

Seems like we'd do well to follow PHP here and also validate the `oid` strings. I'll see if I can do a search for our uses of `bson_init_from_string()` and make sure it follows a validity check.

Utagai: I added a validation check to our call to `bson_init_from_string()`. In order to stay consistent with something like `Decimal128`, I've also added a failable initializer to `ObjectId`.

kmahar: lol oops, forgot we still had this discussion to resolve before approving, new changes also LGTM  tho

jmikola: > ya, the Decimal128 one is static so it can be used from within an init to validate data before actually going ahead with creating the object.

Missed that. Makes sense for Decimal128 to be static then. I'm ambivalent about ObectId.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -549,11 +549,23 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
+    /// Initializes an `ObjectId` from the provided `String`. Assumes that the given string is a valid ObjectId.
+    /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(fromString oid: String) {
         self.oid = oid
     }
 
+    /// Initializes an `ObjectId` from the provided `String`. Returns `nil` if the string is not a valid
+    /// ObjectId.
+    /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
+    public init?(ifValid oid: String) {
```
jmikola: If I understand correctly, the only thing that differentiates the failable and non-failable initializer is the argument name. Is that a common convention?

kmahar: they need to be different in *at least* this way for the compiler to be able to distinguish between them. for a single parameter init, there isn't much else we can do to differentiate

PR #139, opened 2018-11-02
SWIFT-199: Add a throwing getValue method

Utagai:
[SWIFT-199](https://jira.mongodb.org/browse/SWIFT-199)

This PR adds a function to `Document` that is the `get` analog of `setValue()`. This function throws instead of triggering a fatal `preconditionFailure()`, which should help reduce the number of fatal errors. This is implemented via the addition of a throwing function equivalent to `currentValue` of `DocumentIterator`. Eventually, this seemed like the optimal choice, since it lets us package the logic of extracting the current value inside the iterator, where such logic should exist, instead of `Document`.

As usual with these kinds of PRs, let me know if I have missed any instances where we should be using `getValue()`, but this should be all. I used a `git grep` regex to find them all.

### DIFF
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -200,6 +200,11 @@ public class DocumentIterator: IteratorProtocol {
         return bson_iter_next(&self.iter)
     }
 
+    /// Moves the iterator to the specified key. Returns false if the key does not exist. Returns true otherwise.
+    internal func move(to key: String) -> Bool {
+        return bson_iter_find(&self.iter, key.cString(using: .utf8))
+    }
+
     /// Returns the current key. Assumes the iterator is in a valid position.
     internal var currentKey: String {
         return String(cString: bson_iter_key(&self.iter))
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -242,6 +242,17 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    internal func safeCurrentValue() throws -> BSONValue? {
+        switch self.currentType {
+        case .symbol:
+            return try Symbol.asString(from: self)
+        case .dbPointer:
+            return try DBPointer.asDocument(from: self)
+        default:
+            return try DocumentIterator.bsonTypeMap[currentType]?.init(from: self)
+        }
+    }
+
     // uses an iterator to copy (key, value) pairs of the provided document from range [startIndex, endIndex) into a new
     // document. starts at the startIndex-th pair and ends at the end of the document or the (endIndex-1)th index,
     // whichever comes first.
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -242,6 +242,8 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    /// Returns the current value, equivalently to the `currentValue` property, but in the case of error, will throw
+    /// instead of causing a fatal `preconditionFailure`.
     internal func safeCurrentValue() throws -> BSONValue? {
         switch self.currentType {
         case .symbol:
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -247,7 +247,7 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
-    /// Returns the current value (equivalent to the currentValue property) or throws on error
+    /// Returns the current value (equivalent to the `currentValue` property) or throws on error.
     internal func safeCurrentValue() throws -> BSONValue? {
         switch self.currentType {
         case .symbol:
--- a/Sources/MongoSwift/BSON/Document+Sequence.swift
+++ b/Sources/MongoSwift/BSON/Document+Sequence.swift
@@ -242,8 +242,7 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
-    /// Returns the current value, equivalently to the `currentValue` property, but in the case of error, will throw
-    /// instead of causing a fatal `preconditionFailure`.
+    /// Returns the current value (equivalent to the currentValue property) or throws on error
     internal func safeCurrentValue() throws -> BSONValue? {
         switch self.currentType {
         case .symbol:
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -258,7 +258,6 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         guard let iter = DocumentIterator(forDocument: self, advancedTo: key) else {
             throw MongoError.invalidArgument(message: "Failed to construct an iterator advanced to \(key)")
         }
-
         return try iter.safeCurrentValue()
     }
 
@@ -279,7 +278,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      *
      */
     public func get<T: BSONValue>(_ key: String) throws -> T {
-        guard let value = self[key] as? T else {
+        guard let value = try self.getValue(forKey: key) as? T else {
             throw MongoError.typeError(message: "Could not cast value for key \(key) to type \(T.self)")
         }
         return value
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -256,9 +256,13 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
 
     /// Retrieves the value associated with forKey as a BSONValue?, which can be nil.
     internal func getValue(forKey key: String) throws -> BSONValue? {
-        guard let iter = DocumentIterator(forDocument: self, advancedTo: key) else {
+        guard let iter = DocumentIterator(forDocument: self) else {
+            throw MongoError.bsonDecodeError(message: "BSON buffer is unexpectedly too small (< 5 bytes)")
+        }
+        guard iter.move(to: key) else {
             return nil
         }
+
         return try iter.safeCurrentValue()
     }
 
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -254,7 +254,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
     }
 
-    /// Retrieves the value associated with for as a BSONValue?, which can be nil.
+    /// Retrieves the value associated with for as a `BSONValue?`, which can be nil.
     internal func getValue(for key: String) throws -> BSONValue? {
         guard let iter = DocumentIterator(forDocument: self) else {
             throw MongoError.bsonDecodeError(message: "BSON buffer is unexpectedly too small (< 5 bytes)")
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -254,6 +254,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
     }
 
+    /// Retrieves the value associated with forKey as a BSONValue?, which can be nil.
     internal func getValue(forKey key: String) throws -> BSONValue? {
         guard let iter = DocumentIterator(forDocument: self, advancedTo: key) else {
             return nil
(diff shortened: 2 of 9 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -254,6 +254,14 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
     }
 
+    /// Retrieves the value associated with forKey as a BSONValue?, which can be nil.
+    internal func getValue(forKey key: String) throws -> BSONValue? {
+        guard let iter = DocumentIterator(forDocument: self, advancedTo: key) else {
```
Utagai: From what I can tell, we can't really tell _why_ this failed, if it does, so the best we can do here is return `nil`. Let me know if that is not the case, but currently it seems like the `mongoc` function we call simply returns `true` or `false` with no further detail on why.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -373,18 +373,18 @@ public struct BulkWriteResult {
      *   - insertedIds: Map of inserted IDs
      */
     fileprivate init(reply: Document, insertedIds: [Int: BSONValue?]) throws {
-        self.deletedCount = reply["nRemoved"] as? Int ?? 0
-        self.insertedCount = reply["nInserted"] as? Int ?? 0
+        self.deletedCount = try reply.getValue(forKey: "nRemoved") as? Int ?? 0
```
jmikola: Looking at this diff, I wonder if we can remove the label for `getValue()`'s first and only parameter. Perhaps this was done for consistency with `setValue()`, but asking users to repeatedly type `forKey` seems verbose. Defer to @kmahar.

I wonder if there are design patterns from other Swift libs that have getter methods (in lieu of subscript operators) where they've chosen to not used a parameter label. If so, I'd be curious to know if they did the same for their `setValue()` equivalent, too -- unless it's common Swift convention to require labels for anything with multiple required parameters.

AFAIK, the only place we mix unlabeled and labeled params is with things like `insertOne()`, where there is only one required parameter (e.g. the document) and the additional param is optional.

Sources/MongoSwift/BSON/Document.swift
Utagai: Interesting, however, the `init()` we are calling here calls `bson_iter_init_find()`. Does this then call `bson_iter_init()`? In either case, there is also the possibility that it fails because the key we wanted to advance to was invalid/DNE. But there is nothing about the return of `bson_iter_init_find()` that really says anything about this. Perhaps we can use `bson_iter_init()` first, check its return, and then advance to the `key` by ourselves manually. This would allow us to distinguish between the cause of failure, but would probably be substantially slower.

In this case, we return `nil` here because it is possible that the issue is not due to corruption, but because the provided key does not exist. In this case, it is more correct to return `nil`, and in fact, parts of our codebase rely on the fact that when getting the value for a key in a dictionary that does not exist, we will get back `nil`. See [BulkWriteResult](https://github.com/mongodb/mongo-swift-driver/pull/139/files/31eda280b853bf3c4a7fb6cbb380f54586784b58#diff-312fa0fd8cb91f4e116a176fe7b0133eR375)'s `init()`.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -242,6 +242,19 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    /// Returns the current value, equivalently to the `currentValue` property, but in the case of error, will throw
```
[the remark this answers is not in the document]
Utagai: Hm, I suppose you might be right. At the same time, `currentValue` is a computed property, so is it actually a verb? I think computed properties actually blur the line here.

In either case though, I wrote "equivalently" here because it is referring to the action of returning ("Returns ... , equivalently ...").

What do you think about this?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
Utagai: I noticed this same exact thing! However, as you pointed out, this was done for consistency with `setValue()`.

I suppose we'll let @kmahar weigh in on what seems to be potentially a Swift conventions question.

Utagai:
@jmikola The only other documentation block comments that I saw that had subscript notation were ones in which they probably should have them (e.g. the documentation block for `subscript()`), so I think we should be good on that end.

Sources/MongoSwift/BSON/Document.swift
jmikola: Ah, I missed the `advanceTo` argument. In that case, this can certainly fail if we don't find that key in a valid document. It looks like this ultimately calls into `bson_iter_find_w_len()` to do the search.

Two take-aways:

 * The driver can explicitly call `bson_iter_init()` and then `bson_iter_find()` if it wants to differentiate between an invalid buffer and failing to find a key, so that we can raise the appropriate error in the former case.
 * Since this method returns an optional BsonValue, isn't there a general limitation in that we can't differentiate between finding a nil value in the document and not finding it at all? This may relate to earlier discussions I had with @kmahar where we debated use of double-optionals. I think the conclusion there was that users can use `hasKey()` if they need to discern.

Sources/MongoSwift/BSON/Document+Sequence.swift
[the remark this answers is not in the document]
jmikola: How about:

> Returns the current value (equivalent to the `currentValue` property) or throws on error

I don't think we need to say "instead of causing a fatal `preconditionFailure`". This is internal and we probably already know that subscript has no choice but to crash, while we can throw from this method.

jmikola:
> And I don't think our goal is to get rid of subscripting entirely...

@Utagai: Aye, I misunderstood the relationship between these methods and the subscript operators (as noted in #136). I think the doc block examples can remain as-is.

Sources/MongoSwift/BSON/Document+Sequence.swift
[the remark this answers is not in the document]
Utagai: Awesome! Sounds good. Done.

Sources/MongoSwift/BSON/Document.swift
Utagai: Right, so we can distinguish between the two and `bson_iter_find()` makes it so that we don't have to go with the inefficient idea I had proposed in an earlier comment.

The ambiguity is definitely a thing. Actually, my hope is that with [SWIFT-193](https://jira.mongodb.org/browse/SWIFT-193), we should be able to do something non-ambiguous. The double optional idea is one of the potential approaches that I have PoC'd. I haven't pushed that yet because I need to look at MongoKitten's approach to this issue.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: hmmm, I'm torn. I think single parameter labels *can* be useful to better explain what a method is doing. and I think "get value for key" is definitely clearer and consistent. but, I also see that it's a bit verbose. given that this method is internal, we are the only users we need to worry about, and I would choose clarity over us having to type it a few more times. 

that said, maybe it should been`for` in both these places, and not `forKey`... feels more consistent with other times prepositions are used as argument labels. encode(to:), decode(as:from:)... 

in conclusion, idk 😅

Sources/MongoSwift/BSON/Document.swift
kmahar: I like Jeremy's idea of two separate calls so we can differentiate between those cases and throw an error when BSON is corrupt, let's do that

yeah, that ticket is all about trying to resolve these not-present vs. BSON null ambiguities in the driver, and this is one place where we have one. let's see what we decide to go with on that ticket and revisit this then. @Utagai  perhaps you can add a todo here mentioning SWIFT-193?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
Utagai: I actually prefer `for` and I think it saves some keystrokes. Since we're uncertain, how about we just roll with `for`? Seems like a nice middle-of-the-road compromise!

kmahar: good with me

Sources/MongoSwift/BSON/Document.swift

Sources/MongoSwift/MongoCollection+BulkWrite.swift
Utagai: Done (for both `setValue()` and `getValue()`)

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -398,14 +398,15 @@ public struct InsertManyResult {
...
+        self.insertedCount = try reply.getValue(for: "insertedCount") as? Int ?? 0
         self.insertedIds = insertedIds
 
-        if let writeErrors = reply["writeErrors"] as? [Document] {
+        if let writeErrors = try reply.getValue(for: "writeErrors") as? [Document] {
             self.writeErrors = try writeErrors.map { try BSONDecoder().decode(WriteError.self, from: $0) }
         }
 
-        if let writeConcernErrors = reply["writeConcernErrors"] as? [Document], writeConcernErrors.indices.contains(0) {
+        if let writeConcernErrors = try reply.getValue(for: "writeConcernErrors") as? [Document],
```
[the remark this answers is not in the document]
jmikola: Ah, I'm pretty sure I confused `Document.getValue()` with `BSONDecoder.getValue()` near the top of this PR. `Document.getValue()` happily returns nil if the key doesn't exist and would only throw if the document is malformed (e.g. < 5 bytes). If so, this LGTM.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -242,6 +247,18 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    /// Returns the current value (equivalent to the `currentValue` property) or throws on error.
+    internal func safeCurrentValue() throws -> BSONValue? {
+        switch self.currentType {
+        case .symbol:
```
jmikola: I realize the original `currentValue` property accessor had the same logic, but does this mean that the Swift driver cannot round-trip Symbol and DBPointer types w/o converting them to strings and documents? Or perhaps they can be roundtripped so long as we leave them in their Document and don't explicitly get them and set them on a new document to be stored.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -242,6 +247,18 @@ public class DocumentIterator: IteratorProtocol {
...
+            return try Symbol.asString(from: self)
+        case .dbPointer:
+            return try DBPointer.asDocument(from: self)
+        default:
+            return try DocumentIterator.bsonTypeMap[currentType]?.init(from: self)
```
jmikola: Looking at DBPointer and Symbol, I understand they're handled explicitly above. Their `init(from iter: DocumentIterator)` method always throws because we don't want users creating these types anew. Is there a reason we don't handle the `.undefined` case anywhere other than `Document.setValue()` when we check for a no-op replacement?

It's also conspicuously absent from `DocumentIterator.bsonTypeMap`. If `currentType` doesn't exist in `bsonTypeMap` here, does the `return` statement simply evaluate the `nil`? Perhaps this is covered under "Accessing Subscripts of Optional Type" in https://docs.swift.org/swift-book/LanguageGuide/OptionalChaining.html.

Sources/MongoSwift/BSON/Document+Sequence.swift
kmahar: yup, all of that is correct.
if a document was created directly from BSON data and has a Symbol or DBPointer type in it, we won't do anything. but if a value is read out from a doc, it will be upconverted to a different type.

Sources/MongoSwift/BSON/Document+Sequence.swift
kmahar: I *think* the implicit behavior here is that undefined is just upconverted to a nil

Utagai: Yeah, Kaitlin should be write, this should return `nil` for `.undefined`.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -207,7 +207,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
 
     /// Sets key to newValue. if checkForKey=false, the key/value pair will be appended without checking for the key's
     /// presence first.
-    private mutating func setValue(forKey key: String, to newValue: BSONValue?, checkForKey: Bool = true) throws {
+    private mutating func setValue(for key: String, to newValue: BSONValue?, checkForKey: Bool = true) throws {
```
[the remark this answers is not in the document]
Utagai: It may also be due to the label renames, but was still trivial to resolve.

Sources/MongoSwift/MongoCollection+Write.swift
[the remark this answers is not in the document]
Utagai: Yup, that's the intended behavior now that we distinguish between the two kinds of issues.

Utagai:
Seems like there are some issues from the merge not propagating the label changes to places in the codebase where we were actually calling the changed functions. I just pushed a fix for this.

PR #140, opened 2018-11-07
SWIFT-193: Proof-of-concept for various approaches

Utagai:
[SWIFT-193](https://jira.mongodb.org/browse/SWIFT-193)

**_deep breath_**

This PR details nearly 5 (one is a combination of others) approaches for improving BSON usability in the driver. These are sorted in order of least to most promising (1 < 2 < 3 == 4 <? 5):
1. `BSONValue??` - Using a double optional so as to distinguish between `nil` and `Optional(nil)`, where the former suggests a missing value and the latter suggests a true `null` value in the BSON.
2. `BSONMissing` - Using a dedicated object for representing a missing value/non-existent key. This is (somewhat) akin to something like `undefined` for Javascript.
3. `BSONNull` - Similar to `BSONMissing`, but instead of providing a dedicated object for representing a missing value/non-existent key, we provide one for the BSON `null` value.
4. `NSNull` - Analogous to `BSONNull`, but uses the familiar (to Objective-C folks, at least) `NSNull`. This is what [MongoKitten](https://github.com/OpenKitten/MongoKitten) does.
5. `BSONMissing && (BSONNull || NSNull)` - A combination of 2 && (3 || 4) that provides a guarantee of `BSONValue` return from any BSON. In other words, since we would no longer require the use of `nil` as a return from retrieving values from the document, there is no reason to use `BSONValue?` and unwraps.

All of these approaches aim to remove the ambiguity of `nil` in our current BSON usability. This is the immediate improvement, but this improves experience of dealing with `Document` and `BSONValue` and `[BSONValue]` in general.

This PR holds all the approaches mentioned above in a single branch, meaning implementations co-exist, in some sense. Currently, the code is set to use one approach, with other approach-specific logic commented out nearby. However, the semantics/code for how an end-user may use each of the approaches is kept in a test case meant to be a demo area (in `BSONValueTests.swift`). This means however that some of the demo usage code is not actually correct, since the approach-specific logic may be commented out. This is fine, since the test case code is just meant to show how the end-user experience would look like.

Approach 1 is shown via a Swift dictionary holding `BSONValue??` since the semantics should be similar. @kmahar may be able to go further into depth with this, as this was a great suggestion from her.

Approach 5 seems very promising in removing `BSONValue?` hassles, but couldn't be included in part of this PR since it can't co-exist with the other approaches so it includes a question mark in its ranking. I am totally open to making a new PoC branch that branches off this one that tries showcasing the possibility of removing `BSONValue?` via approach 5. Its downside is that we break from using `nil` and Swift dictionary style.

The 'demos' were made with the goal of showing how an end-user would do the following things:
1. Getting an existing key out of the dictionary
2. Checking for key existence
3. Checking to see if the returned value implies a true `nil` or missing.
4. Getting the value of a key for which the value is `null`.

Finally, this is a proof-of-concept branch that _should not be merged_. This exists solely to get the team some visual examples on how the approaches would work, look and feel like. Once we choose an approach, a true `SWIFT-193` branch will be used with the chosen approach and put up as a 'true' PR. This PR will not be merged and therefore, I imagine it needs no reviewers.

With my ramblings, notes and disclaimers out of the way, let me know about your thoughts, use-cases/scenarios I should add, and so on!!! (Sorry for the many words)

### DIFF
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -531,7 +531,7 @@ private class MutableArray: BSONValue {
 
     /// methods required by the BSONValue protocol that we don't actually need/use. MutableArray
     /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
-    required init(from iter: DocumentIterator) {
+    public static func fromIterator(from iter: DocumentIterator) -> BSONValue {
         fatalError("`MutableArray` is not meant to be initialized from a `DocumentIterator`")
     }
     func encode(to encoder: Encoder) throws {
@@ -581,7 +581,7 @@ private class MutableDictionary: BSONValue {
 
     /// methods required by the BSONValue protocol that we don't actually need/use. MutableDictionary
     /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
-    required init(from iter: DocumentIterator) {
+    public static func fromIterator(from iter: DocumentIterator) -> BSONValue {
         fatalError("`MutableDictionary` is not meant to be initialized from a `DocumentIterator`")
     }
     func encode(to encoder: Encoder) throws {
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -531,7 +531,7 @@ private class MutableArray: BSONValue {
 
     /// methods required by the BSONValue protocol that we don't actually need/use. MutableArray
     /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
-    public static func fromIterator(from iter: DocumentIterator) -> BSONValue {
+    public static func from(iterator iter: DocumentIterator) -> BSONValue {
         fatalError("`MutableArray` is not meant to be initialized from a `DocumentIterator`")
     }
     func encode(to encoder: Encoder) throws {
@@ -581,7 +581,7 @@ private class MutableDictionary: BSONValue {
 
     /// methods required by the BSONValue protocol that we don't actually need/use. MutableDictionary
     /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
-    public static func fromIterator(from iter: DocumentIterator) -> BSONValue {
+    public static func from(iterator iter: DocumentIterator) -> BSONValue {
         fatalError("`MutableDictionary` is not meant to be initialized from a `DocumentIterator`")
     }
     func encode(to encoder: Encoder) throws {
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -116,6 +116,33 @@ extension Array: BSONValue {
     }
 }
 
+extension NSNull: BSONValue {
+    public var bsonType: BSONType { return .null }
+
+    public static func fromIterator(from iter: DocumentIterator) throws -> BSONValue {
+        let nullType = bson_type_t(BSONType.null.rawValue)
+        guard bson_iter_type(&iter.iter) != nullType else {
+            throw MongoError.bsonDecodeError(message: "expected iterator type null for encoding a NSNull")
+        }
+
+        return NSNull()
+    }
+
+    public func encode(to storage: DocumentStorage, forKey key: String) throws {
+        guard bson_append_null(storage.pointer, key, Int32(key.count)) else {
+            throw bsonEncodeError(value: self, forKey: key)
+        }
+    }
+
+    public static func == (lhs: BSONValue, rhs: NSNull) -> Bool {
+        return lhs.bsonType == .null
+    }
+
+    public static func == (lhs: NSNull, rhs: BSONValue) -> Bool {
+        return rhs.bsonType == .null
+    }
+}
+
 public struct BSONNull: BSONValue, Equatable {
     public var bsonType: BSONType { return .null }
 
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -163,6 +163,10 @@ public struct BSONNull: BSONValue, Equatable {
         }
     }
 
+    public static func == (lhs: BSONNull, rhs: BSONNull) -> Bool {
+        return true
+    }
+
     public static func == (lhs: BSONValue, rhs: BSONNull) -> Bool {
         return lhs.bsonType == .null
     }
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -109,6 +109,7 @@ extension Array: BSONValue {
         for (i, v) in self.enumerated() {
             try arr.setValue(for: String(i), to: v as? BSONValue)
         }
+
         guard bson_append_array(storage.pointer, key, Int32(key.count), arr.data) else {
             throw bsonEncodeError(value: self, forKey: key)
         }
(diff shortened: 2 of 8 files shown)


Utagai:
So originally, I didn't want to muck around with the tests and dilute the diff, but now I realize that the changes wouldn't be too huge to the tests and it would actually at least show some more examples of using the new types and such.

And more importantly, I did not realize how annoying a red Travis build is to me, so I've fixed it :).

neallester:
Adding 

```swift
    var isMissing: Bool {get}
    var isNull: Bool {get}
```

To the BSONValue protocol would enable users to deal with those cases before unwrapping the (non-optional) BSONValue they retrieve.

Utagai:
@neallester Cool!!! Thanks! I am not sure if this helps a particularly great deal for lots of cases (correct me if I'm wrong, I'm pretty tired 😄), since most of the time, retrieving a `BSONValue` involves getting back a `BSONValue?` anyways, since `nil` is a valid return for us currently, and will be unless we opt for approach 5. However, on a non-optional `BSONValue`, this is a pretty nifty and concise!

This gives me the idea of potentially letting the `==` implementations for `BSONNull` and `BSONMissing` tolerate `BSONValue?`. If we do that, we could get away with things like:

```swift
if optionalBSONValue == BSONMissing() {
    // Missing
}
```
This is flexible enough to handle both `BSONValue` and `BSONValue?` and avoid an unwrap. We already avoid a cast by tolerating `BSONValue`, so why not. Also, we should see if we can utilize `bsonEquals()` to avoid multiple overloads of `==`. I haven't thought about this yet.

Anyways, this gets the same benefits that @neallester is trying to accomplish, without having to deal with quirks or oddities like:
1. `isNull` never being true if we decide to support using `BSONMissing`
2. `isMissing` always being false if we support `BSONNull`/`NSNull`
3. `isNull` being true despite the fact that we should never be `nil` (this makes sense if you think about it: Swift `nil` != MongoDB `null`, but has obvious cognitive friction since `nil` is generally thought of as a synonym for `null` by Swift programmers)
4. `isMissing` being `true` on a clearly concrete, existent type.
These problems disappear when you force an explicit comparison to something like `BSONMissing` or `BSONNull`.

neallester:
I was indeed thinking of the non-optional BSONValue choice. In addition to avoiding unwrapping for these two common (basically universal) cases, I think a non-optional return value would make the API easier to understand. The author and reader doesn't have to learn and remember that the else branch of a failed unwrap means "null" (or "missing" depending on what is ultimately a pretty arbitrary design choice). The meaning (value.isNull, value.isMissing) is right there in the code.

> isNull being true despite the fact that we should never be nil (this makes sense if you think about it: Swift nil != MongoDB null, but has obvious cognitive friction since nil is generally thought of as a synonym for null by Swift programmers)
isMissing being true on a clearly concrete, existent type.

I tend to think the opposite. Having the concrete object reduces cognitive friction because it clarifies that the value you have is a marker for BSONNull or BSONMissing completely distinct from the swift concept of nil.

Utagai:
I didn't mention this in my response, but if you're also thinking of approach 5, then I think your idea is not a bad addition at all to it approach 5, mainly because the optional is gone and we have the new BSON types.

I think the friction comment is minor, like I said in my original response, and it seems like it's really just a matter of personal taste/experience that we differ in. I doubt it's a major issue whether we do `isNull` or something like `== BSONNull` or `is BSONNull`. They are all pretty obvious!

Utagai:
Another realization in favor of approach 5: Potentially eliminating the need for returning things like `BSONValue?` will also make the addition of `Collection` conformance to `Document` much cleaner and easier, which is a great bonus for [SWIFT-192](https://jira.mongodb.org/browse/SWIFT-192).

kmahar:
So in simplest terms, the problem we have right now is that getting back a `nil` from `Document` subscript can mean two different things: that the key is not present, or the key is present and the value is BSON null. We need to provide a way to disambiguate these cases. We can do so by keeping the current behavior for one and changing the other, or changing the behavior for both. 

I am pretty sold on using NSNull for BSON null to disambiguate. I think it's logical to have an actual type, since null *is* a valid BSON type and is a distinct value that just contains no information. it's kinda similar to MinKey and MaxKey in that way. 

I am less convinced about the need for a BSONMissing type. I think that's exactly what optionals are meant to express: that a value is missing.

Utagai:
Right, that is the principal issue as mentioned, and I think all the approaches have a primary goal of disambiguating `nil`.

Your mention about how BSON null is an actual _type_ and therefore should have a type in the code makes sense to me and I agree. `BSONMissing` does not fit this bill in the same way, since 'Missing' is not a type and making a `BSONMissing` implies that it is. Plus, there is the obvious benefit that we get to feel closer to a Swift dictionary.

I think really it boils down to a matter of whether or not we want the safety of removing optionals or the semantic clarity of using only `BSONNull`/`NSNull`. Right now seems like we have a vote for using both as well as a vote for using just a null type.

kmahar:
also cc @jmikola, @mbroadst, not sure if they got emails about this since reviewers weren't requested

jmikola:
Having read the thread I concur with @kmahar's earlier comments:

> I am pretty sold on using NSNull for BSON null to disambiguate. I think it's logical to have an actual type, since null is a valid BSON type and is a distinct value that just contains no information. it's kinda similar to MinKey and MaxKey in that way.
>
> I am less convinced about the need for a BSONMissing type. I think that's exactly what optionals are meant to express: that a value is missing.

If I understand correctly, the only roadblock to using NSNull was BSONValue requiring a custom `init()` implementation, which you've mitigated by switching to a static factory method (`from()`). Extending NSNull as we do other types (e.g. Date, Int) seems preferable to introducing a new BSONNull type to fulfill the same purpose. I'm satisfied with approach 4 in the OP.

While one could argue that BSONMissing is convenient in some cases, I can think of at least two arguments against it:

 * Returning an optional from the subscript method is consistent with Dictionary (for general key access). I realize this isn't an issue for Collection protocol's [subscript method](https://developer.apple.com/documentation/swift/collection/1641358-subscript), since the index passed is always valid.
 * BSONMissing would not actually correspond to a real type code in the BSON specification, unlike all other structs/classes (including whatever we use to represent BSON null).

mbroadst:
I'm inclined to agree with @kmahar here: I think we should maintain the optional return type, and introduce support for using `NSNull` (or `BSONNull` if the former is not supported, though it appears it can be). If you take a look at a basic example of using Dictionaries in Swift:

```
var stringsAsInts: [String:Int] = [
    "zero" : 0,
    "one" : 1,
    "two" : 2
]

stringsAsInts["zero"] // Optional(0)
stringsAsInts["two"] // Optional(2)
stringsAsInts["three"] // nil
```

You will find it very similar to our `Document` class.  I think it would not be idiomatic to return a raw value in this case, simply to support explicitly determining if a value was missing, we'd simply be reinventing the concept of optionals in a special case for our driver (not to mention this would be a huge BC break for existing users, though this should not be a significant concern). 

Instead, it seems we can make `NSNull` extend `BsonValue`, which you have shown here, and I think that achieves all goals:
```
let doc: Document = [
  "hello": NSNull()
]

if let value = doc["hello"] as? NSNull {
  // now we know the value was set, and can use the unboxed value as a null
} else {
  // we know "hello" is not an existing key in the document, and have no value to work with
}
```

Utagai:
I am fine with either approach, and it is pretty obvious from the feedback that the `NSNull`/`BSONNull` is the most preferred route to take here. Personally, I will prefer `NSNull` unless it turns out there is some problems with it later down the road, but this PR seems to show that its possible.

I'm aware of the pros + cons of the approaches, and was personally on the fence (though perhaps interested by the novelty of eliminating optionals 😛). I believe the input has pushed me over to the `NSNull` side, so we'll be going with that.

With all that said, I'll close this PR. I believe you can re-open it if anyone wants to add anything more to the discussion.

Thank you everyone!

Issue #141, opened 2018-11-12
Enable SSl Issue 


iThink32:
I am currently using version 0.0.7 as I saw it in ur releases. I am able to connect and fetch a document from the mlab db but am not able to print the result or access it.

the error that I am getting is 

```
DEBUG:      cluster: Authentication failed: The SCRAM_SHA_1 authentication mechanism requires libmongoc built with ENABLE_SSL
```

mbroadst:
Hi @iThink32, can you provide more information about where you are using the driver? Specifically, what operating system, how have you installed libmongoc, what version libmongoc is installed, etc. Thank you!

iThink32:
@mbroadst I'm using this driver on iOS , on a simulator having a version of 12.0.I did not have to install libmongoc as this version of pod internally adds a mongo-embedded-c-driver folder as a dependancy.This directory contains the mongoc framework having a version of 1.13.0

acmorrow:
The build of the C driver contained in mongo-embedded-c-driver has SSL disabled, because the C drivers native crypto layer doesn't currently build on iOS. Since mongo-embedded-c-driver was envisioned as primarily being used to talk to the embedded server, not a remote server, when deployed on iOS, this was not seen as a problem. In general, our expectation was that mobile applications would not be directly talking to remote databases, but instead to remote app servers that sat in front of a database. So it would be interesting if you could share some additional information on your use case.

We will attempt to get this issue addressed in an upcoming release. The ticket to track is https://jira.mongodb.org/browse/CDRIVER-2850.

acmorrow:
@iThink32 - Well, the fastest way to see this fixed is for https://jira.mongodb.org/browse/CDRIVER-2850 to get fixed, and then we can re-enable SSL in the mongo-embedded-c-driver cocoapod. I think you should definitely reach out on that JIRA ticket and vote for it along with an explanation of why you need it fixed. If C programming on iOS is something you are familiar with and you feel comfortable working with the C driver source, submitting a pull request against the C driver (https://github.com/mongodb/mongo-c-driver) with a fix for the issue is probably the fastest way to see it fixed.

If you are interested in knowing more about how to do a cross compile of the C driver for iOS, you can work from the MongoDB build automation, available here:

- https://github.com/mongodb/mongo/blob/master/etc/evergreen.yml#L3753-L3821
- https://github.com/mongodb/mongo/blob/master/etc/evergreen.yml#L11165-L11177
[...]

PR #142, opened 2018-11-12
SWIFT-251: Create an issue template on the GitHub repo

Utagai:
[SWIFT-251](https://jira.mongodb.org/browse/SWIFT-251)

This PR adds some Github-specific contributing templates. 

The original ticket asks only for an issue template but I've also added one for pull requests. 

I've also added a `CONTRIBUTING.md`.

I've tested this on a tiny private repository of my own, since it seems like this needs to be merged into master before we start seeing changes on the web.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,62 @@
...
+### Operating system
+
+What does this command give you?
+```
+$ uname -a
+```
+
+### Driver version
+
+What does this command give you?
```
kmahar: if they use cocoapods, we can also ask them to include the podspec here.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+### Version of `libmongoc` and `libbson`
+
+What does this command give you?
+```
+$ brew list --versions mongo-c-driver # Only applies if you installed via brew
+```
+
+### How did you install `libmongoc` and `libbson` on your system
```
kmahar: let's move this up before version.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,62 @@
...
+```
+$ cat Package.swift # Only applies if you are using Swift package manager
+```
+
+### Version of `libmongoc` and `libbson`
+
+What does this command give you?
+```
+$ brew list --versions mongo-c-driver # Only applies if you installed via brew
+```
```
kmahar: are there ways people can check their version if they installed it in another way? if so, let's include them.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
+## Contributing
+
+First, thanks for considering contributing to our project! In general, our
+`Development.md` contains a lot of development-related information that could
+prove useful to a contributor!
+
+### Issues
+
+Before creating a new issue, look at current and past issues both on Github
```
kmahar: add links to: GitHub issues, JIRA project, outstanding PRs

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+**AND** MongoDB JIRA. Also look at outstanding PRs, and overall, ensure that
+the issue hasn't already been reported. In particular, MongoDB's JIRA has, at
+the time of writing this, most of the active development information on this
```
kmahar: let's remove "at the time of writing this", I think it will always be true.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+driver.
+
+If you decide that your issue truly is unique and new, then please follow the
+guidelines for submitting an issue, for which there is a template (you'll see
+this template when you go to create the issue). Make sure to write in clear and
+easy-to-understand English. Your issue is more likely to be resolved quicker if
```
kmahar:
```suggestion
easy-to-understand English. Your issue is more likely to be resolved quickly if
```

.github/ISSUE_TEMPLATE.md
Utagai: No idea. If they installed it manually, there is no real universal way. I am not familiar with other methods of installation. Are you?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+-->
+
+Some things to make sure you've done:
+* Ran `swiftlint` and ensured that your additions/deletions are not causing any
+new warnings or errors.
+* You did _not_ include any changes to the generated documentation (we update
+the documentation as part of a separate commit).
+* If you added new functionality, then you've added new tests that correctly
+verify its correctness.
+* You ran `make test` and `make benchmark` and ensured that the driver still
```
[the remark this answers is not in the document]
Utagai: Done.

.github/ISSUE_TEMPLATE.md
Utagai: Done. I have never used Cocoapods, so feel free to critique the addition I've made in response to this.

.github/ISSUE_TEMPLATE.md
Utagai: Done.

.github/CONTRIBUTING.md
Utagai: Done.

.github/CONTRIBUTING.md
Utagai: Done.

.github/ISSUE_TEMPLATE.md
Utagai: Seems like from the recent issue (#141) posted to the repo that pods install things like `libmongoc` locally somehow.

mbroadst: You can check via `pkg-config`, `brew` and `apt` - these are all supported via SwiftPM. So you'll want to include

jmikola: Alternatively, the `mongoc-version.h` and `bson-version.h` files in `/usr/include` (or equivalent) contain relevant information, but I don't think we want to suggest that users go poking around in there.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,66 @@
...
+
+## What is the problem? 
+
+**BE SPECIFIC**:
+* What is the _expected_ behavior and what is _actually_ happening?
+* Do you have any particular output that demonstrates this problem?
+* Do you have any ideas on _why_ this may be happening that could give us a
+clue in the right direction?
+* Did this issue arise out of nowhere, after an update (of the driver? of
+Swift?)?
```
jmikola: I might suggest replacing the parenthetical with: "(of the driver, server, and/or Swift)"

jmikola: I don't see it mentioned elsewhere in the new issue template, but it probably makes sense to request the server version/topology when applicable.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+Some things to make sure you've done:
+* Ran `swiftlint` and ensured that your additions/deletions are not causing any
+new warnings or errors.
+* You did _not_ include any changes to the generated documentation (we update
+the documentation as part of a separate commit).
+* If you added new functionality, then you've added new tests that correctly
+verify its correctness.
+* You ran `make test` and ensured that the driver still passes all of its
+pre-existing tests.
+* You read `CONTRIBUTING.md`!
```
jmikola: I would suggest bumping this to the top.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+<!--
+Thanks for contributing!
+-->
+
+Some things to make sure you've done:
+* Ran `swiftlint` and ensured that your additions/deletions are not causing any
+new warnings or errors.
+* You did _not_ include any changes to the generated documentation (we update
```
jmikola: I would suggest moving this to the bottom.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+<!--
+Thanks for contributing!
+-->
+
+Some things to make sure you've done:
```
jmikola: It might make sense to put all of this into the comment, since it's not text we'd want to show up in the visible PR text. If that's not the case, you could consider turning these into [task lists](https://help.github.com/articles/about-task-lists/#creating-task-lists) with checkbox syntax, which would allow users to check them off manually.

An alternative idea is to ask users to report things such as BC breaks and confirm that tests pass. One example of this is the [Symfony PHP framework's template](https://github.com/symfony/symfony/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

.github/ISSUE_TEMPLATE.md
Utagai: @jmikola I pretty much refrained from that as well because users can technically have such files _anywhere_ as long as their build configuration is aware of it. `pkg-config` is a little better, since it is supposed to manage these things, but even `pkg-config` can omit details if the build configuration the user is using explicitly specifies the location of where to link with the C libraries (anyone: correct me if I'm wrong).

In either case though, most people, myself included, will rely on `pkg-config` so it seems 'ok' to suggest using that. `apt` of course is okay int the same way `brew` is.

.github/ISSUE_TEMPLATE.md
Utagai: @jmikola Great idea about the server topology. Glad to have some more drivers-team perspective here, because I don't really have any experience debugging user issues with drivers, and did not think of asking for topology. Thanks.

Your other comment about the parenthetical is done.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+## What is the problem? 
+
+**BE SPECIFIC**:
+* What is the _expected_ behavior and what is _actually_ happening?
+* Do you have any particular output that demonstrates this problem?
+* Do you have any ideas on _why_ this may be happening that could give us a
+clue in the right direction?
+* Did this issue arise out of nowhere, after an update (of the driver,
```
kmahar:
```suggestion
* Did this issue arise out of nowhere, or after an update (of the driver,
```

kmahar:
could we combine this PR with [SWIFT-147](https://jira.mongodb.org/browse/SWIFT-247), adding a sample project for bug reports? I think it should be pretty trivial, just needs a bare bones SwiftPM project with a test file.

.github/PULL_REQUEST_TEMPLATE.md
Utagai: Done.

.github/PULL_REQUEST_TEMPLATE.md
Utagai: Done.

.github/PULL_REQUEST_TEMPLATE.md
Utagai: Wow didn't even know you could use these "task lists". Incredible, I've done that.

Utagai:
@kmahar Sounds fine to me, and also sounds like a `swift package init` will do most of the work for us.

Does putting it into `.github/` make sense, or should we just put the project in the repo as a subdirectory?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,79 @@
...
+```
+$ mongod --version
+```
+or, running this in a MongoDB shell connected to the relevant node(s):
+```
+> db.version()
+```
+
+### What is your server topology?
+How is your MongoDB deployment configured?
```
jmikola: Did you mean to have line breaks between the title and paragraph text here and in the "How did you install..." question below?

Utagai: Yep, you're right. I've also added more whitespace to separate the each question from each other.

Utagai: (It does not affect the rendering from what I can see).

kmahar:
I don't think people know to look in `.github`, I'd say let's put it in `Examples`.

Utagai:
@kmahar Done. I actually made it so that both the `main.swift` or a testcase suffices.

I think since these are users submitting bugs, it may seem more natural/realistic that they would be familiar with how to use the driver from an application perspective, rather than a test-case, so leaving that option to put in demo code into a `main.swift` seems like a good idea. But let me know if you think this is dumb.

Examples/BugReport/Package.swift
```diff
@@ -0,0 +1,24 @@
...
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "BugReport",
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+         .package(url: "https://github.com/mongodb/mongo-swift-driver", from: "0.0.3")
```
kmahar: this makes me realize, we should actually have people do `cat Package.resolved`. Swift rounds up patches to the latest patch version,  but won't automatically install an update when a new one comes out. so without knowing whether they have recently done `swift package update` we can't tell much from what's stated here -- they could still actually be on 0.0.3 or they could be all the way up to 0.0.7. 

I'm not sure if there is a .resolved equivalent for Cocoapods, but if so switch to that as well

Examples/BugReport/Package.swift
```diff
@@ -0,0 +1,24 @@
+// swift-tools-version:4.2
```
kmahar: we should support the same versions the driver works with, 4.0+. I believe this won't work on lower versions. I think you can provide two separate `Package.swift` files for different swift-tools-versions, I've seen people do it in repos before. probably info in the SPM docs

Examples/BugReport/Tests/BugReportTests/BugReportTests.swift
```diff
@@ -0,0 +1,14 @@
+import MongoSwift
+
+import XCTest
+import class Foundation.Bundle
```
kmahar: do we need this import?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,92 @@
...
+```
+$ uname -a
+```
+
+
+### Driver version
+
+What does this command give you?
+```
+$ cat Package.swift # Only applies if you are using Swift package manager
```
kmahar: looking at the `Package.swift` you added for the sample project made me realize, we should actually have people do `cat Package.resolved`. SPM rounds up patches to the latest patch version when installing (so 0.03 -> 0.0.7), as evidenced by the `Package.resolved` you committed.  

but SPM won't automatically install an update when a new one comes out, so without knowing whether they have recently done `swift package update` or if they very recently added the driver as a dependency, we can't tell much from what's stated here -- they could still actually be on 0.0.3 or they could be all the way up to 0.0.7, or anything in between. 

I'm not sure if there is a `Package.resolved` equivalent for Cocoapods, but if so switch to that as well

.github/ISSUE_TEMPLATE.md
Utagai: I've added a bit for `apt`. However, `pkg-config` doesn't seem to have much going for it besides `-cflags`. Does anyone know what we should be looking for in the output from `-cflags`. Perhaps @jmikola, if you use it on Linux?

Examples/BugReport/Package.swift
jmikola: Does it make sense to include `Package.resolved` in the example?

Utagai: Good catch, I missed this because I've been running Swift 4.2.

Examples/BugReport/Sources/BugReport/main.swift
```diff
@@ -0,0 +1,3 @@
+import MongoSwift
+
+print("Give us a minimal, working example of the bug you've found!")
```
jmikola: If we have `main.swift` for demonstrating the bug, what is the point of `BugReportTests.swift` (and, by extension, including Nimble)?

Examples/BugReport/Tests/BugReportTests/BugReportTests.swift
Utagai: Seems like a no. This was an artifact from `swift package init`.

Examples/BugReport/Package.swift
Utagai: I've just bumped it down to 4.0.

Examples/BugReport/Sources/BugReport/main.swift
kmahar: let's just make them do it one way. I think main.swift is fine, we can keep it simple

Utagai: Sounds fine to me. Done.

.github/ISSUE_TEMPLATE.md
kmahar: does --modversion not work?

Utagai: Yup, sorry, `--modversion` works for this too. But what I don't know is what the output of that looks like. Essentially, I would like to make a complete command that `grep`'s out the libraries we're looking for (`libmongoc` and `libbson`).

E.g., just want someone to confirm that something like:
```
$ pkg-config --modversion | grep -E "(libmongoc|libbson)"
```
Actually does the trick, or if the regex should be different.

kmahar:
```
$ pkg-config --modversion libmongoc-1.0
1.13.1-dev

$ pkg-config --modversion libbson-1.0
1.13.1-dev
```

Utagai: Thanks! Done.

Utagai: BTW: I think this hits on all the options we want to support, so I think we can resolve this.

PR #143, opened 2018-11-12
SWIFT-248: Enable Swiftlint on Tests/ and fix resulting linter issues

Utagai:
[SWIFT-248](https://jira.mongodb.org/browse/SWIFT-248)

This PR enables `swiftlint` to run on our `Tests/` directory, and also fixes all the resulting linter issues from doing so.

I've tested this locally, but we'll see if Travis says anything.


Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -238,7 +238,9 @@ final class MongoCollectionTests: XCTestCase {
         try coll.insertOne(["cat": nil])
         let distinct2 = try coll.distinct(fieldName: "cat", filter: [:])
         expect(distinct2).to(haveCount(3))
+        // swiftlint:disable trailing_closure
```
kmahar: is this another case like that you described in #swift?

Tests/MongoSwiftTests/.swiftlint.yml
```diff
@@ -0,0 +1,20 @@
+disabled_rules:
```
kmahar: what are the diffs from the top-level file and this?
it's too bad that it doesn't do something smart and apply *all* the .swiftlint.yml files as it goes down the tree, giving highest precendence to the lowest level. then we'd only have to include what's different in this file.

Tests/MongoSwiftTests/MongoCollectionTests.swift
Utagai: Yup.

Tests/MongoSwiftTests/.swiftlint.yml
Utagai: Yeah, I thought the same thing. The diff is:
```diff
7a8,9
>   - force_try
>   - force_cast
19,23d20
< 
< excluded:
<   - build
<   - .build
<   - docs
```

In plain English: I've removed the `excluded:` and added ignores for `force_try` and `force_cast`.

P.S.: Did _not_ expect ````` ```diff````` to work but man, it looks great...

kmahar: 😮 nice!

PR #145, opened 2018-11-19
SWIFT-193: Improve handling of BSON nulls

Utagai:
[SWIFT-193](https://jira.mongodb.org/browse/SWIFT-193)

This PR improves our handling of BSON nulls by introducing `NSNull` such that BSON nulls are represented by it, and missing values are Swift `nil`'s.

This PR is pretty big because it has widespread changes to the codebase for moving over to using `NSNull` in place of `nil` for BSON null values. There are _a lot_ of small little changes where we are able to migrate from using `BSONValue?` to `BSONValue`, as well as many changes where we use `NSNull` instead of `nil`. I believe they should all be fine, but would not be surprised whatsoever if a few are not, since there was often some ambiguity of what the past code meant when it had `nil` (was it a missing value? A BSON null value? Both?). This branch does however pass all the tests.

Also, I am now capable of typing `NSNull` in my sleep and have become a master of removing `?` at the end of `BSONValue?`.


Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 /// A struct wrapping a `BSONValue` type that allows for encoding/
 /// decoding `BSONValue`s of unknown type.  
-public struct AnyBSONValue: Codable {
+public struct AnyBSONValue: Codable, Equatable {
```
kmahar: nice

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +192,25 @@ internal struct _BSONDecodingStorage {
 /// Extend _BSONDecoder to add methods for "unboxing" values as various types.
 extension _BSONDecoder {
 
-    fileprivate func unboxBSONValue<T: BSONValue>(_ value: BSONValue?, as type: T.Type) throws -> T? {
+    fileprivate func unboxBSONValue<T: BSONValue>(_ value: BSONValue, as type: T.Type) throws -> T? {
```
kmahar: is there a reason this function returns an optional? I know I'm the one who wrote it that way but I'm not sure why...

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +192,25 @@ internal struct _BSONDecodingStorage {
...
         guard let typed = value as? T else {
             throw DecodingError._typeMismatch(at: self.codingPath, expectation: type, reality: value)
         }
         return typed
     }
 
-    fileprivate func unboxNumber<T: CodableNumber>(_ value: BSONValue?, as type: T.Type) throws -> T? {
-        guard let unwrapped = value else {
+    fileprivate func unboxNumber<T: CodableNumber>(_ value: BSONValue, as type: T.Type) throws -> T? {
```
kmahar: same Q as `unboxBSONValue`

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +192,25 @@ internal struct _BSONDecodingStorage {
...
 
-        guard let primitive = T(from: unwrapped) else {
+        guard let primitive = T(from: value) else {
             throw DecodingError._numberMismatch(at: self.codingPath, expectation: type, reality: value)
         }
         return primitive
     }
 
-    fileprivate func unbox<T: Decodable>(_ value: BSONValue?, as type: T.Type) throws -> T? {
+    fileprivate func unbox<T: Decodable>(_ value: BSONValue, as type: T.Type) throws -> T? {
```
kmahar: same q as `unboxBSONValue`

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -205,7 +205,7 @@ private class CrudTest {
     /// Given the response to a findAndModify command, verify that it matches the expected
     /// results for this `CrudTest`. Meant for use by findAndModify subclasses, i.e. findOneAndX. 
     func verifyFindAndModifyResult(_ result: Document?) {
-        if self.result == nil {
+        if self.result is NSNull {
```
kmahar: is stuff actually stored as null in the test files rather than being missing?

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -716,11 +716,15 @@ final class CodecTests: MongoSwiftTestCase {
                                   from: wrappedMinKey.canonicalExtendedJSON).x.value as? MinKey).to(equal(minKey))
     }
 
-    struct OptionalAnyBSONWrapper: Codable {
-        let val: AnyBSONValue?
+    struct OptionalAnyBSONWrapper: Codable, Equatable {
```
kmahar: maybe we should rename this to get rid of the `Optional`...

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -31,7 +31,11 @@ extension MongoCollection {
             return nil
         }
 
-        return InsertOneResult(insertedId: try document.getValue(for: "_id"))
+        if let insertedId = try document.getValue(for: "_id") {
```
kmahar: nit: I think
```swift
guard let insertedId = ... else { 
   // throw 
}}

 return InsertOneResult(insertedId: insertedId)
```
is more swifty if we're gonna throw in the `else`

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -193,12 +193,32 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
+            return iter.currentValue
+        }
         set(newValue) {
             do {
-                try self.setValue(for: key, to: newValue)
+                if let newValue = newValue {
+                    try self.setValue(for: key, to: newValue)
+                } else {
+                    // TODO SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
+                    self = self.filter {
```
kmahar: nit: one line

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -241,21 +234,25 @@ public class DocumentIterator: IteratorProtocol {
...
-    internal func safeCurrentValue() throws -> BSONValue? {
+    internal func safeCurrentValue() throws -> BSONValue {
         switch self.currentType {
         case .symbol:
             return try Symbol.asString(from: self)
         case .dbPointer:
             return try DBPointer.asDocument(from: self)
         default:
-            return try DocumentIterator.bsonTypeMap[currentType]?.init(from: self)
+            if let curVal = try DocumentIterator.bsonTypeMap[currentType]?.from(iterator: self) {
```
kmahar: similar to my other comment, prefer `guard let...` and then return when the `else` throws.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -814,8 +825,8 @@ public struct Timestamp: BSONValue, Equatable, Codable {
  *  e.g.
  *      4.0 (Double) != 4 (Int)
  *
- *  NOTE: This function will always return `false` if it is used with two arrays that are not of the type `[BSONValue]`
- *  or `[BSONValue?]`, because any arrays that are not of these types are not valid `BSONValue`'s.
+ *  NOTE: This function will always return `false` if it is used with two arrays that are not of the type `[BSONValue]`,
+ *  because any arrays that are not of this type is not a valid `BSONValue`'s.
```
```suggestion
 *  because only arrays composed of solely `BSONValue`s are valid BSON arrays.
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -293,7 +294,7 @@ extension _BSONEncoder {
         }
 
         // The top container should be a new container.
-        guard self.storage.count > depth else { return nil }
+        guard self.storage.count > depth else { return NSNull() }
```
kmahar: is this the right behavior here? similar checks in JSONEncoder seem to return either empty dictionaries or `nil`.

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: Ah good catch. I missed these because I think I only searched for instances of `nil` and `BSONValue?`.

This is essentially trying to return a `BSONValue` for which it can no longer return `nil`. If you look at where this function, it seems like they are trying to catch cases where it returns `nil` as well.

Utagai:
> will this properly consider overwriting an NSNull with an NSNull to be a no-op? (similar to minkey, maxkey)

@kmahar If I understand your question correctly, yes, it will. This should occur because of [this line](https://github.com/mongodb/mongo-swift-driver/blob/bc7e10828776671aba9fce41bd57bde0b0877207/Sources/MongoSwift/BSON/Document.swift#L238), where `.null` is used to return early.

The test [here](https://github.com/mongodb/mongo-swift-driver/blob/bc7e10828776671aba9fce41bd57bde0b0877207/Tests/MongoSwiftTests/DocumentTests.swift#L558) verifies that this is the case because of [this line](https://github.com/mongodb/mongo-swift-driver/blob/bc7e10828776671aba9fce41bd57bde0b0877207/Tests/MongoSwiftTests/DocumentTests.swift#L569).

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: In other words, I think this can return a non-optional now that we have `NSNull`, we're just going to have to change the places at which we call this function to check for `NSNull` instead of checking for `nil` and throwing.

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: See answer to question on `unboxBSONValue()`.

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: See answer to question on `unboxBSONValue()`.

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: Done.

I won't comment "Done." on the others to avoid potential notification spam, already feel bad about "See answer to question on unboxBSONValue()." 😅.

But they are all done, so feel free to resolve them all.

Sources/MongoSwift/MongoCollection+Write.swift
Utagai: You are totally right. Guess I missed some spots.

I realized that it will be weird when I move onto a language without `guard`. Once I form the habit of using `guard`'s like this, I'll really miss it in other languages.

Done.

Sources/MongoSwift/BSON/Document.swift

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: So I thought this was correct since the returned result of `box()` is always added to `containers`, which should be real `BSONValue`'s.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Good question, I'm actually not sure (I clearly assumed that they were at the time of making this change). Is there a way to test this?

My guess is we could take one of the JSON files and mess with it by removing an optional field in the JSON, but I don't actually know which of these fields can be wiped. I've tried a few, but all the fields I've tried just removing have failed both on master and this branch and in the same way, suggesting that those fields are probably mandatory for the test.

kmahar: just look at the find and modify files we are parsing and using this on, and see if they have a result field or have JSON null?

Utagai: Right, but this only shows what is going on with the current files. Would we not want to make sure that in the future any new files added would not omit the result field, or if they do, that it would be represented with `null` instead of just be missing outright?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -560,7 +561,13 @@ private class MutableDictionary: BSONValue {
         }
         set(newValue) {
             keys.append(key)
-            values.append(newValue)
+            if let newValue = newValue {
+                values.append(newValue)
```
jmikola: If the `else` branch below is simply going to search for the key and remove it, why should we bother with `keys.append(key)` for both cases instead of just doing it here?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -560,7 +561,13 @@ private class MutableDictionary: BSONValue {
...
+            } else {
+                guard let index = keys.index(of: key) else { return }
```
jmikola: Since we're always doing `keys.append(key)`, can this lead to a situation where our string exists twice in the `keys` array and we only remove the first occurrence that this `index(of:)` call finds?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -581,7 +588,7 @@ private class MutableDictionary: BSONValue {
 
     /// methods required by the BSONValue protocol that we don't actually need/use. MutableDictionary
     /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
-    required init(from iter: DocumentIterator) {
+    public static func from(iterator iter: DocumentIterator) -> BSONValue {
         fatalError("`MutableDictionary` is not meant to be initialized from a `DocumentIterator`")
```
jmikola: Any reason we use `fatalError()` here instead of preconditionFailure? Grepping through the driver, it looks like this is only used for MutableArray, Mutable Dictionary, and _BSONReferencingEncoder's deinitializer. It's outside the scope of this PR, but just something I'm curious about now that I've noticed it. Probably more a question for @kmahar.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -97,24 +97,36 @@ extension Array: BSONValue {
...
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        // An array is just a document with keys '0', '1', etc. corresponding to indexes
-        if self as? [BSONValue?] == nil {
-            throw MongoError.invalidArgument(message: "Cannot encode a non-BSONValue array element")
-        }
         var arr = Document()
         for (i, v) in self.enumerated() {
-            try arr.setValue(for: String(i), to: v as? BSONValue)
+            guard let val = v as? BSONValue else {
+                throw MongoError.invalidArgument(message: "Cannot encode a non-BSONValue array element")
```
jmikola: Is this still necessary because we don't have conditional conformance before Swift 4.2 and can't ensure that the Array contains only BSONValue instances?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -730,12 +741,12 @@ extension String: BSONValue {
         }
     }
 
-    public init(from iter: DocumentIterator) throws {
+    public static func from(iterator iter: DocumentIterator) throws -> BSONValue {
```
jmikola: I forget if I raised this point elsewhere, but would it make sense to drop the "iterator" label here in favor of `_` since there's only one argument? That would keep the new syntax closer to the old `init(from: ...)` API, since "from" is now the method name.

Sources/MongoSwift/BSON/Document+Codable.swift
```diff
@@ -46,7 +46,7 @@ extension Document: Codable {
         for key in container.allKeys {
             let k = key.stringValue
             if try container.decodeNil(forKey: key) {
-                try output.setValue(for: k, to: nil)
+                try output.setValue(for: k, to: NSNull())
```
jmikola: Is it still possible for `container.decode(AnyBSONValue.self, forKey: key)` below to yield an NSNull? If so, is this just necessary to catch actual `nil` values that might be in the container?

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -211,17 +211,10 @@ public class DocumentIterator: IteratorProtocol {
...
-            switch self.currentType {
-            case .symbol:
-                return try Symbol.asString(from: self)
-            case .dbPointer:
-                return try DBPointer.asDocument(from: self)
-            default:
-                return try DocumentIterator.bsonTypeMap[currentType]?.init(from: self)
-            }
-        } catch {
+            return try self.safeCurrentValue()
```
jmikola: It looks like `safeCurrentValue()` already handled `.symbol` and `.dbPointer`. Was this refactoring unrelated to the PR itself?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -243,26 +261,17 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
-                guard bson_append_null(self.data, key, Int32(key.count)) else {
-                    throw MongoError.bsonEncodeError(message: "Failed to set the value for key \(key) to null")
-                }
-            }
+            try newValue.encode(to: self.storage, forKey: key)
         }
     }
 
-    /// Retrieves the value associated with for as a `BSONValue?`, which can be nil.
+    /// Retrieves the value associated with for as a `BSONValue?`, which can be nil if the key does not exist in the
```
jmikola: I might suggest adding backticks around "for" so this reads more clearly (as referring to the `for` argument).

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -31,7 +31,12 @@ extension MongoCollection {
             return nil
         }
 
-        return InsertOneResult(insertedId: try document.getValue(for: "_id"))
+        guard let insertedId = try document.getValue(for: "_id") else {
+            // This case should not ever really happen, since we handle it above and give the document an _id.
+            throw MongoError.invalidArgument(message: "Failed to get value for _id from document")
```
jmikola: I don't think is an appropriate use of `MongoError.invalidArgument`, since it's a logical assertion occurring _after_ we've executed the operation and doesn't have much to do with the method's input arguments.

In PHP, I would throw an [UnexpectedValueException](https://github.com/mongodb/mongo-php-library/blob/1.4.2/src/Operation/FindAndModify.php#L221) for such a case, which is a [type of runtime exception](http://php.net/unexpectedvalueexception) (it depends on a result only known at runtime). The best equivalent in the Swift driver at the moment is probably `MongoError.commandError`, since you're using that when `distinct()` has a malformed response. Alternatively, a `preconditionFailure()` may be preferable here since a missing `_id` at this point means that either the driver has done something very wrong (i.e. the logic earlier in this same method hasn't done its job).

Tests/MongoSwiftTests/CodecTests.swift
jmikola: I agree that the "Optional" name prefix doesn't make sense, but is there even a point to this class? I imagine we could just use `AnyBSONValue` directly.

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -167,7 +167,7 @@ private class CrudTest {
         self.args = try operation.get("arguments")
         let outcome: Document = try test.get("outcome")
         self.error = outcome["error"] as? Bool
-        self.result = outcome["result"]
+        self.result = outcome["result"]!
```
jmikola: Is this forced unwrap only necessary because we're no longer using an optional for `CrudTest.result`?

The CRUD test's readme file still states:

> This will have some or all of the following fields:

So I would be included to leave `result` as an optional in case it doesn't exist. I could see that being the case for tests where `error` is true and a result is never returned nor asserted.

Tests/MongoSwiftTests/CrudTests.swift
jmikola: "FindOneAndDelete when no documents match" in `findOneAndDelete.json` is an example where `result` is `null` in JSON. In that case, I think the `findOneAndDelete()` method should actually return `nil` instead of NSNull, since NSNull represents a BSON null value within a document. This may be related to my earlier comment about parsing the CRUD test.

If we're parsing the entire CRUD test as a BSON Document (instead of directly from JSON), then I suppose we may have to interpret an NSNull in `outcome["result"]` as a `nil` for this purpose.

Utagai: Yep, in the [`init()` call](https://github.com/mongodb/mongo-swift-driver/pull/145/files#diff-e8428080876ca5bc2e6ecb19e22d6771L163) for the initialization of `self.result`, we parse in the JSON as a `Document`, so we'd use `NSNull` for JSON `null`.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: It may be clearer to ask what exactly this return value here is meant to indicate.

Is the return value from this line supposed to indicate:
1. A true null value?
2. A missing value?
3. An "error" or exceptional situation leading to a general Swift `nil`.

Tagging @kmahar, since she's most knowledgeable about Encoder/Decoder logic.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Yup, that is the only reason for it being there. The decision on what to do here depends on the earlier conversation about `self.result`.

If "missing" values for fields in the JSON are `null` then we can just keep this as is otherwise it makes more sense to revert it back to an optional.

Sources/MongoSwift/MongoCollection+Write.swift
Utagai: I'll switch to a `preconditionFailure`, since this should _never_ happen.

Sources/MongoSwift/BSON/Document.swift
Utagai: Great idea! Done.

Tests/MongoSwiftTests/CrudTests.swift
jmikola: The `null` in the JSON is technically asserting the absence of a Document for the various `findAndModify` helpers (as explained in [these CRUD spec doc blocks](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#find-and-modify)). Since our helpers in `MongoCollection+FindAndModify` return `CollectionType?`, I think `null` here (even if parsed as NSNull since we interpret as a Document) should lead us to assert that those helpers return `nil`.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -272,13 +273,13 @@ extension _BSONEncoder {
...
 
-    fileprivate func box<T: Encodable>(_ value: T) throws -> BSONValue? {
+    fileprivate func box<T: Encodable>(_ value: T) throws -> BSONValue {
         // if it's already a `BSONValue`, just return it, unless if it is an
         // array. technically `[Any]` is a `BSONValue`, but we can only use this
         // short-circuiting if all the elements are actually BSONValues.
         if let bsonValue = value as? BSONValue, !(bsonValue is [Any]) {
             return bsonValue
-        } else if let bsonArray = value as? [BSONValue?] {
+        } else if let bsonArray = value as? [BSONValue] {
```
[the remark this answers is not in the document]
Utagai: Is it actually redundant? Because the first `if` check checks to make sure that the given value is _not_ `[Any]`. That means that even if the value was `[BSONValue]`, it is still also `[Any]`, and would fail the first check.

Then the difference becomes that in the case of `[BSONValue]`, we get back a `[BSONValue]` as opposed to a `BSONValue`. Unsure if perhaps this doesn't make a difference since the return value of the function is a `BSONValue`...

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -116,7 +116,7 @@ internal class _BSONDecoder: Decoder {
 
     // Returns the data stored in this decoder as represented in a container keyed by the given key type.
     public func container<Key>(keyedBy type: Key.Type) throws -> KeyedDecodingContainer<Key> {
-        guard self.storage.topContainer != nil  else {
+        guard !(self.storage.topContainer is NSNull)  else {
```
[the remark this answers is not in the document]
Utagai: So, I know what you mean, but I kept this in just because apparently, from the code, the intended behavior for a `null` is to throw this error with a message tailored for `null`. If we remove this and merge it into the codepath for other `BSONValue` types, then we lose this distinction.

This isn't inherently a bad thing, but we should make sure that this change in behavior is what we want.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +192,25 @@ internal struct _BSONDecodingStorage {
...
         guard let typed = value as? T else {
             throw DecodingError._typeMismatch(at: self.codingPath, expectation: type, reality: value)
         }
         return typed
     }
 
-    fileprivate func unboxNumber<T: CodableNumber>(_ value: BSONValue?, as type: T.Type) throws -> T? {
-        guard let unwrapped = value else {
+    fileprivate func unboxNumber<T: CodableNumber>(_ value: BSONValue, as type: T.Type) throws -> T {
+        guard !(value is NSNull) else {
```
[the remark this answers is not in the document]
Utagai: Ditto my comment to your earlier comment for `_BSONDecoder.container()`.

I believe that when we reach a consensus on `_BSONDecoder.container()`, we can propagate it to other instances.

Utagai:
@jmikola: About `OptionalBSONValueWrapper` (I missed this comment earlier), I think this is to test the usability of the Encoder/Decoder, since the idea behind those things is that you can capture a BSON document as a Swift structure. Technically, `AnyBSONValue` should work too, but I think the wrapper type was made the same way a user may make their own struct conforming to a certain BSON/JSON structure/schema.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: I am not sure I follow, sorry: doesn't this suggest that we should just keep a non-optional `result` if we're expecting a true null anyways in the JSON? Otherwise, we'd have to make it optional but make sure we treat both `nil` and `NSNull` in the same way whenever we use `result` in any logic.

Correct me if I've misunderstood something -- I probably have.

Sources/MongoSwift/BSON/BSONEncoder.swift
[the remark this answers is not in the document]
jmikola: I wasn't looking at the condition expressions themselves, but rather the fact that the first branch returns. Therefore, `else if` can simply be written as its own `if` block (keeping the same conditional expression).

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
jmikola: Is this related to the outstanding questions in https://github.com/mongodb/mongo-swift-driver/pull/145#discussion_r235508617?

Sources/MongoSwift/BSON/BSONEncoder.swift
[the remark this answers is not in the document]
Utagai: Ah, I gotcha. Done!

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
Utagai: Nope, I don't think so. `valueNotFound` is specifically for `nil`'s that correspond to actual `null` values. The questions in the comment are about a situation where we are not actually sure which kind of `nil` we are dealing with, but in this case, we know.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Right. If it is indeed the case that `"result"` may not exist sometime in the future, then we should revert this back to an optional. This is something I was hoping to get an answer for with [this comment](https://github.com/mongodb/mongo-swift-driver/pull/145#discussion_r235063481), so thanks!

I'll move over any new updates on this convo over to the earlier thread. Feel free to resolve this.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Done.

In the case of the `DistinctTest`, I make it so that if the result is non-`nil`, then we proceed as normal. If `self.result` _is_ `nil`, then I expect an empty document instead. This seems to agree with what `distinct()` returns in the Mongo shell when I use it on something like an empty collection, so I imagine someone leaving out `"result"` in the JSON is implying that the result should be empty.

This is what I've done, for quick reference:
```swift
/// A class for executing `distinct` tests
private class DistinctTest: CrudTest {
    override func execute(usingCollection coll: MongoCollection<Document>) throws {
        let filter = self.args["filter"] as? Document
        let fieldName: String = try self.args.get("fieldName")
        let options = DistinctOptions(collation: self.collation)
        // rather than casting to all the possible BSON types, just wrap the arrays in documents to compare them
        let resultDoc: Document = ["result": try coll.distinct(fieldName: fieldName,
                                                               filter: filter ?? [:],
                                                               options: options)]
        let expectedDoc: Document
        if let result = self.result {
            expectedDoc = ["result": result]
        } else {
            expectedDoc = [:]
        }
        expect(resultDoc).to(equal(expectedDoc))
    }
}
```

In particular:
```swift
        let expectedDoc: Document
        if let result = self.result {
            expectedDoc = ["result": result]
        } else {
            expectedDoc = [:]
        }
        expect(resultDoc).to(equal(expectedDoc))
```

Tests/MongoSwiftTests/CodecTests.swift
kmahar: you can, i think my intent here was just to check that encoding and decoding worked as expected when the type wasn't directly encoded/decoded and was optional. which one could argue is something we should just rely on codable to get right and not test ourselves....
however, now that this is a non optional, this test seems redundant with `testAnyBSONValueIsBSONCodable`, which uses an identical `AnyBSONStruct`

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -193,12 +193,30 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      *  d["a"] = 1
      *  print(d["a"]) // prints 1
      *  ```
+     * A nil return suggests that the subscripted key does not exist in the `Document`. A true BSON null is returned as
+     * an `NSNull`.
      */
     public subscript(key: String) -> BSONValue? {
-        get { return DocumentIterator(forDocument: self, advancedTo: key)?.currentValue }
+        get {
+            let iterOptional = DocumentIterator(forDocument: self)
```
kmahar: nit: just unwrap the initializer call directly - 
```swift
guard let iter = DocumentIterator(forDocument: self) else { 
```

Utagai: Oh, thanks! Not sure why I didn't do this before.

Done.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: No reason to just put it in the non-deletion case, so I've done that.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: That is actually possible, I think. But since I've already moved the `append` call inside the non-`nil` case, this should no longer be an issue.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: The only difference between fatalError vs preconditionFailure that I know of is that one of them is silent in release builds?

My guess is that @kmahar followed the Encoder & Decoder implementations from stdlib and they use fatalError here.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Yep! This goes away once we can guarantee conditional conformance to `BSONValue` for arrays of only `[BSONValue]`!!!

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: Hm, perhaps. I actually kept this in because now calls look like:
```swift
from(iterator: myIter)
```

Which looked pretty nice to me cause it reads like English ("from the iterator under the variable: `myIter`").

I don't feel very passionate about this so we can change it too, I think `from(myIter)` looks and reads fine as well!

Sources/MongoSwift/BSON/Document+Codable.swift
Utagai: Yeah, this is because now `AnyBSONValue` can represent a true null via `NSNull`. This is for that case.

Sources/MongoSwift/BSON/Document+Sequence.swift
Utagai: Yes, just removal of code duplication! 😄

Tests/MongoSwiftTests/CodecTests.swift
Utagai: Oh right, good point. Should I just cut this part of the test out?

Sources/MongoSwift/BSON/BSONEncoder.swift
kmahar: so, this method is meant to be analogous to [`JSONEncoder`'s `box_` method.](https://github.com/apple/swift/blob/8b9cc180033d7c786a8646d7f71ae28802f69142/stdlib/public/SDK/Foundation/JSONEncoder.swift#L885) `box_` is called [at the very top level](https://github.com/apple/swift/blob/8b9cc180033d7c786a8646d7f71ae28802f69142/stdlib/public/SDK/Foundation/JSONEncoder.swift#L250) of `JSONEncoder.encode`, and a nil return value there throws, with the error message that the type created no top-level container - I think this would happen if you basically wrote an empty `encode(to:)` method for your type that never requested a container. so, I think a nil return here is meant to indicate that the type encoded *nothing*, not even an empty container.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: can we spice up this error message a bit with info about the offending array element? i.e. index, type, string rep

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I'm fine with either of these

Sources/MongoSwift/BSON/Document+Codable.swift
```diff
@@ -15,10 +15,11 @@ extension Document: Codable {
         var container = encoder.container(keyedBy: _BSONKey.self)
         for (k, v) in self {
             let key = _BSONKey(stringValue: k)!
-            if let val = AnyBSONValue(ifPresent: v) {
-                try container.encode(val, forKey: key)
-            } else {
+            let val = AnyBSONValue(v)
```
kmahar: nit: let's only bother creating this struct if we're gonna encode it (under the else condition)

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -193,12 +193,30 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      *  d["a"] = 1
      *  print(d["a"]) // prints 1
      *  ```
+     * A nil return suggests that the subscripted key does not exist in the `Document`. A true BSON null is returned as
+     * an `NSNull`.
      */
     public subscript(key: String) -> BSONValue? {
-        get { return DocumentIterator(forDocument: self, advancedTo: key)?.currentValue }
+        get {
+            //let iterOptional = DocumentIterator(forDocument: self)
```
kmahar: left in by mistake?

Tests/MongoSwiftTests/CodecTests.swift
kmahar: ya, unless if there's anything we can find here that the previous test isn't checking, we can scrap it

Sources/MongoSwift/BSON/BSONEncoder.swift
kmahar: let's add some tests verifying our behavior matches JSONEncoder?

Tests/MongoSwiftTests/CrudTests.swift
kmahar: all sgtm

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: Ah right, forgot about this. Thanks for bringing it up.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: Makes sense to me.

Sources/MongoSwift/BSON/BSONValue.swift
Utagai: We absolute can! This is something we couldn't do before when we threw this on attempt to converting the array itself to `[BSONValue?]`!

Done.

Sources/MongoSwift/BSON/Document+Codable.swift
Utagai: Done.

Sources/MongoSwift/BSON/Document.swift
Utagai: Oops, yup! Thanks! 😅 

Removed.

Tests/MongoSwiftTests/CodecTests.swift
Utagai: Done.

I migrated over the testing code for `NSNull` in `testOptionalAnyBSONValue()` as well as the test for array that is not all `BSONValue`. That test actually calls `encode()` from the `BSONValue` protocol... which seems a little odd if its not related to `BSONEncoder`/`BSONDecoder`, but oh well.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: Done, although would suggest you check it out in case I did something wrong with writing the test.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -172,13 +172,13 @@ internal struct _BSONDecodingStorage {
     fileprivate var count: Int { return self.containers.count }
 
     /// The container at the top of the stack.
-    internal var topContainer: BSONValue? {
+    internal var topContainer: BSONValue {
```
jmikola: In light of https://github.com/mongodb/mongo-swift-driver/pull/145/files#r235482868, I just want to confirm that the precondition here ensures that we never access `topContainer` when the `containers` array is empty. Therefore, this is always returning an actual BSONValue, which is why we can force-unwrap in the `return` statement.

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
jmikola: Do I understand correctly that JSONDecoder would report `DecodingError.valueNotFound` only for a JSON `null` value, and `DecodingError._typeMismatch` for anything else that isn't a JSON `object` type? Is this related to JSONDecoder using `nil` for JSON `null`, or do they also use NSNull?

If this behavior isn't actually dictated, I personally don't see the benefit of treating BSON `null` differently in this case. Relying on the `DecodingError._typeMismatch` error handling below seems fine and would let us delete a bit of code here.

Looking at the [DecodingError documentation](https://developer.apple.com/documentation/swift/decodingerror), `valueNotFound` says:

> An indication that a nonoptional value of the given type was expected, but a null value was found.

Conversation in [SR-5156](https://bugs.swift.org/browse/SR-5156) leads me to believe this is actually thrown for `nil` values.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -621,12 +634,12 @@ internal struct _BSONKey: CodingKey {
 }
 
 internal extension DecodingError {
-    static func _typeMismatch(at path: [CodingKey], expectation: Any.Type, reality: BSONValue?) -> DecodingError {
+    static func _typeMismatch(at path: [CodingKey], expectation: Any.Type, reality: BSONValue) -> DecodingError {
```
jmikola: Is this a candidate for `fileprivate`? AFAICT, `_typeMismatch` is only called from within `BSONDecoder.swift`.

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
jmikola: I just realized that `_typeMismatch` and `_numberMismatch` are both just convenience methods to create a `DecodingError.typeMismatch`. This may actually be unrelated to my `_BSONDecoder.container()` question above, since that is about two different error codes and JSONDecoder's prior art.

Sources/MongoSwift/BSON/BSONValue.swift
jmikola: @Utagai: Good point that this will lead to self-documenting code.

No objections with leaving as-is.

Tests/MongoSwiftTests/CrudTests.swift
jmikola: If `outcome.result` field is missing from the CRUD spec test case's JSON, it would be correct to not assert anything about the return value of the CRUD method. Consider the case where `outcome.error` is `true` and the spec test only wants to assert that an exception was thrown and there is no return value to examine. The only edge case AFAIK would be a write method, since those exceptions may contain a write result within.

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -401,7 +405,12 @@ private class DistinctTest: CrudTest {
         let resultDoc: Document = ["result": try coll.distinct(fieldName: fieldName,
                                                                filter: filter ?? [:],
                                                                options: options)]
-        let expectedDoc: Document = ["result": self.result]
+        let expectedDoc: Document
+        if let result = self.result {
+            expectedDoc = ["result": result]
```
jmikola: The point I raised in https://github.com/mongodb/mongo-swift-driver/pull/145/files#r236746341 may apply here as well. I think `expect(resultDoc).to(equal(expectedDoc))` should be limited to the condition where `result` is non-nil.

Tests/MongoSwiftTests/CrudTests.swift
jmikola: @Utagai: Your earlier comment in https://github.com/mongodb/mongo-swift-driver/pull/145/files#r235552252 was:

> If self.result is nil, then I expect an empty document instead.

Looking at the current code, if `self.result` is `nil`, `verifyFindAndModifyResult()` returns earlier and doesn't actually assert anything. If the `outcome.result` key is `null` for some reason, I think it's sensible to assert that `distinct()` returns an NSNull value by crafting a `["result": NSNull]` literal. Of course, we know that's not possible so the test would fail.

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: Yep, that's exactly what the `precondition` exists here for. In general, the only case where a force-unwrap is tolerable in good Swift is when its backed by a `precondition` or its something you can reasonably assume to be true in a test, so this seems fine in terms of both correctness and code quality.

Sources/MongoSwift/BSON/BSONDecoder.swift
Utagai: I don't think this is possible because the container extension (`DecodingError`) is of `internal` access level, which is weaker than `fileprivate`, so you can't make `_typeMismatch` any more stricter than `internal`.

Also, I believe this is referenced in other files too, e.g. `Document+Codable.swift` (someone can double check me), so `fileprivate` would not let us be able to use it there.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: @jmikola I am happy to defer to you on this because I'm not too familiar with the specs or the JSON conventions here, but I'd like to make sure I understand your point:

Are you suggesting that instead of asserting against an empty document `[:]` in the `nil` case, we should just avoid asserting whatsoever, or are you saying we should instead assert against `["result": NSNull()]`?

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Just read this comment after posting my question in response to the point raised in that link. It seems like it's actually correct for us to just avoid the assertion if `self.result` is `nil`?

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
Utagai: So I think SR-5156 seems to have been resolved, and so looking at the current code in the Swift stdlib, it seems like they are doing what we are doing here by checking against `NSNull` and throwing essentially the same exact error as we are here. So it seems like their idea of finding a "null value" is `NSNull`?

I think you're right that if it is not really dictated/documented anywhere, we shouldn't need to spend the extra LoC on this particular case. I doubt it is a big issue either way, but perhaps it may end up being helpful to have some ability to distinguish between a null vs. a type mismatch amongst non-null types?

I think I will defer this to @kmahar again as our resident `Encoder`/`Decoder` expert to get her opinion on the matter.

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Done!

Tests/MongoSwiftTests/CrudTests.swift
Utagai: Since [this](https://github.com/mongodb/mongo-swift-driver/pull/145/files#r236788721) clarified what @jmikola wanted and is done, this should also be done.

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -401,8 +405,11 @@ private class DistinctTest: CrudTest {
         let resultDoc: Document = ["result": try coll.distinct(fieldName: fieldName,
                                                                filter: filter ?? [:],
                                                                options: options)]
-        let expectedDoc: Document = ["result": self.result]
-        expect(resultDoc).to(equal(expectedDoc))
+        let expectedDoc: Document
+        if let result = self.result {
+            expectedDoc = ["result": result]
```
jmikola: Should this be rewritten as `let expectedDoc: Document = ["result": result]`? There's probably no longer a reason to declare `expectedDoc` outside of this conditional.

Utagai: Yup, done.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -97,24 +97,40 @@ extension Array: BSONValue {
...
+            try arr.setValue(for: String(i), to: val)
         }
+
         guard bson_append_array(storage.pointer, key, Int32(key.count), arr.data) else {
             throw bsonEncodeError(value: self, forKey: key)
         }
     }
 }
 
+extension NSNull: BSONValue {
```
kmahar: nit: add docstring matching other extension docstrings.

Utagai: Done.

Utagai:
Just tried using the Github web interface to resolve the merge conflicts, hoping that it doesn't break anything... 🤞

kmahar:
there's now a conflict because I merged in the PR removing benchmarks. I would just revert your change to that file and once we release 0.0.8 and the benchmarks use it, we can make that change there.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -714,41 +718,51 @@ final class CodecTests: MongoSwiftTestCase {
...
-        let arr3: [Any?] = [1, "hi", nil, Int16(4)]
-        expect(try arr3.encode(to: DocumentStorage(), forKey: "arr3")).to(throwError())
+    func testIncorrectEncodeFunction() {
+        let encoder = BSONEncoder()
 
-        // nil value
-        expect(try encoder.encode(OptionalAnyBSONWrapper(nil))).to(beNil())
+        // A top-level `encode()` problem should throw an error, but any such issues deeper in the recursion should not.
+        expect(try encoder.encode(IncorrectTopLevelEncode(NSNull()))).to(throwError())
+        expect(try encoder.encode(CorrectTopLevelEncode(NSNull()))).toNot(throwError())
```
kmahar: can we also compare this result against what we expect it to be?

kmahar: perhaps also state that this is the behavior of `JSONEncoder`

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -242,9 +243,9 @@ private class _BSONReferencingEncoder: _BSONEncoder {
 
     /// Finalizes `self` by writing the contents of our storage to the referenced encoder's storage.
     deinit {
-        let value: BSONValue?
+        let value: BSONValue
         switch self.storage.count {
-        case 0: value = nil
+        case 0: value = NSNull()
```
kmahar: it appears that this case returns an empty `NSDictionary` in `JSONEncoder`. should this be an empty document? 

I realize that I am the one who put nil here initially but that may have been wrong. I don't think I ever wrote any tests for _BSONReferencingEncoder or Decoder... which was a poor choice.... just opened [SWIFT-263](https://jira.mongodb.org/browse/SWIFT-263) about that

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
kmahar: let's just use typeMismatch

Utagai:
Just noticed that both our old code and new code have a discrepancy with `JSONEncoder` where we do not encode an empty document when we see our respective `null` from box.

`JSONEncoder` does this thing where in all instances besides the top-level, it will coalesce `nil` and treat it as an empty document. This _includes_ the case of bad `encode()` implementations that are not at the top-level. In the top-level, it throws an error.

We were not doing this at non-top-levels and instead encoding a null value. This should now be fixed.

Tests/MongoSwiftTests/CodecTests.swift
Utagai: Done.

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
Utagai: Done.

Tests/MongoSwiftTests/CodecTests.swift
kmahar: nit: this is redundant with the next check now, since that will also fail if encoding throws
```suggestion
```

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -299,7 +292,8 @@ private struct _BSONKeyedDecodingContainer<K: CodingKey> : KeyedDecodingContaine
     private func decodeNumber<T: CodableNumber>(_ type: T.Type, forKey key: Key) throws -> T {
         let entry = try getValue(forKey: key)
         return try self.decoder.with(pushedKey: key) {
-            guard let value = try decoder.unboxNumber(entry, as: type) else {
+            let value = try decoder.unboxNumber(entry, as: type)
+            guard !(value is NSNull) else {
```
kmahar: is it possible that `unboxNumber` ever returns an `NSNull`? I think it will only return something of the provided type, which must be a `CodableNumber`

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -456,7 +459,8 @@ private struct _BSONUnkeyedDecodingContainer: UnkeyedDecodingContainer {
     private mutating func decodeNumber<T: CodableNumber>(_ type: T.Type) throws -> T {
         try self.checkAtEnd()
         return try self.decoder.with(pushedKey: _BSONKey(index: self.currentIndex)) {
-            guard let typed = try self.decoder.unboxNumber(self.container[currentIndex], as: type) else {
+            let typed = try self.decoder.unboxNumber(self.container[currentIndex], as: type)
```
kmahar: same q about `unboxNumber` ever actually returning `NSNull`?

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -141,17 +133,17 @@ internal class _BSONDecoder: Decoder {
 
     // Returns the data stored in this decoder in a container appropriate for holding values with no keys.
     public func unkeyedContainer() throws -> UnkeyedDecodingContainer {
-        guard self.storage.topContainer != nil else {
+        guard !(self.storage.topContainer is NSNull) else {
```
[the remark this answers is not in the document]
Utagai: Done.

Sources/MongoSwift/BSON/BSONDecoder.swift
kmahar: ya, l agree, not sure why that was returning an optional. oops

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
kmahar: ya, these could probably also just be lumped under the `numberMismatch` error, the errors thrown are almost identical so may as well simplify the logic.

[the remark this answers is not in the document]
Utagai: Done.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -38,14 +38,15 @@ public class BSONEncoder {
 
         let encoder = _BSONEncoder(options: self.options)
 
-        guard let topLevel = try encoder.box(value) else {
+        let boxedValue = try encoder.box_(value)
```
kmahar: nit: just do `guard let boxedValue = try ...`

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +176,21 @@ internal struct _BSONDecodingStorage {
...
     }
 
-    fileprivate func unboxNumber<T: CodableNumber>(_ value: BSONValue?, as type: T.Type) throws -> T? {
-        guard let unwrapped = value else {
-            throw DecodingError._typeMismatch(at: self.codingPath, expectation: type, reality: value)
-        }
-
-        guard let primitive = T(from: unwrapped) else {
+    fileprivate func unboxNumber<T: CodableNumber>(_ value: BSONValue, as type: T.Type) throws -> T {
+        guard !(value is NSNull), let primitive = T(from: value) else {
```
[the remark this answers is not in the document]
Utagai: Done.

Sources/MongoSwift/BSON/BSONEncoder.swift
Utagai: I always forget about `guard let` it seems. Done.

Utagai:
@kmahar No worries. I'd rather clean up the debt now and pay the price, rather than have us pay even more for it in the future.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +176,21 @@ internal struct _BSONDecodingStorage {
...
             throw DecodingError._numberMismatch(at: self.codingPath, expectation: type, reality: value)
         }
         return primitive
     }
 
-    fileprivate func unbox<T: Decodable>(_ value: BSONValue?, as type: T.Type) throws -> T? {
+    fileprivate func unbox<T: Decodable>(_ value: BSONValue, as type: T.Type) throws -> T {
```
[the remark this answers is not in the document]
Utagai: Per our offline discussion, this is better left as-is.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -192,25 +176,21 @@ internal struct _BSONDecodingStorage {
 /// Extend _BSONDecoder to add methods for "unboxing" values as various types.
 extension _BSONDecoder {
 
-    fileprivate func unboxBSONValue<T: BSONValue>(_ value: BSONValue?, as type: T.Type) throws -> T? {
+    fileprivate func unboxBSONValue<T: BSONValue>(_ value: BSONValue, as type: T.Type) throws -> T {
```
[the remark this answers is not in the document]
kmahar: as long as they both still contain the same info, I think that's fine, and even preferable.

PR #146, opened 2018-11-21
SWIFT-192: Make Document conform to Collection protocol

Utagai:
[SWIFT-192](https://jira.mongodb.org/browse/SWIFT-192)

This PR adds `Collection` conformance to `Document`.

This diff is not too big, but this ticket took quite a while due to various and significant issues with trying to get `Collection` conformance working for `Document`!!! I am very happy to finally get this PR out!

Performance improvements for reaching runtime guarantees of `Collection` conformance will be done separately.

I've also refrained from combining `Document+Sequence.swift` with `Document+Collection.swift`, since with the completion of SWIFT-259, I expect a decent amount of LoC increase and I am guessing it would be better to keep them separate for organizational purposes. We can merge them in a later if my guess is wrong.

Finally, I've added some tests that test the methods that are supplied by `Collection` conformance. This is because it turned out that the automatically generated methods were initially causing a crash in Swift because our (automatically generated) range-based subscript was not correct. After fixing it, I found myself still pretty paranoid, so I've added the tests, which are simple but just serve to make sure that nothing is terribly wrong. They can be removed if desired, although some of them also test the new `count` logic and should stay for that purpose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,30 @@ print(doc["a"] ?? "") // prints `1`
 // Set a new value
 doc["d"] = 4
 print(doc) // prints `{"a" : 1, "b" : 2, "c" : 3, "d" : 4}`
+
+// Using functional methods like map, filter:
+let evensDoc = doc.filter { elem in
+    guard let value = elem.value as? Int else {
+        return false
+    }
+    return value % 2 == 0
+}
+print(evensDoc) // prints `{ "b" : 2, "d" : 4 }`
+
+let doubled = doc.map { elem -> Int in
+    guard let value = elem.value as? Int else {
+        return 0
+    }
+
+    return value * 2
+}
+print(doubled) // prints `[2, 4, 6, 8]`
 ```
+Note that `Document` conforms to `Collection`, so useful methods from
+[`Sequence`](https://developer.apple.com/documentation/swift/sequence) and
+[`Collection`](https://developer.apple.com/documentation/swift/collection) are
+all available.
+
 ## Development Instructions
 
 See our [development guide](https://mongodb.github.io/mongo-swift-driver/development.html) for instructions for building and testing the driver.
--- a/README.md
+++ b/README.md
@@ -164,7 +164,8 @@ print(doubled) // prints `[2, 4, 6, 8]`
 Note that `Document` conforms to `Collection`, so useful methods from
 [`Sequence`](https://developer.apple.com/documentation/swift/sequence) and
 [`Collection`](https://developer.apple.com/documentation/swift/collection) are
-all available.
+all available. However, runtime guarantees are not yet met for many of these
+methods.
 
 ## Development Instructions
 
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -17,7 +17,7 @@ extension Document: Collection {
     /// Returns the end index of the Document.
     public var endIndex: Int {
         precondition(self.count > 0)
-        return self.count
+        return self.countFast
     }
 
     private func validIndex(_ i: Int) -> Bool {
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -6,7 +6,6 @@ import Foundation
 extension Document: Collection {
     /// Returns the start index of the Document.
     public var startIndex: Int {
-        precondition(self.count > 0)
         return 0
     }
 
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -20,15 +20,10 @@ extension Document: Collection {
         return self.countFast
     }
 
-    // TODO: Can we use the internal helpers as defined in Collection.swift for this?
-    private func validIndex(_ i: Int) -> Bool {
-        return self.startIndex ... self.endIndex - 1 ~= i
-    }
-
     /// Returns the index after the given index for this Document.
     public func index(after i: Int) -> Int {
         // Index must be a valid one, meaning it must exist somewhere in self.keys.
-        precondition(validIndex(i))
+        _failEarlyRangeCheck(i, bounds: self.startIndex ... self.endIndex)
         return i + 1
     }
 
@@ -38,7 +33,7 @@ extension Document: Collection {
         // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
         // criticism also applies to key-based subscripting via `String`.
         // See SWIFT-250.
-        precondition(validIndex(position))
+        _failEarlyRangeCheck(position, bounds: self.startIndex ... self.endIndex)
         return self.makeIterator().keyValuePairs[position]
     }
 
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -18,13 +18,18 @@ extension Document: Collection {
         return self.count
     }
 
+    private func validIndex(_ i: Int) -> Bool {
+        return self.startIndex ... self.endIndex - 1 ~= i
+    }
+
     public func index(after i: Int) -> Int {
         // Index must be a valid one, meaning it must exist somewhere in self.keys.
-        precondition(self.startIndex ... self.endIndex - 1 ~= i)
+        precondition(validIndex(i))
         return i + 1
     }
 
     public subscript(position: Int) -> Document.KeyValuePair {
-        return (self.keys[position], self.values[position])
+        precondition(validIndex(position))
+        return self.makeIterator().keyValuePairs[position]
     }
 }
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -1,0 +1,30 @@
+//
+//  Document+Collection.swift
+//  MongoSwift
+//
+//  Created by may on 11/8/18.
+//
+
+import Foundation
+
+extension Document: Collection {
+    public subscript(position: Int) -> Document.KeyValuePair {
+        return (self.keys[position], self.values[position])
+    }
+
+    public var startIndex: Int {
+        precondition(self.count > 0)
+        return self.keys.startIndex
+    }
+
+    public var endIndex: Int {
+        precondition(self.count > 0)
+        return self.keys.endIndex
+    }
+
+    public func index(after i: Int) -> Int {
+        // Index must be a valid one, meaning it must exist somewhere in self.keys.
+        precondition(self.startIndex ... self.endIndex - 1 ~= i)
+        return self.keys.index(after: i)
+    }
+}
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -31,7 +31,7 @@ extension Document: Collection {
 
     /// Allows access to a `KeyValuePair` from the `Document`, given the position of the desired `KeyValuePair` held
     /// within. This method does not guarantee constant-time (O(1)) access.
-    public subscript(position: Int) -> Document.KeyValuePair {
+    public subscript(position: Int) -> KeyValuePair {
         // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
         // criticism also applies to key-based subscripting via `String`.
         // See SWIFT-250.
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -17,7 +17,7 @@ extension Document: Collection {
     /// Returns the end index of the Document.
     public var endIndex: Int {
         precondition(self.count > 0)
-        return self.countFast
+        return self.count
     }
 
     /// Returns the index after the given index for this Document.
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -1,12 +1,8 @@
-//
-//  Document+Collection.swift
-//  MongoSwift
-//
-//  Created by may on 11/8/18.
-//
-
 import Foundation
 
+/// An extension of `Document` to make it conform to the `Collection` protocol.
+/// This gives guarantees on non-destructive iteration, and offers an indexed
+/// ordering to the key value pairs in the document.
 extension Document: Collection {
     /// Returns the start index of the Document.
     public var startIndex: Int {
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -11,7 +11,6 @@ extension Document: Collection {
 
     /// Returns the end index of the Document.
     public var endIndex: Int {
-        precondition(self.count > 0)
         return self.count
     }
 
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -36,14 +36,14 @@ extension Document: Collection {
         // criticism also applies to key-based subscripting via `String`.
         // See SWIFT-250.
         failIndexCheck(position)
-        return self.keyValuePairs[position]
+        // Because of our failIndexCheck precondition, this call is guaranteed to provide a non-nil result.
+        return DocumentIterator.subsequence(of: self, startIndex: position, endIndex: position + 1).first!
     }
 
     /// Allows access to a `KeyValuePair` from the `Document`, given a range of indices of the desired `KeyValuePair`'s
     /// held within. This method does not guarantee constant-time (O(1)) access.
     public subscript(bounds: Range<Int>) -> Document {
-        let keyValues = self.keyValuePairs
         // TODO: SWIFT-252 should provide a more efficient implementation for this.
-        return Document(Array(keyValues[bounds]))
+        return DocumentIterator.subsequence(of: self, startIndex: bounds.lowerBound, endIndex: bounds.upperBound)
     }
 }
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -15,7 +15,11 @@ extension Document: Collection {
     }
 
     private func failIndexCheck(_ i: Int) {
-        precondition(self.startIndex ... self.endIndex - 1 ~= i, "Index \(i) is invalid")
+        let invalidIndexMsg = "Index \(i) is invalid"
+        if self.count == 0 {
+            precondition(i == 0, invalidIndexMsg)
+        }
+        precondition(self.startIndex ... self.endIndex - 1 ~= i, invalidIndexMsg)
     }
 
     /// Returns the index after the given index for this Document.
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -28,7 +28,12 @@ extension Document: Collection {
         return i + 1
     }
 
+    /// Allows access to a `KeyValuePair` from the `Document`, given the position of the desired `KeyValuePair` held
+    /// within. This method does not guarantee constant-time (O(1)) access.
     public subscript(position: Int) -> Document.KeyValuePair {
+        // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
+        // criticism also applies to key-based subscripting via `String`.
+        // See SWIFT-250.
         precondition(validIndex(position))
         return self.makeIterator().keyValuePairs[position]
     }
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -8,11 +8,13 @@
 import Foundation
 
 extension Document: Collection {
+    /// Returns the start index of the Document.
     public var startIndex: Int {
         precondition(self.count > 0)
         return 0
     }
 
+    /// Returns the end index of the Document.
     public var endIndex: Int {
         precondition(self.count > 0)
         return self.count
@@ -22,6 +24,7 @@ extension Document: Collection {
         return self.startIndex ... self.endIndex - 1 ~= i
     }
 
+    /// Returns the index after the given index for this Document.
     public func index(after i: Int) -> Int {
         // Index must be a valid one, meaning it must exist somewhere in self.keys.
         precondition(validIndex(i))
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -32,7 +32,7 @@ extension Document: Collection {
         // criticism also applies to key-based subscripting via `String`.
         // See SWIFT-250.
         failIndexCheck(position)
-        return self.makeIterator().keyValuePairs[position]
+        return self.keyValuePairs[position]
     }
 
     /// Allows access to a `KeyValuePair` from the `Document`, given a range of indices of the desired `KeyValuePair`'s
--- a/Sources/MongoSwift/BSON/Document+Collection.swift
+++ b/Sources/MongoSwift/BSON/Document+Collection.swift
@@ -21,7 +21,7 @@ extension Document: Collection {
     public func index(after i: Int) -> Int {
         // Index must be a valid one, meaning it must exist somewhere in self.keys.
         precondition(self.startIndex ... self.endIndex - 1 ~= i)
-        return self.keys.index(after: i)
+        return i + 1
     }
 
     public subscript(position: Int) -> Document.KeyValuePair {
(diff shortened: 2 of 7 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -79,6 +89,9 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         }
 
         self.storage = DocumentStorage()
+        // This is technically not consistent, but the only way this inconsistency can cause an issue is if we fail to
+        // setValue(), in which case we crash anyways.
```
jmikola: Can you elaborate on why this is technically not consistent? Is it because we're initializing it to 0?

I would think that's fine provided that `self.setValue()` increments the count as new keys are assigned. Otherwise, we're correct to initialize to zero if `keyValuePairs` is empty and this `for` loop has nothing to iterate.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -295,11 +335,13 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
     }
 
     /// Appends the key/value pairs from the provided `doc` to this `Document`.
+    /// Note: This function does not check for or clean away duplicate keys.
```
jmikola: Is `merge` our own method or part of some protocol? It looks like we only use it in the `Collection.createIndexes()` helper.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -583,4 +585,116 @@ final class DocumentTests: MongoSwiftTestCase {
...
+            expect(upToPrefixDoc).to(equal(["a": 3, "b": 2]))
+            expect(doc).to(haveCorrectCount())
+
+            // doc.prefix(through:)
+            expect(throughPrefixDoc).to(equal(["a": 3, "b": 2]))
+            expect(doc).to(haveCorrectCount())
+
+            // doc.suffix
+            expect(suffixDoc).to(equal(["b": 2, "c": 5, "d": 4, "e": 3]))
+            expect(doc).to(haveCorrectCount())
```
jmikola: It looks like `haveCorrectCount()` simply re-computes a count with `bson_count_keys()`. Would it make more sense to assert our own literal here? In this case, I would expect `doc.count` to equal `5` since there are five unique keys in the document.

Side note: I'm not sure why the `haveCorrectCount()` checks are repeated in this test, as the three sequences are created immediately after `doc` is assigned. The only thing between these `haveCorrectCount()` assertions are equality checks on those sequences, which I wouldn't expect to change the state at all.

Sources/MongoSwift/BSON/Document.swift
Utagai: That's it. It is not consistent at the time of setting it to 0, because we may actually have more key value pairs. `setValue()` fixes this once it is called enough times.

This is okay because if `setValue()` fails, we crash so it doesn't matter.

Sources/MongoSwift/BSON/Document.swift
Utagai: `merge` seems to be our own function. This note was included because it was something me and Kaitlin realized was not made clear anywhere but may be assumed to be true when its not.

Tests/MongoSwiftTests/DocumentTests.swift
Utagai: That approach would work too, I just chose this approach because it lets the programmer not have to worry about actually keeping track of the document state throughout the test case. It's just "easier" to let it do some extra work and compute it for you each time.

The reason these checks exist is because we are maintaining a stateful `count` property on the Document. This `count` property could actually be incorrect, but the `Document` would still pass the equality check because we don't check to see that two documents have the same value for `count` when comparing them for equality.

In other words, since a simple `Document` to `Document` check does not check for `count` equality, and we can't necessarily assume that our logic for incrementing/decrementing the count is correct, we need to explicitly check that the `count` is indeed the correct amount by itself.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -247,6 +245,14 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    /// Returns an array of KeyValue pairs from the iterator's current position to the end. The iter will be exhausted
+    /// after this property is accessed.
```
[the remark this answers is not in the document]
Utagai: My understanding of this is: 
No, because non-destructive iteration just means that iterating it once does not change the Sequence/Collection itself or its contents, nor does it change the results of future iterations.

The only thing having an exhaustible iterator violates is infinite sequence, which a `Document` is not.

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+
+/// An extension of `Document` to make it conform to the `Collection` protocol.
+/// This gives guarantees on non-destructive iteration, and offers an indexed
+/// ordering to the key value pairs in the document.
+extension Document: Collection {
+    /// Returns the start index of the Document.
+    public var startIndex: Int {
+        precondition(self.count > 0)
```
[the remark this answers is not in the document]
Utagai: Hm, this is probably a small issue in their documentation. I actually distinctly recall reading "The position of the first element in a nonempty collection",  and going "okay, that sounds like a precnodition!". 😆 

It seems like this should actually just return `endIndex` instead... hard to really tell since its contradictory.

Since I don't like `precondition` or `preconditionFailure` and prefer to avoid them whenever possible, I'm going to switch this over to returning `endIndex`. The cool thing here is that due to how I've chosen indices, we can just have `startIndex` always return 0.

Tests/MongoSwiftTests/DocumentTests.swift
jmikola: Would the internal counts actually change between these assertions, though? If I follow your logic, it's the initial calls to `prefix()` and `suffix()` that might alter the count. I don't understand why the count would possibly change between assertions of subsequences that you've already computed at the top of this test method, though.

Should these assertions actually be calling `haveCorrectCount()` on the subsequences themselves instead of repeating the assertion on `doc`?

Utagai: Ah! I see! I actually did not notice this, I see the issue. It _should_ be calling this on `suffixDoc` and friends, not `doc`. This was a mistake. Thanks!

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+/// An extension of `Document` to make it conform to the `Collection` protocol.
+/// This gives guarantees on non-destructive iteration, and offers an indexed
+/// ordering to the key value pairs in the document.
```
kmahar:
```suggestion
/// ordering to the key-value pairs in the document.
```

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,42 @@
...
+extension Document: Collection {
+    /// Returns the start index of the Document.
+    public var startIndex: Int {
+        return 0
+    }
+
+    /// Returns the end index of the Document.
+    public var endIndex: Int {
+        precondition(self.count > 0)
```
kmahar: this seems kind of weird - are you just not supposed to call `endIndex` until you've put something in the doc?

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,42 @@
...
+        return self.count
+    }
+
+    /// Returns the index after the given index for this Document.
+    public func index(after i: Int) -> Int {
+        // Index must be a valid one, meaning it must exist somewhere in self.keys.
+        _failEarlyRangeCheck(i, bounds: self.startIndex ... self.endIndex)
```
kmahar: where does this method come from?

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -247,6 +245,14 @@ public class DocumentIterator: IteratorProtocol {
         return values
     }
 
+    /// Returns an array of KeyValue pairs from the iterator's current position to the end. The iter will be exhausted
```
kmahar:
```suggestion
    /// Returns an array of `KeyValuePair`s from the iterator's current position to the end. The iter will be exhausted
```

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -60,6 +69,7 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
      */
     internal init(fromPointer pointer: UnsafePointer<bson_t>) {
         self.storage = DocumentStorage(fromPointer: pointer)
+        self.count = Int(bson_count_keys(self.storage.pointer))
```
kmahar: I count (ha ha) 5 places now where we do `Int(bson_count_keys(some pointer))`. would it make sense to add an internal computed property on `Document` for this? idk what to name, `countFromData` or something maybe

Sources/MongoSwift/BSON/Document+Collection.swift
Utagai: This is a method from `Collections.swift`. I originally had my own implementation until I realized that most likely Swift should and probably is doing the same exact thing I am in their default implementations. After some digging, I found that they do via uses of this function.

In retrospect, perhaps we shouldn't rely on this function, the leading underscore makes me a bit hesitant. I think at the time I was just happy to remove my own implementation for it, but it may be safer to rely on an implementation we have control over as opposed to this method which may or may not change.

Sources/MongoSwift/BSON/Document+Collection.swift
Utagai: Done.

Sources/MongoSwift/BSON/Document+Collection.swift
jmikola: `_failEarlyRangeCheck` doesn't appear in any published documentation -- most search results are pointing to implementations (or code snippets) from standard library classes. I think this is worth recreating ourself as a private method.

Sources/MongoSwift/BSON/Document.swift
jmikola: Is the idea that the computed property would always recompute? Document already has a `count` property, so would it make more sense for the always-computed property to go on `DocumentStorage` instead?

Utagai: A computed property is recomputed whenever it is accessed, I believe.

I don't think it would make too big of a difference either way, since we'll be making it `private` or `internal` with our two choices.

Sources/MongoSwift/BSON/Document+Collection.swift
Utagai: Yup, as I said, I think in retrospect its safer to rely on our own implementation.

Done!

Sources/MongoSwift/BSON/Document.swift
kmahar: yes, basically this would just be a wrapper for calling `bson_count_keys`. but it's more swifty to make it a computed property instead of a function. 

it could go either place, but thus far we have kept all methods that actually do things with the `bson_t` inside of `Document`, and left `DocumentStorage` as a class that does nothing but handle initialization and cleanup. I think that simplicity is kind of nice.

if we can come up with some sort of established rule for when something is a `DocumentStorage` method vs a `Document` method, I'm not opposed to moving some of our internal methods over there.

Utagai: Ah, I just remembered _why_ I didn't make this a computed property to begin with...

All the uses of `Int(bson_count_keys(self.storage.pointer))` is for setting `self.count`..., but if it is made into a computed property then we'd have to do something like `self.count = self.bsonCount` but this won't work because we are in an initializer before we have initialized all of the variables.

I think going to `DocumentStorage` would make this work, or we can make it a private function that takes a pointer.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -5,6 +5,10 @@ import Foundation
 public class DocumentStorage {
     internal var pointer: UnsafeMutablePointer<bson_t>!
 
+    internal var count: Int {
```
kmahar: nit: let's add a comment so we remember why we put this here.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -583,4 +585,116 @@ final class DocumentTests: MongoSwiftTestCase {
...
+    }
+
+    func testDocumentConformanceToCollection() throws {
+        CollectionConformance.testDocumentIndexLogic()
+        try CollectionConformance.testMutators()
+        CollectionConformance.testPrefixSuffix()
+    }
+}
+
+/// A Nimble matcher for testing that the count of a Document is what it should be.
```
kmahar: let's clarify that this is specifically checking that the count determined by libbson is equal to our own bookkeeping number.

Utagai:
@kmahar Most of the newly added functionality has to do with the associated concept of an "index" when dealing with Swift `Collection`'s. So the new things to show would probably just be things like subscript ranging, first/last element fetching and maybe index-based suffix/prefix stuff.

All of this stuff seems mostly kind of not-too-crazy for users of a `Document` since most of them will not really see it as an ordered data structure.

If all this is totally okay, then I'm down for putting it into the README. We could toss in example usage of things like first/last element and index-based subscripting (with range) to show how one can look at `Document` in an ordered way.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -583,4 +585,116 @@ final class DocumentTests: MongoSwiftTestCase {
 
         expect(noops).to(equal(["null": 5, "maxkey": "hi", "minkey": false]))
     }
+
+    private struct CollectionConformance {
```
kmahar: we already have a Document+SequenceTests.swift. How about we just create a new test class in a new file Document+CollectionTests.swift?

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -1,3 +1,4 @@
+import bson
```
kmahar: import no longer needed.

Utagai: Done.

kmahar:
General question: back when I implemented the `Sequence` protocol, I added a `subsequence` method to `DocumentIterator` that allowed us to get more performant implementations than the defaults for things like `dropFirst` and `dropLast`. I think the asymptotic runtimes remained similar, but it allowed us to avoid stuff like converting the first X values from C types to Swift types just to immediately ignore them.  are there any `Collection` methods where we could/should be overriding the default impls in this way?

in particular, wondering if `popFirst`, `removeFirst`, `prefix(through:)`, `prefix(upTo:)`, etc could be made better. or maybe they already call through to the Sequence protocol methods.

Sources/MongoSwift/BSON/Document.swift

kmahar:
back to the point about the README, ya, that's true. maybe we can just include examples of a couple more useful methods like `map` and `filter`, and then state that `Document` conforms to both `Collection` and `Sequence` and link to the Apple docs on those.

Tests/MongoSwiftTests/DocumentTests.swift
Utagai: Done.

Tests/MongoSwiftTests/DocumentTests.swift
Utagai: Done.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -584,3 +584,4 @@ final class DocumentTests: MongoSwiftTestCase {
         expect(noops).to(equal(["null": 5, "maxkey": "hi", "minkey": false]))
     }
 }
+
```
kmahar: linter? or can we remove line?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -43,9 +53,15 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
         return self.makeIterator().values
     }
 
+    /// Returns a `[KeyValuePair]` containing the key-value pairs stored in this `Document`.
+    public var keyValuePairs: [KeyValuePair] {
```
kmahar: is this part of the protocol, or just something we are adding for the sake of it? I think Array(doc) may return the same thing

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /// Allows access to a `KeyValuePair` from the `Document`, given the position of the desired `KeyValuePair` held
+    /// within. This method does not guarantee constant-time (O(1)) access.
+    public subscript(position: Int) -> Document.KeyValuePair {
+        // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
+        // criticism also applies to key-based subscripting via `String`.
+        // See SWIFT-250.
+        failIndexCheck(position)
+        return self.makeIterator().keyValuePairs[position]
```
kmahar: should we just be using `self.keyValuePairs` here?

Tests/MongoSwiftTests/DocumentTests.swift
kmahar: ah, seems like linting failed on travis due to this and a few other things.

Sources/MongoSwift/BSON/Document.swift
Utagai: I added this because getting a `[KeyValuePair]` was a useful thing for implementing the actual protocol related methods. I transferred it over here because `keys` and `values` already existed, so it seemed only natural to also have it rest alongside the two.

Utagai: re: our offline discussion on `Array(doc)`: After getting help from the xcode debugger I was able to confirm that the initializer for `Array` does indeed call the index-based subscript in order to populate its buffer. What this means is that we will have to rely on the `DocumentIterator` to get the `[KeyValuePair]`, otherwise we'll have a situation where doing `Array(doc)` leads us to call `subscript(index)` which calls `Array(doc)` and so on and so forth.

This should honestly be fine, since if we are going to eventually conform to `Collection`'s runtime guarantees, then we won't even need `keyValuePairs` anymore.

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,49 @@
...
+        // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
+        // criticism also applies to key-based subscripting via `String`.
+        // See SWIFT-250.
+        failIndexCheck(position)
+        return self.keyValuePairs[position]
+    }
+
+    /// Allows access to a `KeyValuePair` from the `Document`, given a range of indices of the desired `KeyValuePair`'s
+    /// held within. This method does not guarantee constant-time (O(1)) access.
+    public subscript(bounds: Range<Int>) -> Document {
```
kmahar: similarly to my comment above, can we use `DocumentIterator.subsequence` here with the provided bounds?

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /// Allows access to a `KeyValuePair` from the `Document`, given the position of the desired `KeyValuePair` held
+    /// within. This method does not guarantee constant-time (O(1)) access.
+    public subscript(position: Int) -> Document.KeyValuePair {
+        // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
+        // criticism also applies to key-based subscripting via `String`.
+        // See SWIFT-250.
+        failIndexCheck(position)
+        return self.keyValuePairs[position]
```
kmahar: could we use `DocumentIterator.subsequence` here instead? or maybe write a variant of that method that uses the same approach to just gets one `KeyValuePair` value given its index?  ideally, we should just advance the iterator to the desired position and return what's there. currently, this requires us to always traverse the whole document and copy every key-value pair over into Swift, but then we just end up only looking at one of them.

Utagai: Yes, I've opted for:
```swift
// Because of our failIndexCheck precondition, this call is guaranteed to provide a non-nil result.
return DocumentIterator.subsequence(of: self, startIndex: position, endIndex: position + 1).first!
```

However, this looks kind of gnarly, the really simple approach one thinks of first would have been:
```swift
let iterator = makeIterator()
if position == 0 { return iterator.next()! }
for _ in 0..<position {
    iterator.advance()
}
// Because of our failIndexCheck precondition, this call is guaranteed to provide a non-nil result.
return iterator.next()!
```
Which is not much prettier. I've opted for the `subsequence` approach because it seems to be a lot easier to understand as being basically a way of getting a single-element subsequence.

Thoughts?

Sources/MongoSwift/BSON/Document+Collection.swift
Utagai: Done.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -12,9 +12,7 @@ import mongoc
 extension Document: Sequence {
     /// The element type of a document: a tuple containing an individual key-value pair.
     public typealias KeyValuePair = (key: String, value: BSONValue?)
-
-    /// Returns the number of (key, value) pairs stored at the top level of this `Document`.
-    public var count: Int { return Int(bson_count_keys(self.data)) }
+    public typealias SubSequence = Document
```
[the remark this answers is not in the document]
Utagai: Done.

Sources/MongoSwift/BSON/Document.swift
Utagai: Yup, we don't need it if we opt for `DocumentIterator.subsequence()`, so I've removed them.

Tests/MongoSwiftTests/Document+CollectionTests.swift
```diff
@@ -0,0 +1,122 @@
+import Foundation
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+final class Document_CollectionTests: MongoSwiftTestCase {
```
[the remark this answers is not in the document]
Utagai: This is a little tricky. Swift `Collection` (including `Dictionary`) will throw fatal error for doing things like `removeFirst()` on an empty collection (and our `Document` does this as well already). This means its pretty much impossible to test for this since fatal errors cannot be caught in tests without considerable hacky work.

I've at least added tests for the things I can check (`startIndex` and `endIndex`), but the other methods like `subscript` and `index(after:)` can't really be tested.

I did at least try them by hand to see if we are consistent, and they both throw a fatal error just like we do. In fact, we do a better job because we actually include a message for why its failing 😄!

Also nit: removed variable name stutter by renaming `testDocumentIndexLogic` -> `testIndexLogic`,

Utagai:
@kmahar Re: your question about efficiency of new `Collection` methods. The implementations for those `Collection` methods actually rely on index-subscripting and index-range-subscripting, or, they use `Sequence`. That is actually where they make guarantees about their runtimes from. The significance of this fact is that regardless of what we do now to shave some time for things like C-to-Swift copy will be made redundant by BSON performance improvements because that includes efficient subscripting for indices (including range).

Currently however, with the `subsequence` implementations of subscript, those functions will be using `subsequence` under the hood anyways.

Sources/MongoSwift/BSON/Document+Sequence.swift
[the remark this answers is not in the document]
kmahar: I think this is more an implementation detail - which is important to note but I'm not sure is super helpful to users. can we make this part a `//` comment and add a `///` line below detailing what's relevant for users to know?
 "the type that will be returned from methods that return parts of documents, like `map` and `filter`", or something like that.

Tests/MongoSwiftTests/Document+CollectionTests.swift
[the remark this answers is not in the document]
kmahar: perfect, that was all I wanted to know.

Sources/MongoSwift/BSON/Document+Collection.swift
kmahar: let's just stick with subsequence, I think it is pretty readable and we may as well not introduce more logic if this is going to change before long anyway.

README.md
```diff
@@ -142,7 +142,30 @@ print(doc["a"] ?? "") // prints `1`
...
+let doubled = doc.map { elem -> Int in
+    guard let value = elem.value as? Int else {
+        return 0
+    }
+
+    return value * 2
+}
+print(doubled) // prints `[2, 4, 6, 8]`
 ```
+Note that `Document` conforms to `Collection`, so useful methods from
```
jmikola: Would this be a good place to mention lack of performance guarantees?

Sources/MongoSwift/BSON/Document+Sequence.swift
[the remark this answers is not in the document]
Utagai: So not exactly sure which parts are better for users and not, but I've opted for this:
```swift
// Since a `Document` is a recursive structure, we want to enforce the use of it as a subsequence of itself.
// instead of something like `Slice<Document>`.
/// The type that is returned from methods such as `dropFirst()` and `split()`.
```

PR #147, opened 2018-11-26
SWIFT-258 Update README to clarify Cocoapods does not work for OS X installation

kmahar:


README.md
```diff
@@ -72,23 +74,21 @@ let package = Package(
 
 Then run `swift build` to download, compile, and link all your dependencies.
 
-### OR: Install the Driver Using CocoaPods
-*Please make sure you have followed the instructions in the previous section on installing the MongoDB C Driver before proceeding.*
-
-CocoaPods is a dependency manager for Swift and Objective-C. You can install it by running `gem install cocoapods`. See [the CocoaPods documentation](https://cocoapods.org/) for more information.
+## iOS, tvOS, and watchOS
+Installation is supported via [CocoaPods](https://cocoapods.org/). The pod includes as a dependency an embedded version of the MongoDB C Driver, meant for use on these OSes.
```
jmikola: Should we mention that this build currently does not support SSL (https://github.com/mongodb/mongo-swift-driver/issues/141)?

README.md
```diff
@@ -32,27 +34,27 @@ Core Server (i.e. SERVER) project are **public**.
...
+Installation on OS X and Linux is supported via [Swift Package Manager](https://swift.org/package-manager/).
+
+#### Step 1: Install the MongoDB C Driver
+The driver wraps the MongoDB C driver, and using it requires having the C driver's two components, `libbson` and `libmongoc`, installed on your system. The minimum required version of the C Driver is **1.13.0**.
 
 On a Mac, you can install both components at once using [Homebrew](https://brew.sh/):
 `brew install mongo-c-driver`.
 
-Or on Linux, use `apt-get` to install `libmongoc` (which includes `libbson` as a dependency) and `pkg-config` (which enables Swift Package Manager to find the components):
+On Linux, use `apt-get` to install `libmongoc` (which includes `libbson` as a dependency) and `pkg-config` (which enables Swift Package Manager to find the components):
```
jmikola: This isn't new to this PR, but I don't think Debian or Ubuntu (assuming `apt-get`) are going to provide the necessary version of libmongoc. It may be easier to just point Linux users to http://mongoc.org/libmongoc/current/installing.html and skip the `apt-get install libmongoc-1.0.0` suggestion entirely. 

When we ultimately get libmongoc builds in MongoDB's own [package repositories](https://repo.mongodb.org), I expect that page will also be updated with instructions accordingly.

README.md
Utagai: Yup, I think this is an important thing to mention in the README.

README.md
```diff
@@ -32,27 +34,27 @@ Core Server (i.e. SERVER) project are **public**.
...
 ```
 sudo apt-get install pkg-config
 sudo apt-get install libmongoc-1.0.0
 ```
 
-Alternatively, see the [installation guide](http://mongoc.org/libmongoc/current/installing.html) from libmongoc's documentation.
+See the [installation guide](http://mongoc.org/libmongoc/current/installing.html) from libmongoc's documentation for more information.
```
Utagai: Should this be `` `libmongoc` `` like in its other mentions? ([example](https://github.com/mongodb/mongo-swift-driver/pull/147/files#diff-04c6e90faac2675aa89e2176d2eec7d8R47))

README.md
kmahar: done.

README.md
kmahar: done.

README.md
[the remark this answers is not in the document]
kmahar: given @jmikola's suggestion above, I'm going to scrap this section and just point users to the libmongoc installation guide, as that will always give the most updated information on this.

kmahar: good point - this is something we have encountered a few times already in issues on the repo. however, the instructions there also point users to `apt-get`. I've updated this say that `apt-get`'s versions are currently insufficient and point users to the instructions for build/install from source instead.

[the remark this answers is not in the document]
Utagai: Sounds like a plan!

jmikola: Using looser language like "If `apt-get`'s versions are too old, follow manual build instructions" may be better. Then this text can remain accurate even if `apt-get` is updated down the line.

jmikola: Ah, just saw https://github.com/mongodb/mongo-swift-driver/pull/147#discussion_r. Simply referring to libmongoc's instructions seems best.

kmahar: ok, I've made this vaguer.

README.md
```diff
@@ -32,26 +34,20 @@ Core Server (i.e. SERVER) project are **public**.
...
+#### Step 1: Install the MongoDB C Driver
+The driver wraps the MongoDB C driver, and using it requires having the C driver's two components, `libbson` and `libmongoc`, installed on your system. **The minimum required version of the C Driver is 1.13.0**.
 
-Next, see instructions for installation with either Swift Package Manager or CocoaPods in the following sections.
+On a Mac, you can install both components at once using [Homebrew](https://brew.sh/):
+`brew install mongo-c-driver`.
 
-### NEXT: Install the Driver Using Swift Package Manager
-*Please make sure you have followed the instructions in the previous section on installing the MongoDB C Driver before proceeding.*
+On Linux: as of now, the `libmongoc` version provided by `apt-get` is < 1.13.0 and is insufficient for use with `MongoSwift`. Please follow the [instructions](http://mongoc.org/libmongoc/current/installing.html#building-on-unix) for building and installing from source from `libmongoc`'s documentation.
```
jmikola:
```suggestion
On Linux, you may be able to install both components using your package manager (e.g. `apt-get install libmongoc-1.0-0`). If the versions provided by your package manager are older than 1.13.0, please follow the [instructions](http://mongoc.org/libmongoc/current/installing.html#building-on-unix) for building and installing from source from `libmongoc`'s documentation.
```

kmahar: oops, I made my own edit and did not realize you were suggesting. now I cannot accept since the old line isn't there.
I can copy paste this in or leave it as what I wrote which is pretty similar but a bit vaguer. I don't care so let me know what you prefer.

README.md
```diff
@@ -32,26 +34,20 @@ Core Server (i.e. SERVER) project are **public**.
...
+#### Step 1: Install the MongoDB C Driver
+The driver wraps the MongoDB C driver, and using it requires having the C driver's two components, `libbson` and `libmongoc`, installed on your system. **The minimum required version of the C Driver is 1.13.0**.
 
-Next, see instructions for installation with either Swift Package Manager or CocoaPods in the following sections.
+On a Mac, you can install both components at once using [Homebrew](https://brew.sh/):
+`brew install mongo-c-driver`.
 
-### NEXT: Install the Driver Using Swift Package Manager
-*Please make sure you have followed the instructions in the previous section on installing the MongoDB C Driver before proceeding.*
+On Linux: please follow the [instructions](http://mongoc.org/libmongoc/current/installing.html#building-on-unix) from `libmongoc`'s documentation. Note that the versions provided by `apt-get` may be too old, in which case you can follow the instructions for building and installing from source.
```
jmikola: I would just suggest replacing `apt-get` with "your package manager" to keep this distribution agnostic. I dunno if we make other assumptions about Linux OS elsewhere in this README. LGTM otherwise.

kmahar: done. I think this section is the only part where linux is discussed.

PR #148, opened 2018-11-28
SWIFT-103: Move benchmarks to separate project

kmahar:
There will be a corresponding PR up shortly on our new benchmarks repo.


Development.md
```diff
@@ -42,19 +42,15 @@ Why is this necessary? The project requires a customized "copy resources" build
 ## Running Tests
 **NOTE**: Several of the tests require a mongod instance to be running on the default host/port, `localhost:27017`.
 
-Additionally, please note that each benchmark test runs for a minimum of 1 minute and therefore **the entire benchmark suite will take around 20-30 minutes to complete**.
-
 You can run tests from Xcode as usual. If you prefer to test from the command line, keep reading.
 
 ### From the Command Line 
-Tests can be run from the command line with `make test`. By default, this will run all the tests excluding the benchmarks.
+Tests can be run from the command line with `make test`. By default, this will run all the tests.
```
Utagai: I may not be understanding the consequences of this PR, but can't we just remove this now and tell users to run `swift test`, since that will no longer run the benchmarks? This has the added benefit of making things more uniform and tied with just Swift as opposed to involving Make targets.

Makefile
```diff
@@ -1,30 +1,30 @@
...
+	FILTERARG =
 endif
 
+define check_for_gem
+	gem list $(1) -i > /dev/null || gem install $(1) || { echo "ERROR: Failed to locate or install the ruby gem $(1); please install yourself with 'gem install $(1)' (you may need to use sudo)"; exit 1; }
+endef
+
 all:
-	swift build -v $(CFLAGS) $(LDFLAGS)
+	swift build -v
```
Utagai: So ` $(CFLAGS) $(LDFLAGS)` is removed here, but looking at the diff, it seems like they never did anything in the first place?:

```make
# CFLAGS = -Xcc -DMONGOC_COMPILATION -Xcc -DBSON_COMPILATION -Xcc -ISources/libbson -Xcc -ISources/libbson/include -Xcc -ISources/libbson/generated -Xcc -ISources/libmongoc/generated	# if provided, FILTER is used as the --filter argument to `swift test`. 
# LDFLAGS = -Xlinker -lsasl2 -Xlinker -lz	
CFLAGS =	
LDFLAGS =
```
Seems like they were both empty, and the substantial assignments above were commented out.

So was this just some clean-up you happened to do and not actually related to the PR?

It's also worth considering that an empty variable is not useless in Makefiles and gives degree of control to the user. Is it possible then that we had empty `CFLAGS` and `LDFLAGS` and users could run something like `make CFLAGS="..."` or similar? They still can do this,  even without the variables in the Makefile, but I _think_ it is better practice to keep such variables as empty in the Makefile for the sake of documentation.

.swiftlint.yml
```diff
@@ -21,3 +21,5 @@ excluded:
   - build
   - .build
   - docs
+  - Examples/*/build
```
Utagai: Nice. I think I ran into this issue before and forgot to bring it up!!! 🎉

Makefile
```diff
@@ -1,30 +1,30 @@
...
 	ruby add_json_files.rb
 
 test:
-	swift test -v $(CFLAGS) $(LDFLAGS) $(FILTERARG)
+	swift test -v $(FILTERARG)
 
-benchmark:
-	swift test -v $(CFLAGS) $(LDFLAGS) --filter MongoSwiftBenchmarks
+test-pretty:
+	@$(call check_for_gem, xcpretty)
```
Utagai: Forgot to include this in my original review, sorry for the extra email spam:

Should we add something to the `README.md` or `Development.md` that notifies users that they can get prettier testing on CLI if they install `xcpretty`, but that it is not mandatory?

Makefile
Utagai: Sounds good to me.

It's odd that it gets replaced because this is not the behavior I've seen in other Makefiles (where an empty target doesn't overwrite the user-supplied version). I was able to get a minimal Makefile working with an empty variable assignment that still took on the user-supplied value on my Macbook just now... Perhaps there is something more subtle going on here... 🤔 

But in either case, sounds good to me. 👍

Development.md
Utagai: Got it!

Makefile
kmahar: lol well you know what they say about assuming... just checked, you were right, I find this behavior very unintuitive though 🤔 I guess these are more of default values, whereas I was reading them as variable assignments. 
in any case, ya, I don't think we need them anymore, if we change our minds we can come back and remember what this looked like

PR #149, opened 2018-11-29
SWIFT-243 Test Swift 4.2 on Travis 

kmahar:
https://jira.mongodb.org/browse/SWIFT-243 

Adds jobs for MacOS and Linux using Swift 4.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,27 @@
-os:
-  - linux
-  - osx
 language: generic
 sudo: required
 dist: trusty
-osx_image: xcode9.4
 
-env:
-  matrix:
-    - MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.0
-    - MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.1
-    - MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.2
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode9.4 
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.0
+    - os: osx
+      osx_image: xcode9.4
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.1
+    - os: osx
+      osx_image: xcode10.1
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.1
+    - os: osx
+      osx_image: xcode10.1
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.2
+    - os: linux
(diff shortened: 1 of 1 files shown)


kmahar:
seems to be some kind of issue with 4.2 on MacOS, will investigate more tomorrow.

Utagai:
@kmahar Just chiming in here because I get email notifications for this: 

You'll love this, but I believe this is to do with XCode version :upside_down_face: :). Travis is running 9.4 we want 10 for Swift 4.2, because 9.4 doesn't support 4.2.

kmahar:
alright, so I've updated this to now test the following combinations:

**OS X:**
- XCode 9, Swift 4.0
- XCode 9, Swift 4.1
- Xcode 10, Swift 4.1
- Xcode 10, Swift 4.2 

**Linux**:
- Swift 4.0
- Swift 4.1
- Swift 4.2

I tried doing XCode 10 + Swift 4.0 for the sake of completeness, but there is some toolchain related issue I haven't quite figured out. I have the same issue building the driver with 4.0 on the command line locally, so if I can come up with a workaround we can introduce it on Travis as well.

regardless, I think this is an improvement over the current state of things.

Utagai:
Perhaps we'll drop 4.0 before having to deal with this 🙃 

Anyways, I can confirm that there is a problem with XCode 10 + Swift 4.0.

Is it possible to install both XCode 10 and 9.4, and then use xcode-select to use one or the other depending on Swift 4.0 vs. 4.1/4.2?

kmahar:
as in, can we do that locally? maybe....

Utagai:
Yup, locally. I'm wondering if its possible for us to run two different XCode versions and switch between the two. I'm too afraid to try it and somehow break my Swift development environment again. 🙃

.travis.yml
```diff
@@ -1,16 +1,27 @@
...
+    - os: osx
+      osx_image: xcode9.4
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.1
+    - os: osx
+      osx_image: xcode10.1
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.1
+    - os: osx
+      osx_image: xcode10.1
+      env: MONGODB_VERSION=3.6.5 SWIFT_VERSION=4.2
+    - os: linux
```
jmikola: Would it make sense to move `dist: trusty` down to these linux entries? I assume `os: osx` simply ignores `dist:` (akin to how a linux entry might ignore `osx_image`), so maybe this doesn't matter.

kmahar: yes, it's just ignored. we'd have to repeat the same line three times, so I think we should just leave it at the top

PR #152, opened 2018-11-29
SWIFT-235 Fix UnsafeMutableRawPointer! deprecation warnings in Swift 4.1+

kmahar:
Just a one line change. Compiles either way, but the compiler starts warning you about this as of 4.1.


kmahar:
yeah, I'm not sure why it was that type before... I'm guessing the 4.0 compiler told me it was that type....

PR #154, opened 2018-11-30
SWIFT-191: Add a way to provide a default value when looking up a key in a Document

Utagai:
[SWIFT-191](https://jira.mongodb.org/browse/SWIFT-191)

**NOTE** This PR can only be merged after [the PR for SWIFT-193](https://github.com/mongodb/mongo-swift-driver/pull/145). The excessive amounts of commits are mostly from that branch, so after it is merged I'll rebase and almost all of them should go away. As a result of this, this PR's diff is basically contaminated by the diff from SWIFT-193. It's probably best to wait for SWIFT-193 to be merged before even trying to review this PR, but if you'd like, the two files of interest are `Document.swift` and `DocumentTests.swift`.

This PR adds functionality to `Document` such that users can now supply a default value to the `subscript()` function. An example of this can be found in `DocumentTests.swift`, at the bottom, but here is a snippet:
```swift
let doc: Document = ["hello": "world"]
doc["DNE", default: 18.2] // Returns 18.2
doc["hello", default: 18.2] // Returns "world"
doc["DNE", default: "this is a string"] // Returns "this is a string"
doc["DNE", default: NSNull()] // Returns NSNull
```

This is added as a second subscript function that just coalesces on a call to the former. We have to do this because `subscript()` implementations cannot take default valued parameters.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -100,6 +100,12 @@ extension Document {
         }
     }
 
+    /// An implementation identical to subscript(key: String), but offers the ability to choose a default value if the
+    /// key is missing.
+    public subscript(key: String, default defaultValue: @autoclosure () -> BSONValue) -> BSONValue {
+        return self[key] ?? defaultValue()
+    }
+
     /// Sets key to newValue. if checkForKey=false, the key/value pair will be appended without checking for the key's
     /// presence first.
     internal mutating func setValue(for key: String, to newValue: BSONValue, checkForKey: Bool = true) throws {
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -329,8 +329,16 @@ extension Document {
         }
     }
 
-    /// An implementation identical to subscript(key: String), but offers the ability to choose a default value if the
-    /// key is missing.
+    /**
+     * An implementation identical to subscript(key: String), but offers the ability to choose a default value if the
+     * key is missing.
+     * For example:
+     *  ```
+     *  let d: Document = ["hello": "world"]
+     *  print(d["hello", default: "foo"]) // prints "world"
+     *  print(d["a", default: "foo"]) // prints "foo"
+     *  ```
+     */
     public subscript(key: String, default defaultValue: @autoclosure () -> BSONValue) -> BSONValue {
         return self[key] ?? defaultValue()
     }
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -357,7 +357,7 @@ extension Document {
     public subscript(key: String, default defaultValue: @autoclosure () -> BSONValue) -> BSONValue {
         return self[key] ?? defaultValue()
     }
-  
+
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -239,20 +241,31 @@ public struct Document: ExpressibleByDictionaryLiteral, ExpressibleByArrayLitera
...
+                }
             } catch {
                 preconditionFailure("Failed to set the value for key \(key) to \(newValue ?? "nil"): \(error)")
             }
         }
     }
 
+    /// An implementation identical to subscript(key: String), but offers the ability to choose a default value if the
+    /// key is missing.
+    public subscript(key: String, default defaultValue: @autoclosure () -> BSONValue) -> BSONValue {
```
jmikola: Is it common for this to use an autoclosure as the way to yield a default value? The tests seem to use basic types (e.g. a float value, newly constructed NSNull), so I'm a bit confused. Is this intended to support anything that would resolve to a BSONValue?

kmahar:
can you rebase please?

Utagai:
@kmahar Forgot to rebase, done now.

I'll get back to any comments some night after work!!!

kmahar:
thanks! lgtm pending adding some tests that utilize the autoclosure thing.

Sources/MongoSwift/BSON/Document.swift
Utagai: @jmikola Technically, the only thing `@autoclosure` does is wrap arguments in closures. Since even single-value arguments are just expressions, we are technically using `@autoclosure`! By the way: even without `@autoclosure`, the function would _still_ support any expression that evaluated to a `BSONValue`, because proving that resolution is the compiler's job. This only makes it so that the actual evaluation happens later. As @kmahar points out, I only kept this because this is what `Dictionary` does.

I've added a more complicated case where we do `floatVal * floatVal` so that we have something that isn't just a trivial value. This isn't much more complex, but is at the very least an expression that will be have its time of computation affected. Let me know if you guys want something even crazier.

kmahar: tests look sufficiently complex to me.

kmahar:
note this will need a rebase given SWIFT-270 to move this into the extension with the public API.

Utagai:
Rebase should be good and the `subscript()` should now be in the extension.

Something seems weird with Travis, or is it just me? Seems like a bunch of things failed but I see no logs from them whatsoever, as if they just never ran?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -328,6 +328,12 @@ extension Document {
             }
         }
     }
+
+    /// An implementation identical to subscript(key: String), but offers the ability to choose a default value if the
+    /// key is missing.
+    public subscript(key: String, default defaultValue: @autoclosure () -> BSONValue) -> BSONValue {
```
kmahar: I just realized we should probably have a little example of usage here, like we do for subscript. Could you add one please?

kmahar: Perfect.

Utagai:
Done.

PR #156, opened 2018-12-02
SWIFT-265: ObjectId should expose timestamp

jsflax:
- Read timestamp from `bson_oid_get_time_t`
- Expose as `timestamp` within `ObjectId` class

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -547,6 +547,9 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     /// This `ObjectId`'s data represented as a `String`.
     public let oid: String
 
+    /// The timestamp used to create this `ObjectId`
+    public let timestamp: TimeInterval
+
     /// Initializes a new `ObjectId`.
     public init() {
         var oid_t = bson_oid_t()
@@ -558,6 +561,9 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(fromString oid: String) {
         self.oid = oid
+        var oid_t = bson_oid_t()
+        bson_oid_init_from_string(&oid_t, oid)
+        self.timestamp = TimeInterval(bson_oid_get_time_t(&oid_t))
     }
 
     /// Initializes an `ObjectId` from the provided `String`. Returns `nil` if the string is not a valid
@@ -579,6 +585,7 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
             bson_oid_to_string(oid_t, bytes)
             return String(cString: bytes)
         }
+        self.timestamp = TimeInterval(bson_oid_get_time_t(oid_t))
     }
 
     /// Returns the provided string as a `bson_oid_t`.
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -548,7 +548,7 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     public let oid: String
 
     /// The timestamp used to create this `ObjectId`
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -558,6 +561,9 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(fromString oid: String) {
         self.oid = oid
+        var oid_t = bson_oid_t()
+        bson_oid_init_from_string(&oid_t, oid)
+        self.timestamp = TimeInterval(bson_oid_get_time_t(&oid_t))
```
jmikola: [TimeInterval](https://developer.apple.com/documentation/foundation/timeinterval) doesn't seem like the correct type. It appears to be an alias of Double and the docs report that it "does not specify a unique point in time, or even a span between specific times." The BSON timestamp is very clearly seconds since the Unix epoch.

Is there another time-related type we could use for this? Otherwise, perhaps an integer (optionally 32-bit) makes sense.

jsflax: Why not? I think there's a good arg to be made for a UInt32, but, Swift date objects produce timestamps as `TimeInterval`s:

```swift
let time: TimeInterval = Date().timeIntervalSince1970
```

I'm amenable to either one. The Java drivers uses an `int`. I think `TimeInterval` would be more idiomatic to Swift, but a UInt32 would be simpler.

jmikola: I didn't realize TimeInterval was idiomatic for reporting offsets from the Unix epoch.

Neither the BSON spec nor the [ObjectID docs](https://docs.mongodb.com/manual/reference/method/ObjectId/) clarify whether the time component is signed. [`bson_oid_get_time_t()`](http://mongoc.org/libbson/current/bson_oid_get_time_t.html) returns a `time_t`, which is likely typedef-ed as a signed 32-bit int on most platforms. Ultimately, the drivers [ObjectID spec](https://github.com/mongodb/specifications/blob/master/source/objectid.rst#timestamp-field) does say this is an unsigned 32-bit integer, so I think that takes priority. If TimeInterval is signed, then perhaps Uint32 makes the most sense.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -547,6 +547,9 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     /// This `ObjectId`'s data represented as a `String`.
     public let oid: String
 
+    /// The timestamp used to create this `ObjectId`
+    public let timestamp: UInt32
```
mbroadst: Any reason this isn't a read-only computed property? 
```
public var timestamp: UInt32 { return UInt32(bson_oid_get_time_t(&oid_t)) }
```

kmahar: I considered that as well, but we don't store a pointer to the `bson_oid_t`, so we'd need to recreate it from the string in here. I guess it depends whether we prefer to always store a `UInt32`, or to recreate the `bson_oid_t` every time this is called. it's not apparent to me which of those is better. what do you think?

mbroadst: I'm trying to remember why we chose to store it as a string. It seems to me that ObjectID's are created enough during an applications lifetime that doing all the extra work here to wrap and unwrap it into the libbson type is likely undesirable overhead. Looking more closely at the implementation, it looks like we should instead be storing the `bson_oid_t` (not a pointer, but just the value type - you won't need a deinit for this) and provide computed properties for whatever we want access to.

kmahar: we do in fact create `ObjectId`s in Swift (see `insertOne` and `insertMany`).
I agree that this change makes sense unless if anyone can remember if there was a good reason why we chose to do it this way. I've opened [SWIFT-268](https://jira.mongodb.org/browse/SWIFT-268) for this. I'm going to merge this as-is, and we can address the change in that ticket.

PR #159, opened 2018-12-04
Fix Linux + Swift 4.2 failure

kmahar:
for some strange reason, this test was failing on Linux with Swift 4.2. don't know why. @mbroadst looked at it for a bit but I don't think he figured out the issue. 

anyway, this seems like a weird Swift on Linux bug, so I've modified the test to check for basically the same thing but in a different way.

@Utagai, including you so you can verify that I correctly interpreted what this assertion was trying to check?


kmahar:
good point @Utagai. I've opened [SWIFT-274](https://jira.mongodb.org/browse/SWIFT-274) about this.

Issue #161, opened 2018-12-06
Conform UUID to BSONValue


valeriomazzeo:
Currently `UUID` are encoded as `String` because they are `Encodable` that way.

From https://docs.mongodb.com/manual/core/document/#the-id-field
>The following are common options for storing values for _id:
>...
>Generate a UUID in your application code. For a more efficient storage of the UUID values in the collection and in the _id index, store the UUID as a value of the BSON BinData type.
Index keys that are of the BinData type are more efficiently stored in the index if:
>* the binary subtype value is in the range of 0-7 or 128-135, and
>* the length of the byte array is: 0, 1, 2, 3, 4, 5, 6, 7, 8, 10, 12, 14, 16, 20, 24, or 32.

It would be therefore nice to have a native extension in the driver, that encodes `UUID` as per mongo documentation recommended way instead.

kmahar:
Closing this issue as we've opened [SWIFT-295](https://jira.mongodb.org/browse/SWIFT-295) to track the work.

PR #162, opened 2018-12-06
Conform UUID to BSONValue

valeriomazzeo:
Fixes https://github.com/mongodb/mongo-swift-driver/issues/161

### DIFF
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -74,6 +74,8 @@ public struct AnyBSONValue: Codable, Equatable {
             self.value = value
         } else if let value = try? container.decode(ObjectId.self) {
             self.value = value
+        } else if let value = try? container.decode(UUID.self) {
+            self.value = value
         } else if let value = try? container.decode(Bool.self) {
             self.value = value
         } else if let value = try? container.decode(RegularExpression.self) {
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -648,7 +648,7 @@ extension UUID: BSONValue {
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         let uuid = self.uuid
         let uuidData = withUnsafePointer(to: uuid) { unsafePointer in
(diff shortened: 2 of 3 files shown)


kmahar:
Hi @valeriomazzeo, thanks so much for the PR. Would you be willing to add a test showcasing how this works in `BSONValueTests.swift`?

valeriomazzeo:
Sure, will do as soon as I can, probably on Monday.

This is how `UUID` are saved right now:
screenshot 2018-12-07 at 17 16 04

This is how they are saved with this PR:
screenshot 2018-12-07 at 17 15 19

valeriomazzeo:
@kmahar I have added a test which is highlighting a major issue.
I was expecting  that by conforming a type to `BSONValue`, `BSONEncoder`/`BSONDecoder` would automatically `encode`/`decode` such type.

What's happening is that just as a coincidence the encoding works, because everything is transformed to a `Document` before being written to the database, de facto triggering a call to `func encode(to storage: DocumentStorage, forKey key: String)`.

Unfortunately, that doesn't happen during decoding for `static func from(iterator iter: DocumentIterator) throws -> BSONValue`.

In fact there's no trace of any call to `BSONValue` encoding/decoding methods from inside `BSONEncoder`/`BSONDecoder`.

I spent some time going through `BSONEncoder`/`BSONDecoder` source code, but at this time I am not sure what was the intended behaviour.

What am I missing?

mbroadst:
Hi @valeriomazzeo, I took a quick look at your problem this afternoon.

The reason you are running into trouble here is that `UUID` provides its own Codable [initializer](https://github.com/apple/swift/blob/master/stdlib/public/Darwin/Foundation/UUID.swift#L165-L175) which expects the data to be a `String`. Since `UUID` is defined as a struct you can't override this method, so we'd probably have to investigate adding something like a `uuidDecodingStrategy` similar to `JSONDecoder`s [dateDecodingStrategy](https://developer.apple.com/documentation/foundation/jsondecoder/2895216-datedecodingstrategy). 

Thoughts @kmahar, @jmikola?

valeriomazzeo:
@mbroadst I totally agree. It seems like the proper way to go. 👍

valeriomazzeo:
Closing this as it's been made obsolete by #176

PR #163, opened 2018-12-06
SWIFT-253 Make test database name accessible from the base class

patrickfreed:
[SWIFT-253](https://jira.mongodb.org/browse/SWIFT-253)

This PR adds new class computed property `testDatabase` to `MongoSwiftTestCase`, which returns the name of the database that particular test case runs against, and replaces all hard-coded database names in tests to use it. 

Since it is static/class, all instances of the same test case share the same test database. It seemed like that was already assumed when the cases were written (e.g. teardown runs in a static context), so I preserved that property here.

### DIFF
--- a/Tests/MongoSwiftTests/CommandMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/CommandMonitoringTests.swift
@@ -16,7 +16,7 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
         self.continueAfterFailure = false
     }
 
-    override class func testDatabase() -> String {
+    override internal class var testDatabase: String {
         return "commandTest"
     }
 
@@ -88,7 +88,7 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
 
     func testAlternateNotificationCenters() throws {
         let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
-        let db = try client.db(type(of: self).testDatabase())
+        let db = try client.db(type(of: self).testDatabase)
         let collection = try db.createCollection("coll1")
         let customCenter = NotificationCenter()
         client.enableMonitoring(forEvents: .commandMonitoring, usingCenter: customCenter)
--- a/Tests/MongoSwiftTests/CommandMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/CommandMonitoringTests.swift
@@ -16,6 +16,10 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
         self.continueAfterFailure = false
     }
 
+    override class func testDatabase() -> String {
+        return "commandTest"
+    }
+
     func testCommandMonitoring() throws {
         let decoder = BSONDecoder()
         let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
@@ -84,7 +88,7 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
 
     func testAlternateNotificationCenters() throws {
         let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
-        let db = try client.db("commandTest")
+        let db = try client.db(type(of: self).testDatabase())
         let collection = try db.createCollection("coll1")
         let customCenter = NotificationCenter()
         client.enableMonitoring(forEvents: .commandMonitoring, usingCenter: customCenter)
--- a/Tests/MongoSwiftTests/CommandMonitoringTests.swift
+++ b/Tests/MongoSwiftTests/CommandMonitoringTests.swift
@@ -16,10 +16,6 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
         self.continueAfterFailure = false
     }
 
-    override internal class var testDatabase: String {
-        return "commandTest"
-    }
-
     func testCommandMonitoring() throws {
         let decoder = BSONDecoder()
         let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
--- a/Tests/MongoSwiftTests/CrudTests.swift
+++ b/Tests/MongoSwiftTests/CrudTests.swift
@@ -19,20 +19,24 @@ final class CrudTests: MongoSwiftTestCase {
         ]
     }
 
+    override class func testDatabase() -> String {
+        return "crudTests"
+    }
+
     // Teardown at the very end of the suite by dropping the "crudTests" db.
     override class func tearDown() {
         super.tearDown()
         do {
-            try MongoClient().db("crudTests").drop()
+            try MongoClient().db(testDatabase()).drop()
         } catch {
-            print("Dropping test db crudTests failed: \(error)")
+            print("Dropping test db \(testDatabase()) failed: \(error)")
         }
     }
 
     // Run tests for .json files at the provided path
     func doTests(forPath: String) throws {
         let client = try MongoClient()
-        let db = try client.db("crudTests")
+        let db = try client.db(type(of: self).testDatabase())
         for (filename, file) in try parseFiles(atPath: forPath) {
 
             if try !client.serverVersionIsInRange(file.minServerVersion, file.maxServerVersion) {
--- a/Tests/MongoSwiftTests/CrudTests.swift
+++ b/Tests/MongoSwiftTests/CrudTests.swift
@@ -19,7 +19,7 @@ final class CrudTests: MongoSwiftTestCase {
         ]
     }
 
-    // Teardown at the very end of the suite by dropping the "crudTests" db.
+    // Teardown at the very end of the suite by dropping the db we tested on.
     override class func tearDown() {
         super.tearDown()
         do {
--- a/Tests/MongoSwiftTests/CrudTests.swift
+++ b/Tests/MongoSwiftTests/CrudTests.swift
@@ -19,7 +19,7 @@ final class CrudTests: MongoSwiftTestCase {
         ]
     }
 
-    override class func testDatabase() -> String {
+    override internal class var testDatabase: String {
         return "crudTests"
     }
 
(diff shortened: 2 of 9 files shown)


Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -9,13 +9,17 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
 
+    override class func testDatabase() -> String {
+        return "testDB"
+    }
+
     override func setUp() {
         continueAfterFailure = false
     }
 
     func testDatabase() throws {
```
Utagai: I just noticed that this has the same name as the added function (or computed property if my early suggestion actually works).

Let's rename this to something like `testMongoDatabase()`?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -12,6 +12,11 @@ class MongoSwiftTestCase: XCTestCase {
         MongoSwift.initialize()
     }
 
+    /// Gets the name of the database the test case is running against.
+    class func testDatabase() -> String {
```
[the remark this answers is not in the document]
patrickfreed: I didn't actually know you could override computed properties in swift. Changed it to that.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
patrickfreed: Good catch, updated to `testMongoDatabase`.

Tests/MongoSwiftTests/TestUtils.swift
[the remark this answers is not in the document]
patrickfreed: The default visibility in swift is internal, but explicit is usually better than explicit.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -5,17 +5,21 @@ import XCTest
 final class MongoDatabaseTests: MongoSwiftTestCase {
     static var allTests: [(String, (MongoDatabaseTests) -> () throws -> Void)] {
         return [
-            ("testDatabase", testDatabase)
+            ("testDatabase", testMongoDatabase)
```
kmahar: I think we also need to update the string here.

Tests/MongoSwiftTests/TestUtils.swift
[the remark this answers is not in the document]
Utagai: Yup, that's why I'd like to see it in the code! Thanks!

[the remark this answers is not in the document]
Utagai: Yup!! 🤓

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -88,11 +88,11 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
             expect(client.readConcern).to(beNil())
 
             // expect that a DB created from this client inherits its unset RC 
-            let db1 = try client.db("test")
+            let db1 = try client.db(type(of: self).testDatabase())
```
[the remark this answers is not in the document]
kmahar: my bad on the unhelpful names 😛

kmahar:
it seems possible that we would use multiple databases at some point, but I can't think of anything off the top of my head. maybe @jmikola would know if there are any spec tests that need multiple.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
patrickfreed: woops, fixed.

Tests/MongoSwiftTests/TestUtils.swift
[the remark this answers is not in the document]
kmahar: there's a swiftlint rule for explicit access control level we can turn on

kmahar:
general comment: unless if we have a compelling reason to override the default name, let's not bother to.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -5,17 +5,21 @@ import XCTest
 final class MongoDatabaseTests: MongoSwiftTestCase {
     static var allTests: [(String, (MongoDatabaseTests) -> () throws -> Void)] {
         return [
-            ("testDatabase", testDatabase)
+            ("testMongoDatabase", testMongoDatabase)
         ]
     }
 
+    override internal class var testDatabase: String {
```
kmahar: do we have a reason to override this, or should we just use the default name?

I have the same question for all the other classes where we are doing this. if the overriding is just to be consistent with the names I originally chose, we don't need to stick to them.

patrickfreed: Yeah, the only reason any overrides are in here is just to preserve what database was used before. I can make them all use the default.

patrickfreed: Done

kmahar:
now that I think about it... is there is a reason this needs to be a `class` property? 

iirc, we made some of the other stuff on `MongoSwiftTestCase` `static`/`class` because we wanted to access it in contexts where we didn't actually have an instance of `MongoSwiftTestCase` available. but it looks like every time we call this we do have an instance of `self` available, since we are doing `type(of: self)`. if that's the case, maybe we should make it non-`class`, and then we can just refer to it as `self.testDatabase`, which is cleaner. 

now that I actually look at the other properties, I'm not clear on what those uses were where we didn't have `self`, as from a quick inspection they all seem to be used in places where `self` is available... so we may need to open a ticket about correcting those

patrickfreed:
> it looks like every time we call this we do have an instance of `self` available

It seems like the two places where we don't are in the suite teardowns of `MongoCollectionTests` (line 86) and `CrudTests` (line 23)

kmahar:
aha, that makes sense. ok, we can leave it as a class property then.

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -23,16 +23,16 @@ final class CrudTests: MongoSwiftTestCase {
     override class func tearDown() {
         super.tearDown()
         do {
-            try MongoClient().db("crudTests").drop()
+            try MongoClient().db(testDatabase).drop()
```
kmahar: so I guess we don't have a `self` here to take the type of. can we at least prefix the property with the class name so we know where this came from? i.e. `CrudTests.testDatabase`

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -23,16 +23,16 @@ final class CrudTests: MongoSwiftTestCase {
...
         } catch {
-            print("Dropping test db crudTests failed: \(error)")
+            print("Dropping test db \(testDatabase) failed: \(error)")
```
kmahar: same q as line 26

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -90,7 +90,7 @@ final class MongoCollectionTests: MongoSwiftTestCase {
                 print("Invalid client")
                 return
             }
-            try client.db("collectionTest").drop()
+            try client.db(testDatabase).drop()
```
kmahar: same q

Tests/MongoSwiftTests/CrudTests.swift
patrickfreed: You actually can use `self` in class contexts, and it refers to the type (i.e. `CrudTests`). It's also nice to use for dynamic dispatch in case a subclass overrides any static properties/methods.

Added.

kmahar: oh right, I think we talked about that last week with the `NSNull` init :) sounds good to me.

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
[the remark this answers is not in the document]
Utagai: @kmahar No worries! I've made far worse variable names before :).

@patrickfreed I may be looking at the wrong diff, but it seems like this hasn't been done yet?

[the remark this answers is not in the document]
kmahar: I think we could just leave them for now - as you said it's outside the scope of this ticket, and I think we can figure out what they do easily enough from the comments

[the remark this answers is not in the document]
patrickfreed: Yeah I thought you meant something like "think about doing this in a separate ticket" by your comment, so I didn't do anything here.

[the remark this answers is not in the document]
Utagai: Sure, sounds fine to me. I think this is a trivial enough change (since its just a set of renames) that we _could_ do it here if we want. Doing it outside of this ticket is fine as well, lets at least file the ticket!

Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -84,7 +84,7 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
 
     func testAlternateNotificationCenters() throws {
         let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
-        let db = try client.db("commandTest")
+        let db = try client.db(type(of: self).testDatabase)
```
jmikola: Why is `type(of:)` required? Wouldn't this be similar to how we call `self.getSpecsPath()` in other tests, even though that method is defined on XCTestCase (via an extension)?

patrickfreed: I don't believe we defined such a method. We do have the static computed property `specsPath` on `MongoSwiftTestCase`, which we refer to by `MongoSwiftTestCase.specsPath` in all our tests. 

Using `type(of:)` to access `testDatabase` is slightly less brittle in the event we rename `MongoSwiftTestCase`, and, more importantly, it allows us to override `testDatabase` in subclasses without updating how we refer to it.

jmikola: Oops, I was looking at an outdated git branch when I mentioned `getSpecsPath()`. It has since been replaced with a static var on MongoSwiftTestCase.

Based on [this article](https://www.hackingwithswift.com/example-code/language/whats-the-difference-between-a-static-variable-and-a-class-variable), my understanding is that `static` and `class` only differ in whether they allow inheritance (unless that's an over simplification). I would have expected that `self` here would refer to the compile-time class where this method is defined (i.e. CommandMonitoringTests). So if CommandMonitoringTests had overwritten `testDatabase`, that new value would be returned. I would think that `type(of: self)` is more useful in a base class (e.g. a test helper method on MongoSwiftTestCase). From my experience with PHP, that would be a form of [late static binding](http://php.net/manual/en/language.oop5.late-static-bindings.php).

I admittedly haven't worked with this in Swift, so I'm trying to wrap my head around it. Happy to defer to the other reviewers that this is correct, though.

PR #166, opened 2018-12-10
SWIFT-270 Move internal Document methods to a separate extension

patrickfreed:
[SWIFT-270](https://jira.mongodb.org/browse/SWIFT-270)

This PR moves all the non-stored properties and functions out of the `Document` struct declaration into separate extensions, one for public API and one for internal/private functionality.

The original ticket suggests simply moving the internal properties/functions out of the declaration and keeping the public API in, but I've moved everything out so as to keep the declaration as bare bones as possible. This maintains the desired readability improvements while also minimizing how much code is duplicated in conditional declarations (which will occur when adding `@dynamicMemberLookup` in SWIFT-219). 

The git diff for this is pretty weird, but all I did was move code around to different extensions.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -119,7 +119,7 @@ extension Document {
                 /// so initialization will succeed and ! is safe.
                 try DocumentIterator(forDocument: self, advancedTo: key)!.overwriteCurrentValue(with: ov)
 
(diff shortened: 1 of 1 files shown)


kmahar:
the new structure lgtm. could we lump in [SWIFT-275](https://jira.mongodb.org/browse/SWIFT-275) with this ticket?

PR #170, opened 2018-12-12
SWIFT-219 Add @dynamicMemberLookup to Document

patrickfreed:
[SWIFT-219](https://jira.mongodb.org/browse/SWIFT-219)

This PR adds conditional support for the `@dynamicMemberLookup` attribute to `Document` for versions of swift >= 4.2. This allows one to access members of a document via dot-notation.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -27,6 +27,9 @@ public class DocumentStorage {
     }
 }
 
+// The attribute @dynamicMemberLookup must immediately precede the type declaration. Since Swift doesn't have a real
+// preprocessor, we can't just conditionally apply the attribute (the #endif between the attribute and the type
+// declaration is apparently a problem) and have to conditionally declare the entire struct.
 #if swift(>=4.2)
 /// A struct representing the BSON document type.
 @dynamicMemberLookup
@@ -341,7 +344,6 @@ extension Document {
         }
     }
 
-#if swift(>=4.2)
     /**
      * Allows setting values and retrieving values using dot-notation syntax.
      * For example:
@@ -355,6 +357,7 @@ extension Document {
      *
      * Only available in Swift 4.2+.
      */
+    @available(swift 4.2)
     public subscript(dynamicMember member: String) -> BSONValue? {
         get {
             return self[member]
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -27,14 +27,26 @@ public class DocumentStorage {
     }
 }
 
+#if swift(>=4.2)
```
kmahar: nit: add a comment explaining the attribute has to be on the type declaration and that's why we did this funky if/else.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -328,6 +340,30 @@ extension Document {
...
+     * For example:
+     *  ```
+     *  let d = Document()
+     *  d.a = 1
+     *  print(d.a) // prints 1
+     *  ```
+     * A nil return suggests that the key does not exist in the `Document`. A true BSON null is returned as
+     * an `NSNull`.
+     *
+     * Only available in Swift 4.2+.
```
kmahar: There is an [`@available` attribute](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#//apple_ref/doc/uid/TP40014097-CH35-ID347) in Swift that might be appropriate for specifying this kind of thing. I'm not totally sure what effect it has though.

Sources/MongoSwift/BSON/Document.swift
patrickfreed: Done

Sources/MongoSwift/BSON/Document.swift
patrickfreed: I added the `@available` tag instead of the `#if`. I don't believe it conditionally compiles, but it does provide a better compiler error if someone tries to use the dynamic lookup subscript on the wrong swift version. Since none of the code in the subscript is actually 4.2 specific, we don't need to conditionally compile anyways.

kmahar: awesome. does the tag mean it's available in 4.2 only, or will be there 4.2+?

patrickfreed: `@available(swift <version>)` is shorthand for `@available(swift, introduced: <version>)`, so it will work for 4.2+

jmikola:
Deferring to @kmahar's review.

Issue #164, opened 2018-12-07
BSONValue  protocol doesn't return Self


valeriomazzeo:
I am not sure this is by design but I would expect this protocol method to return `Self` instead.

From `BSONValue.swift`:
`static func from(iterator iter: DocumentIterator) throws -> BSONValue`

This is what makes sense to me:
`static func from(iterator iter: DocumentIterator) throws -> Self`

mbroadst:
@valeriomazzeo could you elaborate on why that would be preferable? I'm sure it's just something we missed when learning Swift if it's indeed the idiomatic preference.

mbroadst:
Ah, after reading more about it this does make sense. The initialized for a DocumentIterator for a given subclass of `BsonValue` should indeed return an instance of the subclass, not any `BsonValue`

kmahar:
Great point @valeriomazzeo. We've opened [SWIFT-278](https://jira.mongodb.org/browse/SWIFT-278) about this, it should be implemented shortly.

kmahar:
fixed by #171.

PR #171, opened 2018-12-14
SWIFT-278 BSONValue.from(iterator) should return Self, not BSONValue

patrickfreed:
[SWIFT-278](url)

This PR updates `BSONValue.from(iterator)` to return `Self` instead of `BSONValue`, which is semantically nicer. Originally, we returned `BSONValue` to facilitate conversion of deprecated BSON 
types to other ones, but that is handled elsewhere now.

Due to some weird issues involving `NSNull` on Linux, this PR also introduces a new type to represent BSON null values: `BSONNull`.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -324,7 +324,7 @@ private struct _BSONKeyedEncodingContainer<K: CodingKey> : KeyedEncodingContaine
         self.container = container
     }
 
-    public mutating func encodeNil(forKey key: Key) throws { self.container[key.stringValue] = NSNull() }
+    public mutating func encodeNil(forKey key: Key) throws { self.container[key.stringValue] = BSONNull() }
     public mutating func encode(_ value: Bool, forKey key: Key) throws { self.container[key.stringValue] = value }
     public mutating func encode(_ value: Int, forKey key: Key) throws { self.container[key.stringValue] = value }
     public mutating func encode(_ value: Int8, forKey key: Key) throws { try self.encodeNumber(value, forKey: key) }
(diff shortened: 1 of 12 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -119,11 +119,11 @@ extension Array: BSONValue {
     }
 }
 
-/// An extension of `NSNull` to represent BSON null values.
-extension NSNull: BSONValue {
+/// A class to represent the BSON null type
```
```suggestion
/// A struct to represent the BSON null type.
```

kmahar:
This fixes #164 cc @valeriomazzeo

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -324,7 +324,7 @@ private struct _BSONKeyedEncodingContainer<K: CodingKey> : KeyedEncodingContaine
         self.container = container
     }
 
-    public mutating func encodeNil(forKey key: Key) throws { self.container[key.stringValue] = NSNull() }
+    public mutating func encodeNil(forKey key: Key) throws { self.container[key.stringValue] = BSONNull() }
```
mbroadst: nit: can we rename this `encodeNull`?

Sources/MongoSwift/BSON/BSONValue.swift
mbroadst: look at this cool new github feature!

Sources/MongoSwift/BSON/BSONEncoder.swift
patrickfreed: `encodeNil` is actually a requirement of the `Encoder` protocol, so we have to keep the name.

patrickfreed:
As per our offline discussion, I think the flexibility to have classes conform to `BSONValue` in the future is worth keeping around. (re: `init` question)

Sources/MongoSwift/BSON/BSONEncoder.swift
mbroadst: oh d'oh that's right

PR #175, opened 2018-12-18
Added public initializer to BSONNull

### LINKED ISSUE #174: BSONNull init should be public
https://github.com/mongodb/mongo-swift-driver/pull/171 deprecates the use of `NSNull` in favour of `BSONNull`

Unfortunately, it's not possible to instantiate `BSONNull` outside the `MongoSwift` module because the automatically synthesised initialiser is `internal`.

Adding a simple `public init() { }` would solve this issue.

valeriomazzeo:
Fixes #174.

#171 deprecates the use of NSNull in favour of BSONNull

Unfortunately, it's not possible to instantiate BSONNull outside the MongoSwift module because the automatically synthesised initialiser is internal.

Adding a simple public init() { } would solve this issue.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -125,6 +125,7 @@ public struct BSONNull: BSONValue, Codable {
 
     public static func from(iterator iter: DocumentIterator) throws -> BSONNull { return BSONNull() }
 
+    /// Initializes a new, empty `BSONNull` instance.
     public init() { }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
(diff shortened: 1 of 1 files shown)


patrickfreed:
Thanks for finding this and submitting this patch!

I know this is trivial, but it could use a line of documentation (as all our public declarations should). Other than that, it looks fine with me

valeriomazzeo:
Ok, but I am not sure what you'd like me to add?

patrickfreed:
I guess something simple like "Initializes a new, empty \`BSONNull\` instance." is sufficient.

valeriomazzeo:
done 👍

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -125,6 +125,9 @@ public struct BSONNull: BSONValue, Codable {
 
     public static func from(iterator iter: DocumentIterator) throws -> BSONNull { return BSONNull() }
 
+    /// Initializes a new, empty `BSONNull` instance.
```
```suggestion
    /// Initializes a new `BSONNull` instance.
```
nit: I feel like "empty" implies there could be data stored in it, which is never the case.

PR #176, opened 2018-12-19
SWIFT-281 Allow .uuidDecodingStrategy and .dateDecodingStrategy to be set for BSONDecoder

patrickfreed:
[SWIFT-281](https://jira.mongodb.org/browse/SWIFT-281)

This PR adds two options for specifying a `DateDecodingStrategy` and a `UUIDDecodingStrategy` on a given `BSONDecoder`. 

For `Date`, I mimicked the functionality of `JSONDecoder`'s [strategies](https://developer.apple.com/documentation/foundation/jsondecoder/datedecodingstrategy). I also added a `.bsonDateTime` strategy for when the date is encoded as a BSON datetime. 
The default strategy is `.bsonDateTime`.

For `UUID`, I added the following strategies:

- `.defferredToUUID`: uses `UUID`'s implementation of `init(from: decoder)`
- `.binary`: decodes from a BSON binary type

The default is `.binary` as per the encoding in #162 .

The ticket mentions changing the decoder from `public` to `open`, but I didn't see why that is necessary. Openness really just means the class can subclassed / its properties and functions can be overridden outside of the MongoSwift module. Does the introduction of these strategies necessitate that?

### DIFF
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -58,11 +58,8 @@ public struct AnyBSONValue: Codable, Equatable {
     public init(from decoder: Decoder) throws {
         // short-circuit in the `BSONDecoder` case
         if let bsonDecoder = decoder as? _BSONDecoder {
-            if bsonDecoder.storage.topContainer as? Date != nil {
-                switch bsonDecoder.options.dateDecodingStrategy {
-                case BSONDecoder.DateDecodingStrategy.bsonDateTime:
-                    break
-                default:
+            if bsonDecoder.storage.topContainer is Date {
+                guard case .bsonDateTime = bsonDecoder.options.dateDecodingStrategy else {
                     throw MongoError.bsonDecodeError(message: "Got a BSON datetime but was expecting another format. " +
                             "To decode from BSON datetimes, use the default .bsonDateTime DateDecodingStrategy.")
                 }
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -58,6 +58,15 @@ public struct AnyBSONValue: Codable, Equatable {
     public init(from decoder: Decoder) throws {
         // short-circuit in the `BSONDecoder` case
         if let bsonDecoder = decoder as? _BSONDecoder {
+            if bsonDecoder.storage.topContainer as? Date != nil {
+                switch bsonDecoder.options.dateDecodingStrategy {
+                case BSONDecoder.DateDecodingStrategy.bsonDateTime:
+                    break
+                default:
+                    throw MongoError.bsonDecodeError(message: "Got a BSON datetime but was expecting another format. " +
+                            "To decode from BSON datetimes, use the default .bsonDateTime DateDecodingStrategy.")
+                }
+            }
             self.value = bsonDecoder.storage.topContainer
             return
         }
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -329,8 +329,10 @@ extension _BSONDecoder {
     fileprivate func unbox<T: Decodable>(_ value: BSONValue, as type: T.Type) throws -> T {
         // swiftlint:disable force_cast
         if type == Date.self {
+            // We know T is a Date and unboxDate returns a Date or throws, so this cast will always work
             return try unboxDate(value) as! T
         } else if type == UUID.self {
+            // We know T is a Date and unboxUUID returns a UUID or throws, so this cast will always work
             return try unboxUUID(value) as! T
         }
         // swiftlint:enable force_cast
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -15,22 +15,16 @@ public class BSONDecoder {
         /// Decode `Date`s stored as BSON datetimes.
         case bsonDateTime
 
-        /// Decode `Date`s stored as 64 bit integers counting the number of milliseconds since January 1, 1970.
-        case millisecondsSince1970Int64
-
-        /// Decode `Date`s stored as 64 bit integers counting the number of seconds since January 1, 1970.
-        case secondsSince1970Int64
-
-        /// Decode `Date`s stored as `Double`s counting the number of seconds since January 1, 1970.
+        /// Decode `Date`s stored as numbers of seconds since January 1, 1970.
         case millisecondsSince1970
 
-        /// Decode `Date`s stored as BSON doubles counting the number of milliseconds since January 1, 1970.
+        /// Decode `Date`s stored as numbers of milliseconds since January 1, 1970.
         case secondsSince1970
 
         /// Decode `Date`s by deferring to their default decoding implementation.
         case deferredToDate
 
-        /// Decode `Date`s represented by ISO8601 formatted strings.
+        /// Decode `Date`s stored as ISO8601 formatted strings.
         @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
         case iso8601
 
@@ -46,20 +40,20 @@ public class BSONDecoder {
         /// Decode `UUID`s by deferring to their default decoding implementation.
         case deferredToUUID
 
-        /// Decode `UUID`s from strings.
+        /// Decode `UUID`s stored as strings.
         case fromString
 
-        /// Decode `UUID`s from the BSON `Binary` type.
+        /// Decode `UUID`s stored as the BSON `Binary` type.
         case fromBinary
     }
 
     /// Contextual user-provided information for use during decoding.
     public var userInfo: [CodingUserInfoKey: Any] = [:]
 
-    /// The strategy used for decoding dates with this instance.
+    /// The strategy used for decoding `Date`s with this instance.
     public var dateDecodingStrategy: DateDecodingStrategy = .bsonDateTime
 
-    /// The strategy used for decoding UUIDs with this instance.
+    /// The strategy used for decoding `UUID`s with this instance.
     public var uuidDecodingStrategy: UUIDDecodingStrategy = .fromBinary
 
     /// Options set on the top-level decoder to pass down the decoding hierarchy.
@@ -272,17 +266,11 @@ extension _BSONDecoder {
             defer { self.storage.popContainer() }
             return try Date(from: self)
         case .millisecondsSince1970:
-            let val = try self.unbox(value, as: TimeInterval.self)
-            return Date(timeIntervalSince1970: val / 1000.0)
+            let ms = try unboxNumber(value, as: Int64.self)
+            return Date(msSinceEpoch: ms)
         case .secondsSince1970:
-            let val = try self.unbox(value, as: TimeInterval.self)
-            return Date(timeIntervalSince1970: val)
-        case .millisecondsSince1970Int64:
-            let val = try self.unbox(value, as: Int64.self)
-            return Date(msSinceEpoch: val)
-        case .secondsSince1970Int64:
-            let val = try self.unbox(value, as: Int64.self)
-            return Date(timeIntervalSince1970: TimeInterval(val))
+            let seconds = try unboxNumber(value, as: Double.self)
+            return Date(timeIntervalSince1970: seconds)
         case .iso8601:
             guard #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) else {
                 throw MongoError.bsonDecodeError(message: "ISO8601DateFormatter is unavailable on this platform.")
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -14,7 +14,7 @@ public class BSONDecoder {
     ///
     /// As per the BSON specification, the default strategy is to decode `Date`s from BSON datetime objects.
     ///
-    /// See bsonspec.org for more information.
+    /// - SeeAlso: bsonspec.org
     public enum DateDecodingStrategy {
         /// Decode `Date`s stored as BSON datetimes.
         case bsonDateTime
@@ -44,7 +44,7 @@ public class BSONDecoder {
     /// As per the BSON specification, the default strategy is to decode `UUID`s from BSON binary types with the UUID
     /// subtype.
     ///
-    /// See bsonspec.org for more information.
+    /// - SeeAlso: bsonspec.org
     public enum UUIDDecodingStrategy {
         /// Decode `UUID`s by deferring to their default decoding implementation.
         case deferredToUUID
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -284,15 +284,20 @@ extension _BSONDecoder {
             let val = try self.unbox(value, as: Int64.self)
             return Date(timeIntervalSince1970: TimeInterval(val))
         case .iso8601:
-            if #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) {
-                let isoString = try self.unbox(value, as: String.self)
-                guard let date = BSONDecoder.iso8601Formatter.date(from: isoString) else {
-                    throw MongoError.bsonDecodeError(message: "Improperly formatted ISO 8601 Date string")
-                }
-                return date
-            } else {
+            guard #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) else {
                 throw MongoError.bsonDecodeError(message: "ISO8601DateFormatter is unavailable on this platform.")
             }
+            let isoString = try self.unbox(value, as: String.self)
+            guard let date = BSONDecoder.iso8601Formatter.date(from: isoString) else {
+                throw DecodingError.dataCorrupted(
+                        DecodingError.Context(
+                                codingPath: self.codingPath,
+                                debugDescription: "String \"\(isoString)\" is not a properly formatted " +
+                                        "ISO 8601 Date string"
+                        )
+                )
+            }
+            return date
         case .custom(let f):
             self.storage.push(container: value)
             defer { self.storage.popContainer() }
@@ -300,7 +305,13 @@ extension _BSONDecoder {
         case .formatted(let formatter):
             let dateString = try self.unbox(value, as: String.self)
             guard let date = formatter.date(from: dateString) else {
-                throw MongoError.bsonDecodeError(message: "Date string does not match format expected by formatter.")
+                throw DecodingError.dataCorrupted(
+                        DecodingError.Context(
+                                codingPath: self.codingPath,
+                                debugDescription: "String \"\(dateString)\" does not match the format " +
+                                        "expected by formatter."
+                        )
+                )
             }
             return date
         }
@@ -317,7 +328,11 @@ extension _BSONDecoder {
         case .fromString:
             let uuidString = try self.unbox(value, as: String.self)
              guard let uuid = UUID(uuidString: uuidString) else {
-                throw MongoError.bsonDecodeError(message: "Could not decode UUID from \(uuidString)")
+                 throw DecodingError.dataCorrupted(
+                         DecodingError.Context(
+                                 codingPath: self.codingPath,
+                                 debugDescription: "Could not decode UUID from string \"\(uuidString)\"")
+                 )
             }
             return uuid
         case .fromBinary:
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -11,6 +11,10 @@ public class BSONDecoder {
     }()
 
     /// Enum representing the different options for decoding `Date`s from BSON.
+    ///
+    /// As per the BSON specification, the default strategy is to decode `Date`s from BSON datetime objects.
+    ///
+    /// See bsonspec.org for more information.
     public enum DateDecodingStrategy {
         /// Decode `Date`s stored as BSON datetimes.
         case bsonDateTime
@@ -36,6 +40,11 @@ public class BSONDecoder {
     }
 
     /// Enum representing the different options for decoding `UUID`s from BSON.
+    ///
+    /// As per the BSON specification, the default strategy is to decode `UUID`s from BSON binary types with the UUID
+    /// subtype.
+    ///
+    /// See bsonspec.org for more information.
     public enum UUIDDecodingStrategy {
         /// Decode `UUID`s by deferring to their default decoding implementation.
         case deferredToUUID
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -41,7 +41,7 @@ public class BSONDecoder {
         case deferredToUUID
 
         /// Decode `UUID`s stored as the BSON `Binary` type.
-        case fromBinary
+        case binary
     }
 
     /// Contextual user-provided information for use during decoding.
@@ -51,7 +51,7 @@ public class BSONDecoder {
     public var dateDecodingStrategy: DateDecodingStrategy = .bsonDateTime
 
     /// The strategy used for decoding `UUID`s with this instance.
-    public var uuidDecodingStrategy: UUIDDecodingStrategy = .fromBinary
+    public var uuidDecodingStrategy: UUIDDecodingStrategy = .binary
 
     /// Options set on the top-level decoder to pass down the decoding hierarchy.
     fileprivate struct _Options {
@@ -310,7 +310,7 @@ extension _BSONDecoder {
             self.storage.push(container: value)
             defer { self.storage.popContainer() }
             return try UUID(from: self)
-        case .fromBinary:
+        case .binary:
             let binary = try self.unbox(value, as: Binary.self)
             return try UUID(fromBinary: binary)
         }
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -3,7 +3,7 @@ import Foundation
 /// `BSONDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
 public class BSONDecoder {
 
-    @available(OSX 10.12, *)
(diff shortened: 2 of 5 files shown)


Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -680,4 +682,114 @@ final class DocumentTests: MongoSwiftTestCase {
...
+
+        // randomly generated uuid
+        guard let uuid = UUID(uuidString: "2c380a6c-7bc5-48cb-84a2-b26777a72276") else {
+            throw MongoError.bsonDecodeError(message: "Cant create URI")
+        }
+
+        let decoder = BSONDecoder()
+
+        // UUID default decoder expects a string
+        decoder.uuidDecodingStrategy = .deferredToUUID
```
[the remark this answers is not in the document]
patrickfreed: It works fine with and without `@testable`. `open` really only matters when it comes to inheritance and overriding, not simply assigning (from what I understand).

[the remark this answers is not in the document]
kmahar: huh, interesting, I wonder why they bothered to make `JSONDecoder` open then....

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+
+    /// Enum representing the different options for decoding `Date`s from BSON
+    public enum DateDecodingStrategy {
+        /// Decoding `Date`s stored as 64 bit integers counting the number of milliseconds since January 1, 1970
+        case millisecondsSince1970Int64
+
+        /// Decoding `Date`s stored as 64 bit integers counting the number of seconds since January 1, 1970
+        case secondsSince1970Int64
+
+        /// Decoding `Date`s stored as `Double`s counting the number of seconds since January 1, 1970
```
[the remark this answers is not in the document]
patrickfreed: Done

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
 public class BSONDecoder {
 
+    @available(OSX 10.12, *)
+    internal static var iso8601Formatter: ISO8601DateFormatter = {
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = .withInternetDateTime
+        return formatter
+    }()
+
+    /// Enum representing the different options for decoding `Date`s from BSON
```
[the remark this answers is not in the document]
patrickfreed: 1. Done. We actually had a rule on the C++ driver that enforced _no periods_ at the end of comments for whatever reason. 
2. Done

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+        case millisecondsSince1970
+
+        /// Decoding `Date`s stored as `Double`s counting the number of milliseconds since January 1, 1970
+        case secondsSince1970
+
+        /// Decoding `Date`s by deferring to their default decoding implementation
+        case deferredToDate
+
+        /// Decoding `Date`s represented by ISO8601 formatted strings
+        @available(OSX 10.12, *)
```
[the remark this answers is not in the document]
patrickfreed: The `*` means "potential future platforms".

For this specific case, `JSONEncoder` uses `available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)`, which covers everything.

When I looked up the source of `ISO8601DateFormatter` though, its own attribute was `@available(OSX 10.12, *)`, which is what I went with. Now that I'm looking again, I can't find that anywhere... 

At any rate, I'll stick with what `JSONDecoder` does, as it matches what the [docs](https://developer.apple.com/documentation/foundation/iso8601dateformatter) say as well.

As far as Linux support goes, I'm not really sure. The availability of this formatter seems dependent upon platform instead of Swift version, which is pretty odd. Linux users could always use `.formatted` with their own formatter to achieve the same results.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -198,18 +255,81 @@ extension _BSONDecoder {
...
+            defer { self.storage.popContainer() }
+            return try UUID(from: self)
+        case .fromString:
+            let uuidString = try self.unbox(value, as: String.self)
+             guard let uuid = UUID(uuidString: uuidString) else {
+                throw MongoError.bsonDecodeError(message: "Could not decode UUID from \(uuidString)")
+            }
+            return uuid
+        case .fromBinary:
+            let val: Binary = try self.unbox(value, as: Binary.self)
```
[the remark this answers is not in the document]
patrickfreed: Fixed.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+        case fromString
+
+        /// Decoding `UUID`s from the BSON `Binary` type
+        case fromBinary
+    }
+
     /// Contextual user-provided information for use during decoding.
     public var userInfo: [CodingUserInfoKey: Any] = [:]
 
+    /// The strategy used for decoding dates with this instance.
```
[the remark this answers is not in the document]
patrickfreed: Done

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+    public var dateDecodingStrategy: DateDecodingStrategy = .millisecondsSince1970Int64
+
+    /// The strategy used for decoding UUIDs with this instance
```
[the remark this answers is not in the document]
patrickfreed: Done

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+        /// Decoding `Date`s using the provided closure
+        case custom((_ decoder: Decoder) throws -> Date)
+    }
+
+    /// Enum representing the different options for decoding `UUID`s from BSON
+    public enum UUIDDecodingStrategy {
+        /// Decoding `UUID`s by deferring to their default decoding implementation
+        case deferredToUUID
+
+        /// Decoding `UUID`s from strings
```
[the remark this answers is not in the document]
patrickfreed: Done

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+    internal static var iso8601Formatter: ISO8601DateFormatter = {
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = .withInternetDateTime
+        return formatter
+    }()
+
+    /// Enum representing the different options for decoding `Date`s from BSON
+    public enum DateDecodingStrategy {
+        /// Decoding `Date`s stored as 64 bit integers counting the number of milliseconds since January 1, 1970
+        case millisecondsSince1970Int64
```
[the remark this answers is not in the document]
patrickfreed: I think I originally wanted to preserve dates that could fit in an `Int64` but not in a `Double`, but at the end of the day they're all going into a `Date` which holds `Double`s.

Anyways, done.

Tests/MongoSwiftTests/DocumentTests.swift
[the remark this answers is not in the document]
kmahar: thinking about this more.... my guess is that they did it that way so you could subclass `JSONDecoder` and override properties on it for the entirety of your application rather than having to set it every time you create a new decoder. seems potentially valuable. thoughts?

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,74 @@ import Foundation
...
+
+        /// Decoding `UUID`s from the BSON `Binary` type
+        case fromBinary
+    }
+
     /// Contextual user-provided information for use during decoding.
     public var userInfo: [CodingUserInfoKey: Any] = [:]
 
+    /// The strategy used for decoding dates with this instance.
+    public var dateDecodingStrategy: DateDecodingStrategy = .millisecondsSince1970Int64
```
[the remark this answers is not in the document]
kmahar: hmmm, good point - maybe we should specifically look for that case and give them a helpful error telling them to switch strategies? 
it is tempting to just make all of the other options additive (i.e. `secondsSince1970` actually means both `bsonDateTime` *and* `secondsSince1970`), but from an API perspective I think that might be more confusing... what do you think?

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
patrickfreed: I tested it on Sam's machine, and the availability condition passes. I guess Linux is either included under `*` or the Linux version of Foundation is macOS versioning aware. Either way, it seems to work fine.

Tests/MongoSwiftTests/DocumentTests.swift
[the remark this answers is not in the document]
patrickfreed: I'm not really sure. There isn't a _ton_ of value in subclassing `BSONEncoder` because it doesn't really do anything by itself; it's just a wrapper of the real encoder. The only real reason you'd want to do it would be to, as you said, override the default properties, but you don't need a subclass for that (e.g. define a `makeDecoder` factory function which configures it for you).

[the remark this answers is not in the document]
kmahar: Good point. Let’s leave it then, can always add later if someone has a new compelling reason.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -198,18 +255,81 @@ extension _BSONDecoder {
...
+                throw MongoError.bsonDecodeError(message: "ISO8601DateFormatter is unavailable on this platform.")
+            }
+        case .custom(let f):
+            self.storage.push(container: value)
+            defer { self.storage.popContainer() }
+            return try f(self)
+        case .formatted(let formatter):
+            let dateString = try self.unbox(value, as: String.self)
+            guard let date = formatter.date(from: dateString) else {
+                throw MongoError.bsonDecodeError(message: "Date string does not match format expected by formatter.")
```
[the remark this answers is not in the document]
kmahar: what kind of error does `JSONDecoder` throw when there are ISO problems?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -642,6 +642,32 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
+    // TODO: fill the rest of this out for full BSONValue conformance.
+
+    internal init(fromBinary binary: Binary) throws {
+        guard binary.subtype != Binary.Subtype.uuidDeprecated.rawValue else {
+            throw MongoError.bsonDecodeError(message: "Binary subtype \(binary.subtype) is deprecated, " +
+                    "use \(Binary.Subtype.uuid) instead.")
+        }
+        guard binary.subtype == Binary.Subtype.uuid.rawValue else {
+            throw MongoError.bsonDecodeError(message: "Expected a UUID binary type " +
+                    "(\(Binary.Subtype.uuidDeprecated)), got \(binary.subtype) instead.")
```
kmahar: should this be `Binary.Subtype.uuid`?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -642,6 +642,32 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
 
 }
 
+/// Extension to allow a UUID to be initialized from a Binary BSONValue.
+extension UUID {
+    // TODO: fill the rest of this out for full BSONValue conformance.
```
kmahar: let's make a ticket for that to reference here

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -680,4 +682,122 @@ final class DocumentTests: MongoSwiftTestCase {
         expect(doc["autoclosure test", default: "\(stringVal) and \(floatVal)" + stringVal])
             .to(bsonEqual("\(stringVal) and \(floatVal)" + stringVal))
     }
+
+    func testUUIDCodingStrategies() throws {
+        // TODO: add encoding strategy tests SWIFT-282
+
+        // swiftlint:disable nesting
```
kmahar: should we disable this linter rule? I don't really see the point

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -680,4 +682,122 @@ final class DocumentTests: MongoSwiftTestCase {
...
+        struct UUIDWrapper: Codable {
+            let uuid: UUID
+        }
+        // swiftlint:enable nesting
+
+        // randomly generated uuid
+        guard let uuid = UUID(uuidString: "2c380a6c-7bc5-48cb-84a2-b26777a72276") else {
+            throw MongoError.bsonDecodeError(message: "Cant create URI")
```
kmahar: URI -> UUID?

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
patrickfreed: I think it could be useful, but it's probably safer to keep a consistent API. In general, users aren't going to be setting a strategy for `Date`s unless they _know_ their data is stored in a certain nonstandard manner. Even if the `Date`s could be encoded in multiple ways, they always have `.custom` as a fallback.

Also, I don't think we need a custom error, as the current one is pretty decent, though it does say `Double` when really any number would work:
```
Fatal error: Error raised at top level: Swift.DecodingError.typeMismatch(Swift.Double, Swift.DecodingError.Context(codingPath: [CodingKeys
(stringValue: "b", intValue: nil)], debugDescription: "Expected to find a value that can be represented as a Double, but found value 2018-
12-20 17:20:25 +0000 of type Date instead.", underlyingError: nil)): file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang_Fall20
18/swiftlang_Fall2018-1000.11.42/src/swift/stdlib/public/core/ErrorType.swift, line 191
```

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
patrickfreed: They actually `fatalError`.

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
kmahar: yeah, I kinda assume people will try it the normal way first and only set the strategy if that didn't work. error seems fine now.

Tests/MongoSwiftTests/DocumentTests.swift
patrickfreed: Fixed.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: Definitely. nice catch. fixed

Sources/MongoSwift/BSON/BSONDecoder.swift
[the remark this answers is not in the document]
kmahar: weird given that they can presumably throw from the corresponding place that this happens. error seems like the better choice. 👍

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: filed [SWIFT-295](https://jira.mongodb.org/browse/SWIFT-295)

Tests/MongoSwiftTests/DocumentTests.swift
kmahar: I'll do so in my linting PR. for now we can leave this, if this gets merged first then I'll deal with it in my PR.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -680,4 +682,122 @@ final class DocumentTests: MongoSwiftTestCase {
...
+        // UUID default decoder expects a string
+        decoder.uuidDecodingStrategy = .deferredToUUID
+        let stringDoc: Document = ["uuid": uuid.description]
+        let badString: Document = ["uuid": "hello"]
+        let deferredStruct = try decoder.decode(UUIDWrapper.self, from: stringDoc)
+        expect(deferredStruct.uuid).to(equal(uuid))
+        expect(try decoder.decode(UUIDWrapper.self, from: badString)).to(throwError())
+
+        decoder.uuidDecodingStrategy = .fromString
+        let stringStruct = try decoder.decode(UUIDWrapper.self, from: stringDoc)
```
kmahar: is there a difference in functionality between `deferredToUUID` and `fromString`? they seem to work with the same data here.

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -680,4 +682,122 @@ final class DocumentTests: MongoSwiftTestCase {
...
+        decoder.dateDecodingStrategy = .bsonDateTime
+        let bsonDateStruct = try decoder.decode(DateWrapper.self, from: bsonDate)
+        expect(bsonDateStruct.date).to(equal(date))
+
+        decoder.dateDecodingStrategy = .millisecondsSince1970
+        let msInt64: Document = ["date": date.msSinceEpoch]
+        let msInt64Struct = try decoder.decode(DateWrapper.self, from: msInt64)
+        expect(msInt64Struct.date).to(equal(date))
+
+        let msDouble: Document = ["date": TimeInterval(date.msSinceEpoch)]
```
kmahar: I think we should just call this `Double` over `TimeInterval` - although it's just a `typealias` and makes no functional difference, the [docs](https://developer.apple.com/documentation/foundation/timeinterval) state that a `TimeInterval` is always in seconds, so it is a little confusing to be using it for milliseconds here.

Tests/MongoSwiftTests/DocumentTests.swift
patrickfreed: One explicitly invokes the `init(uuidString:)` and one invokes `init(from: decoder)` (which expects a string). So they do the same thing in different ways.

kmahar: is there a good reason to have both then?

patrickfreed: Actually, no, there isn't. I included both because I mistakenly thought `JSONDecoder` similarly duplicated the default in `Date`, but `.deferredToDate` expects seconds since _reference time_ instead of seconds from epoch/1970 which `.secondsSince1970` expects. I assumed they were duplicated for flexibility in case `init(from:)` ever changed, but it turns out there's no duplication at all.

I'll remove the `.fromString` strategy.

Tests/MongoSwiftTests/DocumentTests.swift
patrickfreed: Yep, I have been trying to keep that distinction too but must have missed this one. Good catch, fixed.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -3,17 +3,68 @@ import Foundation
 /// `BSONDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
 public class BSONDecoder {
 
+    @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
```
jmikola: If I understand the `@available` docs on [Attributes](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html) correctly, this means that `iso8601Formatter` is available on all Swift versions but only the above platforms. If so, how does this interact with Linux builds? Do those only use the Swift language version requirement (i.e. always available in this case)?

Side note: I assume these requirements were pulled from the column in [`date(from:)` docs](https://developer.apple.com/documentation/foundation/iso8601dateformatter/1643127-date) (or a similar page).

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -642,6 +642,32 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
+                    "(\(Binary.Subtype.uuid)), got \(binary.subtype) instead.")
+        }
+
+        let data = binary.data
+        let uuid: uuid_t = (
+                data[0], data[1], data[2], data[3],
+                data[4], data[5], data[6], data[7],
+                data[8], data[9], data[10], data[11],
+                data[12], data[13], data[14], data[15]
+        )
```
jmikola: I didn't find much documentation about `uuid_t` in the [UUID docs](https://developer.apple.com/documentation/foundation/uuid), but I just want to confirm that this logic is agnostic about platform endianness. Do I assume correctly that `uuid_t` is simply an array of 16 characters?

patrickfreed: Yep. From `NSUUID` [docs](https://developer.apple.com/documentation/foundation/nsuuid):
>Because a UUID is expressed simply as an array of bytes, there are no endianness considerations for different platforms.

Sources/MongoSwift/BSON/BSONDecoder.swift
patrickfreed: Yeah, it appears to be platform independent on Linux (The formatter exists going back to 4.0 https://github.com/apple/swift-corelibs-foundation/blob/swift-4.0-branch/Foundation/ISO8601DateFormatter.swift). I honestly don't understand why it has platform specific availability instead of Swift version on Apple devices, but it seems to be the case based on the documentation you mentioned and the code in `JSONEncoder`.

I tested it manually on Ubuntu 18.04 and Swift 4.2, and the formatter was available as expected.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -642,6 +642,32 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
 
 }
 
+/// Extension to allow a UUID to be initialized from a Binary BSONValue.
+extension UUID {
+    // TODO: fill the rest of this out for full BSONValue conformance (SWIFT-295).
+
+    internal init(fromBinary binary: Binary) throws {
```
kmahar: a general naming thing we should think about when we write these `from` inits is whether to use `from` or `fromX`. 

Swift standard library tends toward just `from`/`to`, i.e. `init(from decoder: Decoder)`, `encode(to encoder: Encoder)`, etc. 

I think I lean toward just using `from` unless if it is adding more information beyond just the type of the input. for ex. consider the following `Document` initializers:
- `Document.init(fromJSON: Data)`
- `Document.init(fromBSON: Data)`
- `Document.init(fromJSON: String)`

In all cases we are adding more info about the format - it's not just `Data` but BSON data, it's not just a `String` but a JSON string, etc. here, it feels kind of redundant to say `fromBinary`. 

there is the argument that including the type name might make things more readable at the call site, like if I say `UUID(fromBinary: b)`, but I feel like that is more a problem with the user's code if they can't otherwise tell what type `b` is. 

thoughts?

I realize we're not that consistent about this throughout the codebase so I should probably open a ticket about establishing convention here and correcting violations....

patrickfreed: I think you're right. In statically typed languages like Swift it is pretty redundant to have type information encoded in your labels/variable names, since the compiler will send a much louder and clearer message if there's a mismatch anyways. It also makes refactoring a little harder if you have to change the variable names when you decide to use different types.

changed -> `init(from binary: Binary)`

kmahar:
just realized, I wonder if we need to update anything with regard to decoding `AnyBSONValue`s [here](https://github.com/mongodb/mongo-swift-driver/blob/2610f57ac2c937a039b35e331086e225ea5f0cbf/Sources/MongoSwift/BSON/AnyBSONValue.swift#L44)?

patrickfreed:
> just realized, I wonder if we need to update anything with regard to decoding `AnyBSONValue`s [here](https://github.com/mongodb/mongo-swift-driver/blob/2610f57ac2c937a039b35e331086e225ea5f0cbf/Sources/MongoSwift/BSON/AnyBSONValue.swift#L44)?

We still are facing the same issues the comment talks about, even with the strategies. Without the context of the underlying type we're decoding to, it's impossible to infer whether e.g. a `Double` is a `Date` or just a `Double`. The only case that we can now address is if the user encodes a date as a specific type of subdocument (e.g. [day: "2", month: "JAN", year: "2019"]) and wishes to use a custom decoder strategy to make any subdocument of that format into a `Date`. Is that case worth adding here?

Alternatively, we can modify that initializer to take a 'hint' in from the user that would give preference to `Date` over `Double`, for example.

If someone is decoding to `AnyBSONValue` though, they probably have no idea what is stored in that field of their documents, let alone that it is a `Date` and is encoded with a specific format. Instead of supplying a hint, they should just be more explicit with their structures and data.

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -58,6 +58,15 @@ public struct AnyBSONValue: Codable, Equatable {
     public init(from decoder: Decoder) throws {
         // short-circuit in the `BSONDecoder` case
         if let bsonDecoder = decoder as? _BSONDecoder {
+            if bsonDecoder.storage.topContainer as? Date != nil {
+                switch bsonDecoder.options.dateDecodingStrategy {
+                case BSONDecoder.DateDecodingStrategy.bsonDateTime:
```
kmahar: nit: since we only care about one possible strategy, I would probably write this with a `guard` rather than a `switch`:
```swift
guard bsonDecoder.options.dateDecodingStrategy == .bsonDateTime else {
    // throw
}
```

patrickfreed: I had it this way because you can't actually use `==` between two enum types. You can however use `guard case`, which I've updated it to.

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -58,6 +58,15 @@ public struct AnyBSONValue: Codable, Equatable {
     public init(from decoder: Decoder) throws {
         // short-circuit in the `BSONDecoder` case
         if let bsonDecoder = decoder as? _BSONDecoder {
+            if bsonDecoder.storage.topContainer as? Date != nil {
```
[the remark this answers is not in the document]
patrickfreed: Done 👍

Sources/MongoSwift/BSON/AnyBSONValue.swift
kmahar: ohhhh, yeah I guess we would have to explicitly make the type `Equatable` and probably write some custom implementation since it can have associated values. that makes sense.

PR #178, opened 2018-12-21
SWIFT-282 Allow .uuidEncodingStrategy and .dateEncodingStrategy to be set on BSONEncoder

patrickfreed:
[SWIFT-282](https://jira.mongodb.org/browse/SWIFT-282)

This PR adds the ability to specify encoding strategies for `Date`s and `UUID`s. The strategies for both match their corresponding decoding strategies defined in #176 . 

Also, this PR has some missing comments and has some duplication from the preceding one. Once the other merges, I will rebase this off that and put the pieces together. All of the important implementation is complete.

UPDATE: we now follow `JSONEncoder`'s implementation of the `.custom` date encoding strategy for maintaining interoperability with other encoders. Also, some of the pitfalls of doing it this way were a bit overstated.

Outdated:
However, for the `.custom` strategy, I deviated from the implementation that `JSONEncoder` chose. Ours is defined as `.custom((Date) throws -> BSONValue)`, whereas theirs is `.custom((Date, Encoder) throws -> Void)`. 
From our perspective, it makes more sense due to the context in which the closure is used (i.e. within `box(...) -> BSONValue`). In `JSONEncoder`, they use it in a similar context, so they just pop whatever the user added to the encoder and return it to be added on later anyways. If the user didn't add anything, they return an empty `NSContainer`. For us, we would have to either return a `BSONNull` or throw, both of which I think are unsavory. Also, in the event that the closure throws, they manually clean up the encoder state, which I think is needless overhead.
From the user's perspective, it is also nice that they simply need to supply a closure that translates a `Date` to a `BSONValue` instead of having to deal with encoder logic themselves, and this is closer to the whole point of `.custom` anyways.

The same arguments apply for the decoding version of `.custom` as well. I think, for decoding, `.custom((BSONValue) throws -> Date)` would make more sense than the current `.custom((Decoder) throws -> Date)`, and it would parallel what I've proposed here nicely.

e.g. the following would hold
```
let date = Date()
let wrapper = DateWrapper(date: date)
let decoding: (BSONValue) -> Date = { bv in Date(msSinceEpoch: bv as! Int64) }
let encoding: (Date) -> BSONValue = { date in date.msSinceEpoch }
decoder.dateDecodingStrategy = .custom(decoding)
encoder.dateEncodingStrategy = .custom(encoding)

expect(decoding(encoding(date)).to(equal(date))
expect(try decoder.decode(DateWrapper.self, from: try encoder.encode(wrapper)).date).to(equal(date))
```  

Thoughts on this? I know we generally try to stick to what `JSONDecoder` does, but I think this way has both more value and less headache.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -5,6 +5,10 @@ import mongoc
 public class BSONEncoder {
 
     /// Enum representing the various strategies for encoding `Date`s.
+    ///
+    /// As per the BSON specification, the default strategy is to encode `Date`s as BSON datetime objects.
+    ///
+    /// See bsonspec.org for more information.
     public enum DateEncodingStrategy {
         /// Encode the `Date` by deferring to its default encoding implementation.
         case deferToDate
@@ -31,6 +35,11 @@ public class BSONEncoder {
     }
 
     /// Enum representing the various strategies for encoding `UUID`s.
+    ///
+    /// As per the BSON specification, the default strategy is to encode `UUID`s as BSON binary types with the UUID
+    /// subtype.
+    ///
+    /// See bsonspec.org for more information.
     public enum UUIDEncodingStrategy {
         /// Encode the `UUID` by deferring to its default encoding implementation.
         case deferToUUID
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -29,7 +29,7 @@ public class BSONEncoder {
         case custom((Date) throws -> BSONValue)
     }
 
-    /// Enum representing the various strategies for encoding `UUID`s
+    /// Enum representing the various strategies for encoding `UUID`s.
     public enum UUIDEncodingStrategy {
         /// Encode the `UUID` by deferring to its default encoding implementation.
         case deferToUUID
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -25,8 +25,9 @@ public class BSONEncoder {
         /// Encode the `Date` as a string formatted by the given formatter.
         case formatted(DateFormatter)
 
-        /// Encode the `Date` as the value returned by the provided closure.
-        case custom((Date) throws -> BSONValue)
+        /// Encode the `Date` by using the given closure.
+        /// If the closure does not encode a value, an empty document will be encoded in its place.
+        case custom((Date, Encoder) throws -> Void)
     }
 
     /// Enum representing the various strategies for encoding `UUID`s.
@@ -327,7 +328,7 @@ extension _BSONEncoder {
         return try self.box_(value) ?? Document()
     }
 
-    fileprivate func boxDate(_ date: Date) throws -> BSONValue {
+    fileprivate func boxDate(_ date: Date) throws -> BSONValue? {
         switch self.options.dateEncodingStrategy {
         case .bsonDate:
             return date
@@ -347,7 +348,23 @@ extension _BSONEncoder {
                 throw MongoError.bsonEncodeError(message: "ISO8601DateFormatter is unavailable on this platform.")
             }
         case .custom(let f):
-            return try f(date)
+            let depth = self.storage.count
+
+            do {
+                try f(date, self)
+            } catch {
+                if self.storage.count > depth {
+                    _ = self.storage.popContainer()
+                }
+                throw error
+            }
+
+            // If they didn't encode anything, will default to empty subdocument
+            guard self.storage.count > depth else {
+                return nil
+            }
+
+            return self.storage.popContainer()
         }
     }
 
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -71,13 +71,6 @@ public class BSONEncoder {
                         uuidEncodingStrategy: self.uuidEncodingStrategy)
     }
 
-    @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
-    internal static let iso8601Formatter: ISO8601DateFormatter = {
-        let formatter = ISO8601DateFormatter()
-        formatter.formatOptions = .withInternetDateTime
-        return formatter
-    }()
-
     /// Initializes `self`.
     public init() {}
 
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -4,19 +4,70 @@ import mongoc
...
     /// The options set on the top-level encoder.
     fileprivate var options: _Options {
-        return _Options(userInfo: userInfo)
+        return _Options(userInfo: self.userInfo,
+                        dateEncodingStrategy: self.dateEncodingStrategy,
+                        uuidEncodingStrategy: self.uuidEncodingStrategy)
     }
 
+    @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
+    internal static var iso8601Formatter: ISO8601DateFormatter = {
```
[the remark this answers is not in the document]
patrickfreed: While I do think this formatter should be immutable, it already is `static`, so only one instance is ever created. (Note this is not a computed property, the closure is immediately evaluated).

Regardless, will change to `let`.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -4,19 +4,70 @@ import mongoc
...
+        /// Encode the `Date` as an ISO-8601-formatted string (in RFC 339 format).
+        @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
+        case iso8601
+
+        /// Encode the `Date` as a string formatted by the given formatter.
+        case formatted(DateFormatter)
+
+        /// Encode the `Date` as the value returned by the provided closure.
+        case custom((Date) throws -> BSONValue)
+    }
```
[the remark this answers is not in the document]
patrickfreed: Done, thanks for pointing this out.

Sources/MongoSwift/BSON/BSONEncoder.swift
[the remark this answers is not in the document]
valeriomazzeo: You are right, for some reason I missed the `= { }()` 👍🏻

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -680,4 +680,81 @@ final class DocumentTests: MongoSwiftTestCase {
...
+    }
+
+    func testUUIDEncodingStrategies() throws {
+        // swiftlint:disable nesting
+        struct UUIDWrapper: Encodable {
+            let uuid: UUID
+        }
+        // swiftlint:enable nesting
+
+        guard let uuid = UUID(uuidString: "26cd7610-fd5a-4253-94b7-e8c4ea97b6cb") else {
```
kmahar: I'm fine with leaving this as throwing an error, but just as an FYI, in cases like this in tests where you are passing in a valid constant and know initialization will succeed it's fine to just force unwrap the result (force-unwrap rule is disabled for tests for this reason).

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -642,6 +642,20 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
 
 }
 
+extension UUID {
+    /// Copies `self` into a `Binary`.
+    internal func asBinary() throws -> Binary {
```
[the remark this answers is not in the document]
jmikola: That does seem cleaner, as otherwise this is introducing a generically named `asBinary` method on the core UUID class, which returns an instance of _our_ Binary object.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -4,19 +4,70 @@ import mongoc
...
+
+    /// Enum representing the various strategies for encoding `UUID`s.
+    public enum UUIDEncodingStrategy {
+        /// Encode the `UUID` by deferring to its default encoding implementation.
+        case deferToUUID
```
mbroadst: Is it worth discussing if this should be the actual default? What would swift users expect to be working with? @valeriomazzeo maybe you have thoughts here?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -276,7 +328,63 @@ extension _BSONEncoder {
...
+            do {
+                try f(date, self)
+            } catch {
+                if self.storage.count > depth {
+                    _ = self.storage.popContainer()
+                }
+                throw error
+            }
+
+            // If they didn't encode anything, will default to empty subdocument
```
kmahar: nit: while your comment here is true based on the context this is called from, it's confusing because we return `nil`, not empty doc. 
it might be more appropriate to move this comment into the docstring for `box` where we call this, and to add a docstring to this method (and maybe also `_box`) clarifying the overall behavior - "returns the date as a BSONValue, or nil if no values were encoded by custom strategy", or something like that.

Sources/MongoSwift/BSON/BSONEncoder.swift
patrickfreed: We've chosen `.millisecondsSinceEpoch` as the default date encoding/decoding strategy for BSON spec adherence, and Swift users definitely do not expect their dates to be stored as such (all `Date`s in Swift use `Double`s for whatever reason). So we have precedent for setting it to this default.

Either way, I think we should be consistent with our defaults. If we change this one to deferred, we should also change our date encoding/decoding strategy to deferred as well.

Sources/MongoSwift/BSON/BSONEncoder.swift
patrickfreed: Done 👍

Sources/MongoSwift/BSON/BSONEncoder.swift
valeriomazzeo: I suppose most people (including myself), would expect that to be encoded as a `String`, simply because it's not obvious why it should be encoded as `Binary` unless reading the documentation and I doubt that many will know about it.

I believe it goes down to your preference whether you'd like to enforce best practices as per Mongo official documentation by default, or not.

In my specific case, I believe I am going to stick with `String` initially and then optimise it later as I have some data that would need to be migrated first.

patrickfreed: We've decided that we're going to stick with BSON defaults. I added a bit to the docstring explaining that we're doing this with a link to the spec itself, so hopefully that give users some guidance.

On a slightly related note, I also refactored `.bsonDate` to `.bsonDateTime` to match what I did in the decoder.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -5,12 +5,16 @@ import mongoc
 public class BSONEncoder {
 
     /// Enum representing the various strategies for encoding `Date`s.
+    ///
+    /// As per the BSON specification, the default strategy is to encode `Date`s as BSON datetime objects.
+    ///
+    /// See bsonspec.org for more information.
```
kmahar: these will show up more cnicely in the generated docs if we write them as "see alsos".
see [here](https://github.com/mongodb/mongo-swift-driver/blob/2610f57ac2c937a039b35e331086e225ea5f0cbf/Sources/MongoSwift/ReadPreference.swift#L6) for an example of this in code, and [here's](https://mongodb.github.io/mongo-swift-driver/Classes/ReadPreference.html) how it looks in the docs.

patrickfreed: Fixed

PR #180, opened 2019-01-02
SWIFT-255 BSONValue should conform to Hashable

jsflax:


kmahar:
@jsflax Can you look into what happened on Travis? It exited early on all OS X builds due to linter violations, and on Linux (where we don't lint) compilation failed on < 4.2.

jsflax:
Sure thing.

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -103,4 +103,165 @@ public struct AnyBSONValue: Codable, Equatable {
...
+    public lazy var isCodeWithScope: Bool = value is CodeWithScope
+
+    /// True if this is a MaxKey, false otherwise.
+    public lazy var isMaxKey: Bool = value is MaxKey
+
+    /// True if this is a MinKey, false otherwise.
+    public lazy var isMinKey: Bool = value is MinKey
+}
+
+public func ~=(pattern: BSONValue, value: AnyBSONValue) -> Bool {
```
kmahar: please add docstring

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -103,4 +103,165 @@ public struct AnyBSONValue: Codable, Equatable {
...
+        case let value as Binary:
+            value.hash(into: &hasher)
+        case let value as [BSONValue?]:
+            value.forEach({$0.map({AnyBSONValue($0).hash(into: &hasher)})})
+        default: break
+        }
+    }
+
+    /// This value as a BsonDocument if it is one, otherwise nil
+    public lazy var asDocument = value as? Document
```
kmahar: idk if I'm sold on adding all of this to the public API - is `x.asDocument` that much of an improvement over `x.value as? Document`? 
if we keep them I would like to see all the docstrings formatted like this:
```swift
/// The underlying value as a `Document` if it is one, otherwise `nil`.
```

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -103,4 +103,165 @@ public struct AnyBSONValue: Codable, Equatable {
                 message: "Encountered a value that could not be decoded to any BSON type")
         }
     }
+
+    public func hash(into hasher: inout Hasher) {
```
kmahar: are we actually testing hashing at all?

kmahar:
@jsflax, you said in #swift - 
> there’s a case to be made that `BSONValue` itself should be `Hashable`, and that the structure users actually use is `AnyBSONValue`

can you elaborate?

Sources/MongoSwift/BSON/AnyBSONValue.swift
jsflax: I was going by a conversation that was had in #swift– I'm happy to remove it.

jsflax:
Making `BSONValue` conform to `Hashable` would essentially make it non usable due to the generic constraint. However, I do not believe that is a bad thing. Currently, users have the ability to use two different structures representing a `BSONValue`– `BSONValue`, and `AnyBSONValue`. This ends up exposing various implementation details, notably that `BSONValue`s are not `Codable` or `Hashable`. This could be confusing to an end user. With type-erasure being a common enough pattern in Swift (see: `AnyHashable`), it seems like having BSONValue only be used as a generic constraint would be both semantically appropriate and a better UX.

`AnyBSONValue` would be the default "base" class for BSON values. There would be a case to make it into an enum as well so that unwrapping from documents/arrays would be easy enough:

```swift
guard case let .objectId(objectId) = document["_id"] else {
    ...
}

print(objectId.oid)
```

I actually had `AnyBSONValue` converted to an enum early in this PR, but it was a big enough change that it seemed like it would have warranted a discussion, one that we didn't have time for on Stitch right now.

The reason I did not go down the path of genericizing `BSONValue` is because it is a lot of work, and also relies on `Array`'s conditional conformance to `BSONValue`.

Sources/MongoSwift/BSON/AnyBSONValue.swift
jsflax: Now we are 👍

Sources/MongoSwift/BSON/AnyBSONValue.swift
jsflax: It doesn't have much value add. IMO, we should either make this into an enum, or remove these.

jsflax:
@kmahar PTAL

patrickfreed:
Will revisit this when we implement SWIFT-320. For now I'll just conform `AnyBSONValue` to Hashable.

PR #182, opened 2019-01-05
SWIFT-217 Reevaluate linter rules

kmahar:
This implements all of the previously discussed Swiftlint rules, excluding a couple that I felt were overkill once I actually saw what lines they complained about. See [their docs](https://github.com/realm/SwiftLint/blob/master/Rules.md) for further description of all of the rules.

The changes are almost entirely linting based, I'll add inline comments detailing anywhere there was something of more substance.


Sources/MongoSwift/SDAM.swift
```diff
@@ -22,6 +25,7 @@ public struct ConnectionId: Equatable {
     internal init(_ hostAndPort: String = "localhost:27017") {
         let parts = hostAndPort.split(separator: ":")
         self.host = String(parts[0])
+        // swiftlint:disable:next force_unwrapping - should be valid UInt16 unless server response malformed.
```
patrickfreed: Could potentially mark this as `throws` and throw an `invalidResponse` error. I'm not sure how common it would be for the response to be malformed or what the consequences of this initializer being throwable are, though.

Sources/MongoSwift/SDAM.swift
```diff
@@ -226,9 +231,9 @@ public struct TopologyDescription {
         let timeoutValues = self.servers.map { $0.logicalSessionTimeoutMinutes }
         if timeoutValues.contains (where: { $0 == nil }) {
             return nil
-        } else {
-            return timeoutValues.map { $0! }.min()
         }
+        // swiftlint:disable:next force_unwrapping - we filtered out nil values already.
+        return timeoutValues.map { $0! }.min()
```
patrickfreed: Once we drop swift 4.0 we can use `compactMap` to make this a little cleaner and omit the force unwrap.
Alternatively, we can use `reduce` + length comparison above instead of `contains(where:)` which would also allow us to avoid the force unwrap, though I'm not sure there's a ton of readability / style gains there...

Tests/MongoSwiftTests/.swiftlint.yml
```diff
@@ -1,20 +1,6 @@
 disabled_rules:
-  - file_length
```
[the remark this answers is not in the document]
patrickfreed: Sounds good to me. AppCode looks like it handles this okay from what I can tell.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -669,7 +653,7 @@ extension NSRegularExpression {
     }
 
     /// Convert this instance's options object into an alphabetically-sorted string of characters
-    public var stringOptions: String {
```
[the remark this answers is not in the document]
kmahar: oh yes, good catch! I meant to add a comment here. there's not really a reason for us to expose this - it's just a helper used by `RegularExpression.init(from regex: NSRegularExpression)`.
we provide an initializer for `NSRegularExpression` -> `RegularExpression`, and a `nsRegularExpression` property on `RegularExpression`, so users should use those instead. the second one should arguably be changed to be an initializer on `NSRegularExpression` though. opened [SWIFT-313](https://jira.mongodb.org/browse/SWIFT-313) for discussion.

I suspect I inadvertently marked this public back when I made an initial mass edit to make lots of stuff `public` after learning `internal` was Swift's default, lol.

Sources/MongoSwift/SDAM.swift
kmahar: yeah, I considered erroring here as well. I ended up not doing so because everywhere this is called from doesn't throw right now, so there would have been a lot of additional refactoring needed. this is used in the context of SDAM monitoring, so it's also not clear what the behavior should be if parsing isMaster fails. do we just skip publishing the event? do we publish it with a incorrect or missing `ConnectionId`? I opened [SWIFT-314](https://jira.mongodb.org/browse/SWIFT-314) about this.

Sources/MongoSwift/SDAM.swift
kmahar: good points. for now I'll add an `#if swift(..)` check here so we can `compactMap` in 4.1+ and easily switch to it once 4.0 is gone.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -228,14 +231,13 @@ internal struct _BSONDecodingStorage {
 
     /// Pops the top container from the stack. 
     fileprivate mutating func popContainer() {
-        precondition(self.containers.count > 0, "Empty container stack.")
+        precondition(!self.containers.isEmpty, "Empty container stack.")
```
jmikola: `empty_count` rule? Neat.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -6,31 +6,30 @@ public struct SessionOptions: Encodable {
 
 /// A session for ordering sequential operations.
 public class ClientSession: Encodable {
```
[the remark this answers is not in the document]
jmikola: Would it make sense for this to use `:nodoc:`? I saw an internal class described with that doc comment syntax above.

Sources/MongoSwift/BSON/BSONDecoder.swift
kmahar: yup! this does make me wish `isNotEmpty` or something existed though, as I think it is more readable than `!someCollection.isEmpty`, but it's been pitched various times ([1](https://forums.swift.org/t/isnotempty-on-array/15228), [2](https://forums.swift.org/t/isnonempty/16684)) and shot down by standard library authors. though of course, we could just add it ourselves by extending `Sequence`.

Sources/MongoSwift/ClientSession.swift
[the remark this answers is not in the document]
kmahar: that's a good idea, then it won't show up on our docs page either. adding.

PR #183, opened 2019-01-08
SWIFT-299 Introduce new errors and error types

patrickfreed:
[SWIFT-299](https://jira.mongodb.org/browse/SWIFT-299)

This PR introduces the new errors and error detail types that will eventually replace the current `MongoError` cases. It also introduces a function for converting from a `bson_error_t` to a Swift error type. All the errors conform to `Equatable` in ways that seem sensible to me; let me know if you feel otherwise. They all ignore error messages when checking for equality.

I defined an empty protocol `MongoSwiftError` that all of our errors conform to. I originally wanted to define another `Encoding/DecodingError` that would inherit from the standard library one and then implement our protocol, but their error is defined as an `enum`, so no inheritance is possible. We could potentially extend it and conform to our protocol, but that would cause any errors thrown by ANY decoder to be `MongoSwiftErrors`. However, I actually think it makes _more_ sense that we don't throw `MongoSwift` errors in our decoder/encoder. The ethos behind the `Codable` ecosystem is format agnostic, and it seems that the Swift team provides these errors to further that agnosticism. We would be throwing the same types of errors as e.g. `JSONEncoder` in the same contexts, and it leads to a very consistent structure for error handling across encoders. The only thing we need to be careful about is catching encoding/decoding errors that are thrown when we use the encoder/decoder internally, and translate them to `.internalError`s before we emit them to the users. They shouldn't try to catch a `WriteError` and instead get a `DecodingError` because the server returned a malformed response.

I have some questions regarding `WriteConcernError`s in bulk writes, so I've filed a SPECS [ticket](https://jira.mongodb.org/browse/SPEC-1199), the results of which may change the error type definitions a little.

The errors were designed to be as spec compliant as possible and are heavily based upon the [PHP driver's error implementation](http://php.net/manual/en/mongodb.exceptions.tree.php). I refrained from too much specificity in the error cases and tried to keep them as general and consistent as possible (e.g. no `.duplicateKeyError`, that's just contained in either a `WriteError` or a `BulkWriteError`)

The errors themselves are as follows:
- `ServerError`: Represents an error that occured on the server. Every error case here contains at  least one error code from the server.
    - `.commandError(code: Int, message: String, errorLabels: [String]?)`: 
        - Thrown when commands experience errors server side that prevent execution.
        - From the Write Commands Spec: "a command failure [...] indicates the command itself did not complete successfully. Example command failures include failure to authorize, failure to parse, operation aborted by the user, and unexpected errors during execution."
    - `.writeError(writeError: WriteError?, writeConcernError: WriteConcernError?, errorLabels: [String]?)`
        - Thrown when commands that write experience an error not as part of a Bulk Write.
        - Only one of the two optionals will be non-nil
    - `.bulkWriteError(writeErrors: [BulkWriteError]?, writeConcernError: WriteConcernError?, result: BulkWriteResult, errorLabels: [String]?)`
        - Thrown when errors are experienced as part of a Bulk Write
        - If WriteConcernError is populated, writeErrors may not be.
        - This ignores the result when checking for equality.

- `UserError`: Represents errors caused by the user doing something incorrectly with the driver
    - `.logicError(message: String)`
        - Thrown when the user uses the driver incorrectly (i.e. some bad method call).
    - `.invalidArgument(message: String)`
        - Thrown when user passes invalid arguments to some driver function. Should not be thrown when MongoSwift does so behind the scenes.

- `RuntimeError`: Represents errors that occur unexpectedly at runtime not necessarily due to user fault
    - `.internalError(message: String)`
        - Thrown when something is null when it shouldn't be, mongoc has an internal failure, or MongoSwift cannot understand a server response.
        - This is generally indicative of a bug somewhere in the mongod -> mongoc -> MongoSwift stack, or a system related failure (e.g. memory allocation failure)
        - This should be used instead of `preconditionFailure` or `fatalError` in places where the user could reasonably recover
    - `.connectionError(message: String, errorLabels: [String]?)`
        - Thrown during any connection establishment / socket related errors.
        - Also thrown when (MONGOC_ERROR_PROTOCOL/MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION).
            - PHP also throws a connection error in this case which is why I do it here. Should it have a different error?
    - `.authenticationError(message: String)`
        - any auth related failures (MONGOC_ERROR_CLIENT_AUTHENTICATE)
        - This is technically a command error, but mongoc strips away the server error and returns its own for these types of failures


Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,128 @@
...
+public protocol MongoSwiftError: Error {}
+
+/// The possible errors corresponding to types of errors encountered in the MongoDB server.
+/// These errors may contain labels providing additional information on their origin.
+/// - SeeAlso: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err
+public enum ServerError: MongoSwiftError, Equatable {
+    /// Thrown when commands experience errors on the server that prevent execution.
+    case commandError(code: Int, message: String, errorLabels: [String]?)
+
+    /// Thrown when errors occur on the server during commands that write not as part of a bulk write
```
kmahar: this wording is a little confusing... but off the top of my head I do not have a better way to put it... 🤔 maybe see how PHP/other drivers describe this case?
(and period at end of sentence please)

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,128 @@
...
+    /// Note: Only one of writeConcernError or writeError will populated at a time.
+    case writeError(writeError: WriteError?, writeConcernError: WriteConcernError?, errorLabels: [String]?)
+
+    /// Thrown when the server returns errors as part of an executed bulk write.
+    /// Note: writeErrors may not be present if the error experienced was a Write Concern related error.
```
kmahar:
```suggestion
    /// Note: `writeErrors` may not be present if the error experienced was a Write Concern related error.
```

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,128 @@
...
+
+/// The possible errors corresponding to types of errors encountered in the MongoDB server.
+/// These errors may contain labels providing additional information on their origin.
+/// - SeeAlso: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err
+public enum ServerError: MongoSwiftError, Equatable {
+    /// Thrown when commands experience errors on the server that prevent execution.
+    case commandError(code: Int, message: String, errorLabels: [String]?)
+
+    /// Thrown when errors occur on the server during commands that write not as part of a bulk write
+    /// Note: Only one of writeConcernError or writeError will populated at a time.
```
kmahar:
```suggestion
    /// Note: Only one of `writeConcernError` or `writeError` will populated at a time.
```

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,128 @@
...
+                  .connectionError(message: _, errorLabels: rhsLabels)):
+            return sortAndCompareOptionalArrays(lhs: lhsLabels, rhs: rhsLabels, cmp: { $0 < $1 })
+        default:
+            return false
+        }
+    }
+}
+
+/// Internal helper function used to get an appropriate error from a libmongoc error. This should NOT be used to get
+/// `.writeError`s or `.bulkWriteError`s.
```
kmahar: (maybe note how writeErrors and bulkWriteErrors should be gotten then)

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,128 @@
...
+internal func parseMongocError(error: bson_error_t, errorLabels: [String]? = nil) -> MongoSwiftError {
```
kmahar: where will the error labels come from when we are calling this?

Sources/MongoSwift/MongoError.swift
patrickfreed: PHP says:
>Base class for exceptions thrown by a failed write operation. The exception encapsulates a MongoDB\Driver\WriteResult object.

How does "Thrown when a single write command fails on the server." sound?

Sources/MongoSwift/MongoError.swift
patrickfreed: So, I'm not exactly sure yet as the only labels so far are transaction related. They come from mongoc functions that have a `bson_t *reply` parameter, and are returned in the `errorLabels` field on that document. I imagine for now we won't be passing anything into this field, but once we support transactions, we will pass in `reply["errorLabels"]` in places where we can.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -412,38 +412,67 @@ public struct BulkWriteResult {
...
     /// A description of the error.
     public let message: String
 
     private enum CodingKeys: String, CodingKey {
-        case index
         case code
         case message = "errmsg"
     }
+
+    public static func == (lhs: WriteError, rhs: WriteError) -> Bool {
```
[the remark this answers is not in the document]
patrickfreed: It's purely for testing. Moving to to an extension there.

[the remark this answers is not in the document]
patrickfreed: Also, see Jeremy's [comment ](https://jira.mongodb.org/browse/SWIFT-145?focusedCommentId=1995518&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1995518) for the motivation.

Sources/MongoSwift/MongoError.swift
patrickfreed: Done.

Sources/MongoSwift/MongoError.swift
kmahar: sounds good to me.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -412,38 +412,67 @@ public struct BulkWriteResult {
...
-    ///  A description of the error.
+    public static func == (lhs: WriteConcernError, rhs: WriteConcernError) -> Bool {
+        return lhs.code == rhs.code && lhs.details == rhs.details
+    }
+}
+
+/// A struct to represent a write error resulting from an executed bulk write.
+public struct BulkWriteError: Codable, Equatable {
+    /// An integer value identifying the error.
+    public let code: Int
```
kmahar: `code` here (and in other existing error usages) refers to libmongoc error code, right? 

while the upkeep may might be tedious, we should consider whether it's worth replicating all the libmongoc error codes in the driver as an enum with raw type `Int`, so that users don't have to go look at libmongoc docs to figure out what the codes corresponds to. 
alternatively, if we don't do that, we need to at least point users to the documentation for libmongoc error codes. I'm not sure what the best way to do that would be. it seems maybe less than desirable to just copy-paste the link every place we use an error code in a struct or enum case. we might consider introducing `public typealias ErrorCode = Int`, using that throughout the codebase, and just linking to it at the location where we define the `typealias`.
what do PHP and C++ do with respect to libmongoc error codes? 

regardless of what direction we go on that, I think it would be great if we add a ticket to the epic for writing a guide for our docs on error handling. (you can see we currently have one guide page [ in the docs](https://mongodb.github.io/mongo-swift-driver/development.html) already, the development guide.) this can explain the overall hierarchy and the types of errors to look out for, link to relevant pages/docs, and show some examples of best practices for catching and handling the errors.

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,76 @@
...
+
+/// The possible errors corresponding to types of errors encountered in the MongoDB server.
+/// These errors may contain labels providing additional information on their origin.
+/// - SeeAlso: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err
+public enum ServerError: MongoSwiftError {
+    /// Thrown when commands experience errors on the server that prevent execution.
+    case commandError(code: Int, message: String, errorLabels: [String]?)
+
+    /// Thrown when errors occur on the server during commands that write not as part of a bulk write
+    /// Note: Only one of `writeConcernError` or `writeError` will populated at a time.
```
mbroadst: "...will *be* populated..."

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,76 @@
...
+                        errorLabels: [String]?)
+}
+
+/// The possible errors caused by improper use of the driver by the user.
+public enum UserError: MongoSwiftError {
+    /// Thrown when the driver is incorrectly used.
+    case logicError(message: String)
+
+    /// Thrown when the user passes in invalid arguments to a driver method.
+    case invalidArgument(message: String)
```
mbroadst: nit: this is the only error that doesn't end in `Error`

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -58,6 +58,64 @@ class MongoSwiftTestCase: XCTestCase {
     }
 }
 
+extension RuntimeError: Equatable {
```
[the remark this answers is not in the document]
mbroadst: +1

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -58,6 +58,64 @@ class MongoSwiftTestCase: XCTestCase {
...
+    public static func == (lhs: RuntimeError, rhs: RuntimeError) -> Bool {
+        switch (lhs, rhs) {
+        case (.internalError(message: _), .internalError(message: _)),
```
mbroadst: Does this just equate them if they generically match `internalError`, regardless of message and/or code?

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,134 @@
...
+    case connectionError(message: String, errorLabels: [String]?)
+
+    /// Thrown when encountering an authentication related error (e.g. invalid credentials).
+    case authenticationError(message: String)
+}
+
+/// A struct to represent a single write error not resulting from an executed bulk write.
+public struct WriteError: Codable {
+    /// An integer value identifying the error.
+    public let code: Int
```
kmahar: should this also be `ServerErrorCode`?

patrickfreed: Yep, fixed.

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,137 @@
...
+                code: ServerErrorCode(code.rawValue),
+                message: message,
+                errorLabels: errorLabels)
+    case (MONGOC_ERROR_STREAM, _):
+        return RuntimeError.connectionError(message: message, errorLabels: errorLabels)
+    case (MONGOC_ERROR_SERVER_SELECTION, MONGOC_ERROR_SERVER_SELECTION_FAILURE):
+        return RuntimeError.connectionError(message: message, errorLabels: errorLabels)
+    case (MONGOC_ERROR_PROTOCOL, MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION):
+        return RuntimeError.connectionError(message: message, errorLabels: errorLabels)
+    default:
```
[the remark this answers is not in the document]
patrickfreed: I've added the assert. I'd rather not expand the labels to the other types (non-server, non-connection) if they will probably always be nil. From my reading of the transactions spec, we should be relatively safe to keep them only on what we have so far, but I'm not sure how subject that is to change. Do you think there is a reasonable risk that error labels could become more pervasive?

patrickfreed:
While working on some other tickets using these new errors, I discovered that it would be worthwhile to conform all our `MongoSwiftError`s to `LocalizedError` for getting the messages out without switching on various cases. This is also consistent with the legacy `MongoError`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -148,19 +148,25 @@ public class MongoClient {
...
      *
      * - Parameters:
      *   - fromPointer: the `mongoc_client_t` to store and use internally
      */
     public init(fromPointer: OpaquePointer) {
         self._client = fromPointer
 
+        // This call may fail, and if it does, either the error api version was already set or the client was derived
+        // from a pool. In either case, the error handling in MongoSwift will be incorrect unless the correct api
+        // version was set by the caller.
```
kmahar: since this initializer is really only for use in MongoMobile, it might be fine for us to assume that whatever is going on there in terms of setting the error API version (or not) will be the case here. 
I don't think there would ever be a pooled client coming from there. I'm not sure if mongoc_embedded is doing anything in terms of setting the error API when we call `mongoc_embedded_v1_client_create` ([here](https://github.com/mongodb/swift-mongo-mobile/blob/64af359f205860fcdba547c698352412f786eb13/Sources/MongoMobile/MongoMobile.swift#L170)) but I would guess not? maybe worth checking.

if it's not being set there, then we might want to handle the error as we do in the other initializer if there are any other possible reasons this could throw.

side note: we should probably mark this initializer with :nodoc: since it's not meant for consumption by anyone but us.

patrickfreed: marked the initializer with `:nodoc:`. 

After offline discussion, we decided it wasn't worth marking this as `throws` due to the subsequent necessary MongoMobile changes, and since this is only called in MongoMobile, we should know to not be setting the error api version / appropriately handling a pool. So, it is ok that we allow it to fail silently and simply leave a warning comment.

patrickfreed:
@mbroadst @jmikola The other work for this epic is getting a little backed up on this being merged, can you take a final pass on it whenever you get the chance?

Sources/MongoSwift/MongoError.swift
```diff
@@ -1,6 +1,137 @@
...
+    public let code: ServerErrorCode
+
+    /// A description of the error.
+    public let message: String
+
+    /// The index of the request that errored.
+    public let index: Int
+
+    /// The request that errored.
+    public let request: WriteModel? = nil
```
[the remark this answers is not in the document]
jmikola: I recognize it from the CRUD spec. No objection having this here, as it saves the user the trouble of tracking it down from the index.

Sources/MongoSwift/MongoError.swift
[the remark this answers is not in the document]
jmikola: I think connection and server errors are sufficient. I'm hard-pressed to think of a case where an auth error might occur during a transaction (perhaps on a new socket being used for the first time), but even then I don't think an error label would be relevant since the driver would not have accomplished anything with the socket.

This looks fine as-is.

Tests/MongoSwiftTests/MongoError+Equatable.swift
```diff
@@ -0,0 +1,87 @@
...
+
+extension WriteError: Equatable {
+    public static func == (lhs: WriteError, rhs: WriteError) -> Bool {
+        return lhs.code == rhs.code
+    }
+}
+
+extension BulkWriteError: Equatable {
+    public static func == (lhs: BulkWriteError, rhs: BulkWriteError) -> Bool {
+        return lhs.code == rhs.code && lhs.index == rhs.index
```
jmikola: Is it worth checking the index here? Technically, this might consider two BulkWriteErrors equivalent because they have the same code and occurred on the same index within different `bulkWrite()` calls -- nothing about index ties the error back to a specific `bulkWrite()` operation, so that would make the equivalence check a bit coincidental.

patrickfreed: I agree, but this equality operator will only ever be used within our tests and as part of comparing two `.bulkWriteErrors` derived from the same request (note this is defined in `MongoError+Equatable.swift`, which is in our Testing folder). If this were part of the public API then I agree that we should omit the index, (or only conform to `Equatable` at the enum level), but for testing purposes, it is convenient to leave it in.

jmikola: Ah, I missed that this was in the test suite. SGTM!

PR #185, opened 2019-01-09
Download portable SwiftLint binary on OSX builds instead of using Homebrew

patrickfreed:
I have it set to download 0.29.3 which is the current latest one. We can manually bump as necessary.

Will eventually cache this in later work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ install:
 before_script:
   - mkdir ${PWD}/mongodb-${MONGODB_VERSION}/data
   - ${PWD}/mongodb-${MONGODB_VERSION}/bin/mongod --dbpath ${PWD}/mongodb-${MONGODB_VERSION}/data --logpath ${PWD}/mongodb-${MONGODB_VERSION}/mongodb.log --enableMajorityReadConcern --fork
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,8 +47,8 @@ install:
 before_script:
   - mkdir ${PWD}/mongodb-${MONGODB_VERSION}/data
   - ${PWD}/mongodb-${MONGODB_VERSION}/bin/mongod --dbpath ${PWD}/mongodb-${MONGODB_VERSION}/data --logpath ${PWD}/mongodb-${MONGODB_VERSION}/mongodb.log --enableMajorityReadConcern --fork
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew upgrade swiftlint; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then swiftlint --strict; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then mkdir ${PWD}/swiftlint; curl -L https://github.com/realm/SwiftLint/releases/download/0.29.3/portable_swiftlint.zip -o ${PWD}/swiftlint/swiftlint.zip; unzip ${PWD}/swiftlint/swiftlint.zip -d ${PWD}/swiftlint; fi
```
mbroadst: Can we break this up into a more readable line? Additionally, it might be worth considering consolidating all of our build/testing requirements into a single bash script similar to [Realm](https://github.com/realm/realm-cocoa/blob/master/build.sh). I had intended for our `Makefile` to serve this purpose, I think the travis config is getting a little large.

kmahar: I came to same conclusion working on code coverage and started doing so there :)

PR #188, opened 2019-01-10
SWIFT-289 Audit public types to make sure they have public initializers

patrickfreed:
[SWIFT-289](https://jira.mongodb.org/browse/SWIFT-289)

This PR adds public initializers to some public types that don't have them. It also exposes `UUID.init(from binary: Binary)`.

It seems the assumption for the other types that do have public initializers is that the default ACL was public, since the docstrings on all those initializers say they're for convenience. 
e.g.
```
/// Convenience initializer allowing any/all parameters to be optional
```
In reality, these initializers are actually very necessary and are not merely conveniences. The docstrings I added are consistently worded with the old ones but omit "convenience". Let me know if you think I should change the older docstrings and/or the newer ones.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -641,11 +641,10 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     }
 }
 
-/// Extension to allow a UUID to be initialized from a Binary BSONValue.
+/// Extension to allow a `UUID` to be initialized from a `Binary` `BSONValue`.
 extension UUID {
-    // TODO: fill the rest of this out for full BSONValue conformance (SWIFT-295).
-
-    internal init(from binary: Binary) throws {
+    /// Initializes a `UUID` instance from a `Binary` `BSONValue`.
+    public init(from binary: Binary) throws {
         guard binary.subtype != Binary.Subtype.uuidDeprecated.rawValue else {
             throw MongoError.bsonDecodeError(message: "Binary subtype \(binary.subtype) is deprecated, " +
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -641,11 +641,10 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
 }
 
-/// Extension to allow a UUID to be initialized from a Binary BSONValue.
+/// Extension to allow a `UUID` to be initialized from a `Binary` `BSONValue`.
 extension UUID {
-    // TODO: fill the rest of this out for full BSONValue conformance (SWIFT-295).
-
-    internal init(from binary: Binary) throws {
+    /// Initializes a `UUID` instance from a `Binary` `BSONValue`.
+    public init(from binary: Binary) throws {
```
kmahar: lol, good thing we did this ticket before we released UUID stuff....

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -79,7 +79,7 @@ extension MongoCollection {
      *
      * - Returns: The count of the documents that matched the filter
      */
-    public func countDocuments(_ filter: Document = [:], options: CountDocumentsOptions? = nil) throws -> Int {
+    private func countDocuments(_ filter: Document = [:], options: CountDocumentsOptions? = nil) throws -> Int {
```
mbroadst: why were these made private?

patrickfreed: they aren't actually implemented, so we figured it was better to hide them rather than have the user find out at runtime via an error.

PR #190, opened 2019-01-11
SWIFT-255 Conform AnyBSONValue to Hashable

patrickfreed:
[SWIFT-255](https://jira.mongodb.org/browse/SWIFT-255)

This PR adds `Hashable` conformance to `AnyBSONValue` by computing the hashValue of the extended JSON representation of the underlying `BSONValue`. 

I originally tried using the `hashValue` of the wrapped `BSONValue` if it was `Hashable` itself, but I was experiencing some collisions (`hashValue` is only likely to be unique among the same types I guess). Instead, I just resorted to using the extended JSON `hashValue` for everything. It may a bit slower as it has to create a temporary document which is then transformed to JSON, but at least it avoids collisions. There is an argument to be made that tolerating collisions is more performant than converting everything to their extended JSON formats, but I'm not really sure if that is the case or of a good way to quantify that.

We do avoid a copy in the case of the underlying value being a `Document`.

### DIFF
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -5,19 +5,7 @@ import Foundation
 public struct AnyBSONValue: Codable, Equatable, Hashable {
     // swiftlint:disable:next legacy_hashing
     public var hashValue: Int {
-        if let bool = self.value as? Bool {
-            return bool.hashValue
-        } else if let int = self.value as? Int {
-            return int.hashValue
-        } else if let int32 = self.value as? Int32 {
-            return int32.hashValue
-        } else if let int64 = self.value as? Int64 {
-            return int64.hashValue
-        } else if let double = self.value as? Double {
-            return double.hashValue
-        } else if let date = self.value as? Date {
-            return date.hashValue
-        } else if let doc = self.value as? Document {
+        if let doc = self.value as? Document {
             return doc.extendedJSON.hashValue
         } else {
             let doc: Document = ["value": self.value]
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -21,7 +21,8 @@ public struct AnyBSONValue: Codable, Equatable, Hashable {
             return doc.extendedJSON.hashValue
         } else {
             let doc: Document = ["value": self.value]
-            return doc.extendedJSON.hashValue
+            // need to add some string to the beginning to ensure no collisions with the document case.
+            return ("EXT_JSON" + doc.extendedJSON).hashValue
         }
     }
 
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -5,9 +5,7 @@ import Foundation
 public struct AnyBSONValue: Codable, Equatable, Hashable {
     // swiftlint:disable:next legacy_hashing
     public var hashValue: Int {
-        if let value = self.value as? String {
-            return value.hashValue
-        } else if let bool = self.value as? Bool {
+        if let bool = self.value as? Bool {
             return bool.hashValue
         } else if let int = self.value as? Int {
             return int.hashValue
@@ -17,6 +15,8 @@ public struct AnyBSONValue: Codable, Equatable, Hashable {
             return int64.hashValue
         } else if let double = self.value as? Double {
             return double.hashValue
+        } else if let date = self.value as? Date {
+            return date.hashValue
         } else if let doc = self.value as? Document {
             return doc.extendedJSON.hashValue
         } else {
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -5,6 +5,13 @@ import Foundation
 public struct AnyBSONValue: Codable, Equatable, Hashable {
     // swiftlint:disable:next legacy_hashing
     public var hashValue: Int {
+        if let date = self.value as? Date {
+            return "\(self.value.bsonType)-\(date.hashValue)".hashValue
+        } else if let binary = self.value as? Binary {
+            return "\(self.value.bsonType)-\(binary.data.hashValue)-\(binary.subtype)".hashValue
+        } else if let arr = self.value as? [BSONValue] {
+            return "\(self.value.bsonType)-\((["value": arr] as Document).extendedJSON)".hashValue
+        }
         return "\(self.value.bsonType)-\(self.value)".hashValue
     }
 
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -2,8 +2,30 @@ import Foundation
 
 /// A struct wrapping a `BSONValue` type that allows for encoding/
 /// decoding `BSONValue`s of unknown type.  
-public struct AnyBSONValue: Codable, Equatable {
-    /// The `BSONValue` wrapped by this struct. 
+public struct AnyBSONValue: Codable, Equatable, Hashable {
+    // swiftlint:disable:next legacy_hashing
+    public var hashValue: Int {
+        if let value = self.value as? String {
+            return value.hashValue
+        } else if let bool = self.value as? Bool {
+            return bool.hashValue
+        } else if let int = self.value as? Int {
+            return int.hashValue
+        } else if let int32 = self.value as? Int32 {
+            return int32.hashValue
+        } else if let int64 = self.value as? Int64 {
+            return int64.hashValue
+        } else if let double = self.value as? Double {
+            return double.hashValue
+        } else if let doc = self.value as? Document {
+            return doc.extendedJSON.hashValue
+        } else {
+            let doc: Document = ["value": self.value]
+            return doc.extendedJSON.hashValue
+        }
+    }
+
+    /// The `BSONValue` wrapped by this struct.
     public let value: BSONValue
 
     /// Initializes a new `AnyBSONValue` wrapping the provided `BSONValue`.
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -2,8 +2,19 @@ import Foundation
...
-public struct AnyBSONValue: Codable, Equatable {
-    /// The `BSONValue` wrapped by this struct. 
+public struct AnyBSONValue: Codable, Equatable, Hashable {
+    // swiftlint:disable:next legacy_hashing
+    public var hashValue: Int {
+        if let doc = self.value as? Document {
+            return doc.extendedJSON.hashValue
+        } else {
+            let doc: Document = ["value": self.value]
+            // need to add some string to the beginning to ensure no collisions with the document case.
```
kmahar: am I understanding correctly that the problem here is that the hash value for `AnyBSONValue(["value": 1] as Document)` would match the hash value for `AnyBSONValue(1)`?

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -152,4 +153,64 @@ final class BSONValueTests: MongoSwiftTestCase {
...
+                ts: Timestamp(timestamp: 1, inc: 2),
+                int32: 5,
+                int64: 6,
+                dec: Decimal128("1.2E+10"),
+                minkey: MinKey(),
+                maxkey: MaxKey(),
+                regex: RegularExpression(pattern: "^abc", options: "imx"))
+
+        let values = Mirror(reflecting: expected).children.map { child in AnyBSONValue(child.value as! BSONValue) }
+        let set = Set<AnyBSONValue>(values)
```
kmahar: let's use a different variable name here because `set` is already a special keyword in swift... for ex. here: https://github.com/mongodb/mongo-swift-driver/blob/b924785a4f0abc8bc4b9d60f8397c52495af0050/Sources/MongoSwift/BSON/Document.swift#L338

while it seems to work ok here, there could be some kind of unintended consequences, and it makes the syntax highlighting funky.

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -152,4 +153,64 @@ final class BSONValueTests: MongoSwiftTestCase {
...
+                int: 1,
+                ts: Timestamp(timestamp: 1, inc: 2),
+                int32: 5,
+                int64: 6,
+                dec: Decimal128("1.2E+10"),
+                minkey: MinKey(),
+                maxkey: MaxKey(),
+                regex: RegularExpression(pattern: "^abc", options: "imx"))
+
+        let values = Mirror(reflecting: expected).children.map { child in AnyBSONValue(child.value as! BSONValue) }
```
kmahar: is there a reason to use mirror here rather than just making `expected` an `[BSONValue]` to begin with?

Sources/MongoSwift/BSON/AnyBSONValue.swift
patrickfreed: Yep. Since `<any string> + <valid extended JSON>` will never be valid extended JSON, it helps avoid collisions.

Tests/MongoSwiftTests/BSONValueTests.swift
patrickfreed: I switched it to this after looking at Jason's tests, actually.

Using this struct guarantees at the compile time that we cover all `BSONValue` cases in this test. So if we add a new `BSONValue` in the future (and update `AllBSONTypes` accordingly), it will cause this test to stop compiling unless we also update it to use the new type. Using just a list wouldn't cause any errors if a new `BSONValue` were introduced, so we'd likely forget to update this test.

Tests/MongoSwiftTests/BSONValueTests.swift
patrickfreed: fixed

Tests/MongoSwiftTests/BSONValueTests.swift
kmahar: that makes sense, though I wonder if we will remember to update that struct... any ideas for enforcing that?

patrickfreed: We could include a comment near the definition of the `BSONValue` protocol instructing us to do so. 
Alternatively, we could include a test that does some mirroring stuff that verifies the number of children in `AllBSONTypes` is equivalent to the number of cases in `BSONType`.

kmahar:
can't seem to add as a reviewer but cc @jsflax

Tests/MongoSwiftTests/BSONValueTests.swift
patrickfreed: Filed [SWIFT-324](https://jira.mongodb.org/browse/SWIFT-324)

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -2,8 +2,19 @@ import Foundation
...
+            return ("EXT_JSON" + doc.extendedJSON).hashValue
```
jsflax: This seems like it would have a lot of overhead– why not use the naturally primitive hash values?

patrickfreed: It's the only way to guarantee that we minimize hash collisions. In some older versions of swift, using the primitive hash values led to some collisions. (see: https://travis-ci.org/mongodb/mongo-swift-driver/builds/478423525)

Whether or not this overhead is better than a potentially higher rate of hash collisions is unclear.

jsflax: I could be mistaken, but I thought the collisions were intentional so that naturally similar values were represented the same, e.g., `false` would have the same hash as `0`.

patrickfreed: I doubt hash collisions are ever intentional as they incur a performance hit, and hashmaps/dictionaries are generally unordered, so the clustering isn't particularly useful.

e.g. in 4.2 `false.hashValue` doesn't have the same hash as `0` (neither of them hash to 0 as they did in 4.0).

jsflax: I'm not sure managing that error (which is likely an artifact from objc days) warrants the overhead that will come along with having to convert to extended JSON to calculate a hash. Seems like most of the BSONValue classes/structs have simple hashes that wouldn't require calculation. I'm also not sure if there's a use case to use the hash of something like a bool.

Could we at least lazily convert to extendedJSON? Just seems like this is going to get expensive, fast.

mbroadst: @jsflax are you positing that `("EXT_JSON" + doc.extendedJSON).hashValue` has significantly more overhead than `doc.extendedJSON.hashValue` ?

patrickfreed: The following is much faster (for non documents) and avoids most of the collision issues:
```
public var hashValue: Int {
        let prefix = String(UnicodeScalar(UInt8(self.value.bsonType.rawValue)))

        if let document = self.value as? Document {
            return "\(prefix)-\(document.extendedJSON)".hashValue
        } else {
            return "\(prefix)-\(self.value)".hashValue
        }
    }
```
I'm currently running into some collision issues specifically on Linux with Swift 4.0/4.1 that I'm trying to hammer out, but hopefully something along these lines will work. 

It's still converting the `BSONValue` to a string, but it is omitting the document creation and having to convert to extended JSON, for the most part.

patrickfreed: As it turns out, the Swift 4.0/4.1 `hashValue` for `String` on Linux seems to ignore non-ASCII bytes (e.g. the bson type I was prepending). I've opted for just prepending the ASCII representation of the bson type then.

This simple function minimizes collisions and is generally fast AFAIK (simple string conversion, no document creation). If performance is still a concern here, let me know @jsflax

jsflax: @mbroadst No, I'm positing that that converting to extended JSON would have a lot more overhead than, say, having an ObjectID and using the `.oid` hashValue. Or a string use it's own hashValue, and so on

jsflax: @patrickfreed That LGTM. Hope this didn't come off as an over optimization (tough to know without user feedback). Just seemed like there were simpler options

patrickfreed: Not at all, it was a concern I had too. The current method seems to be around ~2x as fast, so the change seems worthwhile.

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -2,8 +2,20 @@ import Foundation
 
 /// A struct wrapping a `BSONValue` type that allows for encoding/
 /// decoding `BSONValue`s of unknown type.  
-public struct AnyBSONValue: Codable, Equatable {
-    /// The `BSONValue` wrapped by this struct. 
+public struct AnyBSONValue: Codable, Equatable, Hashable {
+    // swiftlint:disable:next legacy_hashing
+    public var hashValue: Int {
+        if let date = self.value as? Date {
```
kmahar: can we add comments explaining why these special cases are necessary?

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -2,8 +2,20 @@ import Foundation
 
 /// A struct wrapping a `BSONValue` type that allows for encoding/
 /// decoding `BSONValue`s of unknown type.  
-public struct AnyBSONValue: Codable, Equatable {
-    /// The `BSONValue` wrapped by this struct. 
+public struct AnyBSONValue: Codable, Equatable, Hashable {
+    // swiftlint:disable:next legacy_hashing
+    public var hashValue: Int {
```
kmahar: let's add a todo linking to the ticket for making all underlying BSON types Hashable

Sources/MongoSwift/BSON/AnyBSONValue.swift
patrickfreed: Done.

Sources/MongoSwift/BSON/AnyBSONValue.swift
patrickfreed: Done.

PR #192, opened 2019-01-11
SWIFT-316 Add a .dataEncodingStrategy and a .dataDecodingStrategy to support storing Data in BSON binary format

patrickfreed:
[SWIFT-316](https://jira.mongodb.org/browse/SWIFT-316)

This PR adds the `.deferredToData` and `.binary` decoding/encoding strategies to `BSONEncoder`. 

I decided upon using `.binary` as the default because, at least in my reading of the spec, that is the type specified for this exact purpose. Also, the waste concerns mentioned in the original ticket are worth avoiding by default, I think.

I also think that `Data` is a good candidate for the `.custom` encoding/decoding strategy, since users may be aware of some structure of their buffers and want to encode it in their own way. Messing around with buffers like that wouldn't be the most Swift-y way to do things, but I feel like it may be useful to some users, especially if they're interfacing with their own C libraries. 

I also think there may be some value in providing an `.integerArray` encoding/decoding strategy. It would be the same as `.deferredToData`, except that instead of encoding/decoding each byte to/from a 4 byte integer, it would instead encode/decode groups of 4 bytes to 4 byte integers. It would attempt to adhere to the default implementation but without the waste.

I haven't implemented either `.custom` or `.integerArray` yet, as I wanted to get feedback first. Do you all think either are worthwhile?

### DIFF
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -62,6 +62,11 @@ public class BSONDecoder {
     /// - SeeAlso: bsonspec.org
     public enum DataDecodingStrategy {
         /// Decode `Data`s by deferring to their default decoding implementation.
+        ///
+        /// Note: The default decoding implementation attempts to decode the `Data` from a `[UInt8]`, but because BSON
+        /// does not support integer types other `Int32` and `Int64`, it actually decodes from an `[Int32]` stored
+        /// in BSON. This strategy paired with its corresponding encoding strategy results in an inefficient storage of
+        /// the `Data` in BSON.
         case deferredToData
 
         /// Decode `Data`s stored as the BSON `Binary` type (default).
@@ -361,15 +366,18 @@ extension _BSONDecoder {
 
     fileprivate func unbox<T: Decodable>(_ value: BSONValue, as type: T.Type) throws -> T {
         // swiftlint:disable force_cast
-        if type == Date.self {
+        switch type {
+        case is Date.Type:
             // We know T is a Date and unboxDate returns a Date or throws, so this cast will always work.
             return try unboxDate(value) as! T
-        } else if type == UUID.self {
+        case is UUID.Type:
             // We know T is a UUID and unboxUUID returns a UUID or throws, so this cast will always work.
             return try unboxUUID(value) as! T
-        } else if type == Data.self {
+        case is Data.Type:
             // We know T is a Data and unboxData returns a Data or throws, so this cast will always work.
             return try unboxData(value) as! T
+        default:
+            break
         }
         // swiftlint:enable force_cast
 
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -54,6 +54,20 @@ public class BSONDecoder {
         case binary
     }
 
+    /// Enum representing the various strategies for decoding `Data`s from BSON.
+    ///
+    /// As per the BSON specification, the default strategy is to decode `Data`s from BSON binary types with the generic
+    /// binary subtype.
+    ///
+    /// - SeeAlso: bsonspec.org
+    public enum DataDecodingStrategy {
+        /// Decode `Data`s by deferring to their default decoding implementation.
+        case deferredToData
+
+        /// Decode `Data`s stored as the BSON `Binary` type (default).
+        case binary
+    }
+
     /// Contextual user-provided information for use during decoding.
     public var userInfo: [CodingUserInfoKey: Any] = [:]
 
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -335,6 +367,9 @@ extension _BSONDecoder {
         } else if type == UUID.self {
             // We know T is a UUID and unboxUUID returns a UUID or throws, so this cast will always work.
             return try unboxUUID(value) as! T
+        } else if type == Data.self {
```
kmahar: now that we are checking three different possible values of `type` here, would it make sense to rewrite this as a `switch`?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -381,12 +404,24 @@ extension _BSONEncoder {
...
+        }
+    }
+
     /// Returns the value as a `BSONValue` if possible. Otherwise, returns nil.
     fileprivate func box_<T: Encodable>(_ value: T) throws -> BSONValue? {
         if let date = value as? Date {
             return try boxDate(date)
         } else if let uuid = value as? UUID {
             return try boxUUID(uuid)
+        } else if let data = value as? Data {
```
kmahar: same comment on switching, you could write this like
```swift
switch value {
case let date as Date:
    return try boxDate(date)
case let uuid as UUID:
   ...
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -47,12 +47,32 @@ public class BSONEncoder {
...
+    ///
+    /// As per the BSON specification, the default strategy is to encode `Data`s as BSON binary types with the generic
+    /// binary subtype.
+    ///
+    /// - SeeAlso: bsonspec.org
+    public enum DataEncodingStrategy {
+        /// Encode the `Data` by deferring to its default encoding implementation.
+        ///
+        /// Note: The default encoding implementation attempts to encode the `Data` as a [UInt8], but because BSON does
+        /// not support integer types besides `Int32` or `Int64`, it actually gets encoded to BSON as an `[Int]`.
```
kmahar:
```suggestion
        /// not support integer types besides `Int32` or `Int64`, it actually gets encoded to BSON as an `[Int32]`.
```
nit, it's really a 32-bit int - we just return the value as an `Int` when you retrieve it from a doc since those are more widely used in BSON.

maybe worth also adding a sentence explaining why that is undesirable (space inefficient).

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -47,12 +47,32 @@ public class BSONEncoder {
...
+    /// Enum representing the various strategies for encoding `Data`s.
+    ///
+    /// As per the BSON specification, the default strategy is to encode `Data`s as BSON binary types with the generic
+    /// binary subtype.
+    ///
+    /// - SeeAlso: bsonspec.org
+    public enum DataEncodingStrategy {
+        /// Encode the `Data` by deferring to its default encoding implementation.
+        ///
+        /// Note: The default encoding implementation attempts to encode the `Data` as a [UInt8], but because BSON does
```
kmahar:
```suggestion
        /// Note: The default encoding implementation attempts to encode the `Data` as a `[UInt8]`, but because BSON does
```

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -54,6 +54,20 @@ public class BSONDecoder {
         case binary
     }
 
+    /// Enum representing the various strategies for decoding `Data`s from BSON.
+    ///
+    /// As per the BSON specification, the default strategy is to decode `Data`s from BSON binary types with the generic
+    /// binary subtype.
+    ///
+    /// - SeeAlso: bsonspec.org
+    public enum DataDecodingStrategy {
```
kmahar: maybe worth adding a somewhat symmetric comment in here to what you wrote in the encoding strategy about what `Data`'s default does.

kmahar:
@patrickfreed, oops, I read that but forgot to reply to it in my review.

so, I realized that JSON actually has data [encoding](https://github.com/apple/swift/blob/0c59faf8e311e6d6e4d2a86894a75a3356d3b63c/stdlib/public/Darwin/Foundation/JSONEncoder.swift#L100) and [decoding](https://github.com/apple/swift/blob/0c59faf8e311e6d6e4d2a86894a75a3356d3b63c/stdlib/public/Darwin/Foundation/JSONEncoder.swift#L1033) strategies as well.

the way I see it, the point in providing these strategies in general across all types is to do one or both of:
* store the data in a more efficient/recommended manner 
* make the data better interchangeable with where it came from/other places it may be used

so I'm definitely in favor of adding a custom strategy - it matches JSON, and as you said might allow users to pack their data in more efficiently or make it work with the format they're getting data from an API in or something.

I'm not as convinced that `.integerArray` makes sense, because it wouldn't provide any storage optimization over just using `.binary`, and it's not some sort of standard encoding format that is used elsewhere and would make interchange easier. 

JSON defaults to a strategy of encoding binary data as a base64 string. we could consider replicating that too.

patrickfreed:
@kmahar 
Sounds good, I agree on all points. I'll add `.custom`, and `.base64` for `JSONEncoder` parity.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -329,6 +360,29 @@ extension _BSONEncoder {
         return try self.box_(value) ?? Document()
     }
 
+    fileprivate func handleCustomStrategy<T: Encodable>(
```
kmahar: good idea pulling this out into a separate method! it seems like Apple just copy-pasted this block instead.... lol.

my only comment is that I wonder if we can make the parameter labels a little clearer so it reads better at the call site. I see why you named the first parameter `encode` since it's basically an alternate `encode(to:)` method, but from where we use this I'm not sure that's obvious and it kinda reads more like we're encoding `f`. at the least I think we should change that label to something clearer like `encodeFunc`, or maybe do something like `handleCustomStrategy(f, forValue: data)`? or `handleCustom(strategy: f, forValue: data)`? up to you.

patrickfreed: I like `encodeFunc` and `forValue`; they seem much clearer than what I have. Changing it to them 👍

PR #194, opened 2019-01-17
SWIFT-280, SWIFT-334: Improve handling of documents missing _id in insertOne and insertMany

kmahar:
This makes it so we don't modify the input document(s) in `insertOne` and `insertMany` when they are missing an `_id`. 

Instead, if a doc is missing an `_id`, we make a new document with an `_id` at the start, copy over the data from the original doc, and send the new doc to the server.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -228,8 +228,8 @@ extension Document {
         }
     }
 
-    /// If the document already has an "_id" field, returns it as-is. Otherwise, returns a new document
-    /// containing all the keys from this document, with an _id field prepended.
+    /// If the document already has an _id, returns it as-is. Otherwise, returns a new document
+    /// containing all the keys from this document, with an _id prepended.
     internal func withID() throws -> Document {
         if self.hasKey("_id") {
             return self
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -227,6 +227,18 @@ extension Document {
             self.storage = DocumentStorage(fromPointer: self.data)
         }
     }
+
+    /// If the document already has an "_id" field, returns it as-is. Otherwise, returns a new document
+    /// containing all the keys from this document, with an _id field prepended.
+    internal func withID() throws -> Document {
+        if self.hasKey("_id") {
+            return self
+        }
+
+        var idDoc: Document = ["_id": ObjectId()]
+        try idDoc.merge(self)
+        return idDoc
+    }
 }
 
 /// An extension of `Document` containing its public API.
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -227,6 +227,18 @@ extension Document {
...
     }
+
+    /// If the document already has an _id, returns it as-is. Otherwise, returns a new document
+    /// containing all the keys from this document, with an _id prepended.
+    internal func withID() throws -> Document {
+        if self.hasKey("_id") {
+            return self
+        }
+
+        var idDoc: Document = ["_id": ObjectId()]
```
patrickfreed: Would it be worth it performance-wise to add functionality (internally) for prepending to documents without constructing an entirely new one?

kmahar: so, we might want something like that is when `CollectionType != Document`. 
in those cases, we create a new document when we call `BSONEncoder.encode()`, and then we immediately create a second document here to prepend the `_id`. (when `CollectionType == Document`, the encoder no-ops and returns the same document you passed in. then we have to make a new document anyway in order to avoid modifying user input.) 

however, I think it might actually make more sense to implement that with an internal variant of `encode` that allows you to provide an `inout` starter doc into which we encode the struct. that would also make our [encoding for index creation](https://github.com/mongodb/mongo-swift-driver/blob/fd95efb2d7e6360d706ec227c2c54977eaa049d9/Sources/MongoSwift/MongoCollection%2BIndexes.swift#L221) a bit cleaner - libmongoc expects everything to be in a single struct so we have to make two encode calls and then merge the resulting docs. 

did you have any other ideas in mind for where this would be useful?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -93,7 +93,7 @@ extension MongoCollection {
 
     /// A model for an `insertOne` operation within a bulk write.
     public struct InsertOneModel: WriteModel {
-        private let document: CollectionType
+        internal let document: CollectionType
```
kmahar: this is to allow me to test that the document didn't change.

Sources/MongoSwift/BSON/Document.swift
patrickfreed: Ah, right. So no gain in the `Document` case. 

And the index encoding is the only other place I can find as well where this type of thing would be useful. I don't think we really encode stuff piecemeal and combine it in any other places. Anyways, for this ticket I think were fine as-is can can re-evaluate later (filed [SWIFT-336](https://jira.mongodb.org/browse/SWIFT-336)).

kmahar: sounds good,thanks for filing that!

PR #195, opened 2019-01-18
SWIFT-261 Replace preconditions and preconditionFailures with fatalErrors

patrickfreed:
[SWIFT-261](https://jira.mongodb.org/browse/SWIFT-261)

This PR replaces all `precondition`s and `preconditionFailure`s with `fatalErrors`. 

Comparison:

- In a debug build, `preconditionFailure` is the same as `fatalError`
- In a release build, `preconditionFailure` exits, but does not print a message. `fatalError` does both.
- in an `-Ounchecked` build, the `preconditionFailure` may be skipped. `fatalError` behaves the same.

Because of the silent nature of `preconditionFailure` and the possibility that it doesn't get called in `OUnchecked` builds, I think we should only be using `fatalError`. In the cases where we only want to exit when testing, we should use `assert`. AFAIK there are no such cases outside of the tests where the latter is true in the codebase right now.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -448,10 +448,10 @@ private func postNotification<T: MongoEvent>(type: T.Type,
                                              contextFunc: (OpaquePointer) -> UnsafeMutableRawPointer?
                                             ) where T: InitializableFromOpaquePointer {
     guard let event = _event else {
-        preconditionFailure("Missing event pointer for \(type)")
+        fatalError("Missing event pointer for \(type)")
     }
     guard let context = contextFunc(event) else {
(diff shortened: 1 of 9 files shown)


kmahar:
Were you able to find any information on when standard library developers choose to use one or the other? I seem to recall them using `preconditionFailure`s in `JSONEncoder`/`JSONDecoder` so I'm curious about why they did that instead of `fatalError`s, since they also use `fatalError` in other parts of the std lib.

kmahar:
> I still think that when we fail/crash this way, we will always want to emit the error message, both for user clarity and future bug reports.

I agree with that. I would rather err on the side of too much info (pun semi intended...)

PR #196, opened 2019-01-22
SWIFT-34 Generate and track code coverage

kmahar:
wait to review. just opening PR to see if webhook integration working properly...


kmahar:
also, here's a coverage report for this branch: https://codecov.io/gh/mongodb/mongo-swift-driver/tree/f8e3350fb91717065119a7310d4587439856586e

Makefile
```diff
@@ -16,20 +16,24 @@ all:
...
 	ruby add_json_files.rb
 
 test:
 	swift test -v $(FILTERARG)
 
 test-pretty:
-	@$(call check_for_gem, xcpretty)
-	swift test $(FILTERARG) 2>&1 | xcpretty
+	@$(call check_for_gem,xcpretty)
+	set -o pipefail && swift test $(FILTERARG) 2>&1 | xcpretty
```
kmahar: this is so that when we run on Travis, `test-pretty` exits with a failure code if `swift test` fails.

.travis.yml
```diff
@@ -4,52 +4,45 @@ dist: trusty
...
-  - eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"
-  
-  # clone and build libmongoc
-  - git clone -b r1.13 https://github.com/mongodb/mongo-c-driver /tmp/libmongoc
-  - pushd /tmp/libmongoc
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr/local; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr; fi
-  - sudo make -j8 install
-  - popd
+  - if [[ "$TASK" == "TEST" ]]; then ./install.sh libmongoc; eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"; ./install.sh mongodb; fi
```
patrickfreed: might want to choose another name for our installer so we don't conflate it with the swiftenv installer. (or download the swiftenv installer as something other than install.sh)

kmahar: renamed `travis_install.sh`.

.travis.yml
```diff
@@ -4,52 +4,45 @@ dist: trusty
 
 matrix:
   include:
+    - name: Lint
```
mbroadst: Have you seen Travis' [build stages](https://docs.travis-ci.com/user/build-stages/) before? I think they might be better suited to this task. Rather than hack in testing with environment variables in the matrix, we can just create `lint`, `code coverage`, and `test` stages and run them conditionally.  What do you think?

travis_install.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
mbroadst: nit: I think we should endeavor not to clutter the top level with files. I generally put files like this in something like `Tests/Scripts`. Additionally I might name it `install_dependencies.sh`?

Btw, this will help very much when building on evergreen! Thanks for splitting it out.

travis_install.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Script for installing various dependencies for Travis jobs.
+# Usage: ./travis_install.sh libmongoc|mongodb|swiftlint
+
+if [[ $1 == "libmongoc" ]]
+then
+	git clone -b r1.13 https://github.com/mongodb/mongo-c-driver /tmp/libmongoc
```
mbroadst: I vote for making `r1.13` an environment variable here, so that we can easily switch which version of libmongoc we're testing against

Makefile
```diff
@@ -16,20 +16,24 @@ all:
...
-	@$(call check_for_gem, xcpretty)
-	swift test $(FILTERARG) 2>&1 | xcpretty
+	@$(call check_for_gem,xcpretty)
+	swift test $(FILTERARG) 2>&1 | xcpretty && exit ${PIPESTATUS[0]}
 
 lint:
 	swiftlint autocorrect
 	swiftlint
 
+cov:
```
mbroadst: nit: `cov` => `coverage`

.travis.yml
kmahar: reorganized to use stages 👍

PR #197, opened 2019-01-24
Update CRUD API errors to match the new hierarchy

patrickfreed:
[SWIFT-144](https://jira.mongodb.org/browse/SWIFT-144) / [SWIFT-300](https://jira.mongodb.org/browse/SWIFT-300)

This PR updates reading CRUD operations to throw errors from the new error hierarchy. (SWIFT-144). As part of that, `InsertMany` now both throws a `.bulkWriteError` and is implemented as a bulk write (SWIFT-300). Also, command failure events have been updated to use the new errors as well.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -132,7 +132,7 @@ public struct CommandFailedEvent: MongoEvent, InitializableFromOpaquePointer {
         self.commandName = String(cString: mongoc_apm_command_failed_get_command_name(event))
         var error = bson_error_t()
         mongoc_apm_command_failed_get_error(event, &error)
-        self.failure = parseMongocError(error: error) // should always return a ServerError.commandError
+        self.failure = parseMongocError(error) // should always return a ServerError.commandError
         self.requestId = mongoc_apm_command_failed_get_request_id(event)
         self.operationId = mongoc_apm_command_failed_get_operation_id(event)
         self.connectionId = ConnectionId(mongoc_apm_command_failed_get_host(event))
@@ -347,7 +347,7 @@ public struct ServerHeartbeatFailedEvent: MongoEvent, InitializableFromOpaquePoi
         self.duration = mongoc_apm_server_heartbeat_failed_get_duration(event)
         var error = bson_error_t()
         mongoc_apm_server_heartbeat_failed_get_error(event, &error)
-        self.failure = parseMongocError(error: error)
+        self.failure = parseMongocError(error)
         self.connectionId = ConnectionId(mongoc_apm_server_heartbeat_failed_get_host(event))
     }
 }
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -113,8 +113,8 @@ public struct CommandFailedEvent: MongoEvent, InitializableFromOpaquePointer {
     /// The command name.
     public let commandName: String
 
-    /// The failure, represented as a MongoError.
-    public let failure: MongoError
+    /// The failure, represented as a MongoSwiftError.
+    public let failure: MongoSwiftError
 
     /// The client generated request id.
     public let requestId: Int64
@@ -132,7 +132,7 @@ public struct CommandFailedEvent: MongoEvent, InitializableFromOpaquePointer {
         self.commandName = String(cString: mongoc_apm_command_failed_get_command_name(event))
         var error = bson_error_t()
         mongoc_apm_command_failed_get_error(event, &error)
-        self.failure = MongoError.commandError(message: toErrorString(error))
+        self.failure = parseMongocError(error: error) // should always return a ServerError.commandError
         self.requestId = mongoc_apm_command_failed_get_request_id(event)
         self.operationId = mongoc_apm_command_failed_get_operation_id(event)
         self.connectionId = ConnectionId(mongoc_apm_command_failed_get_host(event))
@@ -337,7 +337,7 @@ public struct ServerHeartbeatFailedEvent: MongoEvent, InitializableFromOpaquePoi
     public let duration: Int64
 
     /// The failure.
-    public let failure: MongoError
+    public let failure: MongoSwiftError
 
     /// The connection ID (host/port pair) of the server.
     public let connectionId: ConnectionId
@@ -347,7 +347,7 @@ public struct ServerHeartbeatFailedEvent: MongoEvent, InitializableFromOpaquePoi
         self.duration = mongoc_apm_server_heartbeat_failed_get_duration(event)
         var error = bson_error_t()
         mongoc_apm_server_heartbeat_failed_get_error(event, &error)
-        self.failure = MongoError.commandError(message: toErrorString(error))
+        self.failure = parseMongocError(error: error)
         self.connectionId = ConnectionId(mongoc_apm_server_heartbeat_failed_get_host(event))
     }
 }
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -164,6 +164,20 @@ public class BSONDecoder {
                                   debugDescription: "Unable to parse JSON string \(json)"))
     }
 
+    /// Internal function used to  convert encountered `DecodingError`s to `.internalErrors`. Use this when using the
+    /// decoder internally on non-user-modified types.
+    public func internalDecode<T: Decodable>(
+            _ type: T.Type,
+            from document: Document,
+            withError errMsg: String = "Failed to decode \(T.self)")
+    throws -> T {
+        do {
+            return try self.decode(type, from: document)
+        } catch is DecodingError {
+            throw RuntimeError.internalError(message: errMsg)
+        }
+    }
+
     /// A struct to wrap a `Decodable` type, allowing us to support decoding to types that
     /// are not inside a wrapping object (for ex., Int or String).
     private struct DecodableWrapper<T: Decodable>: Decodable {
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -164,9 +164,9 @@ public class BSONDecoder {
                                   debugDescription: "Unable to parse JSON string \(json)"))
     }
 
-    /// Internal function used to  convert encountered `DecodingError`s to `.internalErrors`. Use this when using the
+    /// Internal version of decode that throws `.internalErrors` instead of `DecodingErrors`. Use this when using the
     /// decoder internally on non-user-modified types.
-    public func internalDecode<T: Decodable>(
+    internal func internalDecode<T: Decodable>(
             _ type: T.Type,
             from document: Document,
             withError errMsg: String = "Failed to decode \(T.self)")
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -150,6 +150,7 @@ public class MongoClient {
         if options?.eventMonitoring == true { self.initializeMonitoring() }
 
         guard mongoc_client_set_error_api(self._client, MONGOC_ERROR_API_VERSION_2) else {
+            close()
             throw RuntimeError.internalError(message: "Could not configure error handling on client")
         }
     }
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -150,7 +150,7 @@ public class MongoClient {
         if options?.eventMonitoring == true { self.initializeMonitoring() }
 
         guard mongoc_client_set_error_api(self._client, MONGOC_ERROR_API_VERSION_2) else {
-            close()
+            self.close()
             throw RuntimeError.internalError(message: "Could not configure error handling on client")
         }
     }
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -12,17 +12,19 @@ extension MongoCollection {
      * - Returns: a `BulkWriteResult`, or `nil` if the write concern is unacknowledged.
      *
      * - Throws:
-     *   - `MongoError.invalidArgument` if `requests` is empty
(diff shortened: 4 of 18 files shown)


Sources/MongoSwift/APM.swift
```diff
@@ -132,7 +132,7 @@ public struct CommandFailedEvent: MongoEvent, InitializableFromOpaquePointer {
         self.commandName = String(cString: mongoc_apm_command_failed_get_command_name(event))
         var error = bson_error_t()
         mongoc_apm_command_failed_get_error(event, &error)
-        self.failure = MongoError.commandError(message: toErrorString(error))
+        self.failure = parseMongocError(error: error) // should always return a ServerError.commandError
```
mbroadst: nit: should `parseMongocError` s first parameter not be named? Seems redundant to have `error: error` everywhere, since the method also tells us its parsing an error

Sources/MongoSwift/MongoClient.swift
```diff
@@ -150,6 +150,7 @@ public class MongoClient {
         if options?.eventMonitoring == true { self.initializeMonitoring() }
 
         guard mongoc_client_set_error_api(self._client, MONGOC_ERROR_API_VERSION_2) else {
+            close()
```
mbroadst: nice catch

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -164,6 +164,20 @@ public class BSONDecoder {
                                   debugDescription: "Unable to parse JSON string \(json)"))
     }
 
+    /// Internal function used to  convert encountered `DecodingError`s to `.internalErrors`. Use this when using the
```
kmahar: the first sentence reads like this is something that takes in a `DecodingError` and returns an `internalError`. can we swap the sentences or rephrase so it's clearer that it is basically just `decode` with a different error type?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -53,12 +56,17 @@ extension MongoCollection {
         }
 
         /// Adds the `deleteOne` operation to a bulk write
+        ///
+        /// - Throws:
+        ///   - `UserError.invalidArgumentError` if the options form an invalid combination.
+        ///   - `EncodingError` if an error occurs while encoding the options to BSON.
         public func addToBulkWrite(bulk: BulkWriteOperation, index: Int) throws {
-            let opts = try BSONEncoder().encode(self.options)
+            let opts = try BSONEncoder().encode(options)
```
mbroadst: I thought we _were_ doing explicit `self`?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -53,12 +56,17 @@ extension MongoCollection {
         }
 
         /// Adds the `deleteOne` operation to a bulk write
+        ///
```
kmahar: nit: I think @jmikola and I decided a while back that for > 3 lines docstrings we should use `/** ... */` instead

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -81,12 +89,16 @@ extension MongoCollection {
...
+        /// - Throws:
+        ///   - `UserError.invalidArgumentError` if the options form an invalid combination.
+        ///   - `EncodingError` if an error occurs while encoding the options to BSON.
         public func addToBulkWrite(bulk: BulkWriteOperation, index: Int) throws {
             let opts = try BSONEncoder().encode(options)
             var error = bson_error_t()
 
             guard mongoc_bulk_operation_remove_many_with_opts(bulk.bulk, self.filter.data, opts.data, &error) else {
-                throw MongoError.invalidArgument(message: toErrorString(error))
+                throw UserError.invalidArgumentError(message: toErrorString(error))
```
mbroadst: Why isn't this line using `parseMongocError`?  It seems like there are a few more cases like this, I'll restrict my comment to this one

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -239,6 +263,10 @@ extension MongoCollection {
         }
 
         /// Adds the `updateMany` operation to a bulk write
+        ///
```
kmahar: same nit as above about `///` vs `/** ... */`

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -12,20 +12,23 @@ extension MongoCollection {
...
+            throw UserError.invalidArgumentError(message: "requests cannot be empty")
         }
 
         let opts = try BSONEncoder().encode(options)
-        let bulk = BulkWriteOperation(collection: self._collection, opts: opts?.data)
+        let bulk = BulkWriteOperation(collection: self._collection, opts: opts)
 
         try requests.enumerated().forEach { index, model in
             try model.addToBulkWrite(bulk: bulk, index: index)
+            bulk.requests[index] = model
```
mbroadst: Can you explain why this change was needed?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: the swiftlint rule seems to miss a lot of these... 😩

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -197,7 +217,11 @@ extension MongoCollection {
             self.options = UpdateModelOptions(arrayFilters: arrayFilters, collation: collation, upsert: upsert)
         }
 
-        /// Adds the `updateOne` operation to a bulk write
+        /// Adds the `updateOne` operation to a bulk write.
```
kmahar: same nit as above about /// vs /** ... */

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -393,21 +433,30 @@ public struct BulkWriteResult {
...
+    /// Internal initializer used for testing purposes and error handling.
+    internal init(
+            deletedCount: Int? = nil,
+            insertedCount: Int? = nil,
+            insertedIds: [Int: BSONValue]? = nil,
+            matchedCount: Int? = nil,
+            modifiedCount: Int? = nil,
+            upsertedCount: Int? = nil,
+            upsertedIds: [Int: BSONValue]? = nil) {
+        self.deletedCount = deletedCount ?? 0
```
kmahar: would it make sense to just use all of these default values as the defaults for the parameters, rather than setting new default values inside here?

Sources/MongoSwift/BSON/BSONDecoder.swift
patrickfreed: Fixed. Now reads:
>Internal version of `decode` that throws `.internalErrors` instead of `DecodingErrors`. Use this when using the decoder internally on non-user-modified types.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -164,6 +164,20 @@ public class BSONDecoder {
...
+    /// decoder internally on non-user-modified types.
+    public func internalDecode<T: Decodable>(
```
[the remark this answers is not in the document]
patrickfreed: Yep, fixed.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: fixed

Sources/MongoSwift/MongoCollection+BulkWrite.swift
patrickfreed: Fixed.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
patrickfreed: fixed

Sources/MongoSwift/MongoCollection+BulkWrite.swift
patrickfreed: fixed

Sources/MongoSwift/MongoCollection+BulkWrite.swift
patrickfreed: I would still have to unwrap them even if they had default values. I could force unwrap and add a swiftlint ignore if you think that's preferable.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
mbroadst: Where does the requirement to access the requests in the error come from?

patrickfreed: The definition of `BulkWriteError` in the [crud spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#error-handling) includes it.

mbroadst: Well that would certainly justify it! I still think we should try to offload this onto the C driver, we shouldn't have to allocate these twice for potential error scenarios. As it stands, I'm paying the price for something I'll hopefully never see.

Sources/MongoSwift/MongoError.swift
```diff
@@ -169,6 +172,95 @@ internal func parseMongocError(error: bson_error_t, errorLabels: [String]? = nil
     }
 }
 
+/// Internal function used to get an appropriate error from a server reply to a command.
+internal func getErrorFromReply(
```
kmahar: this might be a bit more readable if we broke out the bulk write and write error portions into separate methods.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
jmikola: IIRC, including the write model in the error is optional. The only requirement is that we report the index for individual write errors. I spoke to @patrickfreed about this last week as I saw he added this behavior but I didn't realize at the time that it would add quite a bit of overhead to the common no-error case.

I don't see how this can be offloaded to the C driver, as I don't believe it provides access to operations once they're added to a `mongoc_bulk_operation_t`. Might be best just to remove this altogether.

patrickfreed: The write model field in the error is an optional type (meaning it wouldn't be present in some cases, I guess. I'm not sure what those cases would be, though.), but it is _not_ marked as "NOT REQUIRED" as the fields which are optional for drivers to implement are. It seems, according to spec at least, the field is required. Whether or not it is worth it to us to conform to the spec in this case is another question, though.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -52,13 +55,20 @@ extension MongoCollection {
             self.options = DeleteModelOptions(collation: collation)
         }
 
-        /// Adds the `deleteOne` operation to a bulk write
+        /**
+         * Adds the `deleteOne` operation to a bulk write
```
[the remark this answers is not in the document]
patrickfreed: Done

Sources/MongoSwift/MongoError.swift
patrickfreed: done

Sources/MongoSwift/MongoClient.swift
```diff
@@ -150,6 +150,7 @@ public class MongoClient {
         if options?.eventMonitoring == true { self.initializeMonitoring() }
 
         guard mongoc_client_set_error_api(self._client, MONGOC_ERROR_API_VERSION_2) else {
+            self.close()
```
jmikola: Is this done in other cases, such as failure to initialize monitoring?

Was this added due to code analysis, or would we actually expect `mongoc_client_set_error_api()` to fail in a real use case?

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -233,9 +251,11 @@ extension MongoCollection {
 
         let opts = try encoder.encode(options)
         var error = bson_error_t()
+        let reply = Document()
```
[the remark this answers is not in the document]
jmikola: I don't think that was a problem. On error, you were still reporting the error string from libmongoc's `bson_error_t`. With this change, you're now surfacing the original command error (if any).

The reply was never really needed, since the success case is just returning a list of index names, which can be derived from the input models.

Sources/MongoSwift/MongoClient.swift
patrickfreed: As the driver is now, this call should never fail. If we change our threading/client pooling model, it will (you need to invoke this on the pool rather than a client derived from the pool). I added the check so that if we did start using client pools, tests would break unless this call was updated.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
patrickfreed: Python has a `dict` that represents a `BulkWriteErrror` which has an `'op'` field in it. It looks like Node has similar behavior.

Java, C#, Go, and Ruby don't report the request.

So none of the drivers report an actual `WriteModel`, but Node and Python do report some information about the request that caused the error.

mbroadst: My thought about offloading onto the C driver was that we ask for an accessor for `mongoc_bulk_operation_t` operations. Outside of that, I'd argue that we amend the spec to clarify that this is optional, and not implement it until someone asks for it.

kmahar: I agree with @mbroadst

patrickfreed: To summarize:
I'll remove the request storage from `BulkWriteError` and file a follow-up SPEC ticket regarding making it NOT REQUIRED. If in the future libmongoc provides some sort of accessor on `bulk_operation_t` for getting operations by index or otherwise reports them in the errors, we can revisit this.

patrickfreed: Filed [SPEC-1209](https://jira.mongodb.org/browse/SPEC-1209)

PR #198, opened 2019-01-24
Various error migrations

patrickfreed:
[SWIFT-301](https://jira.mongodb.org/browse/SWIFT-301) / [SWIFT-302 ](https://jira.mongodb.org/browse/SWIFT-302)/ [SWIFT-303](https://jira.mongodb.org/browse/SWIFT-303) / [SWIFT-305](https://jira.mongodb.org/browse/SWIFT-305)

In this PR:

- `.invalidCollection`, `.invalidClient`, `.invalidUri`, and `.invalidResponse` updated to `.internalErrors` or `.fatalErrors` as necessary (SWIFT-301).
    - Most notable change: `MongoDatabase.collection`, `MongoClient.database` no longer throw. They simply `fatalError` when something goes wrong (which it should never).
- `.invalidCursor` converted to new errors as necessary (SWIFT-302)
    - The most notable change here involves making the initializer of `Cursor` throw. This way users will be notified of `.invalidArgumentErrors` when creating the cursor (through a command) instead of checking the error after creation.
    - Errors at creation are `.invalidArgumentError`, server errors are `.commandError`, and iterating a dead cursor yields a `.logicError`.
- `.bsonParseError` changed to `.invalidArgumentError` or `.internalError` (SWIFT-303)
- `.readPreferenceError` changed to `.invalidArgumentError` (SWIFT-305)

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -115,11 +115,16 @@ public class MongoClient {
      *   - options: optional `ClientOptions` to use for this client
      *
      * - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/
+     *
+     * - Throws:
+     *   - A `UserError.invalidArgumentError` if the connection string passed in is improperly formatted.
+     *   - A `UserError.invalidArgumentError` if the connection string specifies the use of TLS but libmongoc was not
+     *     built with TLS support.
      */
     public init(connectionString: String = "mongodb://localhost:27017", options: ClientOptions? = nil) throws {
         var error = bson_error_t()
         guard let uri = mongoc_uri_new_with_error(connectionString, &error) else {
-            throw MongoError.invalidUri(message: toErrorString(error))
+            throw parseMongocError(error: error)
         }
 
         // if retryWrites is specified, set it on the uri (libmongoc does not provide api for setting it on the client).
@@ -129,7 +134,7 @@ public class MongoClient {
 
         self._client = mongoc_client_new_from_uri(uri)
         guard self._client != nil else {
-            throw MongoError.invalidClient()
+            throw UserError.invalidArgumentError(message: "libmongoc not built with TLS support.")
         }
 
         // if a readConcern is provided, set it on the client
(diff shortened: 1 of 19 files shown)


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -4,12 +4,26 @@ import mongoc
...
 
     /// Initializes a new `MongoCursor` instance, not meant to be instantiated directly.
-    internal init(fromCursor: OpaquePointer, withClient: MongoClient) {
+    internal init(fromCursor: OpaquePointer, withClient: MongoClient) throws {
         self._cursor = fromCursor
         self._client = withClient
+
+        if let err = self.error {
+            // Need to explicitly close since deinit will not execute if we throw.
+            close()
```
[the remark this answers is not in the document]
patrickfreed: fixed. also the one in `deinit`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -129,7 +134,7 @@ public class MongoClient {
 
         self._client = mongoc_client_new_from_uri(uri)
         guard self._client != nil else {
-            throw MongoError.invalidClient()
+            throw UserError.invalidArgumentError(message: "libmongoc not built with TLS support.")
```
mbroadst: is this the only possible reason `mongoc_client_new_from_uri` can fail?

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -22,16 +22,16 @@ final class MongoClientTests: MongoSwiftTestCase {
         let connectionString = MongoSwiftTestCase.connStr
         var error = bson_error_t()
         guard let uri = mongoc_uri_new_with_error(connectionString, &error) else {
-            throw MongoError.invalidUri(message: toErrorString(error))
+            throw parseMongocError(error: error)
         }
 
         guard let client_t = mongoc_client_new_from_uri(uri) else {
-            throw MongoError.invalidClient()
+            throw RuntimeError.internalError(message: "creating client from uri failed")
```
mbroadst: Here we use a different error than the TLS one above

Sources/MongoSwift/MongoClient.swift
patrickfreed: Yep. The relevant libmongoc code:

```
mongoc_client_t *
_mongoc_client_new_from_uri (mongoc_topology_t *topology)
{
   BSON_ASSERT (topology);

#ifndef MONGOC_ENABLE_SSL
    if (mongoc_uri_get_ssl (topology->uri)) {
      MONGOC_ERROR ("Can't create SSL client, SSL not enabled in this build.");
      return NULL;
   }
#endif

   client = (mongoc_client_t *) bson_malloc0 (sizeof *client);
```
So if topology (which is created in `mongoc_client_new_from_uri`, which wraps the function above) is null or the malloc fails, the program will crash. Only null return is the TLS case.

Tests/MongoSwiftTests/MongoClientTests.swift
patrickfreed: This is in a test file so I didn't really bother unifying them, but for consistency it makes sense to. Fixed.

PR #199, opened 2019-01-25
SWIFT-304 Convert bsonEncodeError, bsonDecodeError, typeError to new error types

patrickfreed:
[SWIFT-304](https://jira.mongodb.org/browse/SWIFT-304)

This PR replaces `bsonEncodeError`, `bsonDecodeError`, and `typeError` with the new error types. 

The `bsonEncodeErrors` which derived from a document growing too large after being appended to were replaced with `.internalErrors`. In general, I don't like using / documenting `.internalErrors` and would prefer to reserve them for unexpected errors. I think they fit this use-case decently well, but I would not be opposed to considering the addition of one more `RuntimeError`, namely `RuntimeError.documentTooLargeError` or `RuntimeError.documentSizeError` (to account for when the document is too small as well). What do you guys think? Would this be too specific?

`bsonDecodeErrors` generally resulted from iterator methods trying to retrieve types that weren't currently on the iterator. I've replaced these with `logicErrors`, and I believe it fits nicely. I also added new error cases for this to all the `from(DocumentIterator)` methods that come with `BSONValue` conformance.

### DIFF
--- a/Sources/MongoSwift/BSON/AnyBSONValue.swift
+++ b/Sources/MongoSwift/BSON/AnyBSONValue.swift
@@ -66,27 +66,41 @@ public struct AnyBSONValue: Codable, Equatable, Hashable {
         return bsonEquals(lhs.value, rhs.value)
     }
 
-    /// Initializes a new `AnyBSONValue` from a `Decoder`. 
-    ///
-    /// Caveats for usage with `Decoder`s other than MongoSwift's `BSONDecoder` -
-    /// 1) This method does *not* support initializing an `AnyBSONValue` wrapping
-    /// a `Date`. This is because, in non-BSON formats, `Date`s are encoded
-    /// as other types such as `Double` or `String`. We have no way of knowing 
-    /// which type is the intended one when decoding to a `Document`, as `Document`s 
-    /// can contain any `BSONValue` type, so for simplicity we always go with a 
-    /// `Double` or a `String` over a `Date`.
-    /// 2) Numeric values will be attempted to be decoded in the following
-    /// order of types: `Int`, `Int32`, `Int64`, `Double`. The first one
-    /// that can successfully represent the value with no loss of precision will 
-    /// be used.
+    /**
+     * Initializes a new `AnyBSONValue` from a `Decoder`.
+     *
+     * Caveats for usage with `Decoder`s other than MongoSwift's `BSONDecoder` -
+     * 1) This method does *not* support initializing an `AnyBSONValue` wrapping
+     * a `Date`. This is because, in non-BSON formats, `Date`s are encoded
+     * as other types such as `Double` or `String`. We have no way of knowing
+     * which type is the intended one when decoding to a `Document`, as `Document`s
+     * can contain any `BSONValue` type, so for simplicity we always go with a
+     * `Double` or a `String` over a `Date`.
+     * 2) Numeric values will be attempted to be decoded in the following
+     * order of types: `Int`, `Int32`, `Int64`, `Double`. The first one
+     * that can successfully represent the value with no loss of precision will
+     * be used.
+     *
+     * - Throws:
+     *   - `DecodingError` if a `BSONValue` could not be decoded from the given decoder (which is not a `BSONDecoder`).
+     *   - `DecodingError` if a BSON datetime is encountered but a non-default date decoding strategy was set on the
+     *     decoder (which is a `BSONDecoder`).
+     */
     // swiftlint:disable:next cyclomatic_complexity
     public init(from decoder: Decoder) throws {
         // short-circuit in the `BSONDecoder` case
         if let bsonDecoder = decoder as? _BSONDecoder {
             if bsonDecoder.storage.topContainer is Date {
                 guard case .bsonDateTime = bsonDecoder.options.dateDecodingStrategy else {
-                    throw MongoError.bsonDecodeError(message: "Got a BSON datetime but was expecting another format. " +
-                            "To decode from BSON datetimes, use the default .bsonDateTime DateDecodingStrategy.")
+                    throw DecodingError.typeMismatch(
+                            AnyBSONValue.self,
+                            DecodingError.Context(
+                                    codingPath: bsonDecoder.codingPath,
+                                    debugDescription: "Got a BSON datetime but was expecting another format. To " +
+                                            "decode from BSON datetimes, use the default .bsonDateTime " +
+                                            "DateDecodingStrategy."
+                            )
+                    )
                 }
             }
             self.value = bsonDecoder.storage.topContainer
@@ -130,8 +144,12 @@ public struct AnyBSONValue: Codable, Equatable, Hashable {
         } else if let value = try? container.decode(Document.self) {
             self.value = value
         } else {
-            throw MongoError.typeError(
-                message: "Encountered a value that could not be decoded to any BSON type")
+            throw DecodingError.typeMismatch(
+                    AnyBSONValue.self,
+                    DecodingError.Context(
+                            codingPath: decoder.codingPath,
+                            debugDescription: "Encountered a value that could not be decoded to any BSON type")
+            )
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -193,7 +195,7 @@ extension Document {
      */
     internal func get<T: BSONValue>(_ key: String) throws -> T {
         guard let value = self[key] as? T else {
```
kmahar: while you're here, the line above should use `getValue` to be safe, rather than subscript getter, since that could fatal error. can you update it?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -530,14 +571,19 @@ public struct MaxKey: BSONValue, Equatable, Codable {
...
 
     /// Initializes a new `MaxKey` instance.
     public init() {}
 
-    public static func from(iterator iter: DocumentIterator) -> MaxKey { return self.init() }
+    public static func from(iterator iter: DocumentIterator) throws -> MaxKey {
+        guard iter.currentType == .maxKey else {
+            throw wrongIterTypeError(iter, expected: MaxKey.self)
+        }
+        return self.init()
```
kmahar: to be consistent with `BSONNull`'s `from` method, maybe we should we change `self.init()` to `MaxKey()`? (same q for `MinKey`)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -795,18 +846,19 @@ internal struct Symbol: BSONValue {
...
+        throw RuntimeError.internalError(
+                message: "`Symbol`s are deprecated; use `Symbol.asString` to parse as a string instead"
+        )
     }
 
     internal static func asString(from iter: DocumentIterator) throws -> String {
         var length: UInt32 = 0
         guard let strValue = bson_iter_symbol(&iter.iter, &length) else {
-            throw MongoError.bsonDecodeError(message: retrieveErrorMsg(type: "Symbol", key: iter.currentKey))
+            throw wrongIterTypeError(iter, expected: Symbol.self)
```
kmahar: does `bson_iter_symbol` ever actually return `nil`? I guess only when it is called when the iter is not on a symbol?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -776,14 +827,14 @@ extension String: BSONValue {
...
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         guard bson_append_utf8(storage.pointer, key, Int32(key.count), self, Int32(self.count)) else {
-            throw bsonEncodeError(value: self, forKey: key)
+            throw bsonTooLargeError(value: self, forKey: key)
         }
     }
 
     public static func from(iterator iter: DocumentIterator) throws -> String {
         var length: UInt32 = 0
         guard let strValue = bson_iter_utf8(&iter.iter, &length) else {
```
kmahar: same question as that I had about `bson_iter_symbol`. 
also, looking at docs for `bson_iter_utf8`, it says: "It is suggested that the caller validate the content is valid UTF-8 before using this in other places. That can be done by calling `bson_utf8_validate()` or validating the underlying `bson_t` before iterating it."
maybe we should do that here? (or we could open a ticket and defer it to when we add document validation later on.)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -836,13 +888,18 @@ public struct Timestamp: BSONValue, Equatable, Codable {
...
-            throw bsonEncodeError(value: self, forKey: key)
+            throw bsonTooLargeError(value: self, forKey: key)
         }
     }
 
     public static func from(iterator iter: DocumentIterator) throws -> Timestamp {
         var t: UInt32 = 0
         var i: UInt32 = 0
+
+        guard iter.currentType == .timestamp else {
```
kmahar: nit: check before we bother creating `t` and `i`

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -61,12 +61,20 @@ public protocol BSONValue {
     * - Parameters:
     *   - storage: A `DocumentStorage` to write to.
     *   - key: A `String`, the key under which to store the value.
+    *
+    * - Throws:
+    *   - `RuntimeError.internalError` if the document would become too large by encoding this key-value pair.
+    *   - `UserError.invalidArgumentError` in some cases if this `BSONValue` was modified to a state that cannot be
```
kmahar: is this ever anything besides the too large case? 
if not, that seems like more a problem with the `DocumentStorage` than the `BSONValue` itself, so maybe we can rephrase.
I wonder if it is worth providing a way for users to check document size.... do any other drivers do that?

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -420,7 +422,7 @@ extension Document: BSONValue {
         bson_iter_document(&iter.iter, &length, document)
 
         guard let docData = bson_new_from_data(document.pointee, Int(length)) else {
-            throw MongoError.bsonDecodeError(message: "Failed to create a bson_t from document data")
+            throw RuntimeError.internalError(message: "Failed to create a bson_t from document data")
```
kmahar: since you are updating this line anyway, could we give this a better error message? it exposes a libbson internal users shouldn't need to think about. 
not sure how much we want to rely on libbson internals since this is documented as just returning null on an unnamed failure. but at the moment anyway it looks like this just returns null when there are issues with the length (see [here]( https://github.com/mongodb/mongo-c-driver/blob/16be1bf48fe3378a593adcf84fe6bb1c8c8cd3f5/src/libbson/src/bson/bson.c#L2095))

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: makes sense. let's not bother exposing unless someone asks for it then 🙂

kmahar: oops, looking at the diff now, I was actually referring to the second error here - "in some cases if this `BSONValue` was modified to a state that cannot be encoded to BSON"

kmahar: perfect 👍

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -66,27 +66,41 @@ public struct AnyBSONValue: Codable, Equatable, Hashable {
...
     public init(from decoder: Decoder) throws {
         // short-circuit in the `BSONDecoder` case
         if let bsonDecoder = decoder as? _BSONDecoder {
             if bsonDecoder.storage.topContainer is Date {
                 guard case .bsonDateTime = bsonDecoder.options.dateDecodingStrategy else {
-                    throw MongoError.bsonDecodeError(message: "Got a BSON datetime but was expecting another format. " +
-                            "To decode from BSON datetimes, use the default .bsonDateTime DateDecodingStrategy.")
+                    throw DecodingError.typeMismatch(
+                            AnyBSONValue.self,
+                            DecodingError.Context(
```
mbroadst: the addition of the context is really great!

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -345,7 +345,7 @@ extension _BSONDecoder {
             return Date(timeIntervalSince1970: seconds)
         case .iso8601:
             guard #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) else {
-                throw MongoError.bsonDecodeError(message: "ISO8601DateFormatter is unavailable on this platform.")
+                fatalError("ISO8601DateFormatter is unavailable on this platform.")
```
mbroadst: should we ever even get here? Shouldn't we be preventing choosing this value in the first place on this platform?

patrickfreed: I don't believe we ever can get there, but the compiler doesn't know that. It needs to have the `guard` to know it's in a context where that availability holds, despite the fact that the case could only be evaluated in such a context. Without the `guard`, compilation fails.

mbroadst: oh I would have thought `#available` worked like a prepocessor macro like `#ifdef`, so the option wouldn't even be seen in this context. Cool, thanks for clearing this up.

PR #201, opened 2019-01-30
SWIFT-307 Audit invalid argument errors

patrickfreed:
[SWIFT-307](https://jira.mongodb.org/browse/SWIFT-307)

This PR updates some old `MongoError` invalid argument errors and converts any `.invalidArgumentErrors` that were thrown in the wrong places.

### DIFF
--- a/Sources/MongoSwift/WriteConcern.swift
+++ b/Sources/MongoSwift/WriteConcern.swift
@@ -106,6 +106,8 @@ public class WriteConcern: Codable {
     }
 
     /// Initializes a new `WriteConcern`.
+    /// - Throws:
+    ///   - `UserError.invalidArgumentError` if the options form an invalid combination.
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/WriteConcern.swift
```diff
@@ -127,7 +127,7 @@ public class WriteConcern: Codable {
             let journalStr = String(describing: journal)
             let wStr = String(describing: w)
             let timeoutStr = String(describing: wtimeoutMS)
-            throw MongoError.invalidArgument(message:
+            throw UserError.invalidArgumentError(message:
```
kmahar: let's update the docstring here to say this is the type of error thrown. 
same applies to the other places we've switched to this error but I won't duplicate the comment

patrickfreed: Done 👍

PR #208, opened 2019-02-12
SWIFT-337 Support setting of encoding/decoding strategies on client, database, and collection levels

patrickfreed:
[SWIFT-337](https://jira.mongodb.org/browse/SWIFT-337)

This PR allows users to set the encoding/decoding strategies at the client, database, or collection level. Collections inherit these options from databases, and databases inherit them from clients. They also can be specified when getting/creating a client, database, or collection. 

In implementing this, I defined `DateCodingStrategy`, `UUIDCodingStrategy`, and `DataCodingStrategy`, unifying the encoding/decoding pairs into single cases. This reduces some of the boilerplate for us and for users, and it also restricts users from setting incompatible pairs (e.g. encoding uses `bsonDate`, decoding uses `millisecondsSince1970`). They're defined in `BSONEncoder.swift` arbitrarily.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -113,7 +113,8 @@ public class BSONDecoder {
         self.configureWithOptions(options: options)
     }
 
-    /// Initializes `self` by copying the options of another `BSONDecoder`.
+    /// Initializes `self` by using the options of another `BSONDecoder` and the provided options, with preference
+    /// going to the provided options in the case of conflicts.
     internal init(copies other: BSONDecoder, options: CodingStrategyProvider?) {
         self.userInfo = other.userInfo
         self.dateDecodingStrategy = other.dateDecodingStrategy
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -109,12 +109,12 @@ public class BSONDecoder {
     }
 
     /// Initializes `self`.
-    public init(options: CodingStrategyOptions? = nil) {
+    public init(options: CodingStrategyProvider? = nil) {
         self.configureWithOptions(options: options)
     }
 
     /// Initializes `self` by copying the options of another `BSONDecoder`.
-    internal init(copies other: BSONDecoder, options: CodingStrategyOptions?) {
+    internal init(copies other: BSONDecoder, options: CodingStrategyProvider?) {
         self.userInfo = other.userInfo
         self.dateDecodingStrategy = other.dateDecodingStrategy
         self.uuidDecodingStrategy = other.uuidDecodingStrategy
@@ -122,7 +122,7 @@ public class BSONDecoder {
         self.configureWithOptions(options: options)
     }
 
-    internal func configureWithOptions(options: CodingStrategyOptions?) {
+    internal func configureWithOptions(options: CodingStrategyProvider?) {
         self.dateDecodingStrategy = options?.dateCodingStrategy?.rawValue.decoding ?? self.dateDecodingStrategy
         self.uuidDecodingStrategy = options?.uuidCodingStrategy?.rawValue.decoding ?? self.uuidDecodingStrategy
         self.dataDecodingStrategy = options?.dataCodingStrategy?.rawValue.decoding ?? self.dataDecodingStrategy
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -109,14 +109,23 @@ public class BSONDecoder {
     }
 
     /// Initializes `self`.
-    public init() {}
+    public init(options: CodingStrategyOptions? = nil) {
+        self.applyOptions(options: options)
+    }
 
     /// Initializes `self` by copying the options of another `BSONDecoder`.
-    public init(copies other: BSONDecoder) {
+    internal init(copies other: BSONDecoder, options: CodingStrategyOptions?) {
         self.userInfo = other.userInfo
         self.dateDecodingStrategy = other.dateDecodingStrategy
         self.uuidDecodingStrategy = other.uuidDecodingStrategy
         self.dataDecodingStrategy = other.dataDecodingStrategy
+        self.applyOptions(options: options)
+    }
+
+    internal func applyOptions(options: CodingStrategyOptions?) {
+        self.dateDecodingStrategy = options?.dateCodingStrategy?.rawValue.decoding ?? self.dateDecodingStrategy
+        self.uuidDecodingStrategy = options?.uuidCodingStrategy?.rawValue.decoding ?? self.uuidDecodingStrategy
+        self.dataDecodingStrategy = options?.dataCodingStrategy?.rawValue.decoding ?? self.dataDecodingStrategy
     }
 
     /**
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -111,6 +111,14 @@ public class BSONDecoder {
     /// Initializes `self`.
     public init() {}
 
+    /// Initializes `self` by copying the options of another `BSONDecoder`.
+    public init(copies other: BSONDecoder) {
+        self.userInfo = other.userInfo
+        self.dateDecodingStrategy = other.dateDecodingStrategy
+        self.uuidDecodingStrategy = other.uuidDecodingStrategy
+        self.dataDecodingStrategy = other.dataDecodingStrategy
+    }
+
     /**
      * Decodes a top-level value of the given type from the given BSON document.
      *
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -110,7 +110,7 @@ public class BSONDecoder {
 
     /// Initializes `self`.
     public init(options: CodingStrategyOptions? = nil) {
-        self.applyOptions(options: options)
+        self.configureWithOptions(options: options)
     }
 
     /// Initializes `self` by copying the options of another `BSONDecoder`.
@@ -119,10 +119,10 @@ public class BSONDecoder {
         self.dateDecodingStrategy = other.dateDecodingStrategy
         self.uuidDecodingStrategy = other.uuidDecodingStrategy
         self.dataDecodingStrategy = other.dataDecodingStrategy
-        self.applyOptions(options: options)
+        self.configureWithOptions(options: options)
     }
 
-    internal func applyOptions(options: CodingStrategyOptions?) {
+    internal func configureWithOptions(options: CodingStrategyOptions?) {
         self.dateDecodingStrategy = options?.dateCodingStrategy?.rawValue.decoding ?? self.dateDecodingStrategy
         self.uuidDecodingStrategy = options?.uuidCodingStrategy?.rawValue.decoding ?? self.uuidDecodingStrategy
         self.dataDecodingStrategy = options?.dataCodingStrategy?.rawValue.decoding ?? self.dataDecodingStrategy
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -1,192 +1,6 @@
 import Foundation
 import mongoc
 
-/**
- * Enum representing the various encoding/decoding strategy pairs for `Date`s.
- * Set these on a `MongoClient`, `MongoDatabase`, or `MongoCollection` so that the strategies will be applied when
- * interacting with the database.
- *
- * As per the BSON specification, the default strategy is to encode `Date`s as BSON datetime objects.
- *
- * - SeeAlso: bsonspec.org
- */
-public enum DateCodingStrategy: RawRepresentable {
-    public typealias RawValue = (BSONEncoder.DateEncodingStrategy, BSONDecoder.DateDecodingStrategy)
-
-    /// Encode/decode the `Date` by deferring to its default encoding/decoding implementations.
-    case deferredToDate
-
-    /// Encode/decode the `Date` to/from a BSON datetime object (default).
-    case bsonDateTime
-
-    /// Encode/decode the `Date` to/from a 64-bit integer counting the number of milliseconds since January 1, 1970.
-    case millisecondsSince1970
-
-    /// Encode/decode the `Date` to/from a BSON double counting the number of seconds since January 1, 1970.
-    case secondsSince1970
-
-    /// Encode/decode the `Date` to/from an ISO-8601-formatted string (in RFC 339 format).
-    @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
-    case iso8601
-
-    /// Encode/decode the `Date` to/from a string formatted by the given formatter.
-    case formatted(DateFormatter)
-
-    /// Encode the `Date` by using the given `encodeFunc`. Decode the `Date` by using the given `decodeFunc`.
-    /// If `encodeFunc` does not encode a value, an empty document will be encoded in its place.
-    case custom(encodeFunc: (Date, Encoder) throws -> Void, decodeFunc: (Decoder) throws -> Date)
-
-    public init?(rawValue: RawValue) {
-        switch rawValue {
-        case (.deferredToDate, .deferredToDate):
-            self = .deferredToDate
-        case (.bsonDateTime, .bsonDateTime):
-            self = .bsonDateTime
-        case (.millisecondsSince1970, .millisecondsSince1970):
-            self = .millisecondsSince1970
-        case (.secondsSince1970, .secondsSince1970):
-            self = .secondsSince1970
-        case (.iso8601, .iso8601):
-            guard #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) else {
-                fatalError("ISO8601DateFormatter is unavailable on this platform.")
-            }
-            self = .iso8601
-        case let (.formatted(encodingFormatter), .formatted(decodingFormatter)):
-            guard encodingFormatter == decodingFormatter else {
-                return nil
-            }
-            self = .formatted(encodingFormatter)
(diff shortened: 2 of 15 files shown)


Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -281,15 +351,37 @@ public class MongoDatabase {
...
 
         guard let collection = mongoc_database_create_collection(self._database, name, opts?.data, &error) else {
             throw parseMongocError(error)
         }
 
-        return MongoCollection(fromCollection: collection, withClient: self._client)
+        let encoder = BSONEncoder(copies: self.encoder)
+        let decoder = BSONDecoder(copies: self.decoder)
+
+        if let (e, d) = options?.dateCodingStrategy?.rawValue {
```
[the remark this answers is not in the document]
patrickfreed: Thats a good idea. I've added the `CodingStrategyOptions` protocol for that purpose. I realized it may be useful for  users to configure their own encoders / decoders with these options, so I also defined a base `BSONCoderOptions` struct which conforms to that protocol.  

So now e.g. is possible:
```swift
let options = BSONCoderOptions(dateCodingStrategy: .secondsSince1970, dataCodingStrategy: .base64)
let encoder = BSONEncoder(options: options)
let decoder = BSONDecoder(options: options)
```
or, in the old way:
```swift
var encoder = BSONEncoder()
var decoder = BSONDecoder()
encoder.dateEncodingStrategy = .secondsSince1970
decoder.dateDecodingStrategy = .secondsSince1970
encoder.dataEncodingStrategy = .base64
decoder.dateDecodingStrategy = .base64
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -1,13 +1,201 @@
 import Foundation
 import mongoc
 
+/**
```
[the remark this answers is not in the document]
kmahar: hmm. I can see arguments for both. 
maybe it is strange to define them in a different file via extension when they are used throughout the main file. along those lines, maybe it would make sense to move the initializers that use CodingStrategies into extensions in the new file....
I don't feel very strongly either way though, so I'll defer to you. mainly I just wanted to avoid making the encoder file any longer than it already is.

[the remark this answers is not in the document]
patrickfreed: I think the strategy definitions should stay in the respective encoder/decoder files for the reasons you mention. 

Because one of the initializers is part of the public API, I think it's better that they both stay in the respective files as well. That way someone browsing the code can easily see all the possible ways to instantiate them. 

That said, if the files continue to grow, we can evaluate later whether or not its worth splitting them up into extensions, similar to what we did with `Document`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -21,17 +21,35 @@ public struct ClientOptions {
     /// the server's default write concern will be used.
     public let writeConcern: WriteConcern?
 
+    /// Specifies the strategy to use when coding `Date`s to/from BSON.
```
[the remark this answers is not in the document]
kmahar: hmmm.... I wonder if we should be more vague here actually, and just say this is the date coding strategy that will be used for all DBs/colls created from this client. and say "go read about `CollectionOptions.dateCodingStrategy` for more info". and in the collection spot we can actually explain what the strategy does, and talk about `MongoCollection.CollectionType` etc. then we don't have to duplicate so much across the comments for these strategies on the client and DB. I'm not sure what is best really...

kmahar:
another thought is that we probably want to document somewhere that it's the user responsibility to make sure they consistently use the same strategies with the same collections. similarly to how we document it's their responsibility to consistently use the same `CollectionType` each time

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: The options strings I've updated to say basically what you suggested. The collection options ones I added a bit about their responsibility to ensure the data matches. 

I moved the line about converting between non-Document codable types to the actual strategy definitions, since jazzy will link to them from the docstrings for each of the options. For the same reason, I don't think I need to include a reference to `CollectionOptions` in any of the other options.

I added a bit about how `CollectionType` interacts with the strategies in its docstring.

It's safe to say that once this is merged I should get working on that guide...lol

patrickfreed:
@mbroadst just pinging in case you didn't see this one

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -109,7 +109,24 @@ public class BSONDecoder {
...
+    /// Initializes `self` by copying the options of another `BSONDecoder`.
+    internal init(copies other: BSONDecoder, options: CodingStrategyOptions?) {
+        self.userInfo = other.userInfo
+        self.dateDecodingStrategy = other.dateDecodingStrategy
+        self.uuidDecodingStrategy = other.uuidDecodingStrategy
+        self.dataDecodingStrategy = other.dataDecodingStrategy
+        self.applyOptions(options: options)
+    }
+
+    internal func applyOptions(options: CodingStrategyOptions?) {
```
mbroadst: nit: generally `apply` works in the other direction in my experience ("apply the options represented here TO something else"), this seems to be "accepting" or "receiving" the options

Sources/MongoSwift/MongoClient.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 import mongoc
 
 /// Options to use when creating a `MongoClient`.
-public struct ClientOptions {
+public struct ClientOptions: CodingStrategyOptions {
```
mbroadst: This feels weird to me. It's a protocol masquerading as a concrete type almost. Looking through the implementation below I'm conflicted because this design choice forces a few other choices, f.ex not instantiating an encoder/decoder in place, which I think connotes a hard reset of state in its previous form. 

Were there alternatives considered before landing on this approach? Maybe something like `BsonEncoder(fromClientOptions:)` or embedding the `CodingStrategyOptions` in the `ClientOptions`? Maybe even that `MongoDatabase` and `MongoCollection` take type-specific `options` that also have these fields, and e.g. `MongoDatabaseOptions(fromClientOptions:)`

patrickfreed: Doing `BSONEncoder(fromClientOptions:)` would also necessitate other initializers for collection, database, and generic options. Protocols help avoid that duplication nicely, I think.

We could decide to embed the `CodingStrategyOptions` struct within all the options types that conform to `CodingStrategyOptions` protocol (and remove the protocol) at the cost of a bit of verbosity.

e.g. The API would become
```swift
// inline
let client = MongoClient(..., options: ClientOptions(codingStrategyOptions: CodingStrategyOptions(dateCodingStrategy: .deferredToDate)))
// two step
let strategyOpts = CodingStrategyOptions(dateCodingStrategy: .deferredToDate)
let client = MongoClient(..., options: ClientOptions(codingStrategyOptions: strategyOpts))
```

The last approach would still end up requiring either the current design or one of the approaches you mentioned, so it is sort of orthogonal to the issue here. If this were a pure Swift driver and/or if we had a lot of options to pass from client->db->collection, it would definitely be nice to have conversion capability (a la `FindAndModifyOptionsConvertible`), but since the coding strategy stuff is the only options we're passing, I'm not sure how helpful it would be. 

Embedding the `CodingStrategyOptions` within other options types as necessary sounds fine to me if you'd like me to make that change.

Sources/MongoSwift/BSON/BSONDecoder.swift
patrickfreed: How does `configureWithOptions` sound?

Sources/MongoSwift/MongoClient.swift
mbroadst: Maybe it's just a naming thing, like maybe `CodingStrategyProvider` would have made more sense to me on first read. Do you have thoughts on the only observation about keeping instances of encoders around vs instantiating as-needed? It seems to be the pattern in Swift's internal code that encoders/decoders are not long-lived objects.

patrickfreed: Yeah, `CodingStrategyProvider` is a much better name. If we keep the current approach, I'll definitely refactor the protocol to be named that.

>Do you have thoughts on the only observation about keeping instances of encoders around vs instantiating as-needed?

The encoders and decoders as we have implemented them (and as Apple has implemented the JSON ones) are essentially just factories for internal, private encoders and decoders. The top-level ones provide a nicer API, while the internal ones do most of the work (e.g. we have `_BSONEncoder`, they have `_JSONEncoder`). So, our top level encoders are essentially behaving as containers for the encoding strategies that we set on them--they don't have any other state AFAIK. I would say, as I understand it, storing the encoder and decoder is pretty much equivalent to storing a `CodingStrategyOptions` instance as part of a client/db/collection. 

As with the other part, I don't have strong feelings on this and would be happy to implement whichever way you think is best.

While looking at this, I noticed the `MongoClient` encoder/decoder are `internal`. So I will fix them to be `public` like the others.

mbroadst: Cool, I'd like to wait to hear what @kmahar has to say about this.

kmahar: I think `CodingStrategyProvider` is a good name 👍 

I don't see any issue with reusing the encoders/decoders that have been customized with our desired options across operations. each call to `BSONEncoder.encode` will create a new `_BSONEncoder` from scratch. while I agree we often see something like `JSONEncoder().encode(myThing)`, those are cases where the default configuration is being used. 

one thing to keep in mind for the future is that if we add `NumberEncodingStrategy`s they may possibly conflict with how options should be encoded and sent to libmongoc. 
for example, perhaps you would specify via such strategies that `Int64`s should be encoded as `Int32`s if they are small enough. but if libmongoc expects an option as an `Int64` we need to ignore it. 
so possibly we'll need to use separate encoders for options.

kmahar: > While looking at this, I noticed the MongoClient encoder/decoder are internal. So I will fix them to be public like the others.
what's reasoning for having them be public? so you can inspect what the options are?

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -177,9 +231,14 @@ public class MongoDatabase {
     }
 
     /// Initializes a new `MongoDatabase` instance, not meant to be instantiated directly.
-    internal init(fromDatabase: OpaquePointer, withClient: MongoClient) {
+    internal init(fromDatabase: OpaquePointer,
```
kmahar: nit with existing code: I think the slightly more swifty way of naming these would be something like 
```swift
init(fromDatabase database: OpaquePointer,
     withClient client: MongoClient,
     ...
```

Sources/MongoSwift/MongoClient.swift
patrickfreed: >what's reasoning for having them be public? so you can inspect what the options are?

Yeah, and also so that a user can encode/decode stuff in the same manner they would when inserting into/reading from a given collection without actually reaching out to the db. Like if they read/write bson to/from another source (e.g. disk).

Sources/MongoSwift/MongoDatabase.swift
patrickfreed: fixed

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -109,7 +109,24 @@ public class BSONDecoder {
     }
 
     /// Initializes `self`.
-    public init() {}
+    public init(options: CodingStrategyProvider? = nil) {
+        self.configureWithOptions(options: options)
+    }
+
+    /// Initializes `self` by copying the options of another `BSONDecoder`.
```
kmahar: can we add a sentence to this docstring clarifying what options take precedence when `other` conflicts with `options`?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -102,7 +110,25 @@ public class BSONEncoder {
     }
 
     /// Initializes `self`.
-    public init() {}
+    public init(options: CodingStrategyProvider? = nil) {
+        self.configureWithOptions(options: options)
+    }
+
+    /// Initializes `self` by copying the options of another `BSONEncoder`.
```
kmahar: same comment about clarifying the behavior around conflicting options on `other` and `options`.

patrickfreed: done

Sources/MongoSwift/BSON/BSONDecoder.swift
patrickfreed: done

Sources/MongoSwift/MongoClient.swift
```diff
@@ -88,6 +124,12 @@ public class MongoClient {
     /// If command and/or server monitoring is enabled, indicates what event types notifications will be posted for.
     internal var monitoringEventTypes: [MongoEventType]?
 
+    /// Encoder whose options are inherited by databases derived from this client.
+    public let encoder: BSONEncoder
```
mbroadst: what's the benefit to making these publicly accessible?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -260,6 +307,9 @@ public class MongoClient {
             mongoc_database_set_write_concern(db, wc._writeConcern)
         }
 
-        return MongoDatabase(fromDatabase: db, withClient: self)
+        let encoder = BSONEncoder(copies: self.encoder, options: options)
```
mbroadst: follow-on work: `MongoDatabase` should be responsible for setting all these values up, move concerns to the `init` there. It should just be `return MongoDatabase(withClient: self, withOptions: options)`

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -62,8 +62,7 @@ extension MongoCollection {
          *   - `EncodingError` if an error occurs while encoding the options to BSON.
          */
         public func addToBulkWrite(bulk: BulkWriteOperation, index: Int) throws {
-            let opts = try BSONEncoder().encode(self.options)
-
+            let opts = try bulk.encoder.encode(self.options)
```
mbroadst: what's the benefit of having the bulk operation have its own coders? Is it a `CodingStrategyProvider`?

patrickfreed: We need access to the collection's encoder in `addToBulkWrite`, but `BulkWriteOperation` stores a `mongoc_collection_t` rather than the entire collection itself, so it can't access it. Alternatively, we could store a `MongoCollection` on the operation, or pass in the encoder as a parameter to `addToBulkWrite` instead.

Sources/MongoSwift/MongoClient.swift
patrickfreed: Mainly so that a user can encode/decode stuff in the same manner they would when inserting into/reading from a given collection without actually reaching out to the db. Like if they read/write bson to/from another source (e.g. disk). Also, so they can inspect which options are set on a givent client/db/collection.

Sources/MongoSwift/MongoClient.swift
patrickfreed: I actually think this is worth updating now. There's no reason to pass it this way if the database can just get them through the client.

patrickfreed: I now see why this is future work. Filed [SWIFT-377](https://jira.mongodb.org/browse/SWIFT-377) for it.

Sources/MongoSwift/MongoClient.swift
kmahar: > Like if they read/write bson to/from another source (e.g. disk)

should we rewrite bsondump in Swift? 😛

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: I think it should just store the `MongoCollection`. this would also reduce some duplication, for ex. `BulkWrite.isAcknowledged` can just directly call the `MongoCollection`'s `isAcknowledged` method.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -6,16 +6,20 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
+
     /**
      * Initializes a new `MongoCursor` instance, not meant to be instantiated directly.
      *
      * - Throws:
      *   - `UserError.invalidArgumentError` if the options passed to the command that generated this cursor formed an
      *     invalid combination.
      */
-    internal init(fromCursor: OpaquePointer, withClient: MongoClient) throws {
+    internal init(fromCursor: OpaquePointer, withClient: MongoClient, withDecoder: BSONDecoder) throws {
```
kmahar: this signature should probably be
`fromCursor cursor: OpaquePointer, withClient client: MongoClient, withDecoder decoder: BSONDecoder`

Sources/MongoSwift/MongoCollection+BulkWrite.swift
patrickfreed: I now remember why I did it this way. So the `WriteModel`s that we define are actually internal structs of `MongoCollection`. So, `InsertOneModel` is actually a `MongoCollection<MyType>.InsertOneModel`. `BulkWriteOperation`, however, is just a simple wrapper around a `mongoc_bulk_operation_t` with no generic parameters. Because of this, I think in order to store a `MongoCollection<CollectionType>` here, we'd also have to parameterize `BulkWriteOperation`, which I think isn't within scope for this work. Also, as a side note, a user could pass in a specific write concern to the bulk write operation, so we wouldn't be able to use `MongoCollection`'s `isAcknowledged`.

Sources/MongoSwift/MongoCursor.swift
patrickfreed: fixed

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: gotcha. it seems more reasonable to me to just pass it into the bulk operation as you do now one time, rather than doing a bunch of copies passing it in for every operation we add.

PR #210, opened 2019-02-12
SWIFT-313 Move RegularExpression.nsRegularExpression logic into a new NSRegularExpression initializer

patrickfreed:
[SWIFT-313](https://jira.mongodb.com/browse/SWIFT-313)

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -748,8 +748,7 @@ private let regexOptsMap: [Character: NSRegularExpression.Options] = [
     "x": .allowCommentsAndWhitespace
 ]
 
-/// An extension of `NSRegularExpression` to allow it to be initialized from a `RegularExpression` `BSONValue` and to
-/// support converting options to and from strings.
+/// An extension of `NSRegularExpression` to allow it to be initialized from a `RegularExpression` `BSONValue`.
 extension NSRegularExpression {
     /// Convert a string of options flags into an equivalent `NSRegularExpression.Options`
     internal static func optionsFromString(_ stringOptions: String) -> NSRegularExpression.Options {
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -817,25 +817,23 @@ public struct RegularExpression: BSONValue, Equatable, Codable {
...
-    }
-
     /// Returns `true` if the two `RegularExpression`s have matching patterns and options, and `false` otherwise.
     public static func == (lhs: RegularExpression, rhs: RegularExpression) -> Bool {
         return lhs.pattern == rhs.pattern && lhs.options == rhs.options
     }
 }
 
+/// Extension of `NSRegularExpression` allowing it to be initialized from a `RegularExpression` `BSONValue`.
+extension NSRegularExpression {
```
kmahar: I think we could probably combine this with the other `NSRegularExpression` extension we have defined earlier in the file.

patrickfreed: Yep, didn't see that extension the first time. Moved.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -748,7 +748,8 @@ private let regexOptsMap: [Character: NSRegularExpression.Options] = [
     "x": .allowCommentsAndWhitespace
 ]
 
-/// An extension of `NSRegularExpression` to support converting options to and from strings.
+/// An extension of `NSRegularExpression` to allow it to be initialized from a `RegularExpression` `BSONValue` and to
+/// support converting options to and from strings.
```
kmahar: nit: we don't expose the options conversion stuff, but this will be a public docstring now that the extension contains a public method. so maybe just leave it out

PR #212, opened 2019-02-14
SWIFT-351 Correctly encode strings with multi-byte UTF8 characters

patrickfreed:
[SWIFT-351](https://jira.mongodb.com/browse/SWIFT-351)

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -124,7 +124,7 @@ extension Array: BSONValue {
             try arr.setValue(for: String(i), to: val)
         }
 
-        guard bson_append_array(storage.pointer, key, Int32(key.count), arr.data) else {
+        guard bson_append_array(storage.pointer, key, Int32(key.utf8.count), arr.data) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -145,7 +145,7 @@ public struct BSONNull: BSONValue, Codable, Equatable {
     public init() { }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_null(storage.pointer, key, Int32(key.count)) else {
+        guard bson_append_null(storage.pointer, key, Int32(key.utf8.count)) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -242,7 +242,7 @@ public struct Binary: BSONValue, Equatable, Codable {
         let subtype = bson_subtype_t(UInt32(self.subtype))
         let length = self.data.count
         let byteArray = [UInt8](self.data)
-        guard bson_append_binary(storage.pointer, key, Int32(key.count), subtype, byteArray, UInt32(length)) else {
+        guard bson_append_binary(storage.pointer, key, Int32(key.utf8.count), subtype, byteArray, UInt32(length)) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -280,7 +280,7 @@ extension Bool: BSONValue {
     public var bsonType: BSONType { return .boolean }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_bool(storage.pointer, key, Int32(key.count), self) else {
+        guard bson_append_bool(storage.pointer, key, Int32(key.utf8.count), self) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -308,7 +308,7 @@ extension Date: BSONValue {
     public var msSinceEpoch: Int64 { return Int64((self.timeIntervalSince1970 * 1000.0).rounded()) }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_date_time(storage.pointer, key, Int32(key.count), self.msSinceEpoch) else {
+        guard bson_append_date_time(storage.pointer, key, Int32(key.utf8.count), self.msSinceEpoch) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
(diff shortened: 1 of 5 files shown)


kmahar:
@patrickfreed, can you update with a summary of what we learned when looking into this last week? I've already forgotten...

patrickfreed:
In Swift versions < 5, `String`'s underlying storage is either ASCII or UTF-16 encoded, depending on the contents of the string. So, using the `.utf8` view to get the character count, as this patch does, will incur a performance cost. Once we are on Swift 5, the underlying string storage will be UTF-8 encoded, so the performance cost will go away. 

Until then, we can use the UTF-16 view to get the lengths to avoid the conversion cost. We may still be paying the price most of the time if the underling string is ASCII encoded, though. There's also a chance that ASCII storage->UTF-8 view has no cost. I imagine most keys and strings will be in the ASCII range, so it's worth investigating whether or not that is the case. If it is, I think we should just use the `.utf8` view, especially since it won't require a follow-up patch or Swift version specific length handling.

kmahar:
that all makes sense. perhaps we should write some quick benchmarks to test this out?

if you do, I'd suggest we put them in the [mongo-swift-driver-performance](https://github.com/mongodb/mongo-swift-driver-performance) repo. I found the built-in XCTest benchmarking to be a little lackluster when implementing the driver benchmarks (main problem was you can't change the number of iterations... it's hard coded to 10, but driver benchmarks needed 10,000 🤦‍♀️), so I ended up implementing my own measurement code in there (see `Benchmark.swift`.)

kmahar:
given conversation earlier, scratch my previous comment about benchmarking. 

I do think it could be helpful to abstract all of these into a single place though? something like an `internal var byteLength: Int32` on `String`. then we can document in that spot why we are using the calculation that we are, performance considerations, why it's correct, etc.

kmahar:
also FWIW many of the `bson_append_x` docs explicitly state that the keys must be ASCII. so I think it would be fair for us to place the same constraint on users (and libbson may not perform correctly if we don't)

patrickfreed:
I think the libmongoc documentation is incorrect. 

They have a test verifying that a UTF-8 key works (https://github.com/mongodb/mongo-c-driver/blob/master/src/libbson/tests/test-bson.c#L1392).

It would be a pretty egregious violation of the bson spec, so I'd be surprised if the docs were accurate.

patrickfreed:
I've refreshed myself a little bit more on the terminology, and found that we probably couldn't use the UTF-16 view even if we wanted to. You all may already be familiar with this, but I'll admit Unicode was a little hazy to me before I looked into this, and it helped me to write it all down.

Unicode defines a set of symbols to be represented called _code points_ or _unicode scalars_. (e.g.  🤔) One or more code points can be used in combination with each other to form an _extended grapheme cluster_ (e.g. 🇺+ 🇸= 🇺🇸). In Swift, the concept of a character is equivalent to an extended grapheme cluster. 

Each Unicode encoding scheme defines how scalars are represented in terms of combinations of _code units_. In the case of UTF-8, each code unit is 8 bits. Likewise for UTF-16, each code unit is 16 bits. Some code points/unicode scalars may require multiple code units to be represented properly.

So, the `.utf8` view returns an array of UTF-8 code units. Since they are each 1 byte, it gives us the number of bytes in the string. However, if we used `.utf16.count`, we'd get the number of UTF-16 code units in the string, which would translate to roughly 1/2 the bytes in the string. The vanilla `.count` returns the number of characters/grapheme clusters in the string, which, again, doesn't tell us how many bytes are in the string unless it's ASCII only (ASCII characters are encoded with a single code unit in UTF-8).

When passing a `String` to a C function that receives a `char *`, Swift automatically converts it to a UTF-8 encoded null terminated string, by the way. So, `.utf8.count` accurately reflects the number of bytes in the buffers we're passing to libmongoc.

In conclusion, we don't really have a choice in how we calculate the size of strings. If we're going to be using `String` in conjunction with C APIs, then we have to use `.utf8.count`, regardless of performance considerations. Because of this, I think abstracting away `.utf8.count` is perhaps less idiomatic than just using it, since we're not really making a choice anymore. We're just using a feature of the `String` API. What do you think?

Some good reading on Unicode and Swift+Unicode:
- https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html#//apple_ref/doc/uid/TP40014097-CH7-ID297
- https://manishearth.github.io/blog/2017/01/14/stop-ascribing-meaning-to-unicode-code-points/

kmahar:
it sounds like they do support UTF8 keys then - maybe we should open a CDRIVER ticket about updating the docs?

very nice explanation!! I was a bit unclear on the details as well. you could turn that into a nice little blog post 🙂 based on what you've said, I agree that we have no choice but to use `.utf8.count`, and just using it directly everywhere sounds fine to me.

only thing I wonder about: seems like we are going to convert the strings to UTF8 twice: Swift will do it once automatically for the interop, and then we will also do it ourselves when we get the byte count? but I'm not sure how we could get around that.

I see SWIFT-116 is already coming in handy! 😉

mbroadst:
Thanks for the writeup @patrickfreed, I also agree with your conclusions

>only thing I wonder about: seems like we are going to convert the strings to UTF8 twice: Swift will do it once automatically for the interop, and then we will also do it ourselves when we get the byte count? but I'm not sure how we could get around that.

@kmahar this would be resolved by that last option I was discussing in the meeting yesterday: somehow gaining access to the raw data the string is wrapping and directly passing that pointer, and tracked length. I think this would be the most performant route, but it's unclear whether it's even possible, and what the potential benefit would be. Let's create a ticket for further investigation, merge this patch, and punt until it becomes an issue.

patrickfreed:
Filed [SWIFT-359](https://jira.mongodb.org/browse/SWIFT-359)

PR #213, opened 2019-02-18
SWIFT-116 Ensure Linux tests are kept in-sync

kmahar:
I tried out a couple of different approaches for solving this problem: SwiftPM's `--generate-linuxmain` option to `swift test`, as well as [Sourcery](https://github.com/krzysztofzablocki/Sourcery/), a popular Swift "meta-programming" library. I ended up going with the second solution. 

I've added a check on Travis that works as follows:
- runs the command to regenerate the linux test lists 
- use `git diff --exit-code`  to verify nothing changed (i.e. you must have already updated the list yourself if you added a new test)

### SwiftPM's solution
I was hopeful that we would be able to use SwiftPM's built-in solution. Pros of this:
* no extra dependency.
* if XCTest changes in any way, presumably this feature will be updated to work with those changes.

So I tried it out: I deleted all of the manually written code and ran the command. Output:
```
Fatal error: Unsupported test case name MongoSwiftTestCase: file /Users/buildnode/jenkins/workspace/oss-swift-4.2-package-osx/swiftpm/Sources/Commands/GenerateLinuxMain.swift, line 168
Illegal instruction: 4
```
Nice!! With a bit of research, I found out this is because `MongoSwiftTestCase` does not contain any tests of its own. This has been fixed in Swift 5, but it doesn't seem that the fix has been backported to Swift 4.2, as of 4.2.2 anyway. [see PR](https://github.com/apple/swift-package-manager/pull/1701). 

So, I thought maybe in the meantime the fix would be to just add a dummy test case to `MongoSwiftTestCase`, that we could delete once the driver supports Swift 5. I did so, reran the command, and it did not complain. I commit the new code (see what it looks like [here](https://github.com/mongodb/mongo-swift-driver/commit/b0fd4f7f3ce27c86c8e0753985023b3f9ed09dcc)) and go to test it on my new Linux VM... (Yep, `--generate-linuxmain` doesn't work on Linux, because it uses ObjC under the hood. 🤦‍♀️Which I guess kind of makes sense, because if SwiftPM had a way to figure out what all the tests are on Linux then we wouldn't have to do this in the first place...) 

Anyway, running `swift test` over there gives a multitude of errors, all of the form:
```swift
/home/kmahar/mongo-swift-driver/Tests/MongoSwiftTests/XCTestManifests.swift:209:16: error: property '__allTests' with type '[(String, (SDAMTests) -> () throws -> ())]' cannot override a property with type '[(String, (MongoSwiftTestCase) -> () -> ())]'
    static let __allTests = [
               ^
/home/kmahar/mongo-swift-driver/Tests/MongoSwiftTests/XCTestManifests.swift:176:16: note: attempt to override property here
    static let __allTests = [
               ^
```

 With a bit of debugging I determine that the culprit is this auto-generated extension:
```swift
extension MongoSwiftTestCase {
    static let __allTests = [
        ("test1", test1),
    ]
}
```
It seems the issue is that explicitly defining `allTests` at the level of `MongoSwiftTestCase` means that all of its subclasses will then want `allTests` to have the same exact type in their extensions. But they can't, because their test cases are instances of the subclasses, not the base class.

If I delete the extension, the tests will successfully run. However, each subclass still has a dummy `test1` method that it inherits from `MongoSwiftTestCase`, so that's a bit weird. 

In summary, as of Swift 4.2.2 there's no way for us to make using this command work without some manual intervention to correct the file afterward. IMO that defeats the purpose, and it also makes it impossible for us to actually test on CI that the file is up-to-date. Unless if we wrote a script to do the manual intervention for us... At which point we are basically doing codegen anyway 😛 

Even when we do get to Swift 5, a downside of this approach is that running the command requires downloading libmongoc and building the driver and tests, which means it will take ~3 minutes on Travis if we don't combine it with a testing job. I would prefer to leave it as a separate job, given that we've nicely broken out linting and code coverage in a similar way. (once we get around to that ticket for caching build dependencies that would help speed it up.)

### Sourcery
Generating LinuxMain is a very common application of Sourcery, so much so that they have a [template for this in their docs](https://cdn.rawgit.com/krzysztofzablocki/Sourcery/master/docs/linuxmain.html) which I based mine on. I just had to make a couple small tweaks to the template to get it to omit `MongoSwiftTestCase`, and it was good to go.

A downside of this approach is that it introduces an extra dependency for anyone developing the driver, and if anything changes about XCTest, we might have to edit our file in the future. 

An upside is that it runs really fast on Travis (~15 seconds to download sourcery and run the command). 

An additional upside is that it allows a lot of customization. For example, I was able to put some `swiftlint disable`'s in the template. You can imagine more ways we might want to customize in the future, for example if we had an intentionally MacOS-only test suite we could easily programmatically omit it. 

Unfortunately, Sourcery also does not work on Linux. See [discussion](https://github.com/krzysztofzablocki/Sourcery/issues/306) here feat. a couple of our fav Swift community members 🙂. Sigh.

## In conclusion...
Sorry for this very long explanation, but hopefully you have some insight into why I chose Sourcery now. 
As of Swift 4.2.2 the SwiftPM solution simply does not work for us. Whenever we get to Swift 5 we could consider switching to SwiftPM, but even then I think there is an argument to made for Sourcery based on its speed and ability to customize. I also think it could be helpful to us in other parts of the driver down the road, for example the places where we write out error-prone coding key lists.


kmahar:
BTW, I moved the test `.swiftlint.yml` file up a level so that its rules (rather than the top level linting rules) would apply to `LinuxMain.swift`.

Tests/LinuxMain.stencil
```diff
@@ -0,0 +1,16 @@
...
+{% if not type.annotations.disableTests and type.name != "MongoSwiftTestCase" %}extension {{ type.name }} {
+  static var allTests: [(String, ({{ type.name }}) -> () throws -> Void)] = [
+  {% for method in type.methods %}{% if method.parameters.count == 0 and method.shortName|hasPrefix:"test" %}  ("{{ method.shortName }}", {{ method.shortName }}),
+  {% endif %}{% endfor %}]
+}
+
+{% endif %}{% endfor %}
+XCTMain([
+{% for type in types.classes|based:"XCTestCase" %}{% if not type.annotations.disableTests and type.name != "MongoSwiftTestCase" %}  testCase({{ type.name }}.allTests),
+{% endif %}{% endfor %}])
```
patrickfreed: It would be easier to read this file if it was indented similarly to actual code. If it's left this way purposefully so that the code it generates is more readable, then that should trump the readability of this file.

kmahar: good point. I've reformatted the file a bit to make it more readable (some more line breaks etc.) while still preserving the desired output code formatting. also realized the template was using tabs with width 2 instead of 4 spaces for indenting so I fixed that 😱

Makefile
```diff
@@ -19,6 +19,10 @@ project:
 	@$(call check_for_gem,xcodeproj)
 	ruby Tests/Scripts/add_json_files.rb
 
+sourcery:
+	brew list sourcery || brew install sourcery || { echo "ERROR: Failed to locate or install sourcery; please install yourself"; exit 1; }
```
mbroadst: do we need to do the checks using `brew` here? In at least the travis case you will be downloading this from github, not using brew, and then looking for it using brew. Let's just say `sourcery` is a requirement for the `sourcery` make target, and cut the error message out (it should be obvious if it tries to run and it doesn't exist)

kmahar: the travis case doesn't use `make sourcery`, because we download sourcery into a temporary directory and need to call it with the full path (see [here](https://github.com/mongodb/mongo-swift-driver/pull/213/files#diff-354f30a63fb0907d4ad57269548329e3R12)). 

so this only applies if you are doing this locally, in which case I assume all of us will be using brew-installed sourcery.
that said, the error is probably obvious enough, and we aren't performing a similar check for swiftlint, so I will remove this.

PR #216, opened 2019-02-28
SWIFT-291 Cache travis build dependencies

patrickfreed:
[SWIFT-291](SWIFT-291)

This PR caches the libmongoc installation so it doesn't have to be built from source every time.
This saves ~1 minute of time per task, which is decent.

The big bottleneck is still swiftenv which unfortunately can't really be fixed with the travis cache since it's just downloading big binaries.

Note: the cache should be cleared on master whenever we bump the libmongoc version.


Tests/Scripts/install_dependencies.sh
```diff
@@ -14,12 +14,21 @@ install_from_gh () {
 
 if [[ $1 == "libmongoc" ]]
 then
-	git clone -b ${LIBMONGOC_VERSION} https://github.com/mongodb/mongo-c-driver /tmp/libmongoc
-	pushd /tmp/libmongoc
-	if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr/local; fi
-	if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr; fi
-	sudo make -j8 install
-	popd
+	LIBMONGOC_CACHE_DIR=${HOME}/libmongoc
```
mbroadst: Can we add the version to the folder name for posterity?

patrickfreed: That's a really good idea. We won't need to refresh the cache on libmongoc version bump if the folder name itself is versioned.

PR #217, opened 2019-03-11
SWIFT-356 / SWIFT-352 Round trip all BSONValues correctly

patrickfreed:
[SWIFT-356](https://jira.mongodb.org/browse/SWIFT-356) / [SWIFT-352](https://jira.mongodb.org/browse/SWIFT-352)

This PR ensures all BSON types laid out in the bsonspec are correctly round-tripped through `Document` via the bson-corpus spec tests.

The main changes:
- introduction of `BSONUndefined` type
- exposing of `Symbol` and `DBPointer`
- `Decimal128` now stores a `bson_decimal128_t` instead of converting back and forth between strings
  - The default (and only public) constructor now is an `init?`, since we cannot avoid constructing the `bson_decimal_128_t` any longer if it's serving as the storage

### DIFF
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -534,7 +534,7 @@ private struct _BSONKeyedDecodingContainer<K: CodingKey> : KeyedDecodingContaine
         let entry = try getValue(forKey: key)
         return try self.decoder.with(pushedKey: key) {
             let value = try decoder.unbox(entry, as: type)
-            guard !(value is BSONNull) else {
+            guard !(value is BSONNull) || type == BSONNull.self else {
                 throw DecodingError.valueNotFound(
                     type,
                     DecodingError.Context(codingPath: self.decoder.codingPath,
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -1128,8 +1128,6 @@ public struct BSONUndefined: BSONValue, Equatable, Codable {
  * - Returns: `true` if `lhs` is equal to `rhs`, `false` otherwise.
  */
 public func bsonEquals(_ lhs: BSONValue, _ rhs: BSONValue) -> Bool {
-    validateBSONTypes(lhs, rhs)
-
     switch (lhs, rhs) {
     case let (l as Int, r as Int): return l == r
     case let (l as Int32, r as Int32): return l == r
@@ -1174,18 +1172,6 @@ public func bsonEquals(_ lhs: BSONValue?, _ rhs: BSONValue?) -> Bool {
     }
 
     return bsonEquals(left, right)
-}
-
-/// A function for catching invalid BSONTypes that should not ever arise, and triggering a fatalError when it
-/// finds such types.
-private func validateBSONTypes(_ lhs: BSONValue, _ rhs: BSONValue) {
-    let invalidTypes: [BSONType] = [.invalid]
-    guard !invalidTypes.contains(lhs.bsonType) else {
-        fatalError("\(lhs.bsonType) should not be used")
-    }
-    guard !invalidTypes.contains(rhs.bsonType) else {
-        fatalError("\(rhs.bsonType) should not be used")
-    }
 }
 
 /// Error thrown when a BSONValue type introduced by the driver (e.g. ObjectId) is encoded not using BSONEncoder
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -360,6 +360,17 @@ public struct DBPointer: BSONValue, Codable, Equatable {
         self.id = id
     }
 
+    public init(from decoder: Decoder) throws {
+        if decoder is _BSONDecoder {
+            throw bsonDecodingDirectlyError(type: DBPointer.self, at: decoder.codingPath)
+        }
+        throw bsonDecodingUnsupportedError(type: DBPointer.self, at: decoder.codingPath)
+    }
+
+    public func encode(to: Encoder) throws {
+        throw bsonEncodingUnsupportedError(value: self, at: to.codingPath)
+    }
+
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         var oid = try ObjectId.toLibBSONType(self.id.oid) // TODO: use the stored bson_oid_t (SWIFT-268)
         guard bson_append_dbpointer(storage.pointer, key, Int32(key.utf8.count), self.ref, &oid) else {
@@ -405,14 +416,6 @@ public struct Decimal128: BSONValue, Equatable, Codable {
         self.decimal128 = bsonDecimal
     }
 
-    public init(from: Decoder) throws {
-        throw RuntimeError.internalError(message: "Decoding Decimal128 from non-BSON decoder is unsupported")
-    }
-
-    public func encode(to: Encoder) throws {
-        throw RuntimeError.internalError(message: "Encoding Decimal128 to a non-BSON encoder is unsupported")
-    }
-
     /// Initializes a `Decimal128` value from the provided `String`. Returns `nil` if the input is not a valid
     /// Decimal128 string.
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst
@@ -436,6 +439,14 @@ public struct Decimal128: BSONValue, Equatable, Codable {
         throw bsonEncodingUnsupportedError(value: self, at: to.codingPath)
     }
 
+    public func encode(to storage: DocumentStorage, forKey key: String) throws {
+        // TODO: avoid this copy via withUnsafePointer once swift 4.1 support is dropped (SWIFT-284)
+        var copy = self.decimal128
+        guard bson_append_decimal128(storage.pointer, key, Int32(key.utf8.count), &copy) else {
+            throw bsonTooLargeError(value: self, forKey: key)
+        }
+    }
+
     /// Returns the provided string as a `bson_decimal128_t`, or throws an error if initialization fails due an
     /// invalid string.
     /// - Throws:
@@ -450,14 +461,6 @@ public struct Decimal128: BSONValue, Equatable, Codable {
 
     public static func == (lhs: Decimal128, rhs: Decimal128) -> Bool {
         return lhs.decimal128.low == rhs.decimal128.low && lhs.decimal128.high == rhs.decimal128.high
-    }
-
-    public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        // TODO: avoid this copy via withUnsafePointer once swift 4.1 support is dropped (SWIFT-284)
-        var copy = self.decimal128
-        guard bson_append_decimal128(storage.pointer, key, Int32(key.utf8.count), &copy) else {
-            throw bsonTooLargeError(value: self, forKey: key)
-        }
     }
 
     public static func from(iterator iter: DocumentIterator) throws -> Decimal128 {
@@ -1076,6 +1079,17 @@ public final class BSONUndefined: BSONValue, Equatable, Codable {
     public var bsonType: BSONType { return .undefined }
 
     internal init() {}
+
+    public init(from decoder: Decoder) throws {
+        if decoder is _BSONDecoder {
+            throw bsonDecodingDirectlyError(type: BSONUndefined.self, at: decoder.codingPath)
+        }
+        throw bsonDecodingUnsupportedError(type: BSONUndefined.self, at: decoder.codingPath)
+    }
+
+    public func encode(to: Encoder) throws {
+        throw bsonEncodingUnsupportedError(value: self, at: to.codingPath)
+    }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         guard bson_append_undefined(storage.pointer, key, Int32(key.utf8.count)) else {
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -1075,7 +1075,7 @@ public struct Timestamp: BSONValue, Equatable, Codable {
 
 /// A struct to represent the deprecated Undefined type.
 /// Undefined instances cannot be created, but they can be read from existing documents that contain them.
-public final class BSONUndefined: BSONValue, Equatable, Codable {
+public struct BSONUndefined: BSONValue, Equatable, Codable {
     public var bsonType: BSONType { return .undefined }
 
     internal init() {}
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -344,25 +344,27 @@ extension Date: BSONValue {
     }
 }
 
-/// An internal struct to represent the deprecated DBPointer type. While DBPointers cannot
-/// be created, we may need to parse them into `Document`s, and this provides a place for that logic.
-internal struct DBPointer: BSONValue {
+/// An struct to represent the deprecated DBPointer type.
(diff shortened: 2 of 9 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -928,6 +960,31 @@ public struct Timestamp: BSONValue, Equatable, Codable {
     }
 }
 
+/// A struct to represent the deprecated Undefined type.
+/// Undefined instances cannot be created, but they can be read from existing documents that contain them.
+public final class BSONUndefined: BSONValue, Equatable, Codable {
```
patrickfreed: I've made this a `class` because there's no point in ever making copies. `BSONNull` is essentially the same thing--should we make that a `class` as well?

mbroadst: Does making it a class buy us anything? I tend to think of structs as value types, and BSONUndefined and BSONNull fit into that category.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -1080,6 +1151,9 @@ public func bsonEquals(_ lhs: BSONValue, _ rhs: BSONValue) -> Bool {
     case let (l as [BSONValue], r as [BSONValue]): // TODO: SWIFT-242
         return l.count == r.count && zip(l, r).reduce(true, { prev, next in prev && bsonEquals(next.0, next.1) })
     case (_ as [Any], _ as [Any]): return false
+    case let (l as Symbol, r as Symbol): return l == r
```
mbroadst: side thought (perhaps a ticket producing one): can we add an equality operator to the `BsonValue` protocol and depend on that here for deduplication? Something like:
```
public static func == (lhs: Self, rhs: Self) -> Bool
```

Seems like most of these types already have this thing defined, would be cool to use that generically.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -1105,7 +1179,7 @@ public func bsonEquals(_ lhs: BSONValue?, _ rhs: BSONValue?) -> Bool {
 /// A function for catching invalid BSONTypes that should not ever arise, and triggering a fatalError when it
 /// finds such types.
 private func validateBSONTypes(_ lhs: BSONValue, _ rhs: BSONValue) {
-    let invalidTypes: [BSONType] = [.symbol, .dbPointer, .invalid, .undefined]
+    let invalidTypes: [BSONType] = [.invalid]
```
mbroadst: is there value in keeping this function around now?

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -256,18 +256,11 @@ public class DocumentIterator: IteratorProtocol {
...
-            return try Symbol.asString(from: self)
-        case .dbPointer:
-            return try DBPointer.asDocument(from: self)
-        default:
-            guard let curVal = try DocumentIterator.bsonTypeMap[currentType]?.from(iterator: self) else {
-                throw RuntimeError.internalError(message: "Unknown BSONType for iterator's current value.")
-            }
-
-            return curVal
+        guard let curVal = try DocumentIterator.bsonTypeMap[currentType]?.from(iterator: self) else {
```
mbroadst: this looks a little magical, could we split it up into extracting the type (and throwing if the type is not found), and the actual type instantiation?

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -391,17 +380,21 @@ final class CodecTests: MongoSwiftTestCase {
             "ts": Timestamp(timestamp: 1, inc: 2),
             "int32": 5,
             "int64": Int64(6),
-            "dec": Decimal128("1.2E+10"),
+            "dec": Decimal128("1.2E+10")!,
```
mbroadst: Why does this need to be force unwrapped? I don't see us making the initializer for Decimal128 return an optional in this PR, did that change?

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -330,10 +360,28 @@ final class DocumentTests: MongoSwiftTestCase {
 
                 // for cB input:
                 // native_to_bson( bson_to_native(cB) ) = cB
-                expect(Document(fromBSON: cBData).rawBSON).to(equal(cBData))
+                let docFromCB = Document(fromBSON: cBData)
+                expect(docFromCB.rawBSON).to(equal(cBData))
+
+                // test round tripping through documents
+                let testRoundTrip = {
+                    let nativeFromDoc = docFromCB.toArray()
```
mbroadst: can you explain a little bit what's going on here? Why do we need to convert this to an array?

Tests/MongoSwiftTests/CodecTests.swift
patrickfreed: That changed in this PR.

Before this patch, we just stored a `String` within our `Decimal128` type and used libmongoc's string interop for writing to / reading from a document. In order to properly pass the spec tests marked as "lossy" and to properly round trip all `Decimal128`s, we need to store the actual `bson_decimal128_t`.

The old, non-optional initializer didn't bother constructing the `bson_decimal_128_t` since it never stored it. Now, since we always need to store it, the default (and only) String initializer does try to construct one and returns nil if it fails.

Tests/MongoSwiftTests/DocumentTests.swift
patrickfreed: This spec test is designed for drivers like Python that use a `dict` as a `Document` and only write actual BSON bytes when necessary. We actually store everything in a `bson_t`, so we don't have such an abstraction, but we do have a translation layer where bugs can occur.

When we read from the document to the array, we exercise our functions that retrieve the data from the `bson_t` and convert them to Swift types (i.e. `from(iterator:)`). Writing them back from the array to a `Document` exercises our `encode(to:DocumentStorage)` which converts Swift types to BSON data. If they both work properly, the original underlying `bson_t` should be identical to the one wrapped by the document created from the array, even though the types have been round tripped through our translation layer.

I added a comment explaining this more clearly.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -382,34 +398,31 @@ internal struct DBPointer: BSONValue {
...
-        self.data = data
+    internal var decimal128: bson_decimal128_t
+
+    internal init(bsonDecimal: bson_decimal128_t) {
+        self.decimal128 = bsonDecimal
     }
 
     /// Initializes a `Decimal128` value from the provided `String`. Returns `nil` if the input is not a valid
-    /// Decimal128 string. Note that this initializer is less performant than the non-failable initializer `init(_:)`.
+    /// Decimal128 string.
```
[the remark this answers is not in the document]
patrickfreed: It looks...really bad. Nice catch.
```
Decimal128(decimal128: __C.bson_decimal128_t(low: 128, high: 3476778912330022912))
```
Fixed.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -929,56 +938,82 @@ extension String: BSONValue {
         }
     }
 
+    /// Initializer that preserves null bytes embedded in C strings
+    internal init?(cStringWithEmbeddedNulls: UnsafePointer<CChar>, length: Int) {
```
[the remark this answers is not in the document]
patrickfreed: [yep](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/string.json#L32)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -1038,6 +1073,42 @@ public struct Timestamp: BSONValue, Equatable, Codable {
     }
 }
 
+/// A struct to represent the deprecated Undefined type.
+/// Undefined instances cannot be created, but they can be read from existing documents that contain them.
+public struct BSONUndefined: BSONValue, Equatable, Codable {
```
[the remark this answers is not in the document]
patrickfreed: re: `@deprecated`, see my other comment on `DBPointer`.

I'd rather just do it all in one swoop rather than having this one type be separate, if that's okay.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -256,18 +256,11 @@ public class DocumentIterator: IteratorProtocol {
...
-        case .dbPointer:
-            return try DBPointer.asDocument(from: self)
-        default:
-            guard let curVal = try DocumentIterator.bsonTypeMap[currentType]?.from(iterator: self) else {
-                throw RuntimeError.internalError(message: "Unknown BSONType for iterator's current value.")
-            }
-
-            return curVal
+        guard let bsonType = DocumentIterator.bsonTypeMap[currentType] else {
+            throw RuntimeError.internalError(message: "Unknown BSONType for iterator's current value.")
```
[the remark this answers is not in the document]
patrickfreed: done

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -391,17 +380,21 @@ final class CodecTests: MongoSwiftTestCase {
...
             "minkey": MinKey(),
             "maxkey": MaxKey(),
-            "regex": RegularExpression(pattern: "^abc", options: "imx")
+            "regex": RegularExpression(pattern: "^abc", options: "imx"),
+            "symbol": expected.symbol,
```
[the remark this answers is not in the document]
patrickfreed: I've moved this whole construction into a `toDocument` method on `AllBSONTypes`. I kept the extJSON string literal below as its own thing since I assume it's defined separately from our extJSON implementation purposefully.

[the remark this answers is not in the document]
kmahar: yeah, I think that is there to test parsing a doc from ext JSON.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -344,25 +344,41 @@ extension Date: BSONValue {
...
+        throw bsonDecodingUnsupportedError(type: DBPointer.self, at: decoder.codingPath)
     }
 
-    /// Reads DBPointer data from `iter` and converts it to DBRef format
-    ///
-    /// - Throws:
-    ///   - `UserError.logicError` if the current type of the `DocumentIterator` does not correspond to `DBPointer`.
-    internal static func asDocument(from iter: DocumentIterator) throws -> Document {
+    public func encode(to: Encoder) throws {
+        throw bsonEncodingUnsupportedError(value: self, at: to.codingPath)
```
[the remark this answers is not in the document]
patrickfreed: filed [SWIFT-379](https://jira.mongodb.org/browse/SWIFT-379)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -382,34 +398,41 @@ internal struct DBPointer: BSONValue {
...
-
+public struct Decimal128: BSONValue, Equatable, Codable, CustomStringConvertible {
     public var bsonType: BSONType { return .decimal128 }
 
-    /// Initializes a `Decimal128` value from the provided `String`. Assumes that the input string is correctly
-    /// formatted.
-    /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst
-    public init(_ data: String) {
-        self.data = data
+    public var description: String {
```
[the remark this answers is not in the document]
patrickfreed: I think we can punt on deciding that until we do [SWIFT-367](https://jira.mongodb.org/browse/SWIFT-367). Ideally, a user will be able to mutate a `Decimal128` if they so wished, we just currently don't provide the facilities for it. So, I don't think it's worth adding machinery for caching the string version when we may have to get rid of it anyways.

Sources/MongoSwift/BSON/Overwritable.swift
```diff
@@ -43,8 +43,9 @@ extension Double: Overwritable {
 
 extension Decimal128: Overwritable {
     internal func writeToCurrentPosition(of iter: DocumentIterator) throws {
-        var encoded = try Decimal128.toLibBSONType(self.data)
-        bson_iter_overwrite_decimal128(&iter.iter, &encoded)
+        // TODO: avoid this copy via withUnsafePointer once swift 4.1 support is dropped (SWIFT-284)
```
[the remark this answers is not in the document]
patrickfreed: Yeah, I'll just wait for the merge then fix them up in this PR

Sources/MongoSwift/BSON/BSONValue.swift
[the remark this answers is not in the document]
kmahar: sgtm

kmahar:
looks good to me excluding the places that will change given our dropping 4.0 and 4.1.

mbroadst:
> looks good to me excluding the places that will change given our dropping 4.0 and 4.1.

Let's wait until #220 is merged to continue this (should be today)

Sources/MongoSwift/BSON/Overwritable.swift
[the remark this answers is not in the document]
kmahar: I think this spot still needs to be updated.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -534,7 +534,7 @@ private struct _BSONKeyedDecodingContainer<K: CodingKey> : KeyedDecodingContaine
         let entry = try getValue(forKey: key)
         return try self.decoder.with(pushedKey: key) {
             let value = try decoder.unbox(entry, as: type)
-            guard !(value is BSONNull) else {
+            guard !(value is BSONNull) || type == BSONNull.self else {
```
kmahar: what necessitated this change?

patrickfreed: I noticed that I never added `BSONNull` to `AllBSONTypes` (shame), and I was under the impression that I used `AllBSONTypes` for testing the changes as part of this ticket. That's not _technically_ true, though, so I can revert them if it clouds this PR.

The exact change referenced here is to allow the decoder to decode `BSONNull` if it's a member of a struct/class. It's a very uncommon case but is still a bug (which was detected by adding `BSONull` to `AllBSONTypes`).

It's encountered when someone tries decoding to a type like this:
```swift
struct NonsensicalType {
     let a: BSONull
}
```

Sources/MongoSwift/BSON/Overwritable.swift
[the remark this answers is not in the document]
patrickfreed: whoops, fixed.

I have to cast to a mutable pointer here due to a C driver oversight. I filed [CDRIVER-3069](https://jira.mongodb.org/browse/CDRIVER-3069) for it, but I doubt it will be fixed any time soon / ever.

Sources/MongoSwift/BSON/BSONDecoder.swift
kmahar: I'm fine with leaving it, I was just curious about what case would case would lead to this.

Sources/MongoSwift/BSON/Overwritable.swift
```diff
@@ -43,9 +43,11 @@ extension Double: Overwritable {
 
 extension Decimal128: Overwritable {
     internal func writeToCurrentPosition(of iter: DocumentIterator) throws {
-        // TODO: avoid this copy via withUnsafePointer once swift 4.1 support is dropped (SWIFT-284)
-        var copy = self.decimal128
-        bson_iter_overwrite_decimal128(&iter.iter, &copy)
+        withUnsafePointer(to: self.decimal128) { ptr in
+            // bson_iter_overwrite_decimal128 takes in a (non-const) *decimal_128_t, so we need to pass in a mutable
+            // pointer. no mutation of self.decimal128 should occur, however.
```
kmahar: I saw you opened a CDRIVER ticket, can you link it here please?

patrickfreed: done

PR #220, opened 2019-03-27
SWIFT-283, SWIFT-284: Drop Swift 4.0 and 4.1 support

kmahar:
This PR:
- Removes our Swift 4.0 and 4.1 variants on Travis
- Condenses to a single `Package.swift` that uses swift-tools-version 4.2
- Removes all code special-casing lower Swift versions


kmahar:
I think the version info should go in the README as part of our next release. I would like to hold off on writing something though until we figure out what exactly is going on with XCode 10.2 and whether the Swift 5 compiler can build the driver or not. I opened [SWIFT-380] about updating the README (https://jira.mongodb.org/browse/SWIFT-380) with a fix version of 0.0.10.

I'm also opening [SWIFT-381](https://jira.mongodb.org/browse/SWIFT-381) about investigating whether we should use version suffixes.

PR #227, opened 2019-04-01
use internal `_bson_get_len` from swift C bson bindings

mbroadst:
This method is introduced to overcome limitations around field
access when `bson_t` is treated as an `OpaquePointer`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,8 +6,8 @@ let package = Package(
         .library(name: "MongoSwift", targets: ["MongoSwift"])
     ],
     dependencies: [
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -6,7 +6,7 @@ let package = Package(
         .library(name: "MongoSwift", targets: ["MongoSwift"])
     ],
     dependencies: [
-        .package(url: "https://github.com/mongodb/swift-bson", from: "2.0.0"),
+        .package(url: "https://github.com/mongodb/swift-bson", .branch("SWIFT-384/bson-len-accessor")),
```
mbroadst: this will change once https://github.com/mongodb/swift-bson/pull/2 is merged

PR #231, opened 2019-04-08
SWIFT-276: add Swift 5 support

mbroadst:
- Split out `DocumentIterator` into its own file
- Ensure we are no longer passing `bson_t` or `bson_iter_t` as reference, but instead are explicit about when they are being used as unsafe mutable pointers, or just unsafe pointers


kmahar:
one more thing: we have several people watching this ticket, and probably our PRs now, waiting for Swift 5 support. if we're going to merge this in separately first, I don't want to confuse them, so let's open a new ticket "refactor pointer usage" or whatever, and change this PR to be associated with that.

PR #233, opened 2019-04-09
SWIFT-276: Swift 5 support

mbroadst:
This introduces full support for Swift 5 on OSX and Linux.


Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -162,17 +162,8 @@ func clean(json: String?) -> String {
...
-            return String()
-        }
-        let data = try JSONSerialization.data(withJSONObject: object, options: [])
-        guard let string = String(data: data, encoding: .utf8) else {
-            print("Unable to convert JSON data to Data: \(str)")
-            return String()
-        }
-        return string
+        let doc = try Document(fromJSON: str.data(using: .utf8)!)
+        return doc.description
```
patrickfreed: [nit] I think it's more explicit to use `doc.extendedJSON` here instead, even though they're the same thing.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -161,16 +166,34 @@ public class DocumentIterator: IteratorProtocol {
     }
 
     /// Internal helper function for explicitly accessing the `bson_iter_t` as an unsafe pointer
+#if compiler(>=5.0)
     internal func withBSONIterPointer<Result>(_ body: (BSONIterPointer) throws -> Result) rethrows -> Result {
-      return try withUnsafePointer(to: self.iter, body)
+        return try withUnsafePointer(to: self.iter) { iterPtr in
+            try body(OpaquePointer(iterPtr))
```
patrickfreed: In this cast and the one below we can use the typealiases above, unless you meant for the opaque-ness to be explicit.

Sources/MongoSwift/BSON/Document+Sequence.swift
```diff
@@ -17,6 +17,7 @@ extension Document: Sequence {
     // instead of something like `Slice<Document>`.
     /// The type that is returned from methods such as `dropFirst()` and `split()`.
     public typealias SubSequence = Document
+    public typealias Index = Int
```
kmahar: it feels kind of odd that this is defined here, but only used in `Document+Collection`. it looks like it is part of the `Collection` protocol rather than `Sequence` from the [docs](https://developer.apple.com/documentation/swift/collection/2943866-index). should it be defined in that extension instead?

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -161,16 +166,34 @@ public class DocumentIterator: IteratorProtocol {
...
+    }
+#else
+    internal func withBSONIterPointer<Result>(_ body: (BSONIterPointer) throws -> Result) rethrows -> Result {
+        return try withUnsafePointer(to: self.iter, body)
     }
+#endif
 
     /// Internal helper function for explicitly accessing the `bson_iter_t` as an unsafe mutable pointer
+#if compiler(>=5.0)
+    internal func withMutableBSONIterPointer<Result>(
```
kmahar: same q as above about isolating the different line

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -161,16 +166,34 @@ public class DocumentIterator: IteratorProtocol {
     }
 
     /// Internal helper function for explicitly accessing the `bson_iter_t` as an unsafe pointer
+#if compiler(>=5.0)
```
kmahar: I think I would find the differences between this function across compiler versions more readable if we could isolate the 1-2 different lines within the if/else rather than redefining the entire thing. I think you did this a couple places above.  something like
```swift
internal func withBSONIterPointer<Result>(_ body: (BSONIterPointer) throws -> Result) rethrows -> Result {
    return try withUnsafePointer(to: self.iter) { iterPtr in
#if compiler(>=5.0)
        try body(OpaquePointer(iterPtr))
#else 
        try body(iterPtr)
#endif
    }
}
```
makes it really clear that all that's changing is whether we cast to `OpaquePointer`.

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -5,43 +5,46 @@ import Foundation
...
     /// Allows access to a `KeyValuePair` from the `Document`, given the position of the desired `KeyValuePair` held
     /// within. This method does not guarantee constant-time (O(1)) access.
-    public subscript(position: Int) -> KeyValuePair {
+    public subscript(position: Index) -> KeyValuePair {
         // TODO: This method _should_ guarantee constant-time O(1) access, and it is possible to make it do so. This
         // criticism also applies to key-based subscripting via `String`.
         // See SWIFT-250.
         failIndexCheck(position)
+
+        let subsequence = DocumentIterator.subsequence(of: self, startIndex: position, endIndex: position + 1)
```
patrickfreed: this might be a little clearer if we introduced an `init(forDocument: Document, advancedToIndex: Index)` to `DocumentIterator`. It would also mirror how we get an individual element via a string key.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -442,9 +442,13 @@ final class CodecTests: MongoSwiftTestCase {
     // test that Document.init(from decoder: Decoder) works with a non BSON decoder and that
     // Document.encode(to encoder: Encoder) works with a non BSON encoder
     func testDocumentIsCodable() throws {
+        // We presently have no way to control the order of emitted JSON in `cleanEqual`, so this
```
patrickfreed: Is there an SR ticket or bug report that tracks what's going wrong here? If so, it might be worth adding to the comment. Also might be worth filing a SWIFT ticket if there's work that needs to be done by us.

kmahar: honestly, I would be fine with just deleting this test. people should just do what I described in the "note" below if they want to convert between documents and JSON.

patrickfreed: I think we would like to eventually provide real `Codable` conformance for all our types, which is what this test is covering for `Document`. So I'd say it's probably worth leaving in if we expect it to be able to be run again in the future.

kmahar: in that case I would say let's just clean up the comments on why we are ignoring it, as it already says a few lines below that we are skipping on linux for key ordering reasons

Sources/MongoSwift/BSON/DocumentIterator.swift
mbroadst: wow that's weird.. this must have been a result of fighting the compiler, I swear they used to have different signatures 🤔

kmahar: lol I felt like I was doing one of those "spot the differences between these two pictures" game 🕵️‍♀️

Package.swift
```diff
@@ -8,7 +8,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/mongodb/swift-bson", .upToNextMajor(from: "2.0.0")),
         .package(url: "https://github.com/mongodb/swift-mongoc", .upToNextMajor(from: "2.0.0")),
-        .package(url: "https://github.com/Quick/Nimble.git", from: "7.3.0")
+        .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "8.0.0"))
```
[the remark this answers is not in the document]
mbroadst: I'm not sure this is relevant unless we want to use SwiftPM features released during 5.0. This is just a version hint for supported keywords. If we, for instance, wanted to host `swift-bson`/`swift-mongoc` within this repo, then we'd have to update to the 5.0 tools version. Same goes for the clang module projects, they don't need any special features of SwiftPM 4.2+

Sources/MongoSwift/BSON/Document+Collection.swift
mbroadst: How would that be used here?

Sources/MongoSwift/BSON/Document+Sequence.swift
mbroadst: Sure!

Sources/MongoSwift/BSON/Document+Collection.swift
patrickfreed: we'd be able to do:
```swift
let iter = DocumentIterator(forDocument: self, advancedToIndex: position)
return (iter.currentKey, iter.currentValue)
```
As a side note, `currentKeyValuePair` may be a nice computed property on `DocumentIterator` so we could do it in one line:
```swift
return DocumentIterator(forDocument: self, advancedToIndex: position).currentKeyValuePair
```

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -288,11 +293,7 @@ extension Document {
     public var rawBSON: Data {
         // swiftlint:disable:next force_unwrapping - documented as always returning a value.
         let data = bson_get_data(self.data)!
-#if compiler(>=5.0)
-        let length = _bson_get_len(self.data)
-#else
-        let length = self.data.pointee.len
-#endif
+        let length = data.withMemoryRebound(to: Int32.self, capacity: 4) { $0.pointee }
```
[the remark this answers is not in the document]
mbroadst: what about an internal `fun bson_get_len`. I'm not sure its useful to leak this detail, the interface we want people to use is `rawBSON` itself - it has length information, as well as the raw bytes. What would you do with the computed property? 

I agree about the future thoughts, but think we should incorporate those ideas into the pure-swift bson design

Tests/MongoSwiftTests/CodecTests.swift
mbroadst: They do provide a `.sortedKeys` option for `JSONSerialization.writeJSONObject`, but it only works on iOS, and iOS 10.13.x+ at that (and it doesn't seem to work on 10.14.2, afaict). The issue we are left with is that `[String: Any]` isn't sortable. I think we should figure out another way to test what's happening here. We already disabled the test on Linux because I imagine they incorporated changes earlier than OSX that were already on the way.

Sources/MongoSwift/BSON/Document.swift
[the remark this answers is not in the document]
patrickfreed: Yeah, that's a good point. If we're not going to be using length separate from `rawBSON` then I don't see a need for an internal property / function. I do still think it should have a comment, though.

Package.swift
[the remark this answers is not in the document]
kmahar: I had been thinking we needed to do it to start conditionally using 5.0 language features anywhere, but that seems not to be the case based on some testing I just did. so I agree, no reason to update

Sources/MongoSwift/BSON/Document+Collection.swift
mbroadst: I like the first one! I'll add it now.

Sources/MongoSwift/BSON/Document.swift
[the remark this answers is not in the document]
kmahar: if moved out, I think it should be an `internal` computed property, but just leaving here with comment also sounds good if we don't think we will use it elsewhere.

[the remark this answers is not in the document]
kmahar: oh I see, you meant a top level `bson_get_len`, I was imagining something on the document. lgtm as it is now 👍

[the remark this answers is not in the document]
mbroadst: I moved it back into `rawBSON` and copied the note there.

Sources/MongoSwift/BSON/Document+Collection.swift
mbroadst: I ended up just creating a `DocumentIterator` and advancing it it (this is what the convenience init would have done). Using the custom initializer here actually made it difficult to reason about all the potential error cases (should we want to test for all of them):
  - `failIndexCheck` is happening in the subscript, but it _should_ happen in the initializer
  - should the initializer catch an error on `iter.advance()`?
  - it's not even clear to me how one guards a failable initializer, this code does not work:
```
internal convenience init?(forDocument doc: Document, advancedToIndex pos: Document.Index) {
    guard let me = self.init(forDocument: doc) else {
        return nil
    }
    ...
}
```

We only use this in one place, so I chose instead to localize the logic at the call site and not overeagerly DRY the code.

Tests/MongoSwiftTests/CodecTests.swift
mbroadst: merged the two comments, removed the conditional OS check since we're skipping everywhere now.

patrickfreed: This test can just be updated as part of [SWIFT-327](https://jira.mongodb.org/browse/SWIFT-327) / [SWIFT-328](https://jira.mongodb.org/browse/SWIFT-328). Until then it looks good as it is 👍

Sources/MongoSwift/BSON/Document+Collection.swift
```diff
@@ -4,44 +4,56 @@ import Foundation
...
         // criticism also applies to key-based subscripting via `String`.
         // See SWIFT-250.
         failIndexCheck(position)
-        // swiftlint:disable:next force_unwrapping - failIndexCheck precondition ensures non-nil result.
-        return DocumentIterator.subsequence(of: self, startIndex: position, endIndex: position + 1).first!
+        guard let iter = DocumentIterator(forDocument: self) else {
+            fatalError("Failed to initialize an iterator over document \(self)")
+        }
+
+        for pos in 0..<position + 1 {
```
kmahar: nit: maybe clearer to just write this as `for pos in 0...position`?

mbroadst: oh yeah duh, silly copy pasta on my part there!

valeriomazzeo:
That's great, amazing job 👍🏻
Are we getting a new tagged release? 😬

kmahar:
@valeriomazzeo yup, within the next few days 🙂

PR #234, opened 2019-04-09
Enable trailing whitespace rule in comments

kmahar:
So apparently you have to tell SwiftLint that you want this rule on for comments....


kmahar:
my thought was that we should do this first so the swift 5 support commit is less polluted with all those whitespace changes. I thought it would be a clean rebase since no lines are added/deleted, but if it isn't then we can wait

PR #235, opened 2019-04-10
SWIFT-221 Provide better consistency around Int usage with Documents

patrickfreed:
[SWIFT-221](https://jira.mongodb.org/browse/SWIFT-221)

This PR changes the way integers are read from and written with `Document`.

On 64-bit systems:
- BSON int32 will be read to `Int32`
- `Int32` will be written to BSON int32
- BSON int64 will be read to `Int`
- `Int64` will be written to BSON int64
- `Int` will be written to BSON int64

On 32-bit systems:
- BSON int32 will be read to `Int`
- `Int32` will be written to BSON int32
- BSON int64 will be read to `Int64`
- `Int64` will be written to BSON int64
- `Int` will be written to BSON int32

This has the notable effect that `Int` will always be round tripped, though it will be in a size according to the cpu arch.

Also as part of this PR, the protocol `BSONNumber` was introduced. All numeric values that conform to `BSONValue` should conform to `BSONNumber` as well. It facilitates conversion to the numeric types that have native representations in BSON.


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -512,6 +654,15 @@ extension Double: BSONValue {
         }
     }
 
+    /// Initialize this `Double` from another `BSONNumber`.
+    /// If the passed in number cannot be represented exactly as an `Double`, this initializer will return nil.
+    public init?(fromNumber num: BSONNumber) {
```
patrickfreed: Note, I couldn't use `init?(exactly num: BSONNumber)` since it conflicts with the existing `init?(exactly)` as part of `BinaryInteger`. Also `init?(from)` conflicts with the `init(from)` in `CodableNumbers`.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -16,7 +16,7 @@ public struct IndexModel: Encodable {
 
     /// Gets the default name for this index.
     internal var defaultName: String {
-        return String(cString: mongoc_collection_keys_to_index_string(self.keys.data))
```
patrickfreed: libmongoc's keys to index name function only supports `Int32`'s for the directions, so I had to rewrite it in Swift.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -402,24 +402,6 @@ public struct InsertManyResult {
...
-     *   - insertedIds: Map of inserted IDs
-     *
-     * - Throws:
-     *   - `RuntimeError.internalError` if an unexpected error occurs the reading server reply.
-     */
-    fileprivate init(reply: Document, insertedIds: [Int: BSONValue]) throws {
-        self.insertedCount = try reply.getValue(for: "insertedCount") as? Int ?? 0
-        self.insertedIds = insertedIds
-    }
-
```
patrickfreed: Because `InsertMany` is implemented as a bulk write now, this is never used. Instead of updating it, I just decided to get rid of it.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -95,10 +95,13 @@ public class WriteConcern: Codable {
     /// Initializes a new `WriteConcern`.
     /// - Throws:
     ///   - `UserError.invalidArgumentError` if the options form an invalid combination.
-    public init(journal: Bool? = nil, w: W? = nil, wtimeoutMS: Int32? = nil) throws {
+    public init(journal: Bool? = nil, w: W? = nil, wtimeoutMS: Int64? = nil) throws {
```
patrickfreed: The write concern spec defines wtimeoutMS as an `int64`, but libmongoc expects an `int32` here. I've updated it to just truncate if the value is too big. I think it's worth getting our API correct now so that if we ever get off of libmongoc or they update the signature, we won't need to maintain two initializers until a major version bump.

One question though: would it be preferable to just throw an error here?

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -361,16 +363,17 @@ final class CodecTests: MongoSwiftTestCase {
         {
             "double" : 2.0,
             "string" : "hi",
-            "doc" : { "x" : 1 },
+            "doc" : { "x" : { "$numberLong": "1" } },
```
patrickfreed: unfortunately, libmongoc's relaxed extJSON parser reads integers as `Int32`, so I needed to update all these relaxed instances to be canonical.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -524,12 +675,37 @@ extension Double: BSONValue {
...
 
 /// An extension of `Int` to represent the BSON Int32 or Int64 type.
-/// The `Int` will be encoded as an Int32 if possible, or an Int64 if necessary.
-extension Int: BSONValue {
-    public var bsonType: BSONType { return self.int32Value != nil ? .int32 : .int64 }
+/// On 64-bit systems, `Int` corresponds to a BSON Int64. On 32-bit systems, it corresponds to a BSON Int32.
+extension Int: BSONNumber {
+    /// `Int` corresponds to a BSON int32 or int64 depending upon whether the compilation system is 32 or 64 bit.
+    internal static var bsonType: BSONType {
+        return MemoryLayout<Int>.size == 4 ? .int32 : .int64
```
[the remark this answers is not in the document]
mbroadst: Should this be reused in the tests to replace [this code](https://github.com/mongodb/mongo-swift-driver/blob/master/Tests/MongoSwiftTests/TestUtils.swift#L49)?

Sources/MongoSwift/MongoCollection+Indexes.swift
patrickfreed: Filed [CDRIVER-3084](https://jira.mongodb.org/browse/CDRIVER-3804)

Sources/MongoSwift/WriteConcern.swift
patrickfreed: Filed [SWIFT-395](https://jira.mongodb.org/browse/SWIFT-395)

Sources/MongoSwift/BSON/BSONValue.swift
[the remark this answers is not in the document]
patrickfreed: I updated the code you linked to use `Int`'s bsonType to check whether or not the system is 32-bit (as is now done elsewhere in this PR), but I kept the `is32Bit` variable since it is a little more explicit about what it's testing and makes our tests that use it a little easier to read.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -443,18 +443,18 @@ public struct BulkWriteResult {
      *   - `RuntimeError.internalError` if an unexpected error occurs reading the reply from the server.
      */
     fileprivate init(reply: Document, insertedIds: [Int: BSONValue]) throws {
-        self.deletedCount = try reply.getValue(for: "nRemoved") as? Int ?? 0
-        self.insertedCount = try reply.getValue(for: "nInserted") as? Int ?? 0
+        self.deletedCount = (try reply.getValue(for: "nRemoved") as? BSONNumber)?.toInt() ?? 0
```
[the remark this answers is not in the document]
patrickfreed: done

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -103,6 +103,105 @@ extension BSONValue where Self: Equatable {
...
+        return Double(self)
+    }
+}
+
+/// Default implementation of `Decimal128` conversions for all `Numeric`s.
+extension BSONNumber where Self: Numeric {
+    /// Create a `Decimal128` from this `Numeric`.
+    /// This will return nil if the conversion cannot result in an exact representation.
+    public func toDecimal128() -> Decimal128? {
+        return Decimal128(String(describing: self))
```
kmahar: how confident are we that the description strings will always be valid `Decimal128` strings? I seem to recall that the formatting of those is a little tricky. it might be worth testing a few more cases, like negative numbers.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I know I'm the one who proposed some sort of an initializer accepting a `BSONNumber`, but do you think these actually add much value to the API? 
if it could be named `exactly` it might since people are already used to that... but I'm not sure if there's an occasion you'd want to do this rather than just call `toDouble`?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -524,12 +675,39 @@ extension Double: BSONValue {
...
+    /// If the passed in number cannot be represented exactly as an `Int`, this initializer will return nil.
+    public init?(fromNumber num: BSONNumber) {
+        guard let int = num.toInt() else {
+            return nil
+        }
+        self = int
+    }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         if let int32 = self.int32Value {
```
kmahar: I think it might be safer if we explicitly switch on the bson type here, rather than relying on the properties to check it for us.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -649,19 +656,21 @@ final class CodecTests: MongoSwiftTestCase {
         // int32
         let int32 = Int32(5)
 
-        expect(try decoder.decode(AnyBSONValue.self, from: "{ \"$numberInt\" : \"5\" }").value).to(bsonEqual(5))
+        expect(try decoder.decode(AnyBSONValue.self, from: "{ \"$numberInt\" : \"5\" }").value).to(bsonEqual(int32))
 
         let wrappedInt32: Document = ["x": int32]
         expect(try encoder.encode(AnyBSONStruct(int32))).to(equal(wrappedInt32))
         // as int because we convert Int32 -> Int when decoding
```
kmahar: this comment seems out of date.... I don't understand what it's referring to

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -542,21 +720,39 @@ extension Int: BSONValue {
         throw RuntimeError.internalError(message: "`Int` value \(self) could not be encoded as `Int32` or `Int64`")
     }
 
+    public func bsonEquals(_ other: BSONValue?) -> Bool {
```
kmahar: we discussed this friday but I can't quite remember, can you clarify what the behavior is for `bsonEquals` with the integer types / why we separately defining `bsonEquals` for them?

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: I guess it just exists as an alternative way of calling `toDouble`. In the driver, I don't use them anywhere because, as you said, wherever you can use them you could just use the call it's wrapping instead.

I would say that they can just be removed. Let me know if you think they should stay.

As a side note: the reasoning the initializers weren't added to the protocol was that if something conforms to `BSONNumber`, it is likely that it will also conform to at least one of the many `Numeric` descendant protocols, meaning it would probably have an initializer that takes in the result of one of the `toX` converters. So an `init(BSONNumber)` felt redundant. (@mbroadst I know I said friday I'd have a better answer to this question)

Tests/MongoSwiftTests/CodecTests.swift
patrickfreed: removed

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: The thought process here is that the `nil` check constitutes the bson type check, and that we'd still have to perform the nil check after performing the bson type check even though we'd know it would always succeed. Given that, do you still think it's worth the explicit type check?

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: We need separate implementations of `bsonEquals` for `Int`, `Int32`, and `Int64` because they need to compare against both their own type and `Int`, depending on if `Int` represents a BSON int32 or a BSON int64.

e.g. on 64 bit:
```swift
5.bsonEquals(Int64(5)) // true
5.bsonEquals(5) // true
5.bsonEquals(Int32(5)) // false

Int64(5).bsonEquals(5) // true
Int64(5).bsonEquals(Int64) // true
```

So friday I mentioned that I wanted to make some slight changes to how `bsonEquals` works, but having thought about it more, I don't think they're necessary right now (as @mbroadst already determined)

Sources/MongoSwift/BSON/CodableNumber.swift
```diff
@@ -82,7 +82,7 @@ extension UInt8: CodableNumber {
 extension UInt16: CodableNumber {
     internal var bsonValue: BSONValue? {
         // UInt16 always fits in an Int32
-        return Int(exactly: self)
```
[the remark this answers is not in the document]
patrickfreed: yep, I think so

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I'm good with removing them, sorry I sent you down the path of adding of them.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -103,6 +103,105 @@ extension BSONValue where Self: Equatable {
     }
 }
 
+/// A protocol that numeric `BSONValue`s should conform to. It provides functionality for converting to BSON's native
+/// number types.
+public protocol BSONNumber: BSONValue {
+    /// Create an `Int` from this `BSONNumber`.
+    /// This will return nil if the conversion cannot result in an exact representation.
+    func toInt() -> Int?
```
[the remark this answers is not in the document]
kmahar: what you've said makes sense - I'm good with keeping them as methods.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: hm, I guess not. maybe this is just a naming thing... I feel like `int32Value` doesn't convey that we are doing the type check, it kind of sounds like it's just doing the same thing as `toInt32()`.  no immediate ideas for a better name though...

patrickfreed: We could make it simply `int32` and `int64`. Then if `self.int32` is nil, it slightly implies that `self` isn't an `int32`. 

Similarly, `asInt32` and `asInt64` imply cast-like behavior, which is close to what is going on here. If they fail, then `Int` isn't whatever was being "casted" to.

A third option is to provide only a single accessor that returns whatever it actually is (typed as a `BSONNumber`), which can then itself be casted. The property could be called `typedValue` or something along those lines.
e.g. 
```swift
if let int32 = self.typedValue as? Int32 { ... }
else if let int64 = self.typedValue as? Int64 { ... }

switch self.typedValue {
case let int32 as Int32:
    // do stuff with int32
case let int64 as Int64:
    // do stuff with int64
default:
    // should be impossible
}
```

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: No worries at all, it was definitely a path worth exploring.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: As part of updating the protocol to use computed properties, I had to rename `int32Value` and `int64Value` so they wouldn't collide with the protocol. I renamed them to just `int32` and `int64` arbitrarily, and am still open to other rename candidates.

Related to naming collisions: I think `NSNumber` won't be able to conform to `BSONNumber` because its computed properties don't return optionals, but ours do. I don't think this is the biggest problem in the world because `NSNumber` is basically a vehicle for achieving the same stuff as `BSONNumber`, just with the focus on C types instead of BSON types, so I don't think there's any meaningful value in conforming it to `BSONValue`. Honestly, there's not a ton of value in conforming _anything_ to `BSONValue` outside of the types we've done it for since we don't expose any way of writing to the `bson_t`.

kmahar: hmmm..... I kinda like the `typedValue` option just because it's name is the clearest. 

also, since it returns a `BSONNumber` which inherits from `BSONValue`, in a lot of these places it would actually eliminate the need to handle them separately altogether -- for example
```swift
if let int32 = self.int32 {
    return try int32.encode(to: storage, forKey: key)
}
if let int64 = self.int64 {
    return try int32.encode(to: storage, forKey: key)
}
```
could just become: `return try typedValue.encode(to: storage, forKey: key)`.

kmahar: (or maybe you would need to unwrap `typedValue` first, but anyway, saves you the two separate `encode` calls.)

patrickfreed: Since we know `Int` will fit in either `Int32` or `Int64`, we can just return a regular `BSONNumber` and won't need to unwrap.

patrickfreed: (also, I've updated to use the `typedValue` computed property)

Issue #236, opened 2019-04-11
bson_malloc memory leak


valeriomazzeo:
Hi, there. 
I am experiencing massive memory leaks that seems to originate from `bson_malloc`.
These calls are possibly originated by `BSONDecoder`.
Running on macOS 10.14.4 / mongo driver 0.0.9 /  mongo-c-driver 1.13.0

Would be grateful if someone could help debugging the issue. Are you aware or have any suggestion on what could be the cause?

![Screenshot 2019-04-11 at 20 25 37](https://user-images.githubusercontent.com/1882080/55986879-3d047f00-5c98-11e9-9f02-51426db86882.png)

kmahar:
Hey, thanks so much for reporting this! We're not aware of any particular issue that would be causing this - probably, we are not cleaning up a bson_t somewhere that we should be. Could you possibly give us a code sample that reproduces this?

kmahar:
I think I've found the source, actually. Give me a few minutes and I will put up a patch on a branch for you to test.

valeriomazzeo:
Unfortunately it's not that easy, it may be due to the complex data structure on my types.

>I think I've found the source, actually. Give me a few minutes and I will put up a patch on a branch for you to test.

That's great, I can definitely test that against my project.

kmahar:
I've made the changes on top of the 0.0.9 tag on [this branch](https://github.com/mongodb/mongo-swift-driver/tree/leak) -- let us know how it goes. we'll definitely get these fixes into the next tagged release with Swift 5 support.

re: `BSONEncoder`/`BSONDecoder`, you could encounter some race conditions if you are mutating the same encoder or decoder (i.e. by setting encoding/decoding strategies) while using it in multiple threads. They don't preserve any internal state otherwise though - all the storage/data is local to the `encode`/`decode` functions - so if you're just using those functions it should be safe.

valeriomazzeo:
That's what I thought. I will give it a go now.

valeriomazzeo:
@kmahar it's now better but it seems there are still some coming from the client initialisation method.
![Screenshot 2019-04-11 at 22 10 14](https://user-images.githubusercontent.com/1882080/55993296-9f647c00-5ca6-11e9-8a00-57af04697b5d.png)

kmahar:
@valeriomazzeo just pushed another commit to that branch, try now?
thanks so much for helping us find these - we're planning on doing a full audit of the driver for memory leaks in the coming weeks.

[...]

PR #237, opened 2019-04-13
SWIFT-294, SWIFT-394 Fix memory leaks

kmahar:
This fixes several memory leaks in the driver. Kudos to @valeriomazzeo for pointing these out in #236.

I've done the following:
* Rename `Document(fromPointer:)` and `DocumentStorage(fromPointer:)` to `Document(copying:)` and `DocumentStorage(copying:)` to make it clearer what their behavior is 
* Add a new `DocumentStorage(stealing:)` initializer that takes over responsibility for the provided `bson_t`, and a `Document(stealing:)` initializer that calls through to it
* Update every place the `fromPointer` initializers were used to call through to the correct new methods 

I also fixed a leak of the `mongoc_uri_t` in the `MongoClient` initializer, and a leak we'd previously identified in SDAM monitoring (SWIFT-294).

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -20,14 +20,18 @@ public class DocumentStorage {
         return Int(bson_count_keys(self.pointer))
     }
 
+    /// Initializes a new, empty `DocumentStorage`.
     internal init() {
         self.pointer = bson_new()
     }
 
+    /// Initializes a new `DocumentStorage` by copying over the data from the provided `bson_t`.
     internal init(copying pointer: BSONPointer) {
         self.pointer = bson_copy(pointer)
     }
 
+    /// Initializes a new `DocumentStorage` that uses the provided `bson_t` as its backing storage.
+    /// The newly created instance will handle cleaning up the pointer upon deinitialization.
     internal init(stealing pointer: MutableBSONPointer) {
         self.pointer = pointer
     }
@@ -61,8 +65,8 @@ extension Document {
     }
 
     /**
-     * Initializes a `Document` from a pointer to a `bson_t` by making a copy of the
-     * data. The caller is responsible for freeing the original `bson_t`.
+     * Initializes a `Document` from a pointer to a `bson_t` by making a copy of the data. The `bson_t`'s owner is
+     * responsible for freeing the original.
      *
      * - Parameters:
      *   - pointer: a BSONPointer
@@ -75,8 +79,8 @@ extension Document {
     }
 
     /**
-     * Initializes a `Document` from a pointer to a `bson_t`, using the `bson_t` as its
-     * underlying storage. The caller must not modify or free the `bson_t` themselves.
+     * Initializes a `Document` from a pointer to a `bson_t`, "stealing" the `bson_t` to use for underlying storage/
+     * The `bson_t` must not be modified or freed by others after it is used here/
      *
      * - Parameters:
      *   - pointer: a MutableBSONPointer
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -24,8 +24,12 @@ public class DocumentStorage {
         self.pointer = bson_new()
     }
 
-    internal init(fromPointer pointer: BSONPointer) {
+    internal init(copying pointer: BSONPointer) {
         self.pointer = bson_copy(pointer)
+    }
+
+    internal init(stealing pointer: BSONPointer) {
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -24,10 +24,14 @@ public class DocumentStorage {
         self.pointer = bson_new()
     }
 
-    internal init(fromPointer pointer: BSONPointer) {
+    internal init(copying pointer: BSONPointer) {
```
mbroadst: Consider adding documentation here. I think you tend to do it at the call site, but it's probably useful to future readers why these are distinguished

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -57,17 +61,30 @@ extension Document {
     }
 
     /**
-     * Initializes a `Document` from a pointer to a bson_t. Uses a copy
-     * of `bsonData`, so the caller is responsible for freeing the original
-     * memory.
+     * Initializes a `Document` from a pointer to a `bson_t` by making a copy of the
+     * data. The caller is responsible for freeing the original `bson_t`.
```
mbroadst: I don't think this is always correct. For instance, in the APM code above we copy the pointer, and are _not_ responsible for freeing the original `bson_t`

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -24,10 +24,14 @@ public class DocumentStorage {
         self.pointer = bson_new()
     }
 
-    internal init(fromPointer pointer: BSONPointer) {
+    internal init(copying pointer: BSONPointer) {
         self.pointer = bson_copy(pointer)
     }
 
+    internal init(stealing pointer: MutableBSONPointer) {
+        self.pointer = pointer
```
mbroadst: what if `self.pointer` is already set?

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -115,9 +119,13 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
             return nil
         }
-        // swiftlint:disable:next force_unwrapping - if mongoc_cursor_next returned `true`, this is filled out.
-        let doc = Document(fromPointer: out.pointee!)
 
+        guard let pointee = out.pointee else {
+            fatalError("mongoc_cursor_next returned true, but document is nil")
+        }
+
+    // we have to copy because libmongoc owns the pointer.
```
mbroadst: strange formatting here

Sources/MongoSwift/BSON/Document.swift
kmahar: since this is an initializer, I don't think that is possible.

Sources/MongoSwift/BSON/Document.swift
kmahar: added docs for all three `DocumentStorage` initializers.

Sources/MongoSwift/BSON/Document.swift
kmahar: good point, I updated it to just say that whoever owns the `bson_t` is responsible for freeing it.

Sources/MongoSwift/MongoCursor.swift
kmahar: oops, fixed

Sources/MongoSwift/BSON/Document.swift
mbroadst: that's a... really good point 😄 I'm like a hammer looking for nails over here

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -328,11 +349,7 @@ extension Document {
                 throw RuntimeError.internalError(message: toErrorString(error))
             }
 
-#if compiler(>=5.0)
```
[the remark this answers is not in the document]
kmahar: so `DocumentStorage(copying:)` (previously `DocumentStorage(fromPointer:)`) accepts a `BSONPointer` (non-mutable). 
previously, we were passing the result of `bson_new_from_json`, which is a `MutableBSONPointer`, into that initializer. On Swift 4.2, that required a cast to make the 
`UnsafeMutablePointer<bson_t>` into an `UnsafePointer<bson_t>`. 
on 5.0, since `BSONPointer` and `MutableBSONPointer` are both `OpaquePointer`s, no cast was necessary. hence the if/else.

since this code now calls the stealing initializer, which accepts a mutable pointer, no cast is necessary in 4.2.

PR #238, opened 2019-04-15
SWIFT-268 Store bson_oid_t in ObjectId rather than creating one each time we need it

patrickfreed:
[SWIFT-268](https://jira.mongodb.org/browse/SWIFT-268)

As part of this update, I refactored the public `oid` property on `ObjectId` that returns the string representation to `hex`, which is more descriptive. It also aligns with Go driver's `Hex()` and Java driver's `toHexString()`. 

I also refactored `init(fromPointer:)` to `init(copying:)`, similar to the memory leak fix refactors.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -164,7 +164,7 @@ public struct ServerDescriptionChangedEvent: MongoEvent, InitializableFromOpaque
         self.connectionId = ConnectionId(mongoc_apm_server_changed_get_host(event))
         var oid = bson_oid_t()
         mongoc_apm_server_changed_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
         self.previousDescription = ServerDescription(mongoc_apm_server_changed_get_previous_description(event))
         self.newDescription = ServerDescription(mongoc_apm_server_changed_get_new_description(event))
     }
@@ -189,7 +189,7 @@ public struct ServerOpeningEvent: MongoEvent, InitializableFromOpaquePointer {
         self.connectionId = ConnectionId(mongoc_apm_server_opening_get_host(event))
         var oid = bson_oid_t()
         mongoc_apm_server_opening_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
     }
 }
 
@@ -212,7 +212,7 @@ public struct ServerClosedEvent: MongoEvent, InitializableFromOpaquePointer {
         self.connectionId = ConnectionId(mongoc_apm_server_closed_get_host(event))
         var oid = bson_oid_t()
         mongoc_apm_server_closed_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
     }
 }
 
@@ -237,7 +237,7 @@ public struct TopologyDescriptionChangedEvent: MongoEvent, InitializableFromOpaq
     fileprivate init(_ event: OpaquePointer) {
         var oid = bson_oid_t()
         mongoc_apm_topology_changed_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
         self.previousDescription = TopologyDescription(mongoc_apm_topology_changed_get_previous_description(event))
         self.newDescription = TopologyDescription(mongoc_apm_topology_changed_get_new_description(event))
     }
@@ -258,7 +258,7 @@ public struct TopologyOpeningEvent: MongoEvent, InitializableFromOpaquePointer {
     fileprivate init(_ event: OpaquePointer) {
         var oid = bson_oid_t()
         mongoc_apm_topology_opening_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
     }
 }
 
@@ -277,7 +277,7 @@ public struct TopologyClosedEvent: MongoEvent, InitializableFromOpaquePointer {
     fileprivate init(_ event: OpaquePointer) {
         var oid = bson_oid_t()
         mongoc_apm_topology_closed_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
     }
 }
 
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -772,35 +788,27 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
-        self.timestamp = UInt32(bson_oid_get_time_t(oid_t))
+    /// Initializes an `ObjectId` from an `UnsafePointer<bson_oid_t>` by copying the underlying `bson_oid_t`.
+    internal init(copying oid_t: UnsafePointer<bson_oid_t>) {
+        self.oid = oid_t.pointee
     }
 
     /// Returns the provided string as a `bson_oid_t`.
     /// - Throws:
     ///   - `UserError.invalidArgumentError` if the parameter string does not correspond to a valid `ObjectId`.
     internal static func toLibBSONType(_ str: String) throws -> bson_oid_t {
```
kmahar: is this method still used anywhere or can we delete it?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -732,25 +733,40 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
-        bson_oid_init(&oid_t, nil)
-        self.init(fromPointer: &oid_t)
+        var oid = bson_oid_t()
+        bson_oid_init(&oid, nil)
+        self.oid = oid
     }
 
     /// Initializes an `ObjectId` from the provided `String`. Assumes that the given string is a valid ObjectId.
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(fromString oid: String) {
```
kmahar: any opinions on whether we should keep this parameter label? I feel like just `from` is more in line with the rest of the driver, and the `String` isn't needed to disambiguate from the decodable initializer.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: nope, removed.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: I agree `fromString` should probably be replaced. In fact, I kind of think the entire initializer should be removed, and `init?(ifValid: String)` should be the only one, with the `ifValid` label being replaced with either `from` or `hex` or `fromHex`. 

The libbson docs do not mention what happens if `bson_oid_init_from_string` is called with an invalid string, so I guess we can say it's undefined behavior. I really think we should minimize the amount of UB easily available through our API if at all possible. Thoughts?

kmahar: I like the idea of having a single failable initializer. that is consistent with how we handle `Decimal128`s.

patrickfreed: Refactored `init?(ifValid: String)` to just `init?(_:  String)`. Let me know if you think it'd be better with a label. Also, removed the un-failable init.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -732,36 +733,44 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
+    public init() {
+        var oid = bson_oid_t()
+        bson_oid_init(&oid, nil)
         self.oid = oid
-        var oid_t = bson_oid_t()
-        bson_oid_init_from_string(&oid_t, oid)
-        self.timestamp = UInt32(bson_oid_get_time_t(&oid_t))
     }
 
     /// Initializes an `ObjectId` from the provided `String`. Returns `nil` if the string is not a valid
```
kmahar: maybe we should say it's a hex string here?

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I could go either way.... it might be useful as an indicator about the type of string, but maybe that's not that necessary.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: done

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: @mbroadst any strong opinions on this one? I feel like I could go either way too. The one benefit of doing it without a label is that it looks just like it would in the shell, though I doubt we generally make many driver design choices based on how the shell behaves.

Sources/MongoSwift/APM.swift
```diff
@@ -164,7 +164,7 @@ public struct ServerDescriptionChangedEvent: MongoEvent, InitializableFromOpaque
         self.connectionId = ConnectionId(mongoc_apm_server_changed_get_host(event))
         var oid = bson_oid_t()
         mongoc_apm_server_changed_get_topology_id(event, &oid)
-        self.topologyId = ObjectId(fromPointer: &oid)
+        self.topologyId = ObjectId(copying: &oid)
```
mbroadst: Consider adhering to the lessons learned with the Swift 5 DocumentIterator change, and switch to using `with*Pointer` methods here:
```
var oid = bson_oid_t()
withUnsafeMutablePointer(to: oid) { oidPtr in
  mongoc_apm_server_changed_get_topology_id(event, oidPtr)
  self.topologyId = ObjectId(copying: oidPtr)
...
```

I think this makes the access more explicit, as well as the expected lifetime

mbroadst: I see a few recurrences of this pattern below, so I'd say if we change it here let's do it everywhere

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -135,13 +135,13 @@ final class BSONValueTests: MongoSwiftTestCase {
             return
         }
 
-        expect(_id.oid).to(equal(objectId.oid))
+        expect(_id.hex).to(equal(objectId.hex))
```
mbroadst: have we lost equality of the discrete type? Can we still do: `expect(_id).to(equal(objectId))`, and if so should a test be added?

patrickfreed: Nope, it just wasn't tested here probably because the underlying `Equatable` conformance was just comparing the hex strings. Since I've updated the conformance to something else, I added a test here and below.

Sources/MongoSwift/APM.swift
patrickfreed: I've updated all the `get_topology_id` calls to use `withUnsafeMutablePointer` here and below.

I've also just gone ahead and removed the pointer initializer and replaced it with one that just takes in a straight `bson_oid_t`. We can just pass the  `bson_oid_t` in by value.

Tests/MongoSwiftTests/BSONValueTests.swift
mbroadst: Great, thanks

Sources/MongoSwift/APM.swift
mbroadst: Nice!

PR #241, opened 2019-04-17
SWIFT-17: add evergreen linux support

mbroadst:
- update evergreen config
  - install swift using swiftenv
  - add support for skipping tests by topology
  - add `find-cmake.sh` script
  - use patrick's new numeric helpers to avoid topology issues
  - ensure all tests of a commands response use `doubleValue` for `ok`
  - skip command monitoring tests on sharded topologies
  - reduce to a single mongos if multiple are provided
  - add 4.2 and 5.0 variants
  - disable macos tests until we can figure out how install without
    root permissions

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -11,7 +11,7 @@ command_type: system
 # Protect ourself against rogue test case, or curl gone wild, that runs forever
 # Good rule of thumb: the averageish length a task takes, times 5
 # That roughly accounts for variable system performance for various buildvariants
-exec_timeout_secs: 1800 # 6 minutes is the longest we'll ever run
+exec_timeout_secs: 1800 # 30 minutes is the longest we'll ever run
 
 # What to do when evergreen hits the timeout (`post:` tasks are run automatically)
 timeout:
@@ -41,7 +41,7 @@ functions:
               CURRENT_VERSION=latest
            fi
 
-           export DRIVERS_TOOLS="$(pwd)/../drivers-tools"
+           export DRIVERS_TOOLS="$(dirname $(pwd))/drivers-tools"
            export PROJECT_DIRECTORY="$(pwd)"
 
            # Python has cygwin path problems on Windows. Detect prospective mongo-orchestration home directory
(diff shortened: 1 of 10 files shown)


Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -37,10 +38,21 @@ class MongoSwiftTestCase: XCTestCase {
     /// If the environment variable does not exist, this will use a default of "mongodb://127.0.0.1/".
     static var connStr: String {
         if let connStr = ProcessInfo.processInfo.environment["MONGODB_URI"] {
+            if topologyType == .sharded {
+                let uri = mongoc_uri_new(connStr)
```
patrickfreed: This needs to be freed with `mongoc_uri_destory`.

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -37,10 +38,21 @@ class MongoSwiftTestCase: XCTestCase {
     /// If the environment variable does not exist, this will use a default of "mongodb://127.0.0.1/".
     static var connStr: String {
         if let connStr = ProcessInfo.processInfo.environment["MONGODB_URI"] {
+            if topologyType == .sharded {
```
kmahar: nit: prefer explicit `self.topologyType` here (swiftlint rule for `explicit_self` seems to miss like... everything)

Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -10,6 +10,11 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
     }
 
     func testCommandMonitoring() throws {
+        if MongoSwiftTestCase.topologyType == .sharded {
+            print("Skipping test case because of unsupported topology type \(MongoSwiftTestCase.topologyType)")
```
patrickfreed: Might want to abstract this message into a TestUtils function since we print a similar one in another place.

.evergreen/config.yml
```diff
@@ -449,67 +311,27 @@ axes:
...
-     - name: "test-3.2-replica_set"
-     - name: "test-3.2-sharded_cluster"
-     - name: "test-3.2-standalone"
-     - name: "test-3.0-replica_set"
-     - name: "test-3.0-sharded_cluster"
-     - name: "test-3.0-standalone"
-     - name: "test-2.6-replica_set"
-     - name: "test-2.6-sharded_cluster"
-     - name: "test-2.6-standalone"
+     - ".latest"
```
patrickfreed: does this just do matching based on tags then?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -37,10 +38,21 @@ class MongoSwiftTestCase: XCTestCase {
...
+                let uri = mongoc_uri_new(connStr)
+                let hosts = mongoc_uri_get_hosts(uri)!
```
kmahar: >  Returns NULL if this URI’s scheme is “mongodb+srv://”.

do we care about checking for that case?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -37,10 +38,21 @@ class MongoSwiftTestCase: XCTestCase {
...
+                let hostAndPort = withUnsafeBytes(of: hosts.pointee.host_and_port) { rawPtr -> String in
+                    let ptr = rawPtr.baseAddress!.assumingMemoryBound(to: CChar.self)
```
kmahar: baseAddress can be nil if the buffer is empty, but maybe we are confident enough that libmongoc filled this out that we don't care?

Tests/MongoSwiftTests/CommandMonitoringTests.swift
```diff
@@ -10,6 +10,11 @@ final class CommandMonitoringTests: MongoSwiftTestCase {
     }
 
     func testCommandMonitoring() throws {
+        if MongoSwiftTestCase.topologyType == .sharded {
```
kmahar: i think `guard` is more idiomatic in these cases where we return early inside an `if`, but since we will hopefully delete these skips soon anyway i'm fine with leaving as is.

Tests/MongoSwiftTests/TestUtils.swift
kmahar: lol jinx (well you beat me to it but I didn't refresh...)

mbroadst: man, after all that recent business 🤦‍♂️

.evergreen/config.yml
mbroadst: yep!

Tests/MongoSwiftTests/TestUtils.swift
mbroadst: added a few guards, I think it's unlikely we can get to that state anymore

Tests/MongoSwiftTests/TestUtils.swift
mbroadst: added a guard for this

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -312,7 +312,9 @@ final class MongoCollectionTests: MongoSwiftTestCase {
     func testDropIndexByModel() throws {
         let model = IndexModel(keys: ["cat": 1])
         expect(try self.coll.createIndex(model)).to(equal("cat_1"))
-        expect(try self.coll.dropIndex(model)["ok"]).to(bsonEqual(1.0))
+
+        let res = try self.coll.dropIndex(model)
+        expect((res["ok"] as? BSONNumber)?.doubleValue).to(bsonEqual(1.0))
```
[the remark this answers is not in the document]
mbroadst: created https://jira.mongodb.org/browse/SWIFT-398 to track

Tests/MongoSwiftTests/CommandMonitoringTests.swift
mbroadst: I'm going to punt on this for now, I think we should have a _far_ more explicit, DRY way of determining if a test should be skipped in the future. Let's leave this to that project.

mbroadst:
@kmahar https://jira.mongodb.org/browse/SWIFT-399 to track macos builds on EVG

.evergreen/run-tests.sh
```diff
@@ -2,28 +2,25 @@
...
-   export MONGOC_TEST_SSL_CA_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/ca.pem"
-fi
+# override where we look for libmongoc
+export LD_LIBRARY_PATH="${INSTALL_DIR}/lib"
+export DYLD_LIBRARY_PATH="${INSTALL_DIR}/lib"
 
-echo "Running $AUTH tests over $SSL, connecting to $MONGODB_URI"
-make test
+swiftenv local $SWIFT_VERSION
+MONGODB_TOPOLOGY=${TOPOLOGY} MONGODB_URI=$MONGODB_URI make test
```
patrickfreed: So when looking at the evg output you posted in slack, it's pretty hard to find the log output for the tests, and it's not at all possible in the evg ui to navigate at a per-test level. Should it be part of future work to prettify this output, or is it worth doing now as part of this PR?

e.g. in the tools we have:
image

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -37,10 +38,31 @@ class MongoSwiftTestCase: XCTestCase {
...
+            if self.topologyType == .sharded {
+                guard let uri = mongoc_uri_new(connStr) else {
+                    return connStr
+                }
+
+                defer {
+                    mongoc_uri_destroy(uri)
+                }
+
+                guard let hosts = mongoc_uri_get_hosts(uri)! else {
```
kmahar: I think you need to remove the `!` here?

.evergreen/run-tests.sh
mbroadst: I think we should absolutely do this work, but in the future. This work today was meant to prove we could run on these configurations, but there exists an outstanding good deal of work to make our evergreen lives much better.

PR #243, opened 2019-04-18
SWIFT-399: enable evg builds on OSX hosts

mbroadst:
This is dependent on a branch of `swiftenv` on my local github
account to work, that will hopefully be resolved shortly. Other
items of note here are that we only build on the macos-1014 hosts
due to XCode version requirements, and some tricky linker lines
to accommodate rpath linking to libmongoc.

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -9,6 +9,7 @@ SWIFT_VERSION=${SWIFT_VERSION:-4.2}
 INSTALL_DIR="${PROJECT_DIRECTORY}/opt"
 TOPOLOGY=${TOPOLOGY:-single}
 OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+EXTRA_FLAGS="-Xlinker -rpath -Xlinker ${INSTALL_DIR}/lib"
 
 # enable swiftenv
 export SWIFTENV_ROOT="${INSTALL_DIR}/swiftenv"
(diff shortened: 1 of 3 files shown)


.evergreen/run-tests.sh
```diff
@@ -21,6 +22,12 @@ export PKG_CONFIG_PATH="${INSTALL_DIR}/lib/pkgconfig"
 # override where we look for libmongoc
 export LD_LIBRARY_PATH="${INSTALL_DIR}/lib"
 export DYLD_LIBRARY_PATH="${INSTALL_DIR}/lib"
+export DEVELOPER_DIR=/Applications/Xcode10.1.app
 
 swiftenv local $SWIFT_VERSION
-MONGODB_TOPOLOGY=${TOPOLOGY} MONGODB_URI=$MONGODB_URI make test
+
+# build the driver
+swift build $EXTRA_FLAGS
```
patrickfreed: [nit] `swift test` will build the driver before running tests, so this isn't exactly necessary. If you purposefully want a separate stage for building then this is fine, though.

mbroadst: I considered calling just `swift test`, but wanted to make the two calls explicit, especially since only one of the calls requires passing additional environment variables in.

.evergreen/run-tests.sh
```diff
@@ -21,6 +22,12 @@ export PKG_CONFIG_PATH="${INSTALL_DIR}/lib/pkgconfig"
 # override where we look for libmongoc
 export LD_LIBRARY_PATH="${INSTALL_DIR}/lib"
 export DYLD_LIBRARY_PATH="${INSTALL_DIR}/lib"
+export DEVELOPER_DIR=/Applications/Xcode10.1.app
```
kmahar: how are we testing swift 5 if we are always using xcode 10.1?

mbroadst: This is a little misleading I now see. I'm pointing at a `DEVELOPER_DIR` for xcode 10.1, so that we have the right OSX SDK in place (not swift), but we're actually using the Swift 5 toolchain downloaded by swiftenv.

PR #244, opened 2019-04-18
SWIFT-211, SWIFT-353 Update example projects 

kmahar:
This PR updates our example projects to use more recent versions of dependencies, adds READMEs to the Vapor and Kitura examples, and ensures that clients are not used across threads in the Vapor and Kitura examples.


Examples/Kitura/Package.resolved
```diff
@@ -6,116 +6,107 @@
         "repositoryURL": "https://github.com/IBM-Swift/BlueSocket.git",
         "state": {
           "branch": null,
-          "revision": "ce36f72f2c36b88f847de50c8c0ed8021ff80097",
-          "version": "1.0.16"
+          "revision": "ccc671b022bc60e5fffcb2aa15f499c6331bfd18",
```
kmahar: should we even bother keeping the package.resolveds around in these samples?

Examples/loadExampleData.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
-mongo home --eval "db.kittens.insert([{\"name\":\"roscoe\",\"color\":\"tan\"},{\"name\":\"chester\",\"color\":\"orange\"}])"
+mongo home --eval "db.kittens.insert([{\"name\":\"roscoe\",\"color\":\"orange\"},{\"name\":\"chester\",\"color\":\"tan\"}])"
```
kmahar: wanted the example data to be accurate 😛

Examples/Kitura/Package.resolved
mbroadst: I'm down to remove them. The package lock files for node are this sort of endless PITA because there are always conflicts. I don't think we have that problem here, but I'm not attached to them.

Examples/loadExampleData.sh
mbroadst: my apologies to the boys!

kmahar: 🐈🐈

kmahar:
opened https://jira.mongodb.org/browse/SWIFT-403 about Perfect example

Examples/Kitura/Package.resolved
kmahar: I think they're possibly useful for the driver so we can coordinate on exactly what version we've been using, but since these example projects are pretty stagnant and basic it doesn't really add much

PR #245, opened 2019-04-18
SWIFT-373: Introduce Operation protocol, CountOperation, and DistinctOperation 

kmahar:
Part 1 of many. This introduces the `Operation` protocol and migrates the non-cursor `MongoCollection` "read" methods to use it.

### DIFF
--- a/Sources/MongoSwift/Operations/CountOperation.swift
+++ b/Sources/MongoSwift/Operations/CountOperation.swift
@@ -1,34 +1,5 @@
 import mongoc
 
-/// An operation corresponding to a "count" command on a collection.
-internal struct CountOperation<T: Codable>: Operation {
-    private let collection: MongoCollection<T>
-    private let filter: Document
-    private let options: CountOptions?
-
-    internal init(collection: MongoCollection<T>,
-                  filter: Document,
-                  options: CountOptions?) {
-        self.collection = collection
-        self.filter = filter
-        self.options = options
-    }
-
-    internal func execute() throws -> Int {
-        let opts = try collection.encoder.encode(self.options)
-        let rp = self.options?.readPreference?._readPreference
-        var error = bson_error_t()
-        // because we already encode skip and limit in the options,
-        // pass in 0s so we don't get duplicate parameter errors.
-        let count = mongoc_collection_count_with_opts(
-            self.collection._collection, MONGOC_QUERY_NONE, self.filter.data, 0, 0, opts?.data, rp, &error)
-
-        if count == -1 { throw parseMongocError(error) }
-
-        return Int(count)
-    }
-}
-
 /// Options to use when executing a `count` command on a `MongoCollection`.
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/Operations/CountOperation.swift
```diff
@@ -0,0 +1,75 @@
+import mongoc
+
+/// An operation corresponding to a "count" command on a collection.
+internal struct CountOperation<T: Codable>: Operation {
+    private let collection: MongoCollection<T>
+    private let filter: Document
```
kmahar: My one question on this PR is how much we care to duplicate documentation on what these properties are. They are already all documented on the `count` method... is it worth pasting that here too? (same q applies to distinct.)

mbroadst: as long as it doesn't mess with your jazzy count, I don't see a compelling reason to duplicate options here

Sources/MongoSwift/Operations/CountOperation.swift
```diff
@@ -0,0 +1,75 @@
...
+            self.collection._collection, MONGOC_QUERY_NONE, self.filter.data, 0, 0, opts?.data, rp, &error)
+
+        if count == -1 { throw parseMongocError(error) }
+
+        return Int(count)
+    }
+}
+
+/// Options to use when executing a `count` command on a `MongoCollection`.
+public struct CountOptions: Encodable {
```
mbroadst: [nit]: you use the `CountOptions` in the operation above, consider moving the definition before the operation definition.

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -0,0 +1,8 @@
+/// A protocol for operation types to conform to. An `Operation` instance corresponds to any single operation a user
```
mbroadst: I don't remember what the rules are for comments, but I think y'all moved to c-style comments at a certain point?

Sources/MongoSwift/Operations/CountOperation.swift
kmahar: it's all private/internal, so jazzy won't care.

Sources/MongoSwift/Operations/Operation.swift
kmahar: we standardized on (from development guide):
> We use C-style documentation blocks (`/** ... */`) for documentation longer than 3 lines, and triple-slash (`///`) for shorter documentation.

patrickfreed: I've been doing single and double line docstrings as `///` and c-style for everything longer.

Sources/MongoSwift/Operations/CountOperation.swift
kmahar: done here and for distinct

PR #246, opened 2019-04-19
SWIFT-373 Refactor MongoCollection write methods to use operations

kmahar:
This PR does the following:
- Move `insertOne` code into a new `InsertOneOperation` type
- Introduce a `DeleteOperation` type, used for both `deleteOne` and `deleteMany`.
- Introduce an `UpdateOperation` type, used for `updateOne`, `updateMany`, and `replaceOne`.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -21,24 +21,8 @@ extension MongoCollection {
      */
     @discardableResult
     public func insertOne(_ value: CollectionType, options: InsertOneOptions? = nil) throws -> InsertOneResult? {
-        let document = try self.encoder.encode(value).withID()
-        let opts = try self.encoder.encode(options)
-        var error = bson_error_t()
-        let reply = Document()
-        guard mongoc_collection_insert_one(self._collection, document.data, opts?.data, reply.data, &error) else {
-            throw getErrorFromReply(bsonError: error, from: reply)
-        }
-
-        guard isAcknowledged(options?.writeConcern) else {
-            return nil
-        }
-
-        guard let insertedId = try document.getValue(for: "_id") else {
-            // we called `withID()`, so this should be present.
-            fatalError("Failed to get value for _id from document")
-        }
-
-        return InsertOneResult(insertedId: insertedId)
+        let operation = InsertOneOperation(collection: self, value: value, options: options)
+        return try operation.execute()
     }
 
     /**
@@ -88,23 +72,12 @@ extension MongoCollection {
     public func replaceOne(filter: Document,
                            replacement: CollectionType,
                            options: ReplaceOptions? = nil) throws -> UpdateResult? {
-        let replacementDoc = try self.encoder.encode(replacement)
-        let opts = try self.encoder.encode(options)
-        let reply = Document()
-        var error = bson_error_t()
-        guard mongoc_collection_replace_one(
-            self._collection, filter.data, replacementDoc.data, opts?.data, reply.data, &error) else {
-            throw getErrorFromReply(bsonError: error, from: reply)
-        }
-
-        guard isAcknowledged(options?.writeConcern) else {
-            return nil
-        }
-
-        return try self.decoder.internalDecode(
-                UpdateResult.self,
-                from: reply,
-                withError: "Couldn't understand response from the server.")
+        let operation = UpdateOperation(collection: self,
+                                        filter: filter,
+                                        update: try self.encoder.encode(replacement),
+                                        options: options,
+                                        type: .replaceOne)
+        return try operation.execute()
     }
 
     /**
@@ -126,22 +99,12 @@ extension MongoCollection {
      */
     @discardableResult
     public func updateOne(filter: Document, update: Document, options: UpdateOptions? = nil) throws -> UpdateResult? {
-        let opts = try self.encoder.encode(options)
-        let reply = Document()
-        var error = bson_error_t()
-        guard mongoc_collection_update_one(
-            self._collection, filter.data, update.data, opts?.data, reply.data, &error) else {
-            throw getErrorFromReply(bsonError: error, from: reply)
-        }
-
-        guard isAcknowledged(options?.writeConcern) else {
-            return nil
-        }
-
-        return try self.decoder.internalDecode(
-                UpdateResult.self,
-                from: reply,
-                withError: "Couldn't understand response from the server.")
+        let operation = UpdateOperation(collection: self,
+                                        filter: filter,
+                                        update: update,
+                                        options: options,
+                                        type: .updateOne)
+        return try operation.execute()
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -261,7 +187,7 @@ extension MongoCollection {
      *
      * - Returns: Whether the operation will use an acknowledged write concern
      */
-    fileprivate func isAcknowledged(_ writeConcern: WriteConcern?) -> Bool {
+    internal func isAcknowledged(_ writeConcern: WriteConcern?) -> Bool {
```
kmahar: this needed to be `internal` because we use it in operations types now.

Sources/MongoSwift/Operations/UpdateOperation.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    internal func execute() throws -> UpdateResult? {
+        let opts = try self.collection.encoder.encode(self.options)
+        let reply = Document()
+        var error = bson_error_t()
+
+        var result: Bool!
+        switch self.type {
+        case .updateOne:
+            result = mongoc_collection_update_one(
```
kmahar: I did try the approach of defining a lookup map here that would map an `UpdateType` to a libmongoc function which we would retrieve and then call, since all the parameters we pass in are identical and then we could just do something like `result = updateFunc[self.type]!(params...)`.

however, that provided challenging:
- genericized types are not allowed to have static properties, so I could not define the map in here.
- I tried just having it free-floating, but I can't reference the nested `UpdateType` as `UpdateOperation.UpdateType` - it needs to be `UpdateOperation<Type1, Type2>.UpdateType`... which is strange, we'd have to pick arbitrary concrete types, even though they aren't actually relevant to the `UpdateType`. 

if anyone feels strongly I could move `UpdateType` to not be nested, but I do feel like it belongs within the operation type.

Sources/MongoSwift/Operations/UpdateOperation.swift
```diff
@@ -0,0 +1,134 @@
+import mongoc
+
+internal protocol UpdateOperationOptions: Encodable {
```
kmahar: this protocol was necessary because `UpdateOne`/`UpdateMany` use a different options type from `ReplaceOne`.

Sources/MongoSwift/Operations/UpdateOperation.swift
patrickfreed: You could make the enum cases associated with `Int`s and make the map `[Int: <libmongoc function type>]`, then index by the cases raw value.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -88,23 +72,12 @@ extension MongoCollection {
...
-            return nil
-        }
-
-        return try self.decoder.internalDecode(
-                UpdateResult.self,
-                from: reply,
-                withError: "Couldn't understand response from the server.")
+        let operation = UpdateOperation(collection: self,
+                                        filter: filter,
+                                        update: try self.encoder.encode(replacement),
```
mbroadst: should we be more explicit with throwable methods by calling the method outside of the operation initializer?
```
let updateDoc = try self.encoder.encode(replacement)
let operation = UpdateOperation(...
```

Sources/MongoSwift/Operations/DeleteOperation.swift
```diff
@@ -0,0 +1,73 @@
...
+
+        return try self.collection.decoder.internalDecode(
+                DeleteResult.self,
+                from: reply,
+                withError: "Couldn't understand response from the server.")
+    }
+}
+
+/// The result of a `delete` command on a `MongoCollection`.
+public struct DeleteResult: Decodable {
```
mbroadst: consider moving above the operation for the same reason as the options, also feels a little buried down here.

Sources/MongoSwift/Operations/DeleteOperation.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        guard self.collection.isAcknowledged(options?.writeConcern) else {
+            return nil
+        }
+
+        return try self.collection.decoder.internalDecode(
+                DeleteResult.self,
+                from: reply,
+                withError: "Couldn't understand response from the server.")
```
mbroadst: I know this is just copy paste, but maybe we s/decode/understand here? Sounds like we've anthropomorphized the driver 😄

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -21,24 +21,8 @@ extension MongoCollection {
...
-            return nil
-        }
-
-        guard let insertedId = try document.getValue(for: "_id") else {
-            // we called `withID()`, so this should be present.
-            fatalError("Failed to get value for _id from document")
-        }
-
-        return InsertOneResult(insertedId: insertedId)
+        let operation = InsertOneOperation(collection: self, value: value, options: options)
```
mbroadst: did you try to make a `BulkOperation` that serves as the basis for all write operations? It seems to me we could do something like:
```
let ops = [
  InsertOneOperation(collection: self, value: value, options: options)
]

let operation = BulkOperation(operations: ops)
return try operation.execute();
```

Each operation outside of `BulkOperation` would be like a pseudo-operation, and not have a functional `execute` (or even necessarily conform to `Operation`?). `BulkOperation` could then be responsible for switching on the well-defined operation types to determine which libmongoc method to call. 

Food for thought.

Sources/MongoSwift/Operations/UpdateOperation.swift
mbroadst: should we just call this `WriteConcernProvider` and reuse it elsewhere?

Sources/MongoSwift/Operations/UpdateOperation.swift
```diff
@@ -0,0 +1,134 @@
...
+    var writeConcern: WriteConcern? { get }
+}
+
+/// Options to use when executing an `update` command on a `MongoCollection`.
+public struct UpdateOptions: UpdateOperationOptions {
```
mbroadst: This feels like we reduce readability by having to look elsewhere that this struct still conforms to `Encodable`

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: what if we just updated all the write methods to do something like [`insertMany`](https://github.com/mongodb/mongo-swift-driver/blob/c78822f68c0522c701572d96753ed573f10b36e9/Sources/MongoSwift/MongoCollection%2BWrite.swift#L61)? construct a corresponding `BulkWriteModel`, convert the options, and call  `bulkWrite` with that. and then bulkWrite already handles all those different types of writes and we won't be calling all the operation specific methods anymore. (unless if we want to special case when only model is provided, but not sure it matters?) 

then possibly we would move the logic for each of these methods into corresponding operations types. 

we already have a `BulkWriteOperation` that `bulkWrite` constructs and executes, which will be made to conform to `Operation` in a future PR.

patrickfreed: Note that if we update the single write methods to use bulk writes under the hood, we'll need to do some work to parse any `.bulkWriteError`s back into regular `.writeError`s.

kmahar: good point. as the error expert... how much work do you think that is / do you foresee any issues there?

patrickfreed: There's a couple of ways to do it, the easiest being just to catch the bulk write error and translate it to a write error before throwing it again. They're easily convertible since a `.bulkWriteError` is just a `.writeError` that can contain multiple `WriteError`s.

mbroadst: Hah yes, that sounds like _exactly_ what I was proposing. Thanks for pointing out the obvious here, let's definitely explore that path!

kmahar:
closing this PR in favor of #253.

PR #248, opened 2019-04-19
Clean up index methods

kmahar:
This is a precursor to the next operations PR for the index methods.

I've done the following:
- update inaccurate documentation (we had writeConcern listed as a parameter everywhere rather than the options type)
-  route all drop and create commands through single methods, that will be converted to operations in the next PR
- remove the `IndexOptions` parameter from `dropIndex(keys:)`. we were already completely ignoring it and there is no way to send the information the options contain to the server. see Friday afternoon conversation in #drivers for some more context.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -274,6 +274,10 @@ extension MongoCollection {
      */
     @discardableResult
     public func dropIndex(_ name: String, options: DropIndexOptions? = nil) throws -> Document {
+        guard name != "*" else {
+            throw UserError.invalidArgumentError(message:
+                "Invalid index name '*'; use dropIndexes() to drop all indexes")
+        }
         return try _dropIndexes(index: name, options: options)
     }
 
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -175,31 +175,11 @@ extension MongoCollection {
      * Creates an index over the collection for the provided keys with the provided options.
      *
      * - Parameters:
-     *   - model: An `IndexModel` representing the keys and options for the index
-     *   - writeConcern: Optional WriteConcern to use for the command
-     *
-     * - Returns: The name of the created index.
-     *
-     * - Throws:
-     *   - `ServerError.writeError` if an error occurs while performing the write.
-     *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-     *   - `UserError.invalidArgumentError` if the options passed in form an invalid combination.
-     *   - `EncodingError` if an error occurs while encoding the index specification or options.
-     */
-    @discardableResult
-    public func createIndex(_ forModel: IndexModel, options: CreateIndexOptions? = nil) throws -> String {
-        return try createIndexes([forModel], options: options)[0]
-    }
-
-    /**
-     * Creates an index over the collection for the provided keys with the provided options.
-     *
-     * - Parameters:
      *   - keys: a `Document` specifing the keys for the index
      *   - options: Optional `IndexOptions` to use for the index
-     *   - writeConcern: Optional `WriteConcern` to use for the command
+     *   - options: Optional `CreateIndexOptions` to use for the command
      *
-     * - Returns: The name of the created index
+     * - Returns: The name of the created index.
      *
      * - Throws:
      *   - `ServerError.writeError` if an error occurs while performing the write.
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -258,35 +258,33 @@ extension MongoCollection {
...
      *   - `UserError.invalidArgumentError` if the options passed in form an invalid combination.
+     *   - `EncodingError` if an error occurs while encoding the options.
      */
-    public func dropIndex(_ name: String, options: DropIndexOptions? = nil) throws {
-        let opts = try self.encoder.encode(options)
-        var error = bson_error_t()
-        guard mongoc_collection_drop_index_with_opts(self._collection, name, opts?.data, &error) else {
-            throw parseMongocError(error)
-        }
+    @discardableResult
```
kmahar: we had this returning nothing, but according to the spec should return same result as all the other methods.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -339,12 +335,14 @@ extension MongoCollection {
...
     @discardableResult
     public func dropIndexes(options: DropIndexOptions? = nil) throws -> Document {
-        return try _dropIndexes(options: options)
+        return try _dropIndexes(index: "*", options: options)
     }
 
-    private func _dropIndexes(keys: Document? = nil, options: DropIndexOptions? = nil) throws -> Document {
+    /// Internal helper to drop an index. `index` must either be an index specification document or a
+    /// string index name.
+    private func _dropIndexes(index: BSONValue, options: DropIndexOptions? = nil) throws -> Document {
```
kmahar: I considered trying to make the typing of `index` here a bit more strict to only allow string or document but didn't think it was really worth adding a new protocol or something.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -258,35 +258,33 @@ extension MongoCollection {
...
     /**
      * Drops a single index from the collection by the index name.
      *
      * - Parameters:
      *   - name: The name of the index to drop
-     *   - writeConcern: An optional WriteConcern to use for the command
+     *   - options: Optional `DropIndexOptions` to use for the command
      *
      * - Throws:
+     *   - `ServerError.writeError` if an error occurs while performing the command.
```
kmahar: @patrickfreed, these errors were listed for the other drop methods, so I duplicated them here as well.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -258,35 +258,37 @@ extension MongoCollection {
...
+     *   - `EncodingError` if an error occurs while encoding the options.
      */
-    public func dropIndex(_ name: String, options: DropIndexOptions? = nil) throws {
-        let opts = try self.encoder.encode(options)
-        var error = bson_error_t()
-        guard mongoc_collection_drop_index_with_opts(self._collection, name, opts?.data, &error) else {
-            throw parseMongocError(error)
+    @discardableResult
+    public func dropIndex(_ name: String, options: DropIndexOptions? = nil) throws -> Document {
+        guard name != "*" else {
```
kmahar: the spec says:
"If the string passed is '*', the driver MUST raise an error since more than one index would be dropped."

Sources/MongoSwift/MongoCollection+Indexes.swift
patrickfreed: It was originally omitted since we were using the mongoc drop index helper (which doesn't return write errors, I think). Now that we're using the generic write command function, write errors are possible.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -177,8 +177,9 @@ extension MongoCollection {
      * Creates an index over the collection for the provided keys with the provided options.
      *
      * - Parameters:
-     *   - model: An `IndexModel` representing the keys and options for the index
-     *   - writeConcern: Optional WriteConcern to use for the command
+     *   - keys: a `Document` specifing the keys for the index
+     *   - options: Optional `IndexOptions` to use for the index
+     *   - options: Optional `CreateIndexOptions` to use for the command
```
mbroadst: Looks like `options` should not be listed twice here.

Sources/MongoSwift/MongoCollection+Indexes.swift
mbroadst: nice catch

Sources/MongoSwift/MongoCollection+Indexes.swift
kmahar: good catch, the second parameter is named `commandOptions`.

PR #249, opened 2019-04-21
SWIFT-404 Ensure all index options are correctly sent to the server

kmahar:
Another index-related PR.

A user pointed out in [SWIFT-404](https://jira.mongodb.org/browse/SWIFT-404) that `expireAfter` is sent incorrectly because the server expects `expireAfterSeconds`. 

I ended up writing a test that exercises all of the index options and discovered that a number were sent incorrectly because the server expects a slightly different name than what we have.

Note: The spec used to say `expireAfter` but has since been amended via [SPEC-1111](https://jira.mongodb.org/browse/SPEC-1111) to say `expireAfterSeconds` instead. That ticket also led to some other amendments to the spec that are reflected here, such as updating the `storageEngine` field to be a document rather than a string.

This suggests to me we should consider:
- testing all the options for every public API method get correctly sent to and accepted by the server
- index management spec tests @mbroadst proposed way back in [SWIFT-33](https://jira.mongodb.org/browse/SWIFT-33)


kmahar:
I'm going to put the ticket about spec tests back into "needs triage" so we can discuss it more tomorrow.

mhergon:
Hi,

When the fix will be released?

kmahar:
Hi @mhergon, we plan to release next week.

mhergon:
@kmahar thanks!

PR #251, opened 2019-04-23
SWIFT-373 Add CreateIndexes and DropIndexes operations

kmahar:
#248 did some refactoring so that all index creation would go through a single method, and same for dropping indexes. 
This converts those methods into operation types, `CreateIndexesOperation` and `DropIndexesOperation`.

### DIFF
--- a/Sources/MongoSwift/Operations/CreateIndexesOperation.swift
+++ b/Sources/MongoSwift/Operations/CreateIndexesOperation.swift
@@ -26,8 +26,6 @@ internal struct CreateIndexesOperation<T: Codable>: Operation {
     }
 
     internal func execute() throws -> [String] {
-        let collName = String(cString: mongoc_collection_get_name(self.collection._collection))
-
         var indexData = [Document]()
         for index in self.models {
             var indexDoc = try self.collection.encoder.encode(index)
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/Operations/CreateIndexesOperation.swift
```diff
@@ -0,0 +1,56 @@
...
+    internal init(collection: MongoCollection<T>,
+                  models: [IndexModel],
+                  options: CreateIndexOptions?) {
+        self.collection = collection
+        self.models = models
+        self.options = options
+    }
+
+    internal func execute() throws -> [String] {
+        let collName = String(cString: mongoc_collection_get_name(self.collection._collection))
```
mbroadst: why aren't we using the [name property](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/MongoCollection.swift#L30)?

Sources/MongoSwift/Operations/DropIndexesOperation.swift
```diff
@@ -0,0 +1,40 @@
...
+    internal init(collection: MongoCollection<T>,
+                  index: BSONValue,
+                  options: DropIndexOptions?) {
+        self.collection = collection
+        self.index = index
+        self.options = options
+    }
+
+    internal func execute() throws -> Document {
+        let collName = String(cString: mongoc_collection_get_name(self.collection._collection))
```
mbroadst: same question as above

Sources/MongoSwift/Operations/CreateIndexesOperation.swift
kmahar: great point, updated.

Sources/MongoSwift/Operations/DropIndexesOperation.swift
kmahar: done

PR #253, opened 2019-04-23
SWIFT-373 Convert all write methods to use the bulkWrite API

kmahar:
Reimplements all write methods to use the bulk API instead of the libmongoc helpers.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -22,7 +22,8 @@ extension MongoCollection {
     @discardableResult
     public func insertOne(_ value: CollectionType, options: InsertOneOptions? = nil) throws -> InsertOneResult? {
         do {
-            let result = try self.bulkWrite([InsertOneModel(value)], options: options?.asBulkWriteOpts())
+            let model = InsertOneModel(value)
+            let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
             return try InsertOneResult(from: result)
         } catch {
             throw convertBulkWriteError(error)
@@ -47,7 +48,8 @@ extension MongoCollection {
      */
     @discardableResult
     public func insertMany(_ values: [CollectionType], options: InsertManyOptions? = nil) throws -> InsertManyResult? {
-        let result = try self.bulkWrite(values.map { InsertOneModel($0) }, options: options)
+        let models = values.map { InsertOneModel($0) }
+        let result = try self.bulkWrite(models, options: options)
         return InsertManyResult(from: result)
     }
 
@@ -77,7 +79,7 @@ extension MongoCollection {
                                         replacement: replacement,
                                         collation: options?.collation,
                                         upsert: options?.upsert)
-            let result = try self.bulkWrite([model], options: options?.asBulkWriteOpts())
+            let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
             return try UpdateResult(from: result)
         } catch {
             throw convertBulkWriteError(error)
@@ -109,7 +111,7 @@ extension MongoCollection {
                                        arrayFilters: options?.arrayFilters,
                                        collation: options?.collation,
                                        upsert: options?.upsert)
-            let result = try self.bulkWrite([model], options: options?.asBulkWriteOpts())
+            let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
             return try UpdateResult(from: result)
         } catch {
             throw convertBulkWriteError(error)
@@ -141,7 +143,7 @@ extension MongoCollection {
                                         arrayFilters: options?.arrayFilters,
                                         collation: options?.collation,
                                         upsert: options?.upsert)
-            let result = try self.bulkWrite([model], options: options?.asBulkWriteOpts())
+            let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
             return try UpdateResult(from: result)
         } catch {
             throw convertBulkWriteError(error)
@@ -168,7 +170,7 @@ extension MongoCollection {
     public func deleteOne(_ filter: Document, options: DeleteOptions? = nil) throws -> DeleteResult? {
         do {
             let model = DeleteOneModel(filter, collation: options?.collation)
-            let result = try self.bulkWrite([model], options: options?.asBulkWriteOpts())
+            let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
             return try DeleteResult(from: result)
         } catch {
             throw convertBulkWriteError(error)
@@ -195,7 +197,7 @@ extension MongoCollection {
     public func deleteMany(_ filter: Document, options: DeleteOptions? = nil) throws -> DeleteResult? {
         do {
             let model = DeleteManyModel(filter, collation: options?.collation)
-            let result = try self.bulkWrite([model], options: options?.asBulkWriteOpts())
+            let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
             return try DeleteResult(from: result)
         } catch {
             throw convertBulkWriteError(error)
@@ -207,12 +209,12 @@ extension MongoCollection {
 private protocol BulkWriteOptionsConvertible {
     var bypassDocumentValidation: Bool? { get }
     var writeConcern: WriteConcern? { get }
-    func asBulkWriteOpts() -> BulkWriteOptions
+    func asBulkWriteOptions() -> BulkWriteOptions
 }
 
 /// Default implementation of the protocol.
 private extension BulkWriteOptionsConvertible {
-    func asBulkWriteOpts() -> BulkWriteOptions {
+    func asBulkWriteOptions() -> BulkWriteOptions {
         return BulkWriteOptions(bypassDocumentValidation: self.bypassDocumentValidation,
                                 writeConcern: self.writeConcern)
     }
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -21,24 +21,12 @@ extension MongoCollection {
      */
     @discardableResult
     public func insertOne(_ value: CollectionType, options: InsertOneOptions? = nil) throws -> InsertOneResult? {
-        let document = try self.encoder.encode(value).withID()
-        let opts = try self.encoder.encode(options)
-        var error = bson_error_t()
-        let reply = Document()
-        guard mongoc_collection_insert_one(self._collection, document.data, opts?.data, reply.data, &error) else {
-            throw getErrorFromReply(bsonError: error, from: reply)
+        do {
+            let result = try self.bulkWrite([InsertOneModel(value)], options: options?.asBulkWriteOpts())
+            return try InsertOneResult(from: result)
+        } catch {
+            throw convertBulkWriteError(error)
         }
-
-        guard isAcknowledged(options?.writeConcern) else {
-            return nil
-        }
-
-        guard let insertedId = try document.getValue(for: "_id") else {
-            // we called `withID()`, so this should be present.
-            fatalError("Failed to get value for _id from document")
-        }
-
-        return InsertOneResult(insertedId: insertedId)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/MongoError.swift
```diff
@@ -278,6 +278,21 @@ private func getBulkWriteErrorFromReply(
     )
 }
 
+/// Internal function used by write methods that are implemented via the bulk API. Converts a
+/// ServerError.bulkWriteError that resulted from a single write into a ServerError.writeError.
+internal func convertBulkWriteError(_ error: Error) -> Error {
```
kmahar: if you have a better name idea for this please let me know 😛

Sources/MongoSwift/MongoError.swift
```diff
@@ -278,6 +278,22 @@ private func getBulkWriteErrorFromReply(
     )
 }
 
+/// Internal function used by write methods performing single writes that are implemented via the bulk API. If the
+/// input error is a ServerError.bulkWriteError, converts it to a ServerError.writeError. Otherwise, returns the
+/// error as is.
+internal func convertBulkWriteError(_ error: Error) -> Error {
+    if case let ServerError.bulkWriteError(bulkWriteErrors, writeConcernError, _, errorLabels) = error {
```
kmahar: I always forget this `if case let` syntax is a thing, first time I've actually used it 😱

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -432,4 +376,21 @@ public struct UpdateResult: Decodable {
...
+            return nil
+        }
+        self.matchedCount = result.matchedCount
+        self.modifiedCount = result.modifiedCount
+        self.upsertedCount = result.upsertedCount
+        if result.upsertedCount == 1 {
+            guard let id = result.upsertedIds[0] else {
+                throw RuntimeError.internalError(message: "BulkWriteResult missing _id for upserted document")
+            }
+            self.upsertedId = AnyBSONValue(id)
```
kmahar: so we made `upsertedId` an `AnyBSONValue` so the whole struct would be automatically `Decodable`. however, this is not really consistent with the other result types that have `_id`s in them. they just use `BSONValue`.

we no longer decode to the struct in this file since we construct it from a `BulkWriteResult` instead, so I thought about switching to a `BSONValue` and making the struct non-decodable. but, turns out we are using the fact that it's `Decodable` in the CRUD tests to get the expected result from the JSON files. but we could change things to decode it manually if want, which would make this struct consistent with the rest of the API. 

thoughts?

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -59,11 +47,7 @@ extension MongoCollection {
      */
     @discardableResult
     public func insertMany(_ values: [CollectionType], options: InsertManyOptions? = nil) throws -> InsertManyResult? {
-        guard !values.isEmpty else {
```
kmahar: `bulkWrite` already performs this check so it seemed redundant.

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -235,48 +193,35 @@ extension MongoCollection {
...
-         * can safely assume it is acknowledged, since the server requires that
-         * getLastErrorDefaults is acknowledged by at least one member. */
-        guard let wc = writeConcern ?? self.writeConcern else {
-            return true
-        }
+/// Protocol indicating that an options type can be converted to a BulkWriteOptions.
+private protocol BulkWriteOptionsConvertible {
+    var bypassDocumentValidation: Bool? { get }
+    var writeConcern: WriteConcern? { get }
+    func asBulkWriteOpts() -> BulkWriteOptions
```
mbroadst: nit: Is there benefit to dropping the `ion` here? Consider calling it `asBulkWriteOptions`

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -59,11 +47,7 @@ extension MongoCollection {
      */
     @discardableResult
     public func insertMany(_ values: [CollectionType], options: InsertManyOptions? = nil) throws -> InsertManyResult? {
-        guard !values.isEmpty else {
-            throw UserError.invalidArgumentError(message: "values cannot be empty")
-        }
-
-        let result = try self.bulkWrite(values.map { InsertOneModel($0) }, options: BulkWriteOptions(from: options))
+        let result = try self.bulkWrite(values.map { InsertOneModel($0) }, options: options)
```
mbroadst: So this one looks different than the others, is it worth taking the extra effort to make all of these methods look the same? In particular I'm wondering why e.g. an `updateMany`s thrown error will not be a complete `BulkWriteError` like `insertMany`, etc etc

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -21,24 +21,12 @@ extension MongoCollection {
...
     @discardableResult
     public func insertOne(_ value: CollectionType, options: InsertOneOptions? = nil) throws -> InsertOneResult? {
-        let document = try self.encoder.encode(value).withID()
-        let opts = try self.encoder.encode(options)
-        var error = bson_error_t()
-        let reply = Document()
-        guard mongoc_collection_insert_one(self._collection, document.data, opts?.data, reply.data, &error) else {
-            throw getErrorFromReply(bsonError: error, from: reply)
+        do {
+            let result = try self.bulkWrite([InsertOneModel(value)], options: options?.asBulkWriteOpts())
```
mbroadst: nit: I think there is inconsistency between creating the model before the `bulkWrite` call and passing it in, versus creating the model in the `bulkWrite` call itself. Slight preference for creating the model before, making it explicit.

Sources/MongoSwift/MongoCollection+Write.swift
mbroadst: I vote for consistency! Can you explain why it needs to be `AnyBSONValue` for the CRUD tests? This would be a breaking change. IIRC we created `AnyBSONValue` for the `_id` case, since those can be any value. Does it make more sense to standardize on using `AnyBSONValue` in any result struct returning `_id` s?

Sources/MongoSwift/MongoError.swift
mbroadst: 🤘

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: nah, renamed.

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: updated all to create the model(s) beforehand.

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: so the CRUD spec (see section [here](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#basic)) actually says that `insertMany` throws a `BulkWriteException`, whereas the rest of these methods (including `updateMany`) all throw just `WriteException`s (our equivalents are `ServerError.bulkWriteError` and `ServerError.writeError`). so while it's funky it is actually in line with the specs. 

I think an `updateMany` is really considered just a single write -- there's an `UpdateManyModel`. But an `insertMany` is considered multiple writes and has to be translated to multiple `InsertOneModel`s.

@patrickfreed may have more input as he reimplemented this as a bulk write initially.

Sources/MongoSwift/MongoCollection+Write.swift
kmahar: so `AnyBSONValue` exists for the purpose of letting you encode/decode `BSONValue`s without having to know their concrete type. It does the type switching for you in its encode/decode methods. The particular motivation was Stitch having various structs that they stored `_id`s in and they wanted to make those structs `Decodable`. 

so, I *think* we used an `AnyBSONValue` in this struct so we could get free `Decodable` conformance for the `upsertedId` field. The `Decodable` conformance comes into play both when decoding server replies to construct them (we no longer do that as of this PR) and when decoding the CRUD test data into "expected" structs (see [here](https://github.com/mongodb/mongo-swift-driver/blob/088f794b537cdf0e10504e1506d4b16018756661/Tests/MongoSwiftTests/CrudTests.swift#L186)). 

Basically, my guess is that we made this choice for our own convenience, not for the user's. 

While it's a bit more annoying on our end, IMO the result structs should all just use `BSONValue`, not `AnyBSONValue`. Otherwise the user has to do like `let id = result.insertedId.value` instead of `let id = result.insertedId`. 

If we switch to just plain `BSONValue`, we could work around this in the crud tests by writing a manual `Decodable` implementation like this:
```swift
init(from decoder: Decoder) throws {
    let container = try decoder.container(keyedBy: CodingKeys.self)
    self.matchedCount = try container.decode(Int.self, forKey: .matchedCount)
    self.modifiedCount = try container.decode(Int.self, forKey: .modifiedCount)
    let id = try container.decode(AnyBSONValue.self, forKey: .upsertedId)
    self.upsertedId = id.value
    ...
}
```

Sources/MongoSwift/MongoCollection+Write.swift
patrickfreed: I think what you said covers it 👍

kmahar:
Follow on question. Should we still have dedicated operations for all of these? 
Downside: lot of boilerplate for an execute method that boils down to like 3 LOC
Upsides: consistency, and we can move the options/results types to the corresponding files

Sources/MongoSwift/MongoError.swift
patrickfreed: I think the name is fine as-is, but we could consider a different approach altogether to reduce some of the boilerplate of catching and rethrowing.

e.g.
```swift
return try convertBulkWriteErrors {
    let model = InsertOneModel(value)
    let result = try self.bulkWrite([model], options: options?.asBulkWriteOptions())
    return try InsertOneResult(from: result)
}
```
The `convertBulkWriteErrors` function would execute the body and convert any bulk write errors it encounters, otherwise throwing everything else and passing through the return value

kmahar:
As an alternative to adding an operation for every method, we might consider just adding a generic `WriteOperation` instead. Something like:
```swift
protocol InitializableFromBulkWriteResult {
    init?(from result: BulkWriteResult) throws
}

struct WriteOperation<CollectionType: Codable, ResultType: InitializableFromBulkWriteResult>: Operation {
    let collection: MongoCollection<CollectionType>
    let model: WriteModel
    let options: BulkWriteConvertible

    func execute() throws -> ResultType {
        do {
            let result = try collection.bulkWrite([model], options: options?.asBulkWriteOptions())
            return try ResultType(from: result)
        } catch {
            throw convertBulkWriteError(error)
        }
    }
}
```

Then something like `replaceOne` would become:
```swift
public func replaceOne(filter: Document,
                       replacement: CollectionType,
                       options: ReplaceOptions? = nil) throws -> UpdateResult? {
    let model = ReplaceOneModel(filter: filter,
                                replacement: replacement,
                                collation: options?.collation,
                                upsert: options?.upsert)
    return try WriteOperation(collection: self, model: model, options: options).execute()
}
```

Sources/MongoSwift/MongoCollection+Write.swift
mbroadst: Thanks for the detailed writeup. I agree, let's make it a `BSONVaue`

kmahar: done

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -428,8 +362,38 @@ public struct UpdateResult: Decodable {
...
+        } else {
+            self.upsertedId = nil
+        }
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case matchedCount, modifiedCount, upsertedId, upsertedCount
+    }
+
+    public init(from decoder: Decoder) throws {
```
kmahar: I tried to just add the decodable conformance in the tests, but if you add an initializer in another module, it must delegate to an initializer from the defining module. so I would've had to add some kind of initializer here anyway that took in and set all of the properties. seemed not worth it, but open to doing so if anyone feels strongly.

PR #255, opened 2019-04-25
Add pipeline option to createCollection and test all createCollection options

kmahar:
When refactoring the database methods to be operations, I decided to check if we had all of the correct options available for `createCollection`. turns out we were missing `pipeline`.

I've added it and also a test that exercises all the possible createCollection options to make sure the server accepts them.

### DIFF
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -64,7 +64,7 @@ public struct CreateCollectionOptions: Encodable, CodingStrategyProvider {
     /// Maximum number of documents allowed in the collection (if capped).
     public let max: Int64?
 
-    /// Determine which storage engine to use.
+    /// Specifies storage engine configuration for this collection.
     public let storageEngine: Document?
 
     /// What validator should be used for the collection.
@@ -77,17 +77,18 @@ public struct CreateCollectionOptions: Encodable, CodingStrategyProvider {
     /// to be inserted.
     public let validationAction: String?
 
-    /// Allows users to specify a default configuration for indexes when creating a collection.
+    /// Specify a default configuration for indexes created on this collection.
     public let indexOptionDefaults: Document?
 
     /// The name of the source collection or view from which to create the view.
     public let viewOn: String?
 
+    /// An array consisting of aggregation pipeline stages. When used with `viewOn`, will create the view by applying
+    /// this pipeline to the source collection or view.
+    public let pipeline: [Document]?
+
     /// Specifies the default collation for the collection.
     public let collation: Document?
-
-    /// A session to associate with this operation.
-    public let session: ClientSession?
 
     /// A write concern to use when executing this command. To set a read or write concern for the collection itself,
     /// retrieve the collection using `MongoDatabase.collection`.
@@ -110,7 +111,7 @@ public struct CreateCollectionOptions: Encodable, CodingStrategyProvider {
 
     private enum CodingKeys: String, CodingKey {
         case capped, autoIndexId, size, max, storageEngine, validator, validationLevel, validationAction,
-             indexOptionDefaults, viewOn, collation, session, writeConcern
+             indexOptionDefaults, viewOn, pipeline, collation, writeConcern
     }
 
     /// Convenience initializer allowing any/all parameters to be omitted or optional.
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -77,18 +77,19 @@ public struct CreateCollectionOptions: Encodable, CodingStrategyProvider {
...
 
+    /// An array consisting of aggregation pipeline stages. When used with `viewOn`, will create the view by applying
+    /// this pipeline to the source collection or view.
+    public let pipeline: [Document]?
+
     /// Specifies the default collation for the collection.
     public let collation: Document?
 
-    /// A session to associate with this operation.
-    public let session: ClientSession?
```
kmahar: went ahead and removed this while I was here, given the new API. @patrickfreed, there are likely more of these cases in the existing API that we will need to remove.

patrickfreed: Yeah, I've removed them all in my wip sessions implementation

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -34,4 +34,34 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
+            storageEngine: ["wiredTiger": ["configString": "access_pattern_hint=random"] as Document],
+            validationAction: "warn",
+            validationLevel: "moderate",
+            validator: ["phone": ["$type": "string"] as Document],
+            writeConcern: try WriteConcern(w: .majority)
+        )
+        expect(try db.createCollection("foo", options: options)).toNot(throwError())
+
+        // test view options
+        let viewOptions = CreateCollectionOptions(
```
patrickfreed: might be worth inserting a document to the original collection and then ensuring the pipeline is actually being used to make the view.

kmahar: maybe I should just run listCollections and look at the returned info? then we can validate everything has been correctly set.

kmahar: not sure if that might be brittle and vary across server versions? thoughts @mbroadst ?

mbroadst: I don't think you need to worry about brittleness, we can skip sections of tests based on server version already.

patrickfreed: I think a test either way is fine. We generally just have to ensure we're passing the right stuff to libmongoc and can rely on their testbed for anything further than that.

kmahar: I ended up doing as I proposed with testing the returned collection info. `CreateCollectionOptions` is now `Decodable` so that we can directly parse the options document returned by `listCollections` back into options and verify they match the ones that we sent.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -94,20 +94,24 @@ public struct CreateCollectionOptions: Encodable, CodingStrategyProvider {
...
     public let writeConcern: WriteConcern?
 
+    // swiftlint:disable redundant_optional_initialization
+    // to get synthesized decodable conformance for the struct, these strategies need default values.
+
     /// Specifies the `DateCodingStrategy` to use for BSON encoding/decoding operations performed by this collection.
     /// It is the responsibility of the user to ensure that any `Date`s already stored in this collection can be
     /// decoded using this strategy.
-    public let dateCodingStrategy: DateCodingStrategy?
+    public var dateCodingStrategy: DateCodingStrategy? = nil
```
kmahar: giving these default values (for the reason I noted by the "disable" above) also meant they needed to be vars so they can be overwritten. I figured that was OK as all options properties are going to become `var`s soon anyway as of [SWIFT-355](https://jira.mongodb.org/browse/SWIFT-355) which is part of the intern plan.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -7,6 +7,13 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
         continueAfterFailure = false
     }
 
+    override func tearDown() {
```
kmahar: it seems that if a test fails a defer doesn't actually get executed.... I had a `defer db.drop` in the test below but whenever the test failed I would get an error that the collection still existed on the next run.

PR #257, opened 2019-04-25
Swift-28 Implement Sessions

patrickfreed:
This PR adds the implementation of client sessions as per the [design](https://docs.google.com/document/d/1n5N8f1Zgzd5Iow5gxiRSoj6FZ8-Gus8I5CUTrW1Gh3s/edit#heading=h.ky3pv4dt04e).

Deviations from the design:
- `session.sessionId` refactored to `session.id`, the "session" prefix seemed redundant
- `session.endSession` -> `session.end` ditto redundancy
- The test plan prescribed performing spec test 9 (ensure all sessions are cleaned up after each test), but without access to the underlying pool this is difficult
- The test plan states that many of the tests should run every function that takes in a session or against every topology with every read preference. This is overkill for what we're implementing. Assuming libmongoc is performing those tests at their level, all we need to do is ensure that every method that accepts a session properly passes it to libmongoc, which we verify in the `testSessionArguments` test. For the other tests, we just use a single method that takes in a session as a representative.
- I removed the `ClientSessionPointer` typealias, since libmongoc doesn't actually export the complete `mongoc_client_session_t` type. Regardless of compiler version, pointers to it will be imported to swift as `OpaquePointer`s. Since those pointers will always be mutable, I only defined `MutableClientSessionPointer`.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -54,7 +54,7 @@ public struct Document {
     internal var storage: DocumentStorage
 
     /// Returns the number of (key, value) pairs stored at the top level of this `Document`.
-    public fileprivate(set) var count: Int
+    public internal(set) var count: Int
 }
 
 /// An extension of `Document` containing its private/internal functionality.
@@ -252,7 +252,7 @@ extension Document {
      * Therefore, this function should be called just before we are about to modify a document - either by
      * setting a value or merging in another doc.
      */
-    private mutating func copyStorageIfRequired() {
+    internal mutating func copyStorageIfRequired() {
         if !isKnownUniquelyReferenced(&self.storage) {
             self.storage = DocumentStorage(copying: self.data)
         }
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -1,35 +1,121 @@
+import Foundation
+import mongoc
+
+#if compiler(>=5.0)
+internal typealias ClientSessionPointer = OpaquePointer
+internal typealias MutableClientSessionPointer = OpaquePointer
+#else
+internal typealias ClientSessionPointer = UnsafePointer<mongoc_client_session_t>
+internal typealias MutableClientSessionPointer = UnsafeMutablePointer<mongoc_client_session_t>
+#endif
+
 /// Options to use when creating a ClientSession.
-internal struct SessionOptions: Encodable {
-    /// Specifies whether read operations should be causally ordered within the session.
-    private let causalConsistency: Bool?
+public struct ClientSessionOptions {}
+
+/// Private wrapper of a mongoc_session_opt_t.
+private class SessionOptWrapper {
+    /// Opaque pointer to a `mongoc_session_opt_t`
+    fileprivate var _opts: OpaquePointer?
+
+    fileprivate init(from options: ClientSessionOptions?) {
+        guard options != nil else {
+            return
+        }
+        self._opts = mongoc_session_opts_new()
+    }
+
+    deinit {
+        guard let opts = self._opts else {
+            return
+        }
+        mongoc_session_opts_destroy(opts)
+    }
 }
 
-/// :nodoc: A session for ordering sequential operations.
-public class ClientSession: Encodable {
+/// A MongoDB client session.
+public final class ClientSession {
+    /// Error thrown when an inactive session is used.
+    internal static var SessionInactiveError: Error = UserError.logicError(message: "Tried to use an inactive session")
+
+    /// Pointer to the underlying `mongoc_client_session_t`.
+    internal fileprivate(set) var _session: ClientSessionPointer?
+
+    /// Returns whether this session has been ended or not.
+    internal var active: Bool { return self._session != nil }
+
+    /// The client used to start this session.
+    public let client: MongoClient
+
+    /// The session ID of this session.
+    public let id: Document
+
+    /// The most recent cluster time seen by this session.
+    /// If no operations have been executed using this session and `advanceClusterTime` has not been called, this will
+    /// be `nil`.
+    public var clusterTime: Document? {
+        guard let time = mongoc_client_session_get_cluster_time(self._session) else {
+            return nil
+        }
+        return Document(copying: time)
+    }
+
+    /// The options used to start this session.
+    public let options: ClientSessionOptions?
+
     /// Initializes a new client session.
-    internal init() {
+    internal init(client: MongoClient, options: ClientSessionOptions? = nil) throws {
+        self.options = options
+        self.client = client
+
+        let opts = SessionOptWrapper(from: options)
+        var error = bson_error_t()
+        guard let session = mongoc_client_start_session(client._client, opts._opts, &error) else {
+            throw parseMongocError(error)
+        }
+        // swiftlint:disable:next force_unwrapping
+        self.id = Document(copying: mongoc_client_session_get_lsid(session)!) // always returns a value
+        self._session = session
+    }
+
+    /// Destroy the underlying `mongoc_client_session_t` and set this session to inactive.
+    /// Does nothing if this session is already inactive.
+    internal func end() {
+        guard self.active else {
+            return
+        }
+        mongoc_client_session_destroy(self._session)
+        self._session = nil
     }
 
     /// Cleans up internal state.
     deinit {
+        self.end()
     }
 
-    /// Finish the session.
-    private func endSession() {
+    /**
+     * Advances the clusterTime for this session to the given time, if it is greater than the current clusterTime.
+     * If the provided clusterTime is less than the current clusterTime, this method has no effect.
+     *
+     * - Parameters:
+     *   - clusterTime: The session's new cluster time, as a `Document` like `["cluster time": Timestamp(...)]`
+     */
+    public func advanceClusterTime(to clusterTime: Document) {
+        mongoc_client_session_advance_cluster_time(self._session, clusterTime.storage.pointer)
     }
 
-    /// The server session id for this session.
-    private var sessionId: Document {
-        return Document()
-    }
+    /// Appends this provided session to an options document for libmongoc interoperability.
+    /// - Throws:
+    ///   - `UserError.logicError` if this session is inactive
+    internal func append(to doc: inout Document) throws {
+        guard self.active else {
+            throw ClientSession.SessionInactiveError
+        }
 
-    /// The cluster time returned by the last operation executed in this session.
-    private var clusterTime: Int64 {
-        return Int64()
-    }
-
-    /// The operation time returned by the last operation executed in this session.
-    private var operationTime: Int64 {
-        return Int64()
+        doc.copyStorageIfRequired()
+        var error = bson_error_t()
+        guard mongoc_client_session_append(self._session, doc.storage.pointer, &error) else {
+            throw parseMongocError(error)
+        }
+        doc.count = Int(bson_count_keys(doc.storage.pointer))
     }
 }
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -1,13 +1,7 @@
 import Foundation
 import mongoc
 
-#if compiler(>=5.0)
-internal typealias ClientSessionPointer = OpaquePointer
 internal typealias MutableClientSessionPointer = OpaquePointer
-#else
-internal typealias ClientSessionPointer = UnsafePointer<mongoc_client_session_t>
-internal typealias MutableClientSessionPointer = UnsafeMutablePointer<mongoc_client_session_t>
-#endif
 
 /// Options to use when creating a ClientSession.
 public struct ClientSessionOptions {}
@@ -38,7 +32,7 @@ public final class ClientSession {
     internal static var SessionInactiveError: Error = UserError.logicError(message: "Tried to use an inactive session")
 
     /// Pointer to the underlying `mongoc_client_session_t`.
-    internal fileprivate(set) var _session: ClientSessionPointer?
+    internal fileprivate(set) var _session: MutableClientSessionPointer?
 
     /// Returns whether this session has been ended or not.
     internal var active: Bool { return self._session != nil }
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -29,7 +29,7 @@ private class SessionOptWrapper {
 /// A MongoDB client session.
 public final class ClientSession {
     /// Error thrown when an inactive session is used.
-    internal static var SessionInactiveError: Error = UserError.logicError(message: "Tried to use an inactive session")
+    internal static var SessionInactiveError = UserError.logicError(message: "Tried to use an inactive session")
 
     /// Pointer to the underlying `mongoc_client_session_t`.
     internal fileprivate(set) var _session: MutableClientSessionPointer?
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -1,8 +1,6 @@
 import Foundation
 import mongoc
 
-internal typealias MutableClientSessionPointer = OpaquePointer
-
 /// Options to use when creating a ClientSession.
 public struct ClientSessionOptions {}
 
@@ -17,12 +15,13 @@ private func withSessionOpts<T>(wrapping options: ClientSessionOptions?,
 }
 
 /// A MongoDB client session.
+/// This class represents a logical session used for ordering sequential operations.
 public final class ClientSession {
     /// Error thrown when an inactive session is used.
     internal static var SessionInactiveError = UserError.logicError(message: "Tried to use an inactive session")
 
     /// Pointer to the underlying `mongoc_client_session_t`.
-    internal fileprivate(set) var _session: MutableClientSessionPointer?
+    internal fileprivate(set) var _session: OpaquePointer?
 
     /// Returns whether this session has been ended or not.
     internal var active: Bool { return self._session != nil }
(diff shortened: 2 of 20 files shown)


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -54,7 +54,7 @@ public struct Document {
     internal var storage: DocumentStorage
 
     /// Returns the number of (key, value) pairs stored at the top level of this `Document`.
-    public fileprivate(set) var count: Int
+    public internal(set) var count: Int
```
patrickfreed: This change (and the one below) are necessary to use this stuff in `append`, in the future i imagine they'll go back to being private or fileprivate when we work on the memory safety epic.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -270,14 +278,17 @@ public class MongoClient {
...
      *   - `UserError.invalidArgumentError` if the options passed are an invalid combination.
+     *   - `UserError.logicError` if the provided session is inactive.
      *   - `EncodingError` if an error is encountered while encoding the options to BSON.
      */
-    public func listDatabases(options: ListDatabasesOptions? = nil) throws -> MongoCursor<Document> {
-        let opts = try self.encoder.encode(options)
-        guard let cursor = mongoc_client_find_databases_with_opts(self._client, opts?.data) else {
+    public func listDatabases(options: ListDatabasesOptions? = nil,
+                              session: ClientSession? = nil) throws -> MongoCursor<Document> {
+        var opts = try self.encoder.encode(options) ?? Document()
```
patrickfreed: This (and all the others like it) construct a new `Document` every time, regardless of whether any options or a session is passed in. I think `bson_t` does some nifty things where this is a stack allocation if the doc is small enough, so I don't think this is a big performance issue.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -383,6 +391,18 @@ private class FindAndModifyOptions {
...
     }
 
+    fileprivate func setSession(_ session: ClientSession?) throws {
+        guard let session = session else {
+            return
+        }
+        var doc = Document()
+        try session.append(to: &doc)
+
+        guard mongoc_find_and_modify_opts_append(self._options, doc.data) else {
```
patrickfreed: this is a very silly api provided by libmongoc

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -163,13 +163,17 @@ extension MongoCollection {
...
      *   - `UserError.invalidArgumentError` if the options passed in form an invalid combination.
+     *   - `UserError.logicError` if the provided session is inactive.
      *   - `EncodingError` if an error occurs while encoding the index specification or options.
      */
     @discardableResult
     public func createIndex(_ keys: Document,
                             options: IndexOptions? = nil,
-                            commandOptions: CreateIndexOptions? = nil) throws -> String {
-        return try createIndexes([IndexModel(keys: keys, options: options)], options: commandOptions)[0]
+                            commandOptions: CreateIndexOptions? = nil,
```
patrickfreed: I noticed this was called `commandOptions` here, but just `options` in the other index methods. Was that intentional?

Sources/MongoSwift/Operations/DropIndexesOperation.swift
```diff
@@ -14,24 +14,29 @@ public struct DropIndexOptions: Encodable {
 /// An operation corresponding to a "dropIndexes" command.
 internal struct DropIndexesOperation<T: Codable>: Operation {
     private let collection: MongoCollection<T>
+    private let session: ClientSession?
```
patrickfreed: this is a bit out of scope for this work, but I imagine the operations, when we do client pooling, will try to get a collection via the session first, then from the stored mongo collection second if there is no stored session.

Sources/MongoSwift/MongoClient.swift
mbroadst: I know we kind of duked this out on the design already, but something makes me uncomfortable about this approach. Could we instead make all of the options conform to some protocol that provided a `withSession(ClientSession?)`, so you would only add it in the event that the options did exist (and therefore only allocate once)? Something like:
```
guard let cursor = mongoc_client_find_databases_with_opts(self._client, opts?.withSession(session).data) else {
```

Originally I was thinking of `SessionProvider` storing the session with the options, but this would be like an "appending copy" of the options.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
mbroadst: 😞

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -306,13 +327,20 @@ extension MongoCollection {
...
      * - Returns: A `MongoCursor` over the index names.
      *
-     * - Throws: A `userError.invalidArgumentError` if the options passed are an invalid combination.
+     * - Throws: `UserError.logicError` if the provided session is inactive.
      */
-    public func listIndexes() throws -> MongoCursor<Document> {
-        guard let cursor = mongoc_collection_find_indexes_with_opts(self._collection, nil) else {
+    public func listIndexes(session: ClientSession? = nil) throws -> MongoCursor<Document> {
+        var opts: Document?
+        if let session = session {
```
mbroadst: Is this the only place we have a method without options? Perhaps we should encapsulate this in some sort of utility method

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -16,13 +18,19 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
      *   - `UserError.invalidArgumentError` if the options passed to the command that generated this cursor formed an
      *     invalid combination.
      */
-    internal init(fromCursor cursor: OpaquePointer,
-                  withClient client: MongoClient,
-                  withDecoder decoder: BSONDecoder) throws {
+    internal init(from cursor: OpaquePointer,
```
mbroadst: while we're here and changing stuff, can you add a typealias for the cursor pointer?

Sources/MongoSwift/MongoError.swift
```diff
@@ -148,19 +152,26 @@ internal func parseMongocError(_ error: bson_error_t, errorLabels: [String]? = n
     switch (domain, code) {
     case (MONGOC_ERROR_CLIENT, MONGOC_ERROR_CLIENT_AUTHENTICATE):
         return RuntimeError.authenticationError(message: message)
+    case (MONGOC_ERROR_CLIENT, MONGOC_ERROR_CLIENT_SESSION_FAILURE):
+        // If user attempts to start a session against a server that doesn't support it, we throw a compat error.
+        if message.lowercased().contains("support sessions") {
+            return RuntimeError.compatibilityError(message: "Deployment does not support sessions")
```
mbroadst: nice 👌

Sources/MongoSwift/MongoError.swift
```diff
@@ -148,19 +152,26 @@ internal func parseMongocError(_ error: bson_error_t, errorLabels: [String]? = n
     switch (domain, code) {
     case (MONGOC_ERROR_CLIENT, MONGOC_ERROR_CLIENT_AUTHENTICATE):
         return RuntimeError.authenticationError(message: message)
+    case (MONGOC_ERROR_CLIENT, MONGOC_ERROR_CLIENT_SESSION_FAILURE):
+        // If user attempts to start a session against a server that doesn't support it, we throw a compat error.
+        if message.lowercased().contains("support sessions") {
```
mbroadst: any way to check by error code rather than string match?

Sources/MongoSwift/Operations/DropIndexesOperation.swift
mbroadst: I think we should only ever need a namespace, no need to eventually store a collection or a database

Sources/MongoSwift/MongoCollection+Indexes.swift
patrickfreed: Oh whoops, I didn't see the options parameter in there as well. Makes sense.

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -0,0 +1,294 @@
...
+import Nimble
+import XCTest
+
+final class ClientSessionTests: MongoSwiftTestCase {
+    override func tearDown() {
+        do {
+            let client = try MongoClient(MongoSwiftTestCase.connStr)
+            try client.db(type(of: self).testDatabase).drop()
+        } catch {
+            fail("encountered error when tearing down: \(error)")
```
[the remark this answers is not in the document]
patrickfreed: Apparently. I copied it from the bulk write tests: https://github.com/mongodb/mongo-swift-driver/blob/master/Tests/MongoSwiftTests/MongoCollection%2BBulkWriteTests.swift#L38

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -0,0 +1,294 @@
...
+                expect(nestedSession.id).to(bsonEqual(idB))
+            }
+        }
+    }
+
+    struct SessionsArgTest {
+        let name: String
+        let body: (ClientSession?) throws -> Void
+
+        func execute(session: ClientSession) throws {
```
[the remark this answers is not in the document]
patrickfreed: Done

Sources/MongoSwift/ClientSession.swift
```diff
@@ -1,35 +1,115 @@
...
+            return
+        }
+        self._opts = mongoc_session_opts_new()
     }
 
-    /// Cleans up internal state.
     deinit {
+        guard let opts = self._opts else {
+            return
+        }
```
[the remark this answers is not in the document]
patrickfreed: We probably should have, but as part of my changes in response to Matt's comments removed this class.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -1,35 +1,106 @@
...
+    var opts = mongoc_session_opts_new()! // always returns a value
+    defer { mongoc_session_opts_destroy(opts) }
+    return try body(opts)
 }
 
-/// :nodoc: A session for ordering sequential operations.
-public class ClientSession: Encodable {
-    /// Initializes a new client session.
-    internal init() {
+/// A MongoDB client session.
```
mbroadst: Since this is a public type I think we should expand the documentation, or at least link to relevant documentation like [this](https://docs.mongodb.com/manual/reference/server-sessions/).

Sources/MongoSwift/ClientSession.swift
```diff
@@ -1,35 +1,106 @@
...
 /// Options to use when creating a ClientSession.
-internal struct SessionOptions: Encodable {
-    /// Specifies whether read operations should be causally ordered within the session.
-    private let causalConsistency: Bool?
+public struct ClientSessionOptions {}
+
+/// Private helper for providing a `mongoc_session_opt_t` that is only valid within the body of the provided
+/// closure.
+private func withSessionOpts<T>(wrapping options: ClientSessionOptions?,
+                                _ body: (OpaquePointer) throws -> T) rethrows -> T {
```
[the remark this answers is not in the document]
kmahar: I'd actually argued against these before. see [here](https://github.com/mongodb/swift-mongo-mobile/pull/14#discussion_r268257188).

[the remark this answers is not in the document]
kmahar: as an addition to that comment, I would say that it was useful to typealias in the cases where a pointer's type is different across Swift versions, for the bson_ts and iterators. I'm just a bit skeptical about ending up with like 30 typealiases that all are ultimately just `OpaquePointer`s.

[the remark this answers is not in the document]
mbroadst: I think my primary motivation is readability here. We don't actually _lose_ anything from type aliasing, and I think we gain an ability to keep our Swift code in line with what's going on in mongoc. If the goal of keeping it an `OpaquePointer` is to warn the developer that they need to go open the mongoc docs and triple check everything when dealing with it, then we could name it `UnsafeMongocSessionOptsPointer` or something - even that gives me more context than `OpaquePointer`. Are there other benefits to keeping it `OpaquePointer`?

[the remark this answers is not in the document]
kmahar: I agree that it increases readability to know what the pointer is supposedly pointing to. I just feel weird about the fact that any unimported libmongoc type can be implicitly cast to a `SessionOptsPointer`.
I think it's easy to be like "oh, this function takes in a real type, and it compiles when I pass in this variable, so my variable must be the right type" even though you passed in a pointer to a bson_t or something. it seems like if you know it's opaque you might be more likely to look into what exactly it's supposed to be pointing to.... maybe that's just how I think about these though.

maybe we could prefix the typealiases with `Opaque`? `OpaqueSessionOptsPointer` etc.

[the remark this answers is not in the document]
kmahar: another thought I had on these is that in cases we only use them one time, there is probably another way to get the same kind of readability without introducing an extra symbol.
for example, in this case we can just add a parameter label in the closure, like 
```swift
private func withSessionOpts<T>(wrapping options: ClientSessionOptions?,
                                _ body: (_ sessionOptsPtr: OpaquePointer) throws -> T) rethrows -> T {
// ...
}
```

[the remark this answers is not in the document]
kmahar: another thing to consider is actually just naming stored opaque pointer properties in a way that describes what they are -- for ex., `ClientSession` below has 
```swift
 internal fileprivate(set) var _session: MutableClientSessionPointer?
```

if we renamed `_session` -> `mutableClientSessionPointer` then you would get the readability benefit everyplace you use it (for ex. the `init`, in `advanceClusterTime`), not just at the place where you define the property.

then again we generally don't include the type name in other stored properties, so, idk. just a thought.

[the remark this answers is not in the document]
mbroadst: Let's just leave it as it. If the pointer is being used in difficult situations (like the ones needed for Swift 5 support) then we'll type alias them, otherwise use `OpaquePointer`

Sources/MongoSwift/MongoCollection+Indexes.swift
patrickfreed: Updated this to use the new combine helper. It has a weird cast that needs to be there so the compiler can infer the generic type, even though it will be nil.

Sources/MongoSwift/ClientSession.swift
patrickfreed: The docs you linked to deal with server sessions, which the end user of the driver doesn't necessarily need to know about and don't really inform the usage of `ClientSession`. Unfortunately, the mongodb documentation on client sessions is neither detailed nor centralized. The best summary is in the 3.6 patch notes, though they're already outdated since there are no mentions of transactions.

So as it is right now, the client session doesn't actually do anything yet. I think as we add features built on client sessions (e.g. causal consistency, transactions), we can add to the `ClientSession` documentation. I think the [python docs](http://api.mongodb.com/python/current/api/pymongo/client_session.html) have a good page on them documented in that manner.

At any rate, I've added a sentence explaining what their purpose is, which is a little better.

Sources/MongoSwift/ClientSession.swift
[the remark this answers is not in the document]
patrickfreed: In line with this, I've removed the `ClientSessionMutablePointer` typealias.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -306,13 +327,16 @@ extension MongoCollection {
...
      * - Returns: A `MongoCursor` over the index names.
      *
-     * - Throws: A `userError.invalidArgumentError` if the options passed are an invalid combination.
+     * - Throws: `UserError.logicError` if the provided session is inactive.
      */
-    public func listIndexes() throws -> MongoCursor<Document> {
-        guard let cursor = mongoc_collection_find_indexes_with_opts(self._collection, nil) else {
+    public func listIndexes(session: ClientSession? = nil) throws -> MongoCursor<Document> {
+        // need this cast to infer a generic type
+        let opts = try encodeOptions(options: nil as DropIndexOptions?, session: session, using: self.encoder)
```
kmahar: lol TIL you can cast nil... makes sense but kind of silly. did you consider adding a default value of nil for the options parameter to `encodeOptions` instead?

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -1,9 +1,13 @@
 import mongoc
 
+internal typealias MutableCursorPointer = OpaquePointer
```
kmahar: per our other convo should maybe not be a typealias?

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -6,3 +6,16 @@ internal protocol Operation {
     /// Executes this operation and returns its corresponding result type.
     func execute() throws -> OperationResult
 }
+
+/// Internal function for generating an options `Document` for passing to libmongoc.
+internal func encodeOptions<T: Encodable>(options: T?,
```
kmahar: since passing the encoder to this function is going to copy it anyway, and we don't care about the strategies when encoding options (and in the future we may introduce strategies for numbers that could mess up options encoding), maybe consider just creating the encoder here instead?

patrickfreed: yeah, I think that makes sense. The only reason for using the strategies here would be to encode some user data type, and if a user data type ever shows up in an options struct it'll be in the form of a `Document` anyways, meaning they could just encode it themselves. I don't really see us genericizing any options types, which would be the only way a non-`Document` user data element ends up in the options I think.

tl;dr I made it use `BSONEncoder()`

PR #258, opened 2019-04-26
SWIFT-373 Add runCommand operation and switch DB methods to use it

kmahar:
This adds a `RunCommandOperation` and switches `runCommand`, `createCollection`, and `dropDatabase` to use it.

### DIFF
--- a/Sources/MongoSwift/Operations/RunCommandOperation.swift
+++ b/Sources/MongoSwift/Operations/RunCommandOperation.swift
@@ -1,0 +1,53 @@
+import mongoc
+
+/// Options to use when running a command against a `MongoDatabase`.
+public struct RunCommandOptions: Encodable {
+    /// An optional `ReadConcern` to use for this operation.
+    public let readConcern: ReadConcern?
+
+    /// An optional `ReadPreference` to use for this operation.
+    public let readPreference: ReadPreference?
+
+    /// An optional `WriteConcern` to use for this operation.
+    public let writeConcern: WriteConcern?
+
+    /// Convenience initializer allowing any/all parameters to be omitted or optional.
+    public init(readConcern: ReadConcern? = nil,
+                readPreference: ReadPreference? = nil,
+                writeConcern: WriteConcern? = nil) {
+        self.readConcern = readConcern
+        self.readPreference = readPreference
+        self.writeConcern = writeConcern
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case readConcern, writeConcern
+    }
+}
+
+/// An operation corresponding to a `runCommand` call.
+internal struct RunCommandOperation: Operation {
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/Operations/RunCommandOperation.swift
```diff
@@ -0,0 +1,51 @@
...
+        self.writeConcern = writeConcern
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case readConcern, writeConcern
+    }
+}
+
+/// An operation corresponding to a `runCommand` call.
+internal struct RunCommandOperation: Operation {
```
mbroadst: I think we might be jumping the gun a little bit creating the `RunCommandOperation` at this stage, it brings with it many considerations that I think greatly expand the current scope of work. For starters, it's specific to a `MongoDatabase` when it really should be something we can reuse for `MongoClient`/`MongoCollection`/`MongoDatabase`, anything with a namespace really. Secondly, it appears that the mongoc command you chose will [skip database read preference inheritance](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-database.c#L327).

Basically, I think this work moves us into command construction earlier than we want to. Did something necessitate the change, or was it just annoying duplicating code?

kmahar: ok, maybe `RunDatabaseCommandOperation` is a better name for this? it's just copy paste from our existing `MongoDatabase.runCommand` method.
so the read preference thing is an existing bug then - we should check if the `MongoDatabase` has a read preference set and send that along if the user doesn't provide one in the options.

As for rewriting the other methods to use this, yeah, the motivation was just that it was a lot of boilerplate to create new operations and they were simple to rewrite this way. so I was trying to do something similar to the "bulk write everything" refactor.

mbroadst: I think it's probably fine to keep this one named `RunCommandOperation` for now, and use it _only_ for `MongoDatabase.runCommand`. I'm worried in the other cases where you are reusing it here that we are subtly introducing bugs like I pointed to above. As you move into client pooling I think it may be much easier to see patterns for modifying `RunCommandOperation` to support use across all types.

kmahar: I ended up noticing a separate unrelated ReadPreference issue so I just opened a separate PR to fix both of those. Will move dropDatabase and createCollection to use their own operations containing their original logic.

kmahar:
Ok I've updated so that `RunCommandOperation` is only used for `MongoDatabase.runCommand`. The other DB methods have their own operation types.
Also noticed that I missed adding an operation for dropCollection so I threw that one in too.

PR #259, opened 2019-04-29
SWIFT-373 add NextOperation for cursor ops

kmahar:
Adds a `NextOperation` for calls to `mongoc_cursor_next`.

### DIFF
--- a/Sources/MongoSwift/Operations/NextOperation.swift
+++ b/Sources/MongoSwift/Operations/NextOperation.swift
@@ -13,6 +13,10 @@ internal struct NextOperation<T: Codable>: Operation {
             throw UserError.logicError(message: "Tried to iterate a closed cursor.")
         }
 
+        if let session = self.cursor._session, !session.active {
+            throw ClientSession.SessionInactiveError
+        }
+
         let out = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
         defer {
             out.deinitialize(count: 1)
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/Operations/NextOperation.swift
```diff
@@ -0,0 +1,33 @@
+import mongoc
+
+/// An operation corresponding to a `next` call on a MongoCursor.
+internal struct NextOperation<T: Codable>: Operation {
+    private let _cursor: OpaquePointer
```
mbroadst: should we typealias the `OpaquePointer` here?

kmahar: see my comment on the sessions PR on this typealiasing `OpaquePointer`s business. I'll wait to see what resolution we have over there before addressing this

patrickfreed: I think we may want to actually store the `MongoCursor` itself here to ensure it survives for the lifetime of the operation. As it is now, everything is synchronous and this isn't a concern, but in the future the lifetime of the operation may be independent of its generating cursor.

This probably won't be an issue, but in general I think we should try to avoid passing around the raw mongoc pointers as much as we can.

kmahar: good point. that is more consistent with the operations right now anyway, since they store Collections/DBs rather than the pointers. updated.

PR #260, opened 2019-04-29
Fix a couple ReadPreference issues

kmahar:
Stemming from conversation on SWIFT-373 I fixed one ReadPreference issue there and happened to notice another.
I'm surprised we didn't catch the `MongoDatabase.readPreference` issue. I noticed our tests there are very lackluster and only test the type itself and not setting it on clients/DBs/colls or using it for commands or anything like that. I opened [SWIFT-447](https://jira.mongodb.org/browse/SWIFT-447) about remedying this.

### DIFF
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -225,7 +225,7 @@ public class MongoDatabase {
 
     /// The `ReadPreference` set on this database
     public var readPreference: ReadPreference? {
-        return ReadPreference(from: mongoc_collection_get_read_prefs(self._database))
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -225,7 +225,7 @@ public class MongoDatabase {
 
     /// The `ReadPreference` set on this database
     public var readPreference: ReadPreference? {
-        return ReadPreference(from: mongoc_collection_get_read_prefs(self._database))
```
patrickfreed: 😱

PR #261, opened 2019-04-30
SWIFT-402 Resolve MongoSwift namespace issues

kmahar:
Gets rid of the `MongoSwift` class and moves the init/cleanup methods to be free-floating.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -213,7 +213,8 @@ public class MongoClient {
      * :nodoc:
      * Create a new client from an existing `mongoc_client_t`. The new client will destroy the `mongoc_client_t` upon
      * deinitialization.
-     * Do not use this initializer unless you know what you are doing.
+     * Do not use this initializer unless you know what you are doing. You *must* call libmongoc_init *before* using
+     * this initializer for the first time.
      *
      * If this client was derived from a pool, ensure that the error api version was set to 2 on the pool.
      *
@@ -221,9 +222,6 @@ public class MongoClient {
      *   - pointer: the `mongoc_client_t` to store and use internally
      */
     public init(stealing pointer: OpaquePointer) {
-        // Initialize mongoc. Repeated calls have no effect so this is safe to do every time.
-        initializeMongoc()
-
         self._client = pointer
 
         // This call may fail, and if it does, either the error api version was already set or the client was derived
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -164,6 +164,9 @@ public class MongoClient {
      *     built with TLS support.
      */
     public init(_ connectionString: String = "mongodb://localhost:27017", options: ClientOptions? = nil) throws {
+        // Initialize mongoc. Repeated calls have no effect so this is safe to do every time.
+        initializeMongoc()
+
         var error = bson_error_t()
         guard let uri = mongoc_uri_new_with_error(connectionString, &error) else {
             throw parseMongocError(error)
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -208,16 +211,20 @@ public class MongoClient {
...
      * If this client was derived from a pool, ensure that the error api version was set to 2 on the pool.
      *
      * - Parameters:
-     *   - fromPointer: the `mongoc_client_t` to store and use internally
+     *   - pointer: the `mongoc_client_t` to store and use internally
      */
-    public init(fromPointer: OpaquePointer) {
-        self._client = fromPointer
+    public init(stealing pointer: OpaquePointer) {
+        // Initialize mongoc. Repeated calls have no effect so this is safe to do every time.
```
kmahar: do we actually need to call this here? I don't see it being called anywhere in the mongoc_embedded code but it feels like you would call it *before* creating a client there...

Sources/MongoSwift/MongoSwift.swift
```diff
@@ -1,39 +1,29 @@
...
-         * See: https://bugs.swift.org/browse/SR-2500 */
-        mongoc_cleanup()
-    }
+/**
+ * Release all memory and other resources allocated by libmongoc.
+ *
+ * This function should be called once at the end of the application. Users
+ * should not interact with the driver after calling this function.
+ */
+public func cleanupMongoc() {
```
kmahar: open to any other name ideas

Sources/MongoSwift/MongoClient.swift
```diff
@@ -208,16 +211,20 @@ public class MongoClient {
...
      *
      * If this client was derived from a pool, ensure that the error api version was set to 2 on the pool.
      *
      * - Parameters:
-     *   - fromPointer: the `mongoc_client_t` to store and use internally
+     *   - pointer: the `mongoc_client_t` to store and use internally
      */
-    public init(fromPointer: OpaquePointer) {
-        self._client = fromPointer
+    public init(stealing pointer: OpaquePointer) {
```
kmahar: just a drive by change but seemed consistent with other similar inits

kmahar:
I also opened [SWIFT-454](https://jira.mongodb.org/browse/SWIFT-454) about fixing the README to reflect the new API before our next release.

Sources/MongoSwift/MongoClient.swift
patrickfreed: Yeah, it's definitely supposed to be called before a client is created, at least with the normal driver. I'm not sure what the effect is doing so afterwards.

Sources/MongoSwift/MongoSwift.swift
patrickfreed: Even though this function only exists because of the mongoc dependency, I personally think it would be better if we phrased it in terms of MongoSwift. 

i.e. the function would be `cleanupMongoSwift` and the docstring would say something like "Release all memory and other resources allocated by MongoSwift". Note we may want to say "internal memory and resources", since any objects the user holds on to will continue to hold memory even after cleanup is performed.

kmahar: renamed and updated docstring.

Sources/MongoSwift/MongoClient.swift
kmahar: per my messages in #swift-private... I think we should just call `mongoc_init` directly in the embedded library since it needs to happen before this initializer is used. added a note about that to the docstring for this.

kmahar: opened [SWIFT-458](https://jira.mongodb.org/browse/SWIFT-458) about this.

PR #263, opened 2019-05-03
SWIFT-175 Causal consistency support

patrickfreed:
[SWIFT-175](https://jira.mongodb.org/browse/SWIFT-175)

This PR also reworks the previous sessions test so that it's easier for more tests to cover every client/db/collection method that takes in a session.

### DIFF
--- a/Tests/MongoSwiftTests/ClientSessionTests.swift
+++ b/Tests/MongoSwiftTests/ClientSessionTests.swift
@@ -309,7 +309,8 @@ final class ClientSessionTests: MongoSwiftTestCase {
         }
     }
 
-    /// Test that causal consistency guarantees are met.
+    /// Test that causal consistency guarantees are met on deployments that support cluster time.
+    // swiftlint:disable:next cyclomatic_complexity
     func testCausalConsistency() throws {
         guard MongoSwiftTestCase.topologyType != .single else {
             print("Skipping test case because of unsupported topology type \(MongoSwiftTestCase.topologyType)")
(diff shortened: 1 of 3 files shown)


Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -14,7 +14,73 @@ final class ClientSessionTests: MongoSwiftTestCase {
...
-    /// Test that sessions are properly returned to the pool when ended.
+    typealias CollectionSessionOp = (name: String, body: (MongoCollection<Document>, ClientSession?) throws -> Void)
+    typealias DatabaseSessionOp = (name: String, body: (MongoDatabase, ClientSession?) throws -> Void)
+    typealias ClientSessionOp = (name: String, body: (MongoClient, ClientSession?) throws -> Void)
+    typealias SessionOp = (name: String, body: (ClientSession?) throws -> Void)
+
+    typealias InsertOneModel = MongoCollection<Document>.InsertOneModel
+
+    // list of read only operations on MongoCollection that take in a session
+    let collectionSessionReadOps: [CollectionSessionOp] = [
```
patrickfreed: I moved all the operations previously just defined in the args test out here so that they can easily be used in any spec test that requires all methods that take in a session to be tested. This affects some of the old tests a little bit.

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -293,4 +308,207 @@ final class ClientSessionTests: MongoSwiftTestCase {
...
+                seenCommand = true
+            }
+            defer { center.removeObserver(observer) }
+            _ = try collection.find(session: session).next()
+            expect(seenCommand).to(beTrue())
+        }
+    }
+
+    /// Test causal consistent behavior on a topology that doesn't support cluster time.
+    func testCausalConsistencyStandalone() throws {
```
kmahar: it seems like tests 1, 2, and 10 are also ok to run without cluster time, is it worth doing that?

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -293,4 +308,207 @@ final class ClientSessionTests: MongoSwiftTestCase {
...
+        let db = client.db(type(of: self).testDatabase)
+        defer { try? db.drop() }
+        let collection = db.collection(self.getCollectionName())
+
+        // spec test 1
+        let session1 = try client.startSession()
+        expect(session1.operationTime).to(beNil())
+        session1.end()
+
+        // spec test 2 + 3
```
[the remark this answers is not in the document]
patrickfreed: Updated to include error case. I put it in the same test run which isn't exactly according to spec, but I think it tests the same thing. There isn't a reply associated with a `CommandFailedEvent`, so I just check that a failure event was received and that the op time was updated.

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -293,4 +308,207 @@ final class ClientSessionTests: MongoSwiftTestCase {
...
+        }
+
+        let center = NotificationCenter.default
+        let client = try MongoClient(MongoSwiftTestCase.connStr, options: ClientOptions(eventMonitoring: true))
+        client.enableMonitoring()
+        let db = client.db(type(of: self).testDatabase)
+        defer { try? db.drop() }
+        let collection = db.collection(self.getCollectionName())
+
+        // spec test 1
```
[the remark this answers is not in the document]
patrickfreed: Done

Tests/MongoSwiftTests/ClientSessionTests.swift
patrickfreed: Its only a little copypaste, so sure. done

Tests/MongoSwiftTests/ClientSessionTests.swift
mbroadst: very nice refactor!

patrickfreed:
Turns out the error case for spec test 3 did need to be on its own, so I've updated it to that.

PR #264, opened 2019-05-06
SWIFT-411 Make pointer access to bson_ts more explicit

kmahar:
This adds two `withBSONPointer` and `withMutableBSONPointer` methods to `Document`  that can be used to get access to immutable/mutable pointers underlying the document. I've switched to using those everywhere we were just directly using the bson_t pointer before.

Additionally, all errors are now passed in with `withUnsafeMutablePointer` instead of with `&`.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -219,7 +219,7 @@ extension Array: BSONValue {
             try arr.setValue(for: String(i), to: val)
         }
 
-        guard bson_append_array(storage.pointer, key, Int32(key.utf8.count), arr.data) else {
+        guard bson_append_array(storage._bson, key, Int32(key.utf8.count), arr._bson) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -255,7 +255,7 @@ public struct BSONNull: BSONValue, Codable, Equatable {
     }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_null(storage.pointer, key, Int32(key.utf8.count)) else {
+        guard bson_append_null(storage._bson, key, Int32(key.utf8.count)) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -356,7 +356,7 @@ public struct Binary: BSONValue, Equatable, Codable {
         let subtype = bson_subtype_t(UInt32(self.subtype))
         let length = self.data.count
         let byteArray = [UInt8](self.data)
-        guard bson_append_binary(storage.pointer, key, Int32(key.utf8.count), subtype, byteArray, UInt32(length)) else {
+        guard bson_append_binary(storage._bson, key, Int32(key.utf8.count), subtype, byteArray, UInt32(length)) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -392,7 +392,7 @@ extension Bool: BSONValue {
     public var bsonType: BSONType { return .boolean }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_bool(storage.pointer, key, Int32(key.utf8.count), self) else {
+        guard bson_append_bool(storage._bson, key, Int32(key.utf8.count), self) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -422,7 +422,7 @@ extension Date: BSONValue {
     public var msSinceEpoch: Int64 { return Int64((self.timeIntervalSince1970 * 1000.0).rounded()) }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_date_time(storage.pointer, key, Int32(key.utf8.count), self.msSinceEpoch) else {
+        guard bson_append_date_time(storage._bson, key, Int32(key.utf8.count), self.msSinceEpoch) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -464,7 +464,7 @@ public struct DBPointer: BSONValue, Codable, Equatable {
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         try withUnsafePointer(to: id.oid) { oidPtr in
-            guard bson_append_dbpointer(storage.pointer, key, Int32(key.utf8.count), self.ref, oidPtr) else {
+            guard bson_append_dbpointer(storage._bson, key, Int32(key.utf8.count), self.ref, oidPtr) else {
                 throw bsonTooLargeError(value: self, forKey: key)
             }
         }
@@ -538,7 +538,7 @@ public struct Decimal128: BSONNumber, Equatable, Codable, CustomStringConvertibl
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
         try withUnsafePointer(to: self.decimal128) { ptr in
-            guard bson_append_decimal128(storage.pointer, key, Int32(key.utf8.count), ptr) else {
+            guard bson_append_decimal128(storage._bson, key, Int32(key.utf8.count), ptr) else {
                 throw bsonTooLargeError(value: self, forKey: key)
             }
         }
@@ -601,7 +601,7 @@ extension Double: BSONNumber {
     public var bsonType: BSONType { return .double }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_double(storage.pointer, key, Int32(key.utf8.count), self) else {
+        guard bson_append_double(storage._bson, key, Int32(key.utf8.count), self) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
@@ -681,7 +681,7 @@ extension Int32: BSONNumber {
     public var bsonType: BSONType { return .int32 }
 
     public func encode(to storage: DocumentStorage, forKey key: String) throws {
-        guard bson_append_int32(storage.pointer, key, Int32(key.utf8.count), self) else {
+        guard bson_append_int32(storage._bson, key, Int32(key.utf8.count), self) else {
             throw bsonTooLargeError(value: self, forKey: key)
         }
     }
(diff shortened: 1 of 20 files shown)


Sources/MongoSwift/MongoError.swift
```diff
@@ -307,8 +307,7 @@ internal func convertingBulkWriteErrors<T>(_ body: () throws -> T) throws -> T {
 }
 
 internal func toErrorString(_ error: bson_error_t) -> String {
-    var e = error
-    return withUnsafeBytes(of: &e.message) { rawPtr -> String in
+    return withUnsafeBytes(of: error.message) { rawPtr in
```
kmahar: swift 4.2 introduced a non-inout version of this, so we no longer need to copy the error

Tests/MongoSwiftTests/DocumentTests.swift
```diff
@@ -55,6 +55,10 @@ extension Document {
             return DocElem(key: kvp.key, value: kvp.value, type: kvp.value.bsonType)
         }
     }
+
+    internal var data: MutableBSONPointer {
```
kmahar: I've moved this into the tests because various tests in this file checking copy on write behavior were using it. but it is no longer used anywhere in the driver.

Sources/MongoSwift/Operations/RunCommandOperation.swift
```diff
@@ -42,10 +42,22 @@ internal struct RunCommandOperation: Operation {
...
+        self.command.withBSONPointer { cmdPtr in
+            withOptionalBSONPointer(to: opts) { optsPtr in
+                reply.withMutableBSONPointer { replyPtr in
+                    withUnsafeMutablePointer(to: &error) { errPtr in
+                        mongoc_database_command_with_opts(
+                            self.database._database, cmdPtr, rp?._readPreference, optsPtr, replyPtr, errPtr)
+                    }
+                }
+            }
+        }
```
patrickfreed: while I appreciate the added explicitness of what we're doing, this pyramid like growth for each argument we pass to mongoc is pretty large in cases like this.

What do we think about only using such helpers where where they're necessary (e.g. mutable `bson_t` pointers or where we run into compiler issues) and rely on `&` otherwise? I think a large use-case of these helpers is if you want to coerce the pointer into something else. Since we're just passing them as-is, I imagine the same stuff is going on under the hood as if we just used `&`.

kmahar: yeah, all the nesting is a bit annoying. 
I would be ok with making the `.data` property return a `BSONPointer` and using that for read-only access cases. then this one would become
```swift
let success =
reply.withMutableBSONPointer { replyPtr in
    withUnsafeMutablePointer(to: &error) { errPtr in
        mongoc_database_command_with_opts(
            self.database._database, cmd.data, rp?._readPreference, opts?.data, replyPtr, errPtr)
    }
}
```

kmahar: what do you think @mbroadst ?

mbroadst: Discussed offline. We agreed that any class that wraps a pointer must provide a mutable accessor method. Passing local variables by reference is fine (e.g. `&error`)

kmahar:
Ok round 2. we have a new `withMutableBSONPointer` helper used anywhere we mutate a document's storage.
also renamed `.data` to `._bson` and `.storage` to `._storage`.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -219,7 +219,7 @@ extension Array: BSONValue {
             try arr.setValue(for: String(i), to: val)
         }
 
-        guard bson_append_array(storage.pointer, key, Int32(key.utf8.count), arr.data) else {
```
kmahar: it occurred to me as I made these changes that we are kind of doing the same bad thing here where we are mutating this `bson_t`. but maybe it's overkill to do some kind of `withPointer` for the storage, too. 
similar to `DocumentIterator`, `DocumentStorage` could probably use some rethinking...

Sources/MongoSwift/Operations/RunCommandOperation.swift
patrickfreed: Yeah, I think it looks much better now 👍

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -11,57 +11,51 @@ internal typealias MutableBSONPointer = UnsafeMutablePointer<bson_t>
...
 /// A struct representing the BSON document type.
 @dynamicMemberLookup
 public struct Document {
-    /// the storage backing this document
-    internal var storage: DocumentStorage
-
-    /// Returns the number of (key, value) pairs stored at the top level of this `Document`.
-    public internal(set) var count: Int
+    /// the storage backing this document.
+    internal var _storage: DocumentStorage
```
patrickfreed: Just to clarify: what is the convention around underscore naming / what do they signify now?

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: Hmm. It's making me think we should just make `DocumentStorage` a `private` type (not even `internal`) and update all the public stuff that uses it just take a `Document` instead. It seems too dangerous to circumvent the api you're adding to `Document` in this PR to warrant exposing the storage, even at the `internal` level.

Sources/MongoSwift/BSON/Document.swift
kmahar: I read our conversation earlier as that we should use them for anything besides a public property. but I don't feel very strongly so I'm open to any other convention on this

patrickfreed: I think just internal vs public might be too broad. I think it'd be more useful to designate things we truly want as `fileprivate` or `private`. I know a lot of times I'll update some private field to `internal` so I can use it elsewhere, but doing that is not a great practice and leads to problems like what this PR is trying to solve. We could use the underscore naming as a way to say "keep this thing private".

That said, I too don't feel that strongly about this (and have introduced stuff that doesn't align with the above convention myself), so if that convention doesn't make sense I'm fine with not doing it. I do think that if we're going to name some stuff with underscores, we should settle on some convention around them, though.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I considered that.... it seems like it might confuse users that there would be this public method on `BSONValue`:
`encode(to: Document, forKey key: String)` 
but that we don't actually intend for them to set values on documents that way.... 

then again, this existence of that method and `DocumentStorage` in the public API are already confusing as it is.

I wonder if `DocumentStorage` should get a bunch of methods on it: `appendInt32`, `appendString`, and so on... and then we type switch within `Document` (or in some generic `appendBSONValue` method on `DocumentStorage) to decide which method to call? 

that brings us back to same problem where `BSONValue` is a protocol but people can't really actually make arbitrary types conform to it since we wouldn't be switching on them there...

Sources/MongoSwift/BSON/Document.swift
kmahar: yeah, without a real convention on what the underscore means, I don't think it gives us any extra information. that seems like a reasonable convention to use, though yeah it definitely doesn't align with a lot of the driver now... lots of our `_x` properties are actually internal. 

alternatively we can get rid of all of the underscores and say this is what swift's fine-grained ACLs are for.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: Because we call `copyStorageIfRequired` where necessary in `setValue` and we've updated the `count` property, the bad things we're doing here aren't resulting in any buggy behavior (I think). It would be preferable to guarantee that at the mutating call's site, but without big changes that'll be hard to do for the reasons you mention.

As far as big changes go though, I have a fun idea...

We would update `DocumentStorage` to be a wrapper around an internal `Data` buffer that we manage ourselves instead of a `bson_t`. We'd then update the `BSONValue` protocol to require a `bson: Data` property (or something like that) that emits a buffer containing the raw BSON bytes (and remove `encode(to:forKey)`. Then, where we normally call `encode`, we instead get this buffer from the `BSONValue` and append it to our internal buffer (with the necessary bookkeeping for keys, types, etc). When we do need a `bson_t`, we can use `bson_init_static` to create a lightweight, read-only, and stack allocated `bson_t` from the internal buffer.

This gets us like 99% of the way to pure Swift BSON, avoids a lot of memory issues, and makes the `BSONValue` protocol more sensical. This is obviously out of scope for this ticket, but I think it could be a good approach for solving issues we have _now_ as well as moving towards a pure Swift driver.

Sources/MongoSwift/BSON/Document.swift
patrickfreed: Both options are fine with me really.

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: ha, when I started thinking about the pure Swift BSON library design I came up with something very similar where we would have a `bytes` property on `BSONValue` 🙂

patrickfreed: Makes me realize that pure swift BSON library may actually be a breaking change, so we might have to do it pre-GA 🤔

mbroadst: Yeah I think we discussed this some time ago. This is what the C++ driver does, rather than keeping instances of `bson_t` everywhere. `bson_t` initialization isn't zero overhead, but it might not be that bad.  I'd say go ahead and make a ticket for the feature

Sources/MongoSwift/BSON/Document.swift
mbroadst: afaict the underscore is used in Apple's code to indicate a fileprivate

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: Filed [SWIFT-497](https://jira.mongodb.org/browse/SWIFT-497)

Sources/MongoSwift/BSON/Document.swift
patrickfreed: I don't want to hold up this PR for style nits, so it's not super important what happens to them here. I do think we should either revisit the convention in the future or remove all of them from the codebase if we find that such a convention isn't necessary.

kmahar: I've opened a follow up ticket [SWIFT-498](https://jira.mongodb.org/browse/SWIFT-498)

PR #266, opened 2019-05-08
SWIFT-377 Refactor DB and collection initializers to take in options rather than libmongoc types

kmahar:
This defers the creation of `mongoc_database_t`s and `mongoc_collection_t`s to the `MongoDatabase` and `MongoCollection` initializers, respectively.

### DIFF
--- a/Sources/MongoSwift/Operations/CreateCollectionOperation.swift
+++ b/Sources/MongoSwift/Operations/CreateCollectionOperation.swift
@@ -133,14 +133,14 @@ internal struct CreateCollectionOperation<T: Codable>: Operation {
             throw parseMongocError(error)
         }
 
-        let encoder = BSONEncoder(copies: self.database.encoder, options: self.options)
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/Operations/CreateCollectionOperation.swift
```diff
@@ -132,15 +132,12 @@ internal struct CreateCollectionOperation<T: Codable>: Operation {
             self.database._database, self.name, opts?.data, &error) else {
             throw parseMongocError(error)
         }
+        mongoc_collection_destroy(collection)
```
kmahar: this place was a little tricky, because previously we just passed the `mongoc_collection_t` that `mongoc_database_create_collection` returns into the initializer.

I rewrote this so we just throw away that instance and pass on the relevant info to the initializer. 
I considered having a second `MongoCollection` initializer that would still use the pointer, but I think once we add client pooling we will throw away the collection here anyway, so I figured it wasn't worth it.

PR #267, opened 2019-05-13
Convert FindAndModify to use an Operation

kmahar:
Moves the private `findAndModify` method code to a new `FindAndModifyOperation`.

`FindAndModifyOptionsConvertible` needed to become `internal` so we could use it in that file, and I renamed its `asOpts` methods to `asFindAndModifyOptions` for consistency with our `BulkWriteOptionsConvertible` protocol. 
Similarly, `FindAndModifyOptions` had to become `internal` since its now defined in the new file but used in the original file.

### DIFF
--- a/Sources/MongoSwift/Operations/FindAndModifyOperation.swift
+++ b/Sources/MongoSwift/Operations/FindAndModifyOperation.swift
@@ -1,0 +1,162 @@
+import mongoc
+
+/// A class wrapping a `mongoc_find_and_modify_opts_t`, for use with `MongoCollection.findAndModify`
+internal class FindAndModifyOptions {
+    // an `OpaquePointer` to a `mongoc_find_and_modify_opts_t`
+    var _options: OpaquePointer?
+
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/Operations/FindAndModifyOperation.swift
```diff
@@ -0,0 +1,163 @@
...
+        var doc = Document()
+        try session.append(to: &doc)
+
+        guard mongoc_find_and_modify_opts_append(self._options, doc.data) else {
+            throw RuntimeError.internalError(message: "Couldn't read session information")
+        }
+    }
+
+    /// Cleans up internal state.
+    deinit {
```
mbroadst: nit: prefer this at the top of the class declaration like a `defer` block, helps see that we _are_ indeed cleaning things up

kmahar: moved to the top.

PR #268, opened 2019-05-14
SWIFT-178 Implement retryable writes spec test runner

patrickfreed:
[SWIFT-178](https://jira.mongodb.org/browse/SWIFT-178)

This PR implements the spec test runner for the retryable writes tests. As part of that, it adds a generic spec test runner framework (based off of @kmahar 's work on the `crud-experiment` branch) that can be reused for other spec testing. This PR only adds the retryable writes tests and leaves updating the old crud tests to use the runner as later work.

These tests are quite slow (~40s on my macbook) and can be improved tremendously by lowering minHeartbeatFrequency (~10s on my macbook). Unfortunately, there is no way via libmongoc's public API to set this value currently. Follow [CDRIVER-3130](https://jira.mongodb.org/browse/CDRIVER-3130) for updates on progress there.


.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
     - stage: tests
       os: linux
-      env: SWIFT_VERSION=4.2
+      env: SWIFT_VERSION=4.2.4
```
patrickfreed: Swift 4.2.0 on Linux has a difference (a bug, I assume) in the `JSONDecoder` implementation that causes the runner to fail parsing the spec test files. Later versions and swift on macOS don't seem to suffer from it, though. Bumping to 4.2.4 works on all variants, so I've done that here.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -14,24 +14,32 @@ public struct ClientOptions: CodingStrategyProvider {
...
-    public let readPreference: ReadPreference?
-
     /// Specifies a WriteConcern to use for the client. If one is not specified, the server's default write concern
     /// will be used.
     public let writeConcern: WriteConcern?
 
+    // swiftlint:disable redundant_optional_initialization
+
+    /// Specifies a ReadPreference to use for the client.
+    public var readPreference: ReadPreference? = nil
```
patrickfreed: `ReadPreference` isn't `Decodable`, so I needed to provide a default value for it here. swiftlint doesn't detect the semantic value, so I have to disable the rule. The coding strategies are swift specific, so they also need default values.

Sources/MongoSwift/MongoError.swift
```diff
@@ -243,41 +243,40 @@ private func getBulkWriteErrorFromReply(
         withResult result: BulkWriteResult? = nil,
         withWriteConcernError wcErr: WriteConcernError? = nil) throws -> MongoError? {
     let decoder = BSONDecoder()
-    var insertedIds = result?.insertedIds
 
-    var bulkWriteErrors: [BulkWriteError]?
```
patrickfreed: When using fail points, sometimes empty (but non-nil) writeErrors and writeConcernErrors arrays are returned for parsing. The way this function was written assumed that wasn't possible and has been rewritten to accommodate those cases.

Tests/MongoSwiftTests/RetryableWritesTests.swift
```diff
@@ -0,0 +1,168 @@
...
+    let outcome: TestOutcome
+    let operation: AnyTestOperation
+
+    /// Options used to configure the `MongoClient` used for this test.
+    let clientOptions: ClientOptions?
+
+    /// If true, the `MongoClient` for this test should be initialized with multiple mongos seed addresses.
+    /// If false or omitted, only a single mongos address should be specified.
+    /// This field has no effect for non-sharded topologies.
+    let useMultipleMongoses: Bool?
```
patrickfreed: I actually do not process this field at all, since I wasn't sure what our path going forward is with regard to multiple mongoses. I know we filed a SWIFT and maybe a DRIVERS ticket about this, and I figured I could start using this field based on guidance from that or this code review.

Another question is that we provide our URI via environment variables. Would we define a separate one that specifically has one (or two) mongos(es)?

Tests/MongoSwiftTests/TestUtils.swift
```diff
@@ -4,6 +4,70 @@ import mongoc
...
 
+typealias DeleteOneModel = MongoCollection<Document>.DeleteOneModel
+typealias DeleteManyModel = MongoCollection<Document>.DeleteManyModel
+typealias InsertOneModel = MongoCollection<Document>.InsertOneModel
+typealias ReplaceOneModel = MongoCollection<Document>.ReplaceOneModel
+typealias UpdateOneModel = MongoCollection<Document>.UpdateOneModel
+typealias UpdateManyModel = MongoCollection<Document>.UpdateManyModel
+
+/// A struct representing a server version.
+internal struct ServerVersion: Comparable, Decodable {
```
patrickfreed: This was moved out of a `MongoClient` extension so that it could be used on its own inside the spec test runner. I also added `Comparable` conformance so that we can use the comparison operators instead of the english versions.

patrickfreed:
We'll be doing this review in separate, smaller reviews.

PR #269, opened 2019-05-15
SWIFT-178 Conform various options/result types to Decodable

patrickfreed:
This is part of a series of reviews that make up [SWIFT-178](https://jira.mongodb.org/browse/SWIFT-178)

This PR adds `Decodable` conformance to options and result types that will eventually be decoded from JSON spec test files. It also improves the error parser to be more flexible when decoding from server replies.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -2,7 +2,7 @@ import Foundation
 import mongoc
 
 /// Options to use when creating a `MongoClient`.
-public struct ClientOptions: CodingStrategyProvider {
+public struct ClientOptions: CodingStrategyProvider, Decodable {
     /// Determines whether the client should retry supported write operations.
     public let retryWrites: Bool?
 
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/MongoError.swift
```diff
@@ -243,41 +243,40 @@ private func getBulkWriteErrorFromReply(
         withResult result: BulkWriteResult? = nil,
         withWriteConcernError wcErr: WriteConcernError? = nil) throws -> MongoError? {
     let decoder = BSONDecoder()
-    var insertedIds = result?.insertedIds
 
-    var bulkWriteErrors: [BulkWriteError]?
```
patrickfreed: When using fail points, sometimes empty (but non-nil) writeErrors and writeConcernErrors arrays are returned for parsing. The way this function was written assumed that wasn't possible and has been rewritten to accommodate those cases.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -476,6 +476,37 @@ public struct BulkWriteResult {
     /// Map of the index of the operation to the id of the upserted document.
     public let upsertedIds: [Int: BSONValue]
 
+    private enum CodingKeys: CodingKey {
```
kmahar: it always annoys me that the synthesized code is all or none and you can't use the synthesized coding keys if you use a custom `init(from:)` implementation...

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -185,7 +194,8 @@ public struct AggregateOptions: Encodable {
     public let readConcern: ReadConcern?
 
     /// A ReadPreference to use for this operation.
-    public let readPreference: ReadPreference?
+    // swiftlint:disable:next redundant_optional_initialization
```
kmahar: you might want to check if adding this intermediate comment will prevent the docstring from being associated with the property (if you haven't already)

patrickfreed: good catch, they weren't being documented properly. fixed.

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -332,8 +340,15 @@ public struct FindOptions: Encodable {
...
     public let readConcern: ReadConcern?
 
+    // swiftlint:disable redundant_optional_initialization
+
     /// A ReadPreference to use for this operation.
-    public let readPreference: ReadPreference?
+    public var readPreference: ReadPreference? = nil
+
+    /// Indicates the type of cursor to use. This value includes both the tailable and awaitData options.
+    public var cursorType: CursorType? = nil
```
mbroadst: similar to `ReadPreference` here, is it just easier to not make these decodable, or would it not be useful?

patrickfreed: So cursor type is just used as a more ergonomic way to set `tailable` and `awaitData` for users. I think the spec tests would directly specify those options in the tests rather than a cursor type, so we don't need to decode it.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -14,24 +14,32 @@ public struct ClientOptions: CodingStrategyProvider {
...
     public let readConcern: ReadConcern?
 
-    /// Specifies a ReadPreference to use for the client.
-    public let readPreference: ReadPreference?
-
     /// Specifies a WriteConcern to use for the client. If one is not specified, the server's default write concern
     /// will be used.
     public let writeConcern: WriteConcern?
 
+    // swiftlint:disable redundant_optional_initialization
```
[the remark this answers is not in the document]
patrickfreed: It would be pretty easy to implement, but I didn't see any tests that specified one so I didn't bother. I was mainly thinking of only us using this `Decodable` behavior for testing, but this and your other point bring up the question of whether _users_ would ever want to read the various options types from disk or not. I originally didn't think so, but there could be potentially some value in that. It could possibly allow for on the fly configuration without stopping their backend applications and recompiling.

For the purposes of this PR, I think we can leave it as-is. I filed [a ticket](https://jira.mongodb.org/browse/SWIFT-492) for tracking the future work of making all options `Codable`.

PR #271, opened 2019-05-20
Add empty deprecated initialize() and deprecated cleanup()

kmahar:
So I realized we had renamed the actual initialization method to `initializeMongoc`, so I just added an empty `initialize()` method.
I tested this locally, and if you have a call to `MongoSwift.initialize()` you will now get the following compiler warning:
```
/Users/kaitlinmahar/TestProject/Sources/TestProject/main.swift:3:12: warning: 'initialize()' is deprecated: Calling this method no longer has any effect.
MongoSwift.initialize()
           ^
```


kmahar:
Edit: since we renamed `cleanup` I added a deprecated version that calls through to the new one

PR #273, opened 2019-05-21
Fix index test failing on server latest

kmahar:
as of [SERVER-10075](https://jira.mongodb.org/browse/SERVER-10075) server will error if you create an index using `expireAfterSeconds` on more than one key. we had a test that was just trying all the index options at once and creating a multi-key index so it started failing on MacOS which I guess bumped latest in the last few days.

in this PR I:
- moved all the index tests to a new file
- test expireAfterSeconds separately from the other options
- made the test into more of an integration test by calling listIndexes and decoding the results into `IndexOptions` that can be compared against the original options

evg patch confirming this fixes it [here](https://evergreen.mongodb.com/version/5ce476009ccd4e3150c1ff66)


Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -48,7 +48,7 @@ public struct IndexOptions: Encodable {
      *
      * - Example: For an index of name: 1, age: -1, the generated name would be "name_1_age_-1".
      */
-    public let name: String?
+    public var name: String?
```
kmahar: this and `indexVersion` became mutable for the sake of the test. I considered updating all of them but I think that ticket will make it into our next release anyway

Tests/MongoSwiftTests/MongoCollection+IndexTests.swift
```diff
@@ -0,0 +1,237 @@
...
+    func testCreateIndexFromModel() throws {
+        let model = IndexModel(keys: ["cat": 1])
+        expect(try self.coll.createIndex(model)).to(equal("cat_1"))
+        let indexes = try coll.listIndexes()
+        expect(indexes.next()?["name"]).to(bsonEqual("_id_"))
+        expect(indexes.next()?["name"]).to(bsonEqual("cat_1"))
+        expect(indexes.next()).to(beNil())
+    }
+
+    func testIndexOptions() throws {
```
kmahar: this is the only index test that actually changed.

Tests/MongoSwiftTests/MongoCollection+IndexTests.swift
```diff
@@ -0,0 +1,237 @@
...
+
+    func testListIndexes() throws {
+        let indexes = try coll.listIndexes()
+        // New collection, so expect just the _id_ index to exist.
+        expect(indexes.next()?["name"]).to(bsonEqual("_id_"))
+        expect(indexes.next()).to(beNil())
+    }
+}
+
+extension IndexOptions: Equatable {
```
kmahar: this is new too.

PR #277, opened 2019-05-23
SWIFT-134 Write a release script

kmahar:
Adds a release script. For obvious reasons, I have not actually executed the full script in our repo and cannot say with certainty that this will work in its entirety, but I have tested most of it in pieces.


.jazzy.yaml
```diff
@@ -0,0 +1,5 @@
+module: MongoSwift
```
kmahar: There were getting to be a lot of arguments in our call to jazzy in `make documentation` so I pulled them out into here.
we also need to manually specify the authors now because Jazzy was previously reading them out from our podspec.

Makefile
```diff
@@ -19,9 +25,9 @@ project:
 	@$(call check_for_gem,xcodeproj)
 	ruby Tests/Scripts/add_json_files.rb
 
-sourcery:
```
kmahar: Since sourcery is now also used to update `MongoSwift.swift` I renamed this to be more specific.

.travis.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       os: osx
       install: ./Tests/Scripts/install_dependencies.sh sourcery
       before_script: skip
-      script: ${PWD}/sourcery/bin/sourcery; git diff --exit-code Tests/LinuxMain.swift
+      script: export sourcery=${PWD}/sourcery/bin/sourcery; make linuxmain; git diff --exit-code Tests/LinuxMain.swift
```
kmahar: export needed so makefile can find sourcery

kmahar:
Once we have consensus on this I can add a similar script to the mobile repo.

Sources/MongoSwift/MongoSwift.stencil
```diff
@@ -0,0 +1,30 @@
+// swiftlint:disable:previous vertical_whitespace
```
mbroadst: Two thoughts:

- could we just have a `MongoSwiftVersion.swift` (or `Version.swift`) that adds a `private static let versionString` to the right namespace. It seems like there is other stuff going on in this file that we don't want to be templated
- somewhat crazier, but could be nice, can you encode the version into the `Package.swift` and dynamically read it from that file at runtime? `Package.swift` will always be there, in a known fixed location

release/release.sh
```diff
@@ -0,0 +1,59 @@
...
+EOF
+
+# publish new podspec
+pod trunk push ${PWD}/MongoSwift.podspec
+
+# cleanup podspec
+rm ${PWD}/MongoSwift.podspec
+
+# go to GitHub to publish release notes
+open "https://github.com/mongodb/mongo-swift-driver/releases/tag/v${1}"
```
mbroadst: very nice 👏

Sources/MongoSwift/MongoSwift.stencil
kmahar: I went with the first!

PR #280, opened 2019-05-31
Swift-412 Add drop database options

ncarbon:
* Added DropDatabaseOptions with a writeConcern optional that gets sent to mongoc
* Added DropCollectionOptions with a writeConcern optional that gets sent to mongoc
* Made wtimeoutMS in WriteConcern encode as an Int32

### DIFF
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -99,10 +99,10 @@ public class MongoCollection<T: Codable> {
     *   Drops this collection from its parent database.
     * - Parameters:
     *   - session: An optional `ClientSession` to use when executing this command
-    *   - options:  Optional `DropCollectionOptions` to use when executing this command
+    *   - options: An optional `DropCollectionOptions` to use when executing this command
     *
     * - Throws:
-    * - `ServerError.commandError` if an error occurs that prevents the command from executing.
+    *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
     */
     public func drop(session: ClientSession? = nil, options: DropCollectionOptions? = nil) throws {
         let operation = DropCollectionOperation(collection: self, session: session, options: options)
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -104,8 +104,8 @@ public class MongoCollection<T: Codable> {
     * - Throws:
     *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
     */
-    public func drop(session: ClientSession? = nil, options: DropCollectionOptions? = nil) throws {
-        let operation = DropCollectionOperation(collection: self, session: session, options: options)
+    public func drop(options: DropCollectionOptions? = nil, session: ClientSession? = nil) throws {
+        let operation = DropCollectionOperation(collection: self, options: options, session: session)
         try operation.execute()
     }
 }
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -1,5 +1,13 @@
 import mongoc
 
+public struct DropCollectionOptions: Codable {
+    public var writeConcern: WriteConcern?
+
+    public init(writeConcern: WriteConcern? = nil) {
+        self.writeConcern = writeConcern
+    }
+}
+
 /// A MongoDB collection.
 public class MongoCollection<T: Codable> {
     internal var _collection: OpaquePointer?
@@ -87,8 +95,8 @@ public class MongoCollection<T: Codable> {
     /// Drops this collection from its parent database.
     /// - Throws:
     ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropCollectionOperation(collection: self, session: session)
+    public func drop(session: ClientSession? = nil, options: DropCollectionOptions? = nil) throws {
+        let operation = DropCollectionOperation(collection: self, session: session, options: options)
         try operation.execute()
     }
 }
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -1,8 +1,11 @@
 import mongoc
 
+/// Options to use when dropping a collection.
 public struct DropCollectionOptions: Codable {
+    /// An optional `WriteConcern` to use for the command.
     public var writeConcern: WriteConcern?
 
+    /// Initializer allowing any/all parameters to be omitted.
     public init(writeConcern: WriteConcern? = nil) {
         self.writeConcern = writeConcern
     }
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -95,9 +95,15 @@ public class MongoCollection<T: Codable> {
         self._collection = nil
     }
 
-    /// Drops this collection from its parent database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
+    /**
+    *   Drops this collection from its parent database.
+    * - Parameters:
+    *   - session: An optional `ClientSession` to use when executing this command
+    *   - options:  Optional `DropCollectionOptions` to use when executing this command
+    *
+    * - Throws:
+    * - `ServerError.commandError` if an error occurs that prevents the command from executing.
+    */
     public func drop(session: ClientSession? = nil, options: DropCollectionOptions? = nil) throws {
         let operation = DropCollectionOperation(collection: self, session: session, options: options)
         try operation.execute()
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -98,8 +98,8 @@ public class MongoCollection<T: Codable> {
     /**
     *   Drops this collection from its parent database.
     * - Parameters:
-    *   - session: An optional `ClientSession` to use when executing this command
     *   - options: An optional `DropCollectionOptions` to use when executing this command
+    *   - session: An optional `ClientSession` to use when executing this command
     *
     * - Throws:
     *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -144,10 +144,10 @@ public class MongoDatabase {
     *   Drops this database.
     * - Parameters:
     *   - session: An optional `ClientSession` to use for this command
-    *   - options: Optional `DropDatabaseOptions` to use when executing this command
+    *   - options: An optional `DropDatabaseOptions` to use when executing this command
     *
     * - Throws:
-    * - `ServerError.commandError` if an error occurs that prevents the command from executing.
+    *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
     */
     public func drop(session: ClientSession? = nil, options: DropDatabaseOptions? = nil) throws {
         let operation = DropDatabaseOperation(database: self, session: session, options: options)
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -60,6 +60,15 @@ public struct CollectionOptions: CodingStrategyProvider {
     }
 }
 
+/// Options to set when executing dropDatabase command
+public struct DropDatabaseOptions: Codable {
+    public var writeConcern: WriteConcern?
+
+    public init(writeConcern: WriteConcern? = nil) {
+        self.writeConcern = writeConcern
+    }
+}
+
 /// A MongoDB Database.
 public class MongoDatabase {
     internal var _database: OpaquePointer?
@@ -132,8 +141,8 @@ public class MongoDatabase {
     /// Drops this database.
     /// - Throws:
     ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropDatabaseOperation(database: self, session: session)
+    public func drop(session: ClientSession? = nil, options: DropDatabaseOptions? = nil) throws {
+        let operation = DropDatabaseOperation(database: self, session: session, options: options)
         try operation.execute()
     }
 
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -149,8 +149,8 @@ public class MongoDatabase {
     * - Throws:
     *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
     */
-    public func drop(session: ClientSession? = nil, options: DropDatabaseOptions? = nil) throws {
-        let operation = DropDatabaseOperation(database: self, session: session, options: options)
+    public func drop(options: DropDatabaseOptions? = nil, session: ClientSession? = nil) throws {
+        let operation = DropDatabaseOperation(database: self, options: options, session: session)
         try operation.execute()
     }
 
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -62,8 +62,10 @@ public struct CollectionOptions: CodingStrategyProvider {
 
 /// Options to set when executing dropDatabase command
 public struct DropDatabaseOptions: Codable {
+    /// An optional `WriteConcern` to use for the command.
     public var writeConcern: WriteConcern?
 
+    /// Initializer allowing any/all parameters to be omitted.
     public init(writeConcern: WriteConcern? = nil) {
         self.writeConcern = writeConcern
     }
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -143,8 +143,8 @@ public class MongoDatabase {
     /**
     *   Drops this database.
     * - Parameters:
-    *   - session: An optional `ClientSession` to use for this command
     *   - options: An optional `DropDatabaseOptions` to use when executing this command
+    *   - session: An optional `ClientSession` to use for this command
     *
     * - Throws:
     *   - `ServerError.commandError` if an error occurs that prevents the command from executing.
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -60,7 +60,7 @@ public struct CollectionOptions: CodingStrategyProvider {
     }
 }
 
-/// Options to set when executing dropDatabase command
+/// Options to use when executing dropDatabase command.
 public struct DropDatabaseOptions: Codable {
     /// An optional `WriteConcern` to use for the command.
     public var writeConcern: WriteConcern?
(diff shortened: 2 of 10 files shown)


Sources/MongoSwift/WriteConcern.swift
```diff
@@ -145,7 +145,10 @@ public class WriteConcern: Codable {
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encodeIfPresent(self.w, forKey: .w)
-        try container.encodeIfPresent(self.wtimeoutMS, forKey: .wtimeout)
+        if let wtimeout = self.wtimeoutMS {
+            // TODO: change wtimeout to Int64 once tix SWIFT-395 gets fixed
+            try container.encode(Int32(wtimeout), forKey: .wtimeout)
```
kmahar: I think rather than doing `Int32(wtimeout)`, we should use `Int32(truncatingIfNeeded: wtimeout)`. (We already do this in the initializer, see line 103).

With the current approach, the driver will crash if we pass in an Int64 that is too big. With the second one we will just round down.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -110,6 +110,25 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+    func testDropWithWriteConcerns() throws {
+        let encoder = BSONEncoder()
+
+        let center = NotificationCenter.default
+
+        let writeConcern = try WriteConcern(journal: true, w: .number(1))
+        let opts = DropCollectionOptions(writeConcern: writeConcern)
+
+        let observer = center.addObserver(forName: nil, object: nil, queue: nil) { notif in
+            print(notif)
```
kmahar: rather than just printing the notifications, we should inspect them and make sure the commands being sent include the write concern documents.
Specifically we should look for the [`CommandStartedEvent`](https://github.com/mongodb/mongo-swift-driver/blob/cbfa18511567971a86c853b9e0c2ce20cc1bf18f/Sources/MongoSwift/APM.swift#L23)s, and we should make sure that their `command` fields look how we expect. 

The command document might contain some extra info that we don't care about so probably we can just make sure the command name is "drop" and that there is "writeConcern" field that matches the write concern you're sending in the options.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -110,6 +110,25 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+        }
+
+        expect(try self.coll.drop(options: opts)).toNot(throwError())
+        // insert something so we don't error when trying to drop again in teardown
+        expect(try self.coll.insertOne(self.doc1)).toNot(throwError())
+
+        center.removeObserver(observer)
```
kmahar: rather than removing the observer here, we should put this in a `defer` statement right below where we create it, like `defer { center.removeObserver(observer) }`. 

this ensures that if the function doesn't finish running (for ex. that could happen if one of the lines between adding the observer and here throws an error) that we will still remove the observer when the function exits.

you can read a little more about defer statements [here](https://docs.swift.org/swift-book/ReferenceManual/Statements.html) (scroll down). basically they are useful for cleanup code like this 🙂

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -110,6 +110,25 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(try self.coll.insertOne(self.doc1)).toNot(throwError())
     }
 
+    func testDropWithWriteConcerns() throws {
+        let encoder = BSONEncoder()
```
kmahar: are you actually using the encoder anywhere?

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -432,7 +432,7 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
                             expect(try encoder.encode(wc)).to(beNil())
                         } else {
                             if let wtimeoutMS = expected["wtimeout"] as? BSONNumber {
-                                expected["wtimeout"] = wtimeoutMS.int64Value!
+                                expected["wtimeout"] = wtimeoutMS.int32Value!
```
kmahar: let's also add a TODO comment here mentioning the relevant ticket SWIFT-395

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -42,6 +42,35 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
                 .to(throwError(ServerError.commandError(code: 59, message: "", errorLabels: nil)))
     }
 
+    func testDropDatabase() throws {
```
kmahar: the three comments I made on the dropCollection test also apply to the test

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -87,8 +98,8 @@ public class MongoCollection<T: Codable> {
     /// Drops this collection from its parent database.
```
kmahar: I forgot to mention before, we should also update this docstring so that it includes descriptions of the parameters. here's an [example](https://github.com/mongodb/mongo-swift-driver/blob/711a99ea5786ddcf056688b3100bffd457fd76bf/Sources/MongoSwift/MongoCollection%2BWrite.swift#L5) of a docstring where we document parameters

I noticed we never updated this to include the `ClientSession` parameter so I would just document that as something like "An optional session to use for this command".

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -60,6 +60,15 @@ public struct CollectionOptions: CodingStrategyProvider {
     }
 }
 
+/// Options to set when executing dropDatabase command
```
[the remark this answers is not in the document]
kmahar: nit: let's say "use" instead of "set", and end the comment with a period.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -132,8 +143,8 @@ public class MongoDatabase {
     /// Drops this database.
```
kmahar: we should similarly update this docstring like the one for dropCollection

Tests/MongoSwiftTests/MongoCollectionTests.swift
kmahar: let's remove this `print` statement now.
one other thing I would suggest is tracking somehow that we actually saw a `CommandStartedEvent` happen. 
consider if for some reason your call to `collection.drop` below didn't do anything - we would never execute any of the code within this block but the test would still pass.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -42,6 +42,43 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
+        var writeConcern = try WriteConcern(journal: true, w: .number(1))
+
+        let collection = db.collection("collection")
+        try collection.insertOne(["test": "blahblah"])
+
+        var expectedWriteConcerns: [Document] = [
+                                                    ["w": Int32(1), "j": true],
+                                                    ["w": Int32(1), "j": true, "wtimeout": Int32(123)]
+                                                ]
+        let observer = center.addObserver(forName: nil, object: nil, queue: nil) { notif in
```
kmahar: similarly to above, I suggest somehow tracking that we actually saw the `CommandStartedEvent`s happen.
(in this case you could just check that `expectedWriteConcern`s is empty at the end of the test, meaning we saw two events)

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -42,6 +42,43 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
+            guard let event = notif.userInfo?["event"] as? CommandStartedEvent else {
+                return
+            }
+
+            expect(event.command["dropDatabase"]).toNot(beNil())
+            expect(event.command["writeConcern"]).toNot(beNil())
+            expect(event.command["writeConcern"] as? Document).to(equal(expectedWriteConcerns.removeFirst()))
```
kmahar: `removeFirst` crashes if the array you call it on is non-empty, so to be safe we might want to first check that `expectedWriteConcerns.isEmpty` (similar to [here](https://github.com/mongodb/mongo-swift-driver/blob/711a99ea5786ddcf056688b3100bffd457fd76bf/Tests/MongoSwiftTests/CommandMonitoringTests.swift#L62))

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -84,11 +95,17 @@ public class MongoCollection<T: Codable> {
         self._collection = nil
     }
 
-    /// Drops this collection from its parent database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropCollectionOperation(collection: self, session: session)
+    /**
```
kmahar:
```suggestion
   /**
```
super small nit: slightly misaligned

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -129,11 +140,17 @@ public class MongoDatabase {
         self._database = nil
     }
 
-    /// Drops this database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropDatabaseOperation(database: self, session: session)
+    /**
```
kmahar:
```suggestion
   /**
```
same alignment thing

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -42,6 +42,47 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
+        client.enableMonitoring(forEvents: .commandMonitoring)
+
+        var db = client.db(type(of: self).testDatabase)
+        var writeConcern = try WriteConcern(journal: true, w: .number(1))
+
+        let collection = db.collection("collection")
+        try collection.insertOne(["test": "blahblah"])
+
+        var expectedWriteConcerns: [Document] = [
+                                                    ["w": Int32(1), "j": true],
```
patrickfreed: To make this test a little less brittle, I think it would be better to define the array of expected write concerns as an array of the actual type (e.g. `[WriteConcern]`). Then, in the event handler we can encode the popped one and compare (e.g. `encoder.encode(expectedWriteConcerns.removeFirst())`). This removes the dependency on the encoded format of `WriteConcern` from this test, so that when the C driver fixes the bug requiring int32's for wtimeoutms, this test won't have to be updated. 

With this change, below when passing in the write concerns, we can index on `expectedWriteConcerns` itself instead of constructing write concerns that correspond to its entries.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -84,11 +95,17 @@ public class MongoCollection<T: Codable> {
...
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropCollectionOperation(collection: self, session: session)
+   /**
+    *   Drops this collection from its parent database.
+    * - Parameters:
+    *   - session: An optional `ClientSession` to use when executing this command
+    *   - options:  Optional `DropCollectionOptions` to use when executing this command
+    *
+    * - Throws:
+    * - `ServerError.commandError` if an error occurs that prevents the command from executing.
```
patrickfreed: [nit] this line should be indented the same as the options and session line above.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -84,11 +95,17 @@ public class MongoCollection<T: Codable> {
...
-    /// Drops this collection from its parent database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropCollectionOperation(collection: self, session: session)
+   /**
+    *   Drops this collection from its parent database.
+    * - Parameters:
+    *   - session: An optional `ClientSession` to use when executing this command
+    *   - options:  Optional `DropCollectionOptions` to use when executing this command
```
patrickfreed: [nit] This line has an extra space preceding "Optional" that can be removed. Also, the first part can be slightly rewritten to correspond with the "session" description ("An optional `DropCollectionOptions` ... <rest the same as it is>")

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -129,11 +140,17 @@ public class MongoDatabase {
...
-    /// Drops this database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropDatabaseOperation(database: self, session: session)
+   /**
+    *   Drops this database.
+    * - Parameters:
+    *   - session: An optional `ClientSession` to use for this command
+    *   - options: Optional `DropDatabaseOptions` to use when executing this command
```
patrickfreed: ditto nit above

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -129,11 +140,17 @@ public class MongoDatabase {
...
+    *
+    * - Throws:
+    * - `ServerError.commandError` if an error occurs that prevents the command from executing.
```
patrickfreed: ditto nit above

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -129,11 +140,17 @@ public class MongoDatabase {
...
+    */
+    public func drop(session: ClientSession? = nil, options: DropDatabaseOptions? = nil) throws {
```
patrickfreed: The order of these parameters should be switched so that session is the last one to correspond with our other methods.

Sources/MongoSwift/Operations/DropCollectionOperation.swift
```diff
@@ -4,15 +4,17 @@ import mongoc
 internal struct DropCollectionOperation<T: Codable>: Operation {
     private let collection: MongoCollection<T>
     private let session: ClientSession?
+    private let options: DropCollectionOptions?
 
-    internal init(collection: MongoCollection<T>, session: ClientSession?) {
+    internal init(collection: MongoCollection<T>, session: ClientSession?, options: DropCollectionOptions?) {
```
patrickfreed: ditto switching of the session/options ordering here.

Sources/MongoSwift/Operations/DropDatabaseOperation.swift
```diff
@@ -4,15 +4,17 @@ import mongoc
 internal struct DropDatabaseOperation: Operation {
     private let database: MongoDatabase
     private let session: ClientSession?
+    private let options: DropDatabaseOptions?
 
-    internal init(database: MongoDatabase, session: ClientSession?) {
+    internal init(database: MongoDatabase, session: ClientSession?, options: DropDatabaseOptions?) {
```
patrickfreed: ditto ordering

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -110,6 +110,39 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(try self.coll.insertOne(self.doc1)).toNot(throwError())
     }
 
+    func testDropWithWriteConcerns() throws {
```
patrickfreed: I think this test can be combined with `testDrop`, similar to how you have a single test for the database drop.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -110,6 +110,39 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+        let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
+        client.enableMonitoring(forEvents: .commandMonitoring)
+
+        var db = client.db(type(of: self).testDatabase)
+        let writeConcern = try WriteConcern(journal: true, w: .number(1))
+
+        let collection = db.collection("collection")
+        try collection.insertOne(["test": "blahblah"])
+
+        var expectedWriteConcern: Document = ["w": Int32(1), "j": true]
```
patrickfreed: the comments on the other test about not defining the document structure of the write concerns can apply here as well.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -84,11 +95,17 @@ public class MongoCollection<T: Codable> {
...
 
-    /// Drops this collection from its parent database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropCollectionOperation(collection: self, session: session)
+    /**
+    *   Drops this collection from its parent database.
+    * - Parameters:
+    *   - session: An optional `ClientSession` to use when executing this command
```
kmahar: nit: switch order of session and options here to match order in the function signature

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -129,11 +140,17 @@ public class MongoDatabase {
...
 
-    /// Drops this database.
-    /// - Throws:
-    ///   - `ServerError.commandError` if an error occurs that prevents the command from executing.
-    public func drop(session: ClientSession? = nil) throws {
-        let operation = DropDatabaseOperation(database: self, session: session)
+    /**
+    *   Drops this database.
+    * - Parameters:
+    *   - session: An optional `ClientSession` to use for this command
```
kmahar: same nit about switching order of sessions and options in docstring.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -105,9 +105,38 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+            guard let event = notif.userInfo?["event"] as? CommandStartedEvent else {
+                return
+            }
+
+            commandStarted = true
+
+            expect(event.command["drop"]).toNot(beNil())
+            expect(event.command["writeConcern"]).toNot(beNil())
+            expect(event.command["writeConcern"] as? Document)
+                .to(sortedEqual(try? encoder.encode(expectedWriteConcern)))
```
kmahar: do we need to `try?` because we can't throw within this code block?

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -42,6 +42,49 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
+
+        var expectedWriteConcerns: [WriteConcern] = [
+                                                    try WriteConcern(journal: true, w: .number(1)),
+                                                    try WriteConcern(journal: true, w: .number(1), wtimeoutMS: 123)
+                                                ]
+        let observer = center.addObserver(forName: nil, object: nil, queue: nil) { notif in
+            guard let event = notif.userInfo?["event"] as? CommandStartedEvent else {
+                return
+            }
+
```
patrickfreed: One more suggestion: instead of removing elements from the array, I think it might be better to instead just count the number of events seen and compare it to the length of the array at the end. That way, we can iterate over the array itself without worrying about race conditions.

this will allow us to do e.g.:
```swift
for wc in expectedWriteConcerns {
     expect(try db.drop(options: DropDatabaseOptions(writeConcern: wc))).toNot(throwError())
}
```
We can't do this as-is since we'd be removing elements from the array while iterating over it.

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -105,9 +105,38 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
-        // insert something so we don't error when trying to drop again in teardown
-        expect(try self.coll.insertOne(self.doc1)).toNot(throwError())
+        let encoder = BSONEncoder()
+        let center = NotificationCenter.default
+
+        let client = try MongoClient(options: ClientOptions(eventMonitoring: true))
+        client.enableMonitoring(forEvents: .commandMonitoring)
+
+        var db = client.db(type(of: self).testDatabase)
+        let writeConcern = try WriteConcern(journal: true, w: .number(1))
```
patrickfreed: `writeConcern` here is identical to `expectedWriteConcern`, so we can remove it and just use `expectedWriteConcern`.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
kmahar: good point. i'm the one who suggested the removal, we do that in the command monitoring tests. oops

PR #282, opened 2019-06-03
Swift-320 Make custom types conform to hashable

ncarbon:
* Make custom types that conform to BSONValue conform to Hashable
* Update AnyBSONValue Hashable implementation


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -387,6 +394,14 @@ public struct Binary: BSONValue, Equatable, Codable {
     }
 }
 
+// An extension of `BSONUndefined` to add capability to be hashed
+extension Binary: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(data)
```
patrickfreed: In cases where all the members of a struct / class conform to `Hashable`, the compiler actually "synthesizes" the conformance for us. For example, in this case you can simply update the declaration of `Binary` to `Binary: BSONValue, Equatable, Hashable, Codable` and remove this hash function and it be generated for you at compile time. You may actually omit `Equatable` in the declaration I think since `Hashable` implies it.

The same thing applies to `Equatable`. You'll notice there's no `==` defined on `Binary`, we just get it for free since `data` and `subtype` conform to `Equatable`.

This synthesization applies to any type whose `hash` function is just combining all it's fields (e.g. `Timestamp`)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -949,6 +1001,15 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     }
 }
 
+// An extension of `ObjectId` to add the capability to be hashed
+extension ObjectId: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(hex)
```
patrickfreed: libbson provides a [way](http://mongoc.org/libbson/current/bson_oid_hash.html) to generate hashes for `bson_oid_t`'s already, so I think it makes sense to leverage that.

If you look at the `timestamp` computed property, you'll see an example of how we use the underlying libmongoc type. 

In this case, we could do something like: 
```swift
let hashValue = withUnsafePointer(to: self.oid) { oid in
    return bson_oid_hash(oid)
}
hasher.combine(hashValue)
```

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -540,3 +540,10 @@ extension Document: ExpressibleByArrayLiteral {
         self.init(elements)
     }
 }
+
+// An extension of `Document` to add the capability to be hashed
+extension Document: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(self.rawBSON)
+    }
+}
```
patrickfreed: [nit] add a newline at the end of the file.

As part of the workflow, you should make sure both that the tests pass locally and that swiftlint didn't detect any violations before opening a PR. It helps reduce the number of round trips in code review and the time it takes code to get merged.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -572,6 +595,13 @@ public struct Decimal128: BSONNumber, Equatable, Codable, CustomStringConvertibl
      }
 }
 
+// An extension of `Decimal128` to add capability to be hashed
+extension Decimal128: Hashable {
```
patrickfreed: [nit] This is my own fault for giving you potentially bad advice, but I think these should actually be moved inwards into the definitions themselves instead of in extensions. I only say that because it seems that everywhere else in the file, that's the way things are done.

What do you think, @kmahar?

kmahar: I think I'm pro extension actually. while this doesn't quite line up with the rest of this file I think it does line up with the trend we've followed elsewhere in the driver.

I think at some point in the future (probably when I get around to splitting up this massive file into separate files by type) the other conformances should be moved out to extensions, too.

`Codable` conformance often can't be in an extension because of rules about required initializers, but we can put conformance for stuff like `Equatable` and `CustomStringConvertible` in extensions. 

I like having it in the extensions because it makes it very clear what parts of the API are relevant to which protocol.

patrickfreed: Okay, my original comment can be disregarded then.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -572,6 +587,13 @@ public struct Decimal128: BSONNumber, Equatable, Codable, CustomStringConvertibl
...
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(description)
```
kmahar: nit: let's say `self.description` - although the compiler doesn't require it, we try to always explicitly use `self` so it's clearer where the value is coming from.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -496,6 +503,14 @@ public struct DBPointer: BSONValue, Codable, Equatable {
     }
 }
 
+// An extension of `DBPointer` to add capability to be hashed
+extension DBPointer: Hashable {
+    public func hash(into hasher: inout Hasher) {
```
kmahar: I think the comment Patrick made last time about the synthesized implementation applies here - since this is just hashing all the properties, we don't need to write the function ourselves and can just mark the type as `Hashable`.
this should also be true for a few more types below

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -3,24 +3,46 @@ import Foundation
...
+        case let v as Decimal128:
+            hasher.combine(v)
+        case let v as MinKey:
+            hasher.combine(v)
+        case let v as MaxKey:
+            hasher.combine(v)
+        case let v as Document:
+            hasher.combine(v)
+        default:
         hasher.combine("\(self.value)")
```
patrickfreed: i think this line needs an indent.

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -3,24 +3,46 @@ import Foundation
 /// A struct wrapping a `BSONValue` type that allows for encoding/
 /// decoding `BSONValue`s of unknown type.
 public struct AnyBSONValue: Codable, Equatable, Hashable {
-    // TODO: conform all `BSONValue` types to `Hashable` (SWIFT-320).
     public func hash(into hasher: inout Hasher) {
```
patrickfreed: the linter is complaining about "cyclomatic complexity" because there are so many branches in this function. There's no way to avoid the complexity when we have to switch like this, so you just need to add a swiftlint ignore for it. You can examples of that ignore elsewhere in the driver by grepping for "cyclomatic".

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: note that there is a swiftlint rule for this but it is _very_ bad at actually catching violations of the rule...

Sources/MongoSwift/BSON/AnyBSONValue.swift
```diff
@@ -2,25 +2,49 @@ import Foundation
...
-            hasher.combine(date.timeIntervalSince1970)
-        } else if let binary = self.value as? Binary {
-            // `Binary`'s string representation omits the data itself, so we include its hashValue.
-            hasher.combine(binary.data)
-            hasher.combine(binary.subtype)
-        } else if let arr = self.value as? [BSONValue] {
-            // To factor in every item in the array, we include the arrays extended JSON representation.
-            hasher.combine((["value": arr] as Document).extendedJSON)
+
+        switch self.value {
```
kmahar: I think you are missing `BSONNull`, `Timestamp`, `BSONUndefined`, and `DBPointer` here.

kmahar: Now that I look at the rest of this file, I see we are missing the last 3 of those from the `init(from decoder: Decoder)` method below where we try all of the different BSON types. So I'm guessing that's why you missed them here since I told you to look at that for an example... That is a bad bug!! 😩 
I opened a JIRA ticket about this: [SWIFT-508](https://jira.mongodb.org/browse/SWIFT-508)

PR #284, opened 2019-06-04
SWIFT-503/validate w and wtimeoutMS

ncarbon:
* Check that the values for w and wtimeoutMS in WriteConcern are non-negative.


ncarbon:
> Everything you have here looks good!
> In the original ticket, I mentioned that there were some spec tests we are currently skipping because we weren't performing this validation. The place we are skipping them is [here](https://github.com/mongodb/mongo-swift-driver/blob/711a99ea5786ddcf056688b3100bffd457fd76bf/Tests/MongoSwiftTests/ReadWriteConcernTests.swift#L406).
> we should remove that line so we run those cases, and in theory because of your changes here they will pass now.

Do we still want to check for the non-negative validation here though? I added some tests in my previous commits that check for this but I was wondering if we should have more somewhere else.

PR #286, opened 2019-06-07
SWIFT-374 Add ConnectionString type

kmahar:
First of many pooling PRs. Introduces a new `ConnectionString` type which `MongoClient` now uses.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -47,7 +47,7 @@ internal class ConnectionString {
         }
         set(rc) {
             rc.withMongocReadConcern { rcPtr in
-                mongoc_client_set_read_concern(self._uri, rcPtr)
+                mongoc_uri_set_read_concern(self._uri, rcPtr)
             }
         }
     }
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -1,0 +1,86 @@
+import mongoc
+
+/// Class representing a connection string for connecting to MongoDB.
+internal class ConnectionString {
+    /// Pointer to the underlying `mongoc_uri_t`.
+    internal let _uri: OpaquePointer
+
+    /// Initializes a new `ConnectionString` with the provided options. Updates `options` to correctly reflect the
+    /// final options set on the `ConnectionString`.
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -171,41 +171,20 @@ public class MongoClient {
...
-        // if a writeConcern is provided, set it on the client
-        if let wc = options?.writeConcern {
-            mongoc_client_set_write_concern(self._client, wc._writeConcern)
-        }
-
         self.encoder = BSONEncoder(options: options)
         self.decoder = BSONDecoder(options: options)
 
-        if options?.eventMonitoring == true { self.initializeMonitoring() }
+        if options.eventMonitoring == true { self.initializeMonitoring() }
```
patrickfreed: this `== true` can now be omitted.

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -0,0 +1,82 @@
+import mongoc
+
+/// Class representing a connection string for connecting to MongoDB.
+internal class ConnectionString {
+    /// Pointer to the underlying `mongoc_uri_t`.
+    internal let _uri: OpaquePointer
+
+    /// Initializes a new `ConnectionString` with the provided options. Updates `options` to correctly reflect the
+    /// final options set on the `ConnectionString`.
+    internal init(_ connectionString: String, options: inout ClientOptions) throws {
```
patrickfreed: this internal api is pretty weird, but I assume the plan is to re-do it with the "rationalization" ticket. That's totally fine with me in order to get the client pooling stuff rolling.

Sources/MongoSwift/MongoClient.swift
kmahar: good catch, I'm surprised there isn't a compiler warning for this. fixed.

Sources/MongoSwift/ConnectionString.swift
kmahar: yeah, agreed. 
I thought about maybe adding a `ConnectionStringOptions` type and adding a property to `ConnectionString` to retrieve an instance of that that contained the most up to date options, or something like that. but I figured I would just leave that to figure out later.

kmahar:
btw I know this has a conflict now because of the ReadConcern changes, but I'm going to wait to resolve that until we get the WriteConcern changes in too (#288)

PR #287, opened 2019-06-10
Swift-509 Rewrite ReadConcern in pure Swift

ncarbon:
Part 1 of SWIFT-509: Made changes to ReadConcern so that it no longer saves a pointer to monogc type.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -191,10 +191,8 @@ public class MongoClient {
         self.decoder = BSONDecoder(options: options)
 
         // if a readConcern is provided, set it on the client
-        if let rc = options?.readConcern {
-            try? rc.withMongocReadConcern { tmpReadConcernPtr in
-                mongoc_client_set_read_concern(self._client, tmpReadConcernPtr)
-            }
+        options?.readConcern?.withMongocReadConcern { tmpReadConcernPtr in
+            mongoc_client_set_read_concern(self._client, tmpReadConcernPtr)
         }
 
         // if a readPreference is provided, set it on the client
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -187,9 +187,14 @@ public class MongoClient {
             throw UserError.invalidArgumentError(message: "libmongoc not built with TLS support")
         }
 
+        self.encoder = BSONEncoder(options: options)
+        self.decoder = BSONDecoder(options: options)
+
         // if a readConcern is provided, set it on the client
         if let rc = options?.readConcern {
-            mongoc_client_set_read_concern(self._client, rc._readConcern)
+            try? rc.withMongocReadConcern { tmpReadConcernPtr in
+                mongoc_client_set_read_concern(self._client, tmpReadConcernPtr)
+            }
         }
 
         // if a readPreference is provided, set it on the client
@@ -201,9 +206,6 @@ public class MongoClient {
         if let wc = options?.writeConcern {
             mongoc_client_set_write_concern(self._client, wc._writeConcern)
         }
-
-        self.encoder = BSONEncoder(options: options)
-        self.decoder = BSONDecoder(options: options)
 
         if options?.eventMonitoring == true { self.initializeMonitoring() }
 
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -103,7 +103,9 @@ public class MongoDatabase {
         }
 
         if let rc = options?.readConcern {
-            mongoc_database_set_read_concern(db, rc._readConcern)
+            try? rc.withMongocReadConcern { tmpReadConcernPtr in
+                mongoc_database_set_read_concern(db, tmpReadConcernPtr)
+            }
         }
 
         if let rp = options?.readPreference {
@@ -262,6 +264,7 @@ public class MongoDatabase {
     public func runCommand(_ command: Document,
                            options: RunCommandOptions? = nil,
                            session: ClientSession? = nil) throws -> Document {
+        print("level: \(options?.readConcern?.level)")
         let operation = RunCommandOperation(database: self, command: command, options: options, session: session)
         return try operation.execute()
     }
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -102,10 +102,8 @@ public class MongoDatabase {
             fatalError("Couldn't get database '\(name)'")
         }
 
-        if let rc = options?.readConcern {
-            try? rc.withMongocReadConcern { tmpReadConcernPtr in
-                mongoc_database_set_read_concern(db, tmpReadConcernPtr)
-            }
+        options?.readConcern?.withMongocReadConcern { tmpReadConcernPtr in
+            mongoc_database_set_read_concern(db, tmpReadConcernPtr)
         }
 
         if let rp = options?.readPreference {
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -264,7 +264,6 @@ public class MongoDatabase {
     public func runCommand(_ command: Document,
                            options: RunCommandOptions? = nil,
                            session: ClientSession? = nil) throws -> Document {
-        print("level: \(options?.readConcern?.level)")
         let operation = RunCommandOperation(database: self, command: command, options: options, session: session)
         return try operation.execute()
     }
--- a/Sources/MongoSwift/ReadConcern.swift
+++ b/Sources/MongoSwift/ReadConcern.swift
@@ -66,11 +66,6 @@ public struct ReadConcern: Codable {
         }
     }
 
-    /// Initializes a new `ReadConcern` by copying an existing `ReadConcern`.
-    public init(from readConcern: ReadConcern) {
-        self.level = readConcern.level
-    }
-
     /// Initialize a new `ReadConcern` with a `Level`.
     public init(_ level: Level) {
         self.level = level
@@ -108,9 +103,7 @@ public struct ReadConcern: Codable {
         if let level = self.level {
             mongoc_read_concern_set_level(readConcern, level.rawValue)
         }
-        defer {
-            mongoc_read_concern_destroy(readConcern)
-         }
+        defer { mongoc_read_concern_destroy(readConcern) }
         return try body(readConcern)
     }
 }
--- a/Sources/MongoSwift/ReadConcern.swift
+++ b/Sources/MongoSwift/ReadConcern.swift
@@ -90,12 +90,8 @@ public struct ReadConcern: Codable {
         }
     }
 
-    private enum CodingKeys: String, CodingKey {
-        case level
-    }
-
     /**
-     * Creates a new `ReadConcern` with the provided options and passes it to the provided closure.
+     * Creates a new `mongoc_rad_concern_t` based on this `ReadConcern` and passes it to the provided closure.
      * The read concern is only valid within the body of the closure and will be ended after the body completes.
      */
     internal func withMongocReadConcern<T>(_ body: (OpaquePointer) throws -> T) rethrows -> T {
--- a/Sources/MongoSwift/ReadConcern.swift
+++ b/Sources/MongoSwift/ReadConcern.swift
@@ -3,7 +3,7 @@ import mongoc
 /// A struct to represent a MongoDB read concern.
 public struct ReadConcern: Codable {
     /// An enumeration of possible ReadConcern levels.
-    public enum Level: String, Encodable {
+    public enum Level: RawRepresentable, Encodable, Equatable {
         /// See https://docs.mongodb.com/manual/reference/read-concern-local/
         case local
         /// See https://docs.mongodb.com/manual/reference/read-concern-available/
@@ -14,6 +14,39 @@ public struct ReadConcern: Codable {
         case linearizable
         /// See https://docs.mongodb.com/master/reference/read-concern-snapshot/
         case snapshot
+        case custom(level: String)
+
+        public var rawValue: String {
+            switch self {
+            case .local:
+                return "local"
+            case .available:
+                return "available"
+            case .majority:
+                return "majority"
+            case .linearizable:
+                return "linearizable"
+            case .snapshot:
+                return "snapshot"
+            case .custom(let l):
+                return l
+            }
+        }
+
+        public init?(rawValue: String) {
+            switch rawValue {
+            case "local":
+                self = .local
+            case "available":
+                self = .available
+            case "majority":
+                self = .majority
+            case "linearizable":
+                self = .linearizable
+            default:
+                self = .custom(level: rawValue)
+            }
+        }
     }
 
     /// The level of this `ReadConcern`, or `nil` if the level is not set.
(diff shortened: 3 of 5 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -187,9 +187,14 @@ public class MongoClient {
             throw UserError.invalidArgumentError(message: "libmongoc not built with TLS support")
         }
 
+        self.encoder = BSONEncoder(options: options)
```
kmahar: was there a reason for moving this and the next line up here?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -187,9 +187,14 @@ public class MongoClient {
...
+        self.decoder = BSONDecoder(options: options)
+
         // if a readConcern is provided, set it on the client
         if let rc = options?.readConcern {
-            mongoc_client_set_read_concern(self._client, rc._readConcern)
+            try? rc.withMongocReadConcern { tmpReadConcernPtr in
```
kmahar: I don't think we want to `try?` here. if there is an error, we should propagate it to the user using `try`.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -262,6 +264,7 @@ public class MongoDatabase {
     public func runCommand(_ command: Document,
                            options: RunCommandOptions? = nil,
                            session: ClientSession? = nil) throws -> Document {
+        print("level: \(options?.readConcern?.level)")
```
kmahar: let's remove this print statement

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,65 +14,94 @@ public class ReadConcern: Codable {
...
-        return String(cString: level)
     }
 
+    /// The level of this `ReadConcern`, or `nil` if the level is not set.
+    public var level: Level?
+
     /// Indicates whether this `ReadConcern` is the server default.
     public var isDefault: Bool {
-        return mongoc_read_concern_is_default(self._readConcern)
+       if level == nil {
```
kmahar: could we just do  `return level == nil` instead?

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,65 +14,94 @@ public class ReadConcern: Codable {
...
+           return true
+       }
+       return false
     }
 
-    /// Initialize a new `ReadConcern` from a `ReadConcern.Level`.
-    public convenience init(_ level: Level) {
-        self.init(level.rawValue)
+    /// Initializes a new `ReadConcern` by copying a `mongoc_read_concern_t`.
```
kmahar: Since we don't actually do a `copy` of the passed-in `mongoc_read_concern_t` anymore, it might be clearer to say something like "Initializes a new `ReadConcern` with the same level as the provided `mongoc_read_concern_t` or something like that.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -187,9 +187,14 @@ public class MongoClient {
             throw UserError.invalidArgumentError(message: "libmongoc not built with TLS support")
         }
 
+        self.encoder = BSONEncoder(options: options)
+        self.decoder = BSONDecoder(options: options)
+
         // if a readConcern is provided, set it on the client
         if let rc = options?.readConcern {
```
kmahar: now that we don't need to access `rc._readConcern` I think we could rewrite this using optional chaining as

`try options?.readConcern?.withMongocReadConcern { ....`

same applies for where you call the equivalent methods in `MongoDatabase` and `MongoCollection` initializers

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -86,13 +115,19 @@ public class ReadConcern: Codable {
...
-    /// Cleans up internal state.
-    deinit {
-        guard let readConcern = self._readConcern else {
-            return
+    /**
+     * Creates a new `ReadConcern` with the provided options and passes it to the provided closure.
+     * The read concern is only valid within the body of the closure and will be ended after the body completes.
+     */
+    public func withMongocReadConcern<T>(_ body: (OpaquePointer) throws -> T) throws -> T {
+        var readConcern: OpaquePointer = mongoc_read_concern_new()
```
kmahar: 1. does this need to be mutable or can we declare it with `let`? (while we do mutate the value that it points to, I don't think we mutate the pointer itself.)
2. I think you can omit the `: OpaquePointer` type annotation. The compiler should infer it based on the return type of `mongoc_read_concern_new`.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -86,13 +115,19 @@ public class ReadConcern: Codable {
...
 
-    /// Cleans up internal state.
-    deinit {
-        guard let readConcern = self._readConcern else {
-            return
+    /**
+     * Creates a new `ReadConcern` with the provided options and passes it to the provided closure.
+     * The read concern is only valid within the body of the closure and will be ended after the body completes.
+     */
+    public func withMongocReadConcern<T>(_ body: (OpaquePointer) throws -> T) throws -> T {
```
kmahar: So it appears this function only throws an error if the closure passed to it (`body`) also throws. None of the work you do here to create the `mongoc_read_concern_t` throws, you just propagate the error from `body`.

for that reason, we should mark this function as `rethrows` rather than `throws`. here's a nice little blog post on [`rethrows`](https://www.hackingwithswift.com/example-code/language/how-to-use-the-rethrows-keyword).

this should mean that you don't actually need the `try?` / `try` at all where you are using this helper to call `mongoc_client_set_read_concern` etc.

Sources/MongoSwift/MongoClient.swift
kmahar: I think my comment below about using `rethrows` will make this comment obsolete ^

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -30,19 +30,19 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
     func testReadConcernType() throws {
         // check level var works as expected
         let rc = ReadConcern(.majority)
-        expect(rc.level).to(equal("majority"))
+        expect(rc.level).to(equal(ReadConcern.Level.majority))
```
kmahar: I think you can just say `.majority` rather than `ReadConcern.Level.majority`. Since `rc.level` has type `ReadConcern.Level`, the compiler should be able to infer the type.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,65 +14,94 @@ public class ReadConcern: Codable {
         case linearizable
         /// See https://docs.mongodb.com/master/reference/read-concern-snapshot/
         case snapshot
-    }
-
-    /// A pointer to a `mongoc_read_concern_t`.
-    internal var _readConcern: OpaquePointer?
+        case custom(level: String)
```
kmahar: hmm, I'm not sure whether I agree with adding a `custom` case here. I think it implies that a custom level is some feature MongoDB has, when really this is just a catch-all way to provide a value that the server supports but our driver didn't add to the enum yet. Was there a reason for not just leaving the `level` as a `String?` ?

maybe the issue is just the name - maybe something like `other` might better communicate what this case is for?

we should also add a docstring to this case.

it may also be worth seeing how other drivers that provide an enum for the read concern level handle the unknown case.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,65 +14,94 @@ public class ReadConcern: Codable {
...
-    internal init(from readConcern: OpaquePointer?) {
-        self._readConcern = mongoc_read_concern_copy(readConcern)
-    }
-
     private enum CodingKeys: String, CodingKey {
         case level
     }
 
-    public required convenience init(from decoder: Decoder) throws {
+    public init(from decoder: Decoder) throws {
```
kmahar: Since `level` is now a stored property rather than computed, is it possible to delete this `init` and take advantage of the compiler-synthesized implementation?

kmahar: same question applies about the `encode` method.

Sources/MongoSwift/ReadConcern.swift
patrickfreed: This method is only used by us internally and isn't part of the public API, so it can have an `internal` access level control modifier. Also, since the outer portion of this method doesn't throw, it can be marked as `rethrows`, indicating it will only throw if the closure does. This will allow us to omit `try` in cases in which the closure we pass in doesn't throw. See the "rethrowing functions and methods" portion of https://docs.swift.org/swift-book/ReferenceManual/Declarations.html for more info.

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -30,19 +30,19 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
...
 
         // test copy init
         let rc2 = ReadConcern(from: rc)
-        expect(rc2.level).to(equal("majority"))
+        expect(rc2.level).to(equal(ReadConcern.Level.majority))
 
         // test empty init
         let rc3 = ReadConcern()
         expect(rc3.level).to(beNil())
```
patrickfreed: I know this is existing code, but I think it makes sense to add a test for `isDefault` here.

Sources/MongoSwift/MongoClient.swift
ncarbon: Swift wanted all the other fields to be initialized before using self._client inside the closure.
` error: 'self' captured by a closure before all members were initialized`

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -30,19 +30,19 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
     func testReadConcernType() throws {
```
patrickfreed: It's probably worth duplicating all these tests to use the string initializer too (and to cover the `custom` case)

Sources/MongoSwift/ReadConcern.swift
patrickfreed: This initializer can also be internal.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,85 +14,104 @@ public class ReadConcern: Codable {
...
-        var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encodeIfPresent(self.level, forKey: .level)
-    }
-
-    /// Cleans up internal state.
-    deinit {
-        guard let readConcern = self._readConcern else {
-            return
+    /**
+     * Creates a new `ReadConcern` with the provided options and passes it to the provided closure.
```
patrickfreed: I would reword this to say "Creates a new `mongoc_read_concern_t` based on this `ReadConcern` and passes it to the provided closure."

Also "The pointer is only valid within the body of the closure and will be freed after the body completes."

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,85 +14,104 @@ public class ReadConcern: Codable {
         case linearizable
         /// See https://docs.mongodb.com/master/reference/read-concern-snapshot/
         case snapshot
-    }
-
-    /// A pointer to a `mongoc_read_concern_t`.
-    internal var _readConcern: OpaquePointer?
+        /// Unknown level
```
patrickfreed: I think this docstring could be a bit more descriptive. Something like "Any other read concern level not covered by the other cases. This case is present to provide forwards compatibility with any future read concerns which may be added to new versions of MongoDB."

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,85 +14,104 @@ public class ReadConcern: Codable {
...
+        if let level = mongoc_read_concern_get_level(readConcern) {
+            self.level = Level(rawValue: String(cString: level))
+        }
     }
 
-    /// Initialize a new `ReadConcern` from a `ReadConcern.Level`.
-    public convenience init(_ level: Level) {
-        self.init(level.rawValue)
+    /// Initializes a new `ReadConcern` by copying an existing `ReadConcern`.
+    public init(from readConcern: ReadConcern) {
```
kmahar: do we use this initializer anywhere?
now that ReadConcern is a value type, I'm not sure how useful it really is, since if you need a copy you can just do `let copyOfRC = oldRC`. (if you do that with a class, both variables still point to the same object.)

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,85 +14,104 @@ public class ReadConcern: Codable {
...
+     * The read concern is only valid within the body of the closure and will be ended after the body completes.
+     */
+    internal func withMongocReadConcern<T>(_ body: (OpaquePointer) throws -> T) rethrows -> T {
+        let readConcern: OpaquePointer = mongoc_read_concern_new()
+        if let level = self.level {
+            mongoc_read_concern_set_level(readConcern, level.rawValue)
         }
-        mongoc_read_concern_destroy(readConcern)
-        self._readConcern = nil
+        defer {
```
kmahar: tiny nit: I would prefer the `defer` happens in the line right after initialization, before we set the level. then it's very clear at a quick glance that we are cleaning this thing up correctly.

Sources/MongoSwift/ReadConcern.swift
patrickfreed: This still needs to be updated.

Sources/MongoSwift/ReadConcern.swift
patrickfreed: The second part about why the case is present (for forwards compatibility) is important to include also. Otherwise users will wonder why we have this last case instead of just enumerating all the possibilities. 

Also, a small style nit: make sure to end all the docstrings in periods. I know I always forget to do that...

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,85 +14,97 @@ public class ReadConcern: Codable {
...
-        self._readConcern = mongoc_read_concern_copy(readConcern._readConcern)
-    }
-
-    /// Initializes a new `ReadConcern` by copying a `mongoc_read_concern_t`.
-    /// The caller is responsible for freeing the original `mongoc_read_concern_t`.
-    internal init(from readConcern: OpaquePointer?) {
-        self._readConcern = mongoc_read_concern_copy(readConcern)
-    }
-
     private enum CodingKeys: String, CodingKey {
```
kmahar: I think we could get rid of this `CodingKeys` type too, since what we have here is what the compiler will synthesize, as `level` is the only stored property.

kmahar:
a couple of comments from my previous review and Patrick's note about the `other` case for level still need to be addressed. otherwise looks good.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -14,85 +14,95 @@ public class ReadConcern: Codable {
...
-        var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encodeIfPresent(self.level, forKey: .level)
-    }
-
-    /// Cleans up internal state.
-    deinit {
-        guard let readConcern = self._readConcern else {
-            return
+    /**
+     * Creates a new `mongoc_rad_concern_t` based on this `ReadConcern` and passes it to the provided closure.
```
kmahar:
```suggestion
     * Creates a new `mongoc_read_concern_t` based on this `ReadConcern` and passes it to the provided closure.
```

PR #288, opened 2019-06-12
SWIFT-504 Rewrite WriteConcern in pure Swift

ncarbon:


Sources/MongoSwift/WriteConcern.swift
```diff
@@ -113,15 +105,14 @@ public class WriteConcern: Codable {
                 if wNumber < 0 {
                     throw UserError.invalidArgumentError(message: "Invalid value: w=\(w) cannot be negative.")
                 }
-                mongoc_write_concern_set_w(self._writeConcern, wNumber)
```
kmahar: so in this switch statement, all we really care about now is the case where it's `.number`, since we need to validate the stored number is >= 0. 

otherwise, all we end up doing in every case is setting `self.w` to `w`. we don't need to reconstruct the appropriate cases like we are now with `self.w = .tag(wTag)` and `self.w = .majority`.

I think we can simplify this logic and get rid of the switch statement using a `case let` (kinda funky but cool syntax to match a single case):
```swift
if case let .number(wNumber) = w {
    // error if wNumber < 0
} 

self.w = w
```

note this should also handle the case where the `w` parameter is nil - the `if` will evaluate to false and `self.w` will correctly be set to `nil`.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -131,17 +122,42 @@ public class WriteConcern: Codable {
...
     /// The caller is responsible for freeing the original `mongoc_write_concern_t`.
     internal init(from writeConcern: OpaquePointer?) {
-        self._writeConcern = mongoc_write_concern_copy(writeConcern)
+        if mongoc_write_concern_journal_is_set(writeConcern) {
+            self.journal = mongoc_write_concern_get_journal(writeConcern)
+        } else {
+            self.journal = nil
+        }
+
+       let number = mongoc_write_concern_get_w(writeConcern)
```
kmahar: the alignment looks funky around here, maybe this needs to be one space forward?

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -39,72 +36,67 @@ public class WriteConcern: Codable {
...
-    public var wtimeoutMS: Int64? {
-        let timeout = mongoc_write_concern_get_wtimeout(self._writeConcern)
-        return timeout == 0 ? nil : Int64(timeout)
-    }
+    public var wtimeoutMS: Int64?
 
     /// Indicates whether this is an acknowledged write concern.
     public var isAcknowledged: Bool {
-        return mongoc_write_concern_is_acknowledged(self._writeConcern)
+        if let w = self.w {
```
kmahar: similar to my other comment below, I think we could use `if case let` to just match the `.number` case and handle that logic.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -131,17 +122,42 @@ public class WriteConcern: Codable {
...
+           self.wtimeoutMS = wtimeout
+       }
     }
 
     private enum CodingKeys: String, CodingKey {
         case w, j, wtimeout
     }
 
-    public required convenience init(from decoder: Decoder) throws {
+    public init(from decoder: Decoder) throws {
```
kmahar: rather than having this manual `decode` implementation, I think we can delete this and use the synthesized one now that we store all the properties.

*but*, in order for the synthesized one to be correct, we probably need to tell the compiler about how the key names don't quite match up. WriteConcern documents use `j` for `journal` and `wtimeout` for `wtimeoutMS`. 

we can do that by updating the `CodingKeys` to reflect this:
```swift
   case w, journal = "j", wtimeoutMS = "wtimeout"
}
```
This tells the compiler that the value for our `w` property will be stored in a document under the key `"w"`, but that `journal` will be stored under `"j"` and `wtimeoutMS` will be stored under `"wtimeout"`.

you will then need to also update the `encode` method to use those new CodingKey cases, e.g. the last line becomes `try container.encodeIfPresent(self.journal, forKey: .journal)`

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -158,14 +174,30 @@ public class WriteConcern: Codable {
         }
         try container.encodeIfPresent(self.journal, forKey: .j)
     }
-
-    /// Cleans up internal state.
-    deinit {
-        guard let writeConcern = self._writeConcern else {
-            return
+    /**
+    *  Creates a new `mongoc_write_concern_t` based on this `WriteConcern` and passes it to the provided closure.
```
kmahar: nit: this asterisk and the next one should both be forward one space, aligned with the first asterisk in the opening line.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -158,14 +174,30 @@ public class WriteConcern: Codable {
...
+    *  The pointer is only valid within the body of the closure and will be freed after the body completes.
+    **/
```
kmahar: nit: only one asterisk for the closing line, so just `*/`

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -39,72 +36,67 @@ public class WriteConcern: Codable {
...
     /// Indicates whether this is the default write concern.
     public var isDefault: Bool {
-        return mongoc_write_concern_is_default(self._writeConcern)
+        return self.w == nil && self.journal == nil && self.wtimeoutMS == nil
     }
 
     /// Indicates whether the combination of values set on this `WriteConcern` is valid.
     private var isValid: Bool {
-        return mongoc_write_concern_is_valid(self._writeConcern)
+        if let w = self.w {
```
kmahar: same comment applies about using `if case let` to check the one case we care about

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -131,17 +122,42 @@ public class WriteConcern: Codable {
...
     }
 
-    /// Initializes a new `WriteConcern` by copying a `mongoc_write_concern_t`.
+    /// Initializes a new `WriteConcern` with the same values as the provided `mongoc_write_concern_t`.
     /// The caller is responsible for freeing the original `mongoc_write_concern_t`.
     internal init(from writeConcern: OpaquePointer?) {
-        self._writeConcern = mongoc_write_concern_copy(writeConcern)
+        if mongoc_write_concern_journal_is_set(writeConcern) {
+            self.journal = mongoc_write_concern_get_journal(writeConcern)
+        } else {
```
kmahar: do we need the `else`? I would think `self.journal` is automatically nil if we don't execute the `if`

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -131,17 +122,42 @@ public class WriteConcern: Codable {
...
+        case MONGOC_WRITE_CONCERN_W_TAG:
+            if let wTag = mongoc_write_concern_get_wtag(writeConcern) {
+                self.w = .tag(String(cString: wTag))
+            }
+        default:
+            self.w = .number(number)
+        }
+
+       let wtimeout = Int64(mongoc_write_concern_get_wtimeout(writeConcern))
+       if wtimeout == 0 {
```
kmahar: similar to comment above, do we need to explicitly set it to nil or can we have an `if` that sets the property when it's non-zero?

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -39,72 +36,67 @@ public class WriteConcern: Codable {
...
+        self.wtimeoutMS = nil
     }
 
     /// Initializes a new `WriteConcern`.
     /// - Throws:
     ///   - `UserError.invalidArgumentError` if the options form an invalid combination.
     public init(journal: Bool? = nil, w: W? = nil, wtimeoutMS: Int64? = nil) throws {
-        self._writeConcern = mongoc_write_concern_new()
-        if let journal = journal { mongoc_write_concern_set_journal(self._writeConcern, journal) }
+        if let journal = journal {
```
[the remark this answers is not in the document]
ncarbon: Oops, that's right, we're just initializing journal here, unlike the values for wtimeoutMS and w which are checked before initializing them.

Updated.

Sources/MongoSwift/WriteConcern.swift
ncarbon: Oh, yeah, I did find it kind of awkward having to use switch statement to extract the  `.number`, just didn't know you could use `case let` by itself. That's neat, thanks! 😄 

Updated.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -187,6 +187,9 @@ public class MongoClient {
             throw UserError.invalidArgumentError(message: "libmongoc not built with TLS support")
         }
 
+        self.encoder = BSONEncoder(options: options)
```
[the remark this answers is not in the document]
ncarbon: Yes, I have rebased on master now.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -39,89 +36,64 @@ public class WriteConcern: Codable {
...
-            if let wTag = mongoc_write_concern_get_wtag(self._writeConcern) {
-                return .tag(String(cString: wTag))
-            }
-        default:
-            break
-        }
-
-        return .number(number)
-    }
+    public var w: W?
```
patrickfreed: If these properties are mutable, it's possible that a user could modify them to get to an invalid write concern, bypassing the checks we have in the initializer. Because of that, let's make `w`, `journal`, and `wtimeoutMS` immutable as they are today, and we can figure out later if we want to change the API to accommodate mutability.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -131,41 +103,69 @@ public class WriteConcern: Codable {
...
+        case w, journal = "j", wtimeoutMS = "wtimeout"
     }
 
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encodeIfPresent(self.w, forKey: .w)
         if let wtimeout = self.wtimeoutMS {
             // TODO: change wtimeout to Int64 once tix SWIFT-395 gets fixed
-            try container.encode(Int32(truncatingIfNeeded: wtimeout), forKey: .wtimeout)
+            try container.encode(Int32(truncatingIfNeeded: wtimeout), forKey: .wtimeoutMS)
```
kmahar: Note that this line is now the only reason we need to have a manual `encode` function. Otherwise we could just encode the stored properties.
For that reason I think it is worth updating the TODO comment here noting that this is the reason for our manual implementation and that we can remove this `encode` method once SWIFT-395 is fixed.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -39,89 +36,64 @@ public class WriteConcern: Codable {
...
-        let timeout = mongoc_write_concern_get_wtimeout(self._writeConcern)
-        return timeout == 0 ? nil : Int64(timeout)
-    }
+    public var wtimeoutMS: Int64?
 
     /// Indicates whether this is an acknowledged write concern.
     public var isAcknowledged: Bool {
-        return mongoc_write_concern_is_acknowledged(self._writeConcern)
+        if let w = self.w, case let .number(wNumber) = w {
+            return !((journal == nil || journal == false ) && wNumber == 0)
```
kmahar: for places like this where we are now implementing this kind of logic dictated by the spec, I think it can be helpful to a future reader if you add a comment (possibly just copy-pasting a sentence from the spec) indicating what this is supposed to be doing.

for example this could have a comment taken from the spec:
`// An Unacknowledged WriteConcern is when (w equals 0) AND (journal is not set or is false)`

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -39,89 +36,64 @@ public class WriteConcern: Codable {
...
     public var isDefault: Bool {
-        return mongoc_write_concern_is_default(self._writeConcern)
+        return self.w == nil && self.journal == nil && self.wtimeoutMS == nil
     }
 
     /// Indicates whether the combination of values set on this `WriteConcern` is valid.
     private var isValid: Bool {
-        return mongoc_write_concern_is_valid(self._writeConcern)
+        if let w = self.w, case let .number(wNumber) = w {
+            return journal == nil || journal == false || wNumber != 0
```
kmahar: could maybe use a comment explaining what this is checking too

Sources/MongoSwift/WriteConcern.swift
ncarbon: If we make the journal, w, and wtimeoutMS immutable as @patrickfreed suggests, Swift will complain about the values not being initialized because there's no `else`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -201,8 +201,8 @@ public class MongoClient {
         }
 
         // if a writeConcern is provided, set it on the client
-        if let wc = options?.writeConcern {
-            mongoc_client_set_write_concern(self._client, wc._writeConcern)
+        options?.writeConcern?.withMongocWriteConcern { tmpWriteConcern in
```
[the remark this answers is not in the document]
ncarbon: Updated

Sources/MongoSwift/WriteConcern.swift
ncarbon: @kmahar Making the WriteConcern values immutable will require having the else statements that initialize them to nil in the `init` with `mongoc_write_concern_t`. Swift complains because it thinks if statements could fail and possibly the switch statements too?
`Error: return from initializer without initializing all stored properties`

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -76,8 +76,8 @@ public class MongoCollection<T: Codable> {
             mongoc_collection_set_read_prefs(collection, rp._readPreference)
         }
 
-        if let wc = options?.writeConcern {
-            mongoc_collection_set_write_concern(collection, wc._writeConcern)
+        options?.writeConcern?.withMongocWriteConcern { tmpWriteConcern in
```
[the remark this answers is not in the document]
ncarbon: Updated.

Sources/MongoSwift/WriteConcern.swift
kmahar: ok we can explicitly set them to nil in those cases then 👍

Sources/MongoSwift/WriteConcern.swift
ncarbon: Updated

kmahar: this still needs to be fixed
```suggestion
        let number = mongoc_write_concern_get_w(writeConcern)
```

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -131,41 +105,76 @@ public class WriteConcern: Codable {
...
+            } else {
+                self.w = nil
+            }
+        default:
+            self.w = .number(number)
+        }
 
-    private enum CodingKeys: String, CodingKey {
-        case w, j, wtimeout
+       let wtimeout = Int64(mongoc_write_concern_get_wtimeout(writeConcern))
```
kmahar: I think this line down through the end of the `else` is also misaligned, looks like it all needs to be forward one space.

PR #291, opened 2019-06-24
SWIFT-445 Add Atlas connectivity tests

kmahar:
Adds a new `AtlasConnectivity` executable target to the driver and updates evergreen config to run that on Ubuntu.

[sample evg patch](https://evergreen.mongodb.com/version/5d1147d77742ae061c0162bf) (huge backup on MacOS patches rn but Ubuntu has completed anyway)


.evergreen/config.yml
```diff
@@ -172,101 +172,122 @@ functions:
 
 pre:
   - func: "fetch source"
-  - func: "prepare resources"
```
kmahar: `prepare resources` downloads drivers evergreen tools and starts mongo orchestration, so we don't want to do it before all tasks anymore as its irrelevant for atlas tests.

.evergreen/config.yml
```diff
@@ -172,101 +172,122 @@ functions:
...
   - func: "install dependencies"
 
-post:
```
kmahar: similarly, both of these tasks are irrelevant for atlas tests. we haven't started orchestration, and cleanup just removes the drivers evg tools we haven't downloaded.

.evergreen/config.yml
```diff
@@ -326,11 +347,44 @@ axes:
 
 buildvariants:
 
-- matrix_name: "tests-all"
```
kmahar: i couldn't figure out a way to still use the matrix but have it skip the atlas tests on macOS, let me know if you have any ideas

Package.swift
```diff
@@ -12,6 +12,7 @@ let package = Package(
     ],
     targets: [
         .target(name: "MongoSwift", dependencies: ["mongoc", "bson"]),
-        .testTarget(name: "MongoSwiftTests", dependencies: ["MongoSwift", "Nimble"])
+        .target(name: "AtlasConnectivity", dependencies: ["MongoSwift"]),
+        .testTarget(name: "MongoSwiftTests", dependencies: ["MongoSwift", "Nimble", "mongoc"])
```
kmahar: I think that this has just worked without mongoc there since the driver always gets built first, but I think technically mongoc should be a dependency here as we directly `import libmongoc` in a few test files.

.evergreen/run-atlas-tests.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# override where we look for libmongoc
+export LD_LIBRARY_PATH="${INSTALL_DIR}/lib"
+export DYLD_LIBRARY_PATH="${INSTALL_DIR}/lib"
+export DEVELOPER_DIR=/Applications/Xcode10.1.app
+
+swiftenv local $SWIFT_VERSION
+
+# run the tests
+ATLAS_REPL="$ATLAS_REPL" ATLAS_SHRD="$ATLAS_SHRD" ATLAS_FREE="$ATLAS_FREE" ATLAS_TLS11="$ATLAS_TLS11" ATLAS_TLS12="$ATLAS_TLS12" swift run AtlasConnectivity $EXTRA_FLAGS
```
patrickfreed: Does `swift run` work on targets too, not just executable products? Also, will we be actually running this / populating these environment variables from a private evg project?

kmahar: yup, `swift run AtlasConnectivity` runs this. I've omitted this from the products to prevent someone from doing `import AtlasConnectivity`. 
the output of `swift run --help` is confusing in that it says "Build and run an executable product" when in reality it builds and runs an executable target.
I've just set it up on our normal evg project, and all of these are private project variables added to the project settings via the UI. 
that said, from discussion in the comments on [DRIVERS-382](https://jira.mongodb.org/browse/DRIVERS-382) it sounds like Bernie wants us to use separate projects, though from a quick look at other drivers on evg it seems Node, Java, Python, maybe more all are just using their normal projects. thoughts @mbroadst ?

mbroadst: yeah I think no need to set up a separate project, as long as you make sure never to `set -x` on the atlas shell script (which would leak the credentials)

.evergreen/config.yml
```diff
@@ -172,101 +172,122 @@ functions:
...
 
-post:
-  - func: "stop mongo-orchestration"
-  - func: "cleanup"
-
 tasks:
     - name: "test-3.6-standalone"
       tags: ["3.6", "standalone"]
       commands:
+        - func: "prepare resources"
```
mbroadst: dang, this is annoying but I guess unavoidable. I see the python driver leaves "cleanup" tasks (including stopping mongo orchestration) in the `post` to avoid some of this duplication, maybe worth considering here?

.evergreen/config.yml
mbroadst: see here: https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/config.yml#L1520

you'll just make a matrix entry that specifies a specific host

kmahar: ah, yeah i was trying to embed these tasks within the existing matrix but adding a new one works too.

.evergreen/config.yml
kmahar: ya, we could do that, as any errors in `post` won't cause a task to fail. 
I had that originally, however, it makes the log output kind of confusing as it will always contain an error about not being able to find mongo-orchestration. 
cleanup should always succeed anyway as all it does is `rm -rf` a directory so we don't need to duplicate that one, I've removed it

PR #292, opened 2019-06-24
SWIFT-374 Operation layer changes

kmahar:
This PR:
- Introduces the `OperationExecutor` protocol and a `DefaultOperationExecutor` type. `MongoClient` stores an executor which is used for executing all operations that rely on the client. 
- Updates the `Operation` protocol so that `execute` now takes in a `Connection` and `ClientSession` with which to execute an `Operation`. 
- Updates all `Operation` types to conform to the protocol 
    - At the moment, they all ignore the passed-in `Connection`. In the next PR they will start using those to construct the short-lived DB and client pointers.
    - The operations no longer need to store their `ClientSession`s since they get passed into `execute`.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -125,7 +125,7 @@ public class MongoClient {
 
     internal let connectionPool: ConnectionPool
 
-    private let executor = DefaultOperationExecutor()
+    private let operationExecutor: OperationExecutor = DefaultOperationExecutor()
 
     /// If command and/or server monitoring is enabled, stores the NotificationCenter events are posted to.
     internal var notificationCenter: NotificationCenter?
@@ -280,6 +280,6 @@ public class MongoClient {
     /// Executes an `Operation` using this `MongoClient` and an optionally provided session.
     internal func executeOperation<T: Operation>(_ operation: T,
                                                  session: ClientSession? = nil) throws -> T.OperationResult {
-        return try self.executor.execute(operation, client: self, session: session)
+        return try self.operationExecutor.execute(operation, client: self, session: session)
     }
 }
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -125,6 +125,8 @@ public class MongoClient {
 
     internal let connectionPool: ConnectionPool
 
+    private let executor = DefaultOperationExecutor()
+
     /// If command and/or server monitoring is enabled, stores the NotificationCenter events are posted to.
     internal var notificationCenter: NotificationCenter?
 
@@ -273,5 +275,11 @@ public class MongoClient {
      */
     public func db(_ name: String, options: DatabaseOptions? = nil) -> MongoDatabase {
         return MongoDatabase(name: name, client: self, options: options)
+    }
+
+    /// Executes an `Operation` using this `MongoClient` and an optionally provided session.
+    internal func executeOperation<T: Operation>(_ operation: T,
+                                                 session: ClientSession? = nil) throws -> T.OperationResult {
+        return try self.executor.execute(operation, client: self, session: session)
     }
 }
--- a/Sources/MongoSwift/MongoCursor.swift
+++ b/Sources/MongoSwift/MongoCursor.swift
@@ -119,7 +119,7 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
         do {
             let operation = NextOperation(cursor: self)
             // TODO SWIFT-374: we temporarily use a fake connection here. eventually MongoCursor will store its source
-            // connection and we can pass that in here instead (though the execute method still won't use it.) 
+            // connection and we can pass that in here instead (though the execute method still won't use it.)
             // we want to pass in a specific connection and call execute() ourselves rather than using an operation
             // executor as that would require checking out a separate, unused connection for every single next() call.
             // the force unwrap is safe as this bitPattern is always valid.
--- a/Sources/MongoSwift/MongoCursor.swift
+++ b/Sources/MongoSwift/MongoCursor.swift
@@ -4,7 +4,7 @@ import mongoc
 public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     internal var _cursor: OpaquePointer?
     private var _client: MongoClient?
-    internal var _session: ClientSession?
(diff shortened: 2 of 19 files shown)


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -118,7 +118,13 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     public func next() -> T? {
         do {
             let operation = NextOperation(cursor: self)
-            let out = try operation.execute()
+            // TODO SWIFT-374: we temporarily use a fake connection here. eventually MongoCursor will store its source
```
kmahar: (I deferred this til a future PR as it will require all the methods returning cursors to start checking out connections and I think the diff will be somewhat significant)

Sources/MongoSwift/Operations/DistinctOperation.swift
```diff
@@ -33,34 +33,28 @@ public struct DistinctOptions: Codable {
 }
 
 /// An operation corresponding to a "distinct" command on a collection.
-internal struct DistinctOperation<T: Codable> {
+internal struct DistinctOperation<T: Codable>: Operation {
```
kmahar: this never actually conformed 🤦‍♀

Sources/MongoSwift/MongoClient.swift
```diff
@@ -125,6 +125,8 @@ public class MongoClient {
 
     internal let connectionPool: ConnectionPool
 
+    private let executor = DefaultOperationExecutor()
```
mbroadst: nit: `operationExecutor`

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -4,7 +4,7 @@ import mongoc
 public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     internal var _cursor: OpaquePointer?
     private var _client: MongoClient?
-    internal var _session: ClientSession?
+    private var _session: ClientSession?
```
mbroadst: are `_client` and `_session` `fileprivate` now?

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -118,7 +118,13 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         do {
             let operation = NextOperation(cursor: self)
-            let out = try operation.execute()
+            // TODO SWIFT-374: we temporarily use a fake connection here. eventually MongoCursor will store its source
+            // connection and we can pass that in here instead (though the execute method still won't use it.)
+            // we want to pass in a specific connection and call execute() ourselves rather than using an operation
+            // executor as that would require checking out a separate, unused connection for every single next() call.
+            // the force unwrap is safe as this bitPattern is always valid.
+            // swiftlint:disable:next force_unwrapping
+            let out = try operation.execute(using: Connection(OpaquePointer(bitPattern: 1)!), session: self._session)
```
mbroadst: Consider making a base `Connection` protocol so that we can pass in a `FakeConnection`/`DummyConnection` here, and possibly mock out connections passed in for operation execution.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -274,4 +276,10 @@ public class MongoClient {
     public func db(_ name: String, options: DatabaseOptions? = nil) -> MongoDatabase {
         return MongoDatabase(name: name, client: self, options: options)
     }
+
+    /// Executes an `Operation` using this `MongoClient` and an optionally provided session.
+    internal func executeOperation<T: Operation>(_ operation: T,
+                                                 session: ClientSession? = nil) throws -> T.OperationResult {
+        return try self.executor.execute(operation, client: self, session: session)
```
mbroadst: not sure where we landed on this, but is the label for `session` superfluous here?

Sources/MongoSwift/MongoCursor.swift
kmahar: we will need to be able to read the underlying `mongoc_client_t` property out of real passed in `Connection`s, so I guess we would end up with something like this then? : 

```swift
/// A connection to the database.
internal protocol Connection {
    /// Pointer to underlying state.
    let handle: OpaquePointer
}

/// A connection to mongoc.
internal struct MongocConnection: Connection {
    /// Pointer to the underlying `mongoc_client_t`.
    internal let handle: OpaquePointer

    internal init(_ handle: OpaquePointer) {
        self.handle = handle
    }
}

/// A mock connection.
internal struct DummyConnection: Connection {
    internal let handle = OpaquePointer(bitPattern: 1)!
}
```

Sources/MongoSwift/MongoCursor.swift
kmahar: they're `private` - we only use both of them within `MongoCursor` which is all contained in this file. 
`fileprivate` would be mildly more permissive and allow them to be used within this file outside of the class definition

Sources/MongoSwift/MongoCursor.swift
kmahar: something slightly weird about this is that our `ConnectionPool` is really then just a `MongocConnectionPool` 🤔 I guess maybe eventually `ConnectionPool` would become a protocol and then we could have a mock pool implementation?
since this line is going to disappear in a subsequent PR anyway i wonder if we should wait to introduce this stuff until we will actually need/use a `DummyConnection`

patrickfreed: If one of your next PRs is going to remove the dummy connection stuff anyways then I don't see a big need for including any of the added machinery around it. If as part of the design we'll continue to use dummy connections until say, we remove libmongoc, then I'm for including the `DummyConnection` thing.

Sources/MongoSwift/MongoClient.swift
patrickfreed: I personally try to limit how long I think about internal parameter stuff, so I don't really feel strongly either way also (I spent _way_ too long thinking about what to call the error getters). My general approach these days is to choose something acceptable that I can update either in code review if someone else does feel strongly one way or, if after using it in the code for a while another way feels more natural, just refactor it.

Sources/MongoSwift/MongoCursor.swift
mbroadst: Let's defer this decision

PR #293, opened 2019-06-26
SWIFT-178 Implement retryable writes runner

patrickfreed:
This is the last PR in a series of them for implementing the spec test runner for retryable writes. 

This PR adds the retryable writes runner itself.

I forgot to rebase this off master so the commit history includes some stuff from the previous PR, so you can ignore that. The diff appears correct, though.


Tests/MongoSwiftTests/SpecTestRunner/SpecTest.swift
```diff
@@ -65,9 +65,8 @@ extension SpecTest {
             expect(seenError).to(beNil(), description: self.description)
         }
 
-        if let expectedResult = self.outcome.result {
-            expect(result).toNot(beNil(), description: self.description)
-            expect(result).to(equal(expectedResult), description: self.description)
+        if let expectedResult = self.outcome.result, let receivedResult = result {
```
patrickfreed: [One of the tests](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/bulkWrite.yml#L287) consistently sees a connection error instead of a bulk write error (which may or may not be intended, I couldn't really tell), so I updated this to only compare results when it did receive a bulk write error. Let me know if you think this is the right approach.

Tests/MongoSwiftTests/RetryableWritesTests.swift
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    func testRetryableWrites() throws {
+        let testFilesPath = MongoSwiftTestCase.specsPath + "/retryable-writes/tests"
+        let testFiles: [String] = try FileManager.default.contentsOfDirectory(atPath: testFilesPath)
+
+        let tests: [RetryableWritesTestFile] = try testFiles.map { fileName in
+            let url = URL(fileURLWithPath: "\(testFilesPath)/\(fileName)")
+            let data = try String(contentsOf: url).data(using: .utf8)!
+            var testFile = try JSONDecoder().decode(RetryableWritesTestFile.self, from: data)
```
kmahar: since we are decoding various documents, it might be more performant to use the `BSONDecoder` here rather than `JSONDecoder`, since using `JSONDecoder` will necessitate decoding each value in a doc individually via `AnyBSONValue` which uses that long switch statement to test every type.
OTOH we have minimal tests for how `JSONDecoder` works with `Document`, so maybe it's good to do that for once 🤷‍♀

Tests/MongoSwiftTests/SpecTestRunner/SpecTest.swift
patrickfreed: Woops, that link was wrong. [This](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/bulkWrite.yml#L212) is the correct one. The failpoint is configured to fail twice which seems to me like this write is supposed to fail, but I'm not sure if it's supposed to fail with a connection error. I'll check with other drivers.

patrickfreed: So after talking with Jeremy, I have made the following changes:
- `ServerError.bulkWriteError` now has an `otherError` field corresponding to any error that falls outside of the write error / write concern error definitions (e.g. connection)
- The only `ServerError` or `RuntimeError` case that `MongoCollection.bulkWrite` will throw is `.bulkWriteError`, even in the case of connection / auth errors.
    - This is a little undesirable, but it is pretty much the only way to remain consistent when dealing with bulk write errors.
    - We _could_ throw top level other errors when the result is empty and there aren't any write / write concern errors, but, again, this would be pretty inconsistent with how theyd be thrown after anything succeeded.
- We can now just compare results as we originally had it and error if a result isn't present when the spec wants one.
- While looking through this code, I improved the description getter.

Sources/MongoSwift/MongoError.swift
```diff
@@ -260,33 +269,26 @@ internal func extractMongoError(error bsonError: bson_error_t, reply: Document?
...
-        if let writeErrors = serverReply["writeErrors"] as? [Document] {
+        if let writeErrors = reply["writeErrors"] as? [Document] {
             bulkWriteErrors = try writeErrors.map { try BSONDecoder().decode(BulkWriteError.self, from: $0) }
         }
 
         // Need to create new result that omits the ids that failed in insertedIds.
         var errResult: BulkWriteResult?
         if let result = partialResult {
-            let ordered = try bulkOp.opts?.getValue(for: "ordered") as? Bool ?? true
+            let ordered = try `for`.opts?.getValue(for: "ordered") as? Bool ?? true
```
kmahar: rather than doing this kinda funky thing to use a keyword as a variable name, could we just update the method signature so it says like `for op: BulkWriteOperation`?

Tests/MongoSwiftTests/SpecTestRunner/SpecTest.swift
kmahar: that all sounds good to me 👍

Tests/MongoSwiftTests/SpecTestRunner/SpecTest.swift
```diff
@@ -3,6 +3,101 @@ import Foundation
 import Nimble
 import XCTest
 
+/// Protocol that test cases which configure fail points during their execution conform to.
+internal protocol FailPointConfigured: class {
```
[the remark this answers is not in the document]
kmahar: seems to have broken Swift 4 😩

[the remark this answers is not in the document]
kmahar: i'm fine with just leaving it as class 👍

Tests/MongoSwiftTests/RetryableWritesTests.swift
```diff
@@ -0,0 +1,109 @@
...
+    /// This field has no effect for non-sharded topologies.
+    let useMultipleMongoses: Bool?
+
+    /// The optional fail point to configure before running this test.
+    /// This option and useMultipleMongoses: true are mutually exclusive.
+    let failPoint: FailPoint?
+}
+
+/// Struct representing a single retryable-writes spec test JSON file.
+private struct RetryableWritesTestFile: Decodable {
```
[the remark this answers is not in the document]
patrickfreed: The idea was to keep the framework minimal (at this point) so that crud v1 tests can also use it. I think I'd prefer to defer deciding on moving which parts into the framework once we implement v2 if that's okay. That work will probably involve modularizing the runner framework so that both v2 style and v1 style tests can use the parts they need without duplicating e.g. test file structs or unified test format based structs.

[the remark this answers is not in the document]
mbroadst: SGTM

PR #294, opened 2019-06-28
SWIFT-174 Implement DB enumeration spec

patrickfreed:
[SWIFT-174](https://jira.mongodb.org/browse/SWIFT-174)

This PR implements the full DB enumeration spec. As part of this change, there are now 3 methods on `MongoClient` in our public API for listing databases:
- `listDatabases -> [DatabaseSpecification]`
- `listDatabaseNames -> [String]`
- `listMongoDatabases -> [MongoDatabase]`

There were two different paths we could've chosen here:
- Implement _just_ `listDatabases` and provide an options struct for specifying `nameOnly`
- implement both `listDatabases` and `listDatabaseNames` and omit the options structs.

I opted to go with the latter for a couple reasons:
- Other drivers (e.g. Java, Go, even libmongoc) do
- It allows us to define the fields of `DatabaseSpecification` as non-optionals
- it seems more ergonomic. I imagine most users would be filtering it down to just names anyways

I also implemented the `listMongoDatabases` method for similar reasons, but I think the distinction is less important. 

Another thing worth mentioning: I moved filter out of the options struct to be a top level argument, since it felt more in line with our `find` wrapper. Other drivers do this as well.

I didn't include the `totalSize` part of the command result because the spec explicitly calls out including it as not super valuable. I kind of lean towards `listDatabases` returning a `ListDatabasesResult` that includes that within it, but I don't hold this opinion strongly enough to go against the spec though.

Anyways, these decisions were a little opinionated, so let me know what you guys think.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -61,17 +61,6 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     }
 }
 
-/// Options to use when listing available databases.
-public struct ListDatabasesOptions: Encodable {
-    /// Optionally indicate whether only names should be returned.
-    public var nameOnly: Bool?
-
-    /// Convenience constructor for basic construction
-    public init(nameOnly: Bool? = nil) {
-        self.nameOnly = nameOnly
-    }
-}
-
 /// Options to use when retrieving a `MongoDatabase` from a `MongoClient`.
 public struct DatabaseOptions: CodingStrategyProvider {
     /// A read concern to set on the retrieved database. If one is not specified, the database will inherit the
@@ -253,20 +242,11 @@ public class MongoClient {
      */
     public func listDatabases(_ filter: Document? = nil,
                               session: ClientSession? = nil) throws -> ListDatabasesResult {
-        var opts = try encodeOptions(options: Document(), session: session)
-        if let filter = filter {
-            if opts != nil {
-                opts?["filter"] = filter
-            } else {
-                opts = ["filter": filter]
-            }
+        let operation = ListDatabasesOperation(client: self, filter: filter, options: nil, session: session)
+        guard case let .full(result) = try operation.execute() else {
+            throw RuntimeError.internalError(message: "Invalid result")
         }
-
-        guard let cursor = mongoc_client_find_databases_with_opts(self._client, opts?._bson) else {
-            fatalError("Couldn't get cursor from the server")
-        }
-
-        return Array(try MongoCursor(from: cursor, client: self, decoder: self.decoder, session: session))
+        return result
     }
 
     /**
@@ -296,20 +276,14 @@ public class MongoClient {
      *   - `UserError.logicError` if the provided session is inactive.
      */
     public func listDatabaseNames(_ filter: Document? = nil, session: ClientSession? = nil) throws -> [String] {
-        var opts: Document = ["nameOnly": true]
-        if let filter = filter {
-            opts["filter"] = filter
+        let operation = ListDatabasesOperation(client: self,
+                                               filter: filter,
+                                               options: ListDatabasesOptions(nameOnly: true),
+                                               session: session)
+        guard case let .names(result) = try operation.execute() else {
+            throw RuntimeError.internalError(message: "Invalid result")
         }
-
-        guard let cursor = mongoc_client_find_databases_with_opts(self._client, opts._bson) else {
-            fatalError("Couldn't get cursor from the server")
-        }
-        return try MongoCursor<Document>(from: cursor, client: self, decoder: self.decoder, session: session).map {
-            guard let name = $0["name"] as? String else {
-                throw RuntimeError.internalError(message: "Server response didn't contain a database name: \($0)")
-            }
-            return name
-        }
+        return result
     }
 
     /**
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -247,7 +247,7 @@ public class MongoClient {
         let operation = ListDatabasesOperation(client: self,
                                                options: ListDatabasesOptions(filter: filter, nameOnly: nil),
                                                session: session)
-        guard case let .specs(result) = try operation.execute() else {
+        guard case let .specs(result) = try self.executeOperation(operation) else {
             throw RuntimeError.internalError(message: "Invalid result")
         }
         return result
@@ -283,7 +283,7 @@ public class MongoClient {
         let operation = ListDatabasesOperation(client: self,
                                                options: ListDatabasesOptions(filter: filter, nameOnly: true),
                                                session: session)
-        guard case let .names(result) = try operation.execute() else {
+        guard case let .names(result) = try self.executeOperation(operation) else {
             throw RuntimeError.internalError(message: "Invalid result")
         }
         return result
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -63,15 +63,11 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
 
 /// Options to use when listing available databases.
 public struct ListDatabasesOptions: Encodable {
-    /// An optional filter for the returned databases.
-    public var filter: Document?
-
     /// Optionally indicate whether only names should be returned.
     public var nameOnly: Bool?
 
     /// Convenience constructor for basic construction
-    public init(filter: Document? = nil, nameOnly: Bool? = nil) {
-        self.filter = filter
+    public init(nameOnly: Bool? = nil) {
         self.nameOnly = nameOnly
     }
 }
@@ -243,10 +239,10 @@ public class MongoClient {
     }
 
     /**
-     * Get a list of databases.
+     * Run the `listDatabases` command.
      *
      * - Parameters:
-     *   - options: Optional `ListDatabasesOptions` to use when executing the command
+     *   - filter: Optional `Document` specifying a filter that the listed databases must pass.
      *
      * - Returns: A `MongoCursor` over `Document`s describing the databases matching provided criteria
      *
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    internal func execute() throws -> ListDatabasesResults {
+        let cmd: Document = ["listDatabases": 1]
+        var opts = try self.client.encoder.encode(self.options)
+        if let filter = self.filter {
+            if opts == nil {
+                opts = Document()
+            }
+            opts = ["filter": filter] as Document
```
kmahar: will this overwrite `nameOnly` if it was in the `opts`?

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -7,8 +7,34 @@ import XCTest
...
+
+        let databases = [
+            "db1",
+            "empty",
+            "db3"
+        ]
+
+        databases.forEach {
+            try? client.db($0).drop()
+            _ = try? client.db($0).createCollection("c")
```
kmahar: will this line ever fail?

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -7,8 +7,34 @@ import XCTest
...
+        }
+        try client.db("db1").collection("c").insertOne(["a": 1])
+        try client.db("db3").collection("c").insertOne(["a": 1])
+
+        let dbInfo = try client.listDatabases()
+        expect(dbInfo.map { $0.name }).to(contain(databases))
```
kmahar: didn't realize you could pass an array into `contain`, nice 👌

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -0,0 +1,91 @@
+import mongoc
+
+/// Options to use when listing available databases.
+public struct ListDatabasesOptions: Encodable {
```
kmahar: since users can't actually pass this in anywhere, should it be internal?

kmahar: I think you could also keep `filter` in this struct then, and it would simplify your encoding logic in the operation below.

patrickfreed: Done and done

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
patrickfreed: Yep, good catch. was supposed to be a field assignment, not the whole doc. Due to including filter in the options struct, this all goes away.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -0,0 +1,91 @@
...
+                opts = Document()
+            }
+            opts = ["filter": filter] as Document
+        }
+        opts = try encodeOptions(options: opts, session: self.session)
+        var reply = Document()
+        var error = bson_error_t()
+
+        let success = withMutableBSONPointer(to: &reply) { replyPtr in
+            mongoc_client_read_command_with_opts(self.client._client,
```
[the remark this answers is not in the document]
patrickfreed: Very good catch. I've added a `.primary` read preference to the command, but I'm still getting errors when running directly against a secondary. I need to do some more investigation to figure out why that is.

Tests/MongoSwiftTests/MongoClientTests.swift
patrickfreed: it might. I originally ignored the errors in case of any weirdness with mongoses, but I guess it makes more sense to just run the commands and fail on any errors. it seems preferable to failing in one of the later assertions and making debugging harder.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -241,25 +226,65 @@ public class MongoClient {
     }
 
     /**
-     * Get a list of databases.
+     * Run the `listDatabases` command.
      *
      * - Parameters:
-     *   - options: Optional `ListDatabasesOptions` to use when executing the command
+     *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
```
patrickfreed: I added the clarification here since apparently the filter only applies to fields in the output, so `nameOnly` (which is set in the other two) restricts the filter to just "name" based filtration. Here, you can filter on anything.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    /// The amount of disk space consumed by this database.
+    public let sizeOnDisk: Int
+
+    /// Whether or not this database is empty.
+    public let empty: Bool
+
+    /// For sharded clusters, this field includes a document which maps each shard to the size in bytes of the database
+    /// on disk on that shard. For non sharded environments, this field is nil.
+    public let shards: Document?
```
patrickfreed: Accidentally omitted this field before. Because the structure of this document is based on the sharded topology, we can't use a `Codable` struct here in place of a `Document`.

kmahar:
no further comments 👍 lmk when you have an update on the secondary stuff

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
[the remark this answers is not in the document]
patrickfreed: It actually works fine against direct connections to secondaries; I was just testing it incorrectly 🙃

Sources/MongoSwift/MongoClient.swift
```diff
@@ -241,25 +226,65 @@ public class MongoClient {
...
      * - Returns: A `MongoCursor` over `Document`s describing the databases matching provided criteria
      *
      * - Throws:
-     *   - `UserError.invalidArgumentError` if the options passed are an invalid combination.
      *   - `UserError.logicError` if the provided session is inactive.
      *   - `EncodingError` if an error is encountered while encoding the options to BSON.
+     *
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/command/listDatabases/
+     */
+    public func listDatabases(_ filter: Document? = nil,
```
mbroadst: I think its worth noting/discussing what we're gaining and giving up with this approach. Not using the options struct might feel more ergonomic, however it does move us towards a few potential risks:
  - It makes extension of these methods more difficult. What if I wanted to specify a client-side timeout (`maxTimeMS`) for this operation? Should it be added as an additional parameter?
  - Lets say we just add it as a parameter, what do we do with the old method? Mark it deprecated, and delete it sometime in the future in a major version release?
  - It definitely expands the API surface we need to maintain

Its a bit weird to bring this up in a PR _about_ adding a helper, but one of the drivers mantras is to [Minimize Administrative Helpers](https://github.com/mongodb/specifications#minimize-administrative-helpers). It could be argued that adding three methods, when one would suffice, goes against this principle. 

Anyway, I don't feel _very_ strongly about this, but I think its worth mentioning.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -11,8 +11,8 @@ final class CodecTests: MongoSwiftTestCase {
             DecodingError.Context(codingPath: [], debugDescription: "dummy error"))
 
     func testEncodeListDatabasesOptions() throws {
-        let options = ListDatabasesOptions(filter: ["a": 10], nameOnly: true)
-        let expected: Document = ["filter": ["a": 10] as Document, "nameOnly": true]
+        let options = ListDatabasesOptions(filter: ["x": 1], nameOnly: true)
```
mbroadst: morbid curiosity: what's going on with this change?

Sources/MongoSwift/MongoClient.swift
kmahar: > Lets say we just add it as a parameter, what do we do with the old method?
I don't think we'd need a new method, it would just be an additional optional parameter that defaults to `nil` and all existing calls to it would keep working 

that said I feel like that kind of stuff goes in an options struct, so we would end up needing a public options struct again 🤔

kmahar: FWIW note that `listCollections` also sticks `filter` into an options struct so if we don't have a ticket about making that line up with this stuff we should make one

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    /// Convenience constructor for basic construction
+    internal init(filter: Document?, nameOnly: Bool?) {
+        self.filter = filter
+        self.nameOnly = nameOnly
+    }
+}
+
+/// A struct modeling the information returned from the `listDatabases` command about a single database.
+public struct DatabaseSpecification: Codable {
```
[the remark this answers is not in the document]
patrickfreed: I guess it does give us issues with forward compatibility, but in general we'd want to be encouraging our users to be using `Codable` over `Document` wherever possible, including in their own structs and what not. Providing APIs that return `Document` sort of goes against this principle. I personally would like to avoid a situation where we supply two ways to do something and prefer if users only use one of them; I'd rather just omit the other case.

Also, I think it's worth noting that the spec specifically calls out using a model as an acceptable return value and does not give the warning "format subject to change" that I have seen in other specs.

Finally, having two overloads only differ in return value makes type inference a little annoying for users.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        guard success else {
+            throw extractMongoError(error: error, reply: reply)
+        }
+
+        guard let databases = reply["databases"] as? [Document] else {
+            throw RuntimeError.internalError(message: "Invalid server response: \(reply)")
+        }
+
+        if self.options?.nameOnly ?? false {
```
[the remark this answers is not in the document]
patrickfreed: So yeah, this is a little weird but the rationale is as follows:
- `listDatabases` and `listDatabase` with `nameOnly` have significantly different performance characteristics due to db level locking in the former
- For that reason, we should probably provide two different APIs to communicate this (e.g. include `listDatabaseNames`)
    - Other drivers (e.g. go, libmongoc) provide separate helpers, so we get cross-driver consistency which is always a plus imo
- If we're going to provide a `listDatabaseNames`, then there is no reason a user should ever use the full `listDatabases` and provide a `nameOnly` option, so let's not allow them to.

Not allowing `nameOnly` also has the added benefit of allowing all fields on `DatabaseSpecification` to be non-optional.

Tests/MongoSwiftTests/CodecTests.swift
patrickfreed: So in a previous commit I removed filter from the options, then in a later one I added it back and needed to fix this test. Instead of reverting it I just threw a random document in there. I'll fix this so this diff goes away.

kmahar:
BTW this isn't showing conflicts but make sure to rebase as the `Operation` protocol has changed now

Sources/MongoSwift/MongoClient.swift
mbroadst: We've debated this thoroughly in person :) I'm going to resolve it now.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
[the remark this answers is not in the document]
mbroadst: @patrickfreed we discussed this only briefly in person, but remember that the `nameOnly` is currently being serialized into the `with_opts` options, but should instead be in the direct command document above. Did you decide against that, or did it just get lost in the mix?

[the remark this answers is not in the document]
patrickfreed: Yep, totally forgot about that, good point. I've actually just gone ahead and removed the options struct since it isn't part of the public API anyways and the values in it aren't even being sent as options.

PR #295, opened 2019-07-03
SWIFT-185 Add support for maxTimeMS option for createIndex and dropIndex command

ncarbon:


Sources/MongoSwift/Operations/CreateIndexesOperation.swift
```diff
@@ -5,9 +5,13 @@ public struct CreateIndexOptions: Encodable {
     /// An optional `WriteConcern` to use for the command.
     public var writeConcern: WriteConcern?
 
+    /// The maximum amount of time to allow the query to run.
```
mbroadst: nit: let's specify that the timeout is enforced server-side

Tests/MongoSwiftTests/MongoCollection+IndexTests.swift
```diff
@@ -208,6 +208,52 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
...
+        let indexes = try coll.listIndexes()
+        expect(indexes).toNot(beNil())
+        expect(indexes.next()?["name"]).to(bsonEqual("_id_"))
+        expect(indexes.next()).to(beNil())
+
+        // test that maxTimeMS is an accepted option for createIndex and dropIndex
+        expect(receivedEvents.count).to(equal(2))
+        expect(receivedEvents[0].command["createIndexes"]).toNot(beNil())
+        expect(receivedEvents[0].command["maxTimeMS"]).toNot(beNil())
+        expect(100).to(bsonEqual(receivedEvents[0].command["maxTimeMS"]))
```
kmahar: nit: flip `100` and the `receivedEvents...` variables here. 
when we do `expect(x).to(bsonEqual(y))` and the assertion fails, the error message we get is "expected to equal y, got [whatever we got]". so in this case we would like the expected value to be 100.

Tests/MongoSwiftTests/MongoCollection+IndexTests.swift
```diff
@@ -208,6 +208,52 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
...
+        expect(receivedEvents[1].command["dropIndexes"]).toNot(beNil())
+        expect(receivedEvents[1].command["maxTimeMS"]).toNot(beNil())
+        expect(100).to(bsonEqual(receivedEvents[1].command["maxTimeMS"]))
```
kmahar: (same nit about ordering)

PR #296, opened 2019-07-03
SWIFT-447 Add more comprehensive ReadPreference tests

ncarbon:


Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -89,4 +89,99 @@ final class ReadPreferenceTests: MongoSwiftTestCase {
...
+    }
+
+    func testClientReadPreference() throws {
+        let primaryPreferred = ReadPreference(.primaryPreferred)
+        let secondary = ReadPreference(.secondary)
+
+        do {
+            // expect that a client with an unset read preference has it default to primary
+            let client = try MongoClient()
+            expect(client.readPreference.mode).to(equal(.primary))
```
kmahar: I think it would be better to directly compare client.readPreference to a `ReadPreference(.primary)`. and similar for the other comparisons throughout this test

I know the ReadConcern and WriteConcern tests currently for some reason just read particular properties but I don't really know why as we made the types equatable. so I am fixing that in one of my pooling PRs where i added more tests for ReadConcern and WriteConcern

patrickfreed:
I think we could use SDAM monitoring to get an idea of the topology from which we can construct a map of connection id => server type. Then we'd use command monitoring to get the connection id of a specific command and verify its server type matches what we were trying to select with the read preference.

kmahar:
i'm indifferent about whether it's worth testing that, defer to you and @mbroadst. we could also just make a follow up ticket about it

patrickfreed:
Either of those outcomes is fine with me. I'll also defer to @mbroadst

mbroadst:
This looks great to me, I don't think we need to verify using monitoring

PR #297, opened 2019-07-09
SWIFT-374 MongoDatabase and MongoCollection no longer wrap libmongoc types

kmahar:
Apologies for the size of this, it got bigger than expected as I ended up doing a lot of refactoring in the read concern and write concern tests.

Changes:
- Added `MongoNamespace` type which DBs and collections use to track their namespaces.
- MongoDatabase and MongoCollection no longer wrap their corresponding mongoc types. They now have helper methods to create the mongoc types as needed.
- All operations now use the connections passed into `execute` to call those helpers and get the temp pointers.
- Methods returning `MongoCursor`s now check out `Connection`s and use them to call mongoc methods. The connection is then handed off to the `MongoCursor`, which will return it to the pool upon deinitialization.
    - As part of the work to store a connection in the cursor, I ended up getting rid of the manual `close` method on `MongoCursor`. I considered switching to the approach we discussed before with an enum wrapping the values we need to nil out but ultimately this seemed simpler.

Once the new read preference tests in #296 get merged in I will augment those tests with similar checks for the read preference set on the temporary pointers.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -63,7 +63,7 @@ extension MongoCollection {
             guard let cursor = mongoc_collection_aggregate(collPtr,
                                                            MONGOC_QUERY_NONE,
                                                            pipeline._bson,
-                                                           opts?._bson, 
+                                                           opts?._bson,
                                                            rp) else {
                 fatalError("Couldn't get cursor from the server")
             }
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -25,7 +25,7 @@ extension MongoCollection {
         let conn = try self._client.connectionPool.checkOut()
         let cursor: OpaquePointer = self.withMongocCollection(from: conn) { collPtr in
             guard let cursor = mongoc_collection_find_with_opts(collPtr, filter._bson, opts?._bson, rp) else {
-                fatalError("Couldn't get cursor from the server")
+                fatalError(failedToRetrieveCursorMessage)
             }
             return cursor
         }
@@ -65,7 +65,7 @@ extension MongoCollection {
                                                            pipeline._bson,
                                                            opts?._bson,
                                                            rp) else {
-                fatalError("Couldn't get cursor from the server")
+                fatalError(failedToRetrieveCursorMessage)
             }
             return cursor
         }
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -22,10 +22,19 @@ extension MongoCollection {
         let opts = try encodeOptions(options: options, session: session)
         let rp = options?.readPreference?._readPreference
 
-        guard let cursor = mongoc_collection_find_with_opts(self._collection, filter._bson, opts?._bson, rp) else {
-            fatalError("Couldn't get cursor from the server")
+        let conn = try self._client.connectionPool.checkOut()
+        let cursor: OpaquePointer = self.withMongocCollection(from: conn) { collPtr in
+            guard let cursor = mongoc_collection_find_with_opts(collPtr, filter._bson, opts?._bson, rp) else {
+                fatalError("Couldn't get cursor from the server")
+            }
+            return cursor
         }
-        return try MongoCursor(from: cursor, client: self._client, decoder: self.decoder, session: session)
+
+        return try MongoCursor(from: cursor,
+                               client: self._client,
+                               connection: conn,
+                               decoder: self.decoder,
+                               session: session)
     }
 
     /**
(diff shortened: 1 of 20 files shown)


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -20,21 +23,21 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         self._session = session
         self.decoder = decoder
+        self.swiftError = nil
 
         if let session = session, !session.active {
             throw ClientSession.SessionInactiveError
         }
 
         if let err = self.error {
-            // Need to explicitly close since deinit will not execute if we throw.
```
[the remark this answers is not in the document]
patrickfreed: whoops 😛

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -70,213 +71,330 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
                 .to(throwError(UserError.invalidArgumentError(message: "")))
     }
 
+    /// Checks that a `MongoClient` as well as the underlying mongoc clients have the expected read concern.
+    func checkClientReadConcern(_ client: MongoClient,
```
patrickfreed: These could probably be condensed into a single `checkConcern` that takes in the client/db/collection read/write concern as well as a closure to extract the temporary one.

e.g.
```swift 
func checkConcern<T: Equatable>(topLevel: T, expected: T, derived: () -> T, desc: String) -> Bool
```

Then for client e.g.
```swift
let clientRCDerivation = { client in
    return { 
        try client.connectionPool.withConnection { conn in
            let connRC = ReadConcern(from: mongoc_client_get_read_concern(conn.clientHandle))
            expect(connRC).to(equal(expected), description: description)
     }
}

checkConcern(client.readConcern, empty, clientRCDerivation(client), "client created with no RC provided") 
```
Totally up to you whether you find this approach preferable. I know in general I have no aversion to code duplication in tests, especially if it makes the test clearer / simpler, so this might not be worth doing. It just seemed like a fun/functional way to do it 😄

kmahar: I ended up condensing things a bit using a somewhat similar approach. the main difference is rather than using closures, the methods for getting RCs/WCs for pointers are protocol methods. this saves me having to pass in a closure every time I call the compare method.

patrickfreed: Nice, much more swifty.

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -49,11 +58,23 @@ extension MongoCollection {
...
-            self._collection, MONGOC_QUERY_NONE, pipeline._bson, opts?._bson, rp) else {
-            fatalError("Couldn't get cursor from the server")
+        let conn = try self._client.connectionPool.checkOut()
+        let cursor: OpaquePointer = self.withMongocCollection(from: conn) { collPtr in
+            guard let cursor = mongoc_collection_aggregate(collPtr,
+                                                           MONGOC_QUERY_NONE,
+                                                           pipeline._bson,
+                                                           opts?._bson,
+                                                           rp) else {
+                fatalError("Couldn't get cursor from the server")
```
mbroadst: starting to feel like this message should be a constant

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -39,62 +42,46 @@ public class MongoCollection<T: Codable> {
...
+        return self.namespace.collection! // swiftlint:disable:this force_unwrapping
     }
 
     /// The `ReadConcern` set on this collection, or `nil` if one is not set.
-    public var readConcern: ReadConcern? {
-        // per libmongoc docs, we don't need to handle freeing this ourselves
-        let rc = ReadConcern(from: mongoc_collection_get_read_concern(self._collection))
-        return rc.isDefault ? nil : rc
-    }
+    public let readConcern: ReadConcern?
```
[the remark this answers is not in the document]
mbroadst: good idea

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -48,18 +51,8 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
-    /// Closes the cursor.
-    public func close() {
-        guard let cursor = self._cursor else {
-            return
-        }
-        mongoc_cursor_destroy(cursor)
-        self._cursor = nil
-        self._client = nil
-        self._session = nil
+        self._client.connectionPool.checkIn(self._connection)
```
mbroadst: something worth considering in the future: we should return the connection as soon as the cursor is exhausted. Right now I could fully iterate a cursor, print out its contents, but accidentally keep some reference to it, and run the risk of exhausting our connection pool. At very least we should thoroughly document the interplay of connections and cursors.

patrickfreed: I agree we should probably just be returning the connections upon exhausting the cursor. It will require us to keep track of whether the cursor is exhausted internally, which is sorta equivalent to the cursor being closed early (which we were trying to avoid with these changes unfortunately)

kmahar: I agree , I've opened [SWIFT-517](https://jira.mongodb.org/browse/SWIFT-517) about doing this. 
there's some trickiness we need to handle with checking for errors after exhaustion, and also with knowing when the cursor is actually exhausted, since we don't currently track cursor type in `MongoCursor`. so I think we should do this shortly but I'm going to defer it to another ticket.

Sources/MongoSwift/MongoCollection+Read.swift
kmahar: added a constant `failedToRetrieveCursorMessage`

PR #298, opened 2019-07-11
SWIFT-176 Add ChangeStream, ChangeStreamDocument, and ChangeStreamOptions structs.

ncarbon:
This is the first PR for [SPEC-176/ChangeStreams API](https://jira.mongodb.org/browse/SWIFT-176).
It implements:
* ChangeStream
* ChangeStreamOptions
* ChangeStreamDocument

I added some tests just for what's been added so far by calling `mongoc_collection_watch`, but these will likely change once `MongoClient.watch`, `MongoDatabase.watch`, and `MongoDatabase.watch` are implemented and just use those instead?

### DIFF
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -17,25 +17,25 @@ public struct ChangeStreamToken: Codable {
 /// A MongoDB ChangeStream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
 public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
-    // A `ChangeStreamToken` used for manually resuming a change stream.
+    /// A `ChangeStreamToken` used for manually resuming a change stream.
     public private(set) var resumeToken: ChangeStreamToken
 
-    // A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
+    /// A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
     private let client: MongoClient
 
-    // A `Connection` stored to make sure the client connection stays valid until the change stream is destroyed.
+    /// A `Connection` stored to make sure the client connection stays valid until the change stream is destroyed.
     private let connection: Connection
 
-    // A `ClientSession` stored to make sure the session stays valid until the change stream is destroyed.
+    /// A `ClientSession` stored to make sure the session stays valid until the change stream is destroyed.
     private let session: ClientSession?
 
-    // A reference to the `mongoc_change_stream_t` pointer.
+    /// A reference to the `mongoc_change_stream_t` pointer.
     private let changeStream: OpaquePointer
 
-    // Decoder for decoding documents into type `T`.
+    /// Decoder for decoding documents into type `T`.
     private let decoder: BSONDecoder
 
-    // Used for storing Swift errors.
+    /// Used for storing Swift errors.
     private var swiftError: Error?
 
     /// The error that occurred while iterating the change stream, if one exists. This should be used to check
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -14,6 +14,12 @@ public struct ChangeStreamToken: Codable {
     }
 }
 
+extension ChangeStreamToken: Equatable {
+    public static func == (lhs: ChangeStreamToken, rhs: ChangeStreamToken) -> Bool {
+        return lhs.resumeToken == rhs.resumeToken
+    }
+}
+
 /// A MongoDB ChangeStream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
 public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -137,19 +137,15 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
                 client: MongoClient,
                 session: ClientSession? = nil,
                 decoder: BSONDecoder) throws {
-      self.changeStream = changeStream
-      self.session = session
-      self.client = client
-      self.decoder = decoder
-      self.resumeToken = ChangeStreamToken(resumeToken: [])
+    self.changeStream = changeStream
+    self.session = session
+    self.client = client
+    self.decoder = decoder
+    self.resumeToken = ChangeStreamToken(resumeToken: [])
 
     if let err = self.error {
       throw err
     }
-  }
-
-  func foobar() {
-    fgdf
   }
 
   /// Cleans up internal state.
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -2,7 +2,7 @@ import mongoc
 /// A token used for manually resuming a change stream. Pass this to the `resumeAfter` field of
 /// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
-public struct ChangeStreamToken: Codable {
+public struct ChangeStreamToken: Codable, Equatable {
     private let resumeToken: Document
 
     internal init(resumeToken: Document) {
@@ -11,12 +11,6 @@ public struct ChangeStreamToken: Codable {
 
     public init(from decoder: Decoder) throws {
         self.resumeToken = try Document(from: decoder)
-    }
-}
-
-extension ChangeStreamToken: Equatable {
-    public static func == (lhs: ChangeStreamToken, rhs: ChangeStreamToken) -> Bool {
-        return lhs.resumeToken == rhs.resumeToken
     }
 }
 
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -93,8 +93,6 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
     // we have to copy because libmongoc owns the pointer.
     let doc = Document(copying: pointee)
 
-    print(doc)
-
     // Update the resumeToken with the `_id` field from the document.
     if let resumeToken = doc["_id"] as? Document {
         self.resumeToken = ChangeStreamToken(resumeToken: resumeToken)
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -17,22 +17,25 @@ public struct ChangeStreamToken: Codable {
 /// A MongoDB ChangeStream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
 public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
-    /// A `ChangeStreamToken` used for manually resuming a change stream.
+    // A `ChangeStreamToken` used for manually resuming a change stream.
     public private(set) var resumeToken: ChangeStreamToken
 
-    /// A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
+    // A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
     private let client: MongoClient
 
-    /// A `ClientSession` stored to make sure the session stays valid until the change stream is destroyed.
+    // A `Connection` stored to make sure the client connection stays valid until the change stream is destroyed.
+    private let connection: Connection
+
+    // A `ClientSession` stored to make sure the session stays valid until the change stream is destroyed.
     private let session: ClientSession?
 
-    /// A reference to the `mongoc_change_stream_t` pointer.
+    // A reference to the `mongoc_change_stream_t` pointer.
     private let changeStream: OpaquePointer
 
-    /// Decoder for decoding documents into type `T`.
+    // Decoder for decoding documents into type `T`.
     private let decoder: BSONDecoder
 
-    /// Used for storing Swift errors.
+    // Used for storing Swift errors.
     private var swiftError: Error?
 
     /// The error that occurred while iterating the change stream, if one exists. This should be used to check
@@ -135,21 +138,24 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
      */
     internal init(stealing changeStream: OpaquePointer,
                   client: MongoClient,
+                  connection: Connection,
                   session: ClientSession? = nil,
                   decoder: BSONDecoder) throws {
-        self.changeStream = changeStream
-        self.session = session
-        self.client = client
-        self.decoder = decoder
         self.resumeToken = ChangeStreamToken(resumeToken: [])
+        self.client = client
+        self.connection = connection
+        self.session = session
+        self.changeStream = changeStream
+        self.decoder = decoder
 
         if let err = self.error {
             throw err
         }
     }
 
-    /// Cleans up internal state.
-    deinit {
-        mongoc_change_stream_destroy(self.changeStream)
-    }
+     /// Cleans up internal state.
+     deinit {
+         mongoc_change_stream_destroy(self.changeStream)
+         self.client.connectionPool.checkIn(self.connection)
+     }
 }
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -1,6 +1,8 @@
 import mongoc
 
-/// A `ChangeStreamToken` used for wrapping the resume token.
+/// A token used for manually resuming a change stream. Pass this to the `resumeAfter` or `startAfter` fields of
+/// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
 public struct ChangeStreamToken: Codable {
     private let resumeToken: Document
 
@@ -16,7 +18,7 @@ public struct ChangeStreamToken: Codable {
 /// A MongoDB ChangeStream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
 public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
-  /// The cached resume token.
+  /// A `ChangeStreamToken` used to manually resume a change stream.
   public private(set) var resumeToken: ChangeStreamToken
 
   /// A `MongoClient` stored to make sure the source client stays valid
@@ -33,9 +35,8 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
   /// Used for storing Swift errors.
   private var swiftError: Error?
 
-  /// The error that occurred while iterating the ChangeStream cursor,
-  /// if one exists. This should be used to check for errors after
-  /// `next()` returns `nil`.
+  /// The error that occurred while iterating the ChangeStream cursor, if one exists.
+  /// This should be used to check for errors after `next()` returns `nil`.
   public var error: Error? {
     if let err = self.swiftError {
         return err
@@ -68,26 +69,25 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
   /// Decoder for decoding documents into type `T`.
   internal let decoder: BSONDecoder
 
-  /// Returns the next `T` in the change stream or nil if there is no next value.
-  /// Will block for a maximum of `maxAwaitTimeMS` milliseconds as specified in the
-  /// `ChangeStreamOptions`, or for the server default timeout if omitted.
+  /// Returns the next `T` in the change stream or nil if there is no next value. Will block for a maximum of
+  /// `maxAwaitTimeMS` milliseconds as specified in the `ChangeStreamOptions`, or for the server default timeout
+  /// if omitted.
   public func next() -> T? {
     let cursor = self.changeStream
     // Allocate space for a reference to a BSON pointer.
     let out = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
+    defer {
+        out.deinitialize(count: 1)
+        out.deallocate()
+    }
 
     // Check that there’s another document in the stream.
     guard mongoc_change_stream_next(cursor, out) else {
       return nil
     }
 
-    defer {
-        out.deinitialize(count: 1)
-        out.deallocate()
-    }
-
     guard let pointee = out.pointee else {
-        fatalError("mongoc_change_stream_next returned true, but document is nil")
+        fatalError("Could not advance the change stream.")
     }
 
     // we have to copy because libmongoc owns the pointer.
@@ -102,30 +102,25 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
     }
 
     do {
-      return try decoder.decode(T.self, from: doc)
+        return try decoder.decode(T.self, from: doc)
     } catch {
-      self.swiftError = error
+        self.swiftError = error
     }
 
     return nil
   }
 
   /**
-   * Returns the next `T` in this change stream or
-   * `nil`, or throws an error if one occurs -- compared to `next()`,
-   * which returns `nil` and requires manually checking for an error
-   * afterward.
-   * Will block for a maximum of `maxAwaitTimeMS` milliseconds as specified in the
-   * `ChangeStreamOptions`, or for the server default timeout if omitted.
-   * - Returns: the next `T` in this change stream,
-   *            or `nil` if at the end of the change stream cursor.
+   * Returns the next `T` in this change stream or `nil`, or throws an error if one occurs -- compared to `next()`,
+   * whichreturns `nil` and requires manually checking for an error afterward. Will block for a maximum of
+   * `maxAwaitTimeMS` milliseconds as specified in the `ChangeStreamOptions`, or for the server default timeout if
+   * omitted.
+   * - Returns: the next `T` in this change stream, or `nil` if at the end of the change stream cursor.
    * - Throws:
-   *   - `ServerError.commandError` if an error occurs on the
-   *     server while iterating the change stream cursor.
-   *   - `UserError.logicError` if this function is called and the
-   *     session associated with this change stream is inactive.
-   *   - `DecodingError` if an error occurs decoding the server's
-   *     response.
+   *   - `ServerError.commandError` if an error occurs on the server while iterating the change stream cursor.
+   *   - `UserError.logicError` if this function is called and the session associated with this change stream is
+   *     inactive.
+   *   - `DecodingError` if an error occurs decoding the server's response.
    */
   public func nextOrError() throws -> T? {
     if let next = self.next() {
@@ -141,10 +136,8 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
   /**
    * Initializes a `ChangeStream`.
    * - Throws:
-   *   - `ServerError.commandError` if an error occurred on the
-   *     server when creating the `mongoc_change_stream_t`.
-   *   - `UserError.invalidArgumentError` if the `mongoc_change_stream_t`
-   *     was created with invalid options.
+   *   - `ServerError.commandError` if an error occurred on the server when creating the `mongoc_change_stream_t`.
+   *   - `UserError.invalidArgumentError` if the `mongoc_change_stream_t` was created with invalid options.
    */
   internal init(stealing changeStream: OpaquePointer,
                 client: MongoClient,
@@ -163,12 +156,12 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
 
   /// Cleans up internal state.
   deinit {
-     guard let cursor = self.changeStream else {
-          return
-      }
-      mongoc_change_stream_destroy(changeStream)
-      self.changeStream = nil
-      self.session = nil
-      self.client = nil
+    guard let cursor = self.changeStream else {
+        return
+    }
+    mongoc_change_stream_destroy(changeStream)
+    self.changeStream = nil
+    self.session = nil
+    self.client = nil
   }
 }
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -110,7 +110,7 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
 
   /**
    * Returns the next `T` in this change stream or `nil`, or throws an error if one occurs -- compared to `next()`,
-   * whichreturns `nil` and requires manually checking for an error afterward. Will block for a maximum of
+   * which returns `nil` and requires manually checking for an error afterward. Will block for a maximum of
    * `maxAwaitTimeMS` milliseconds as specified in the `ChangeStreamOptions`, or for the server default timeout if
    * omitted.
    * - Returns: the next `T` in this change stream, or `nil` if at the end of the change stream cursor.
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -92,7 +92,6 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
 
     // we have to copy because libmongoc owns the pointer.
     let doc = Document(copying: pointee)
-    print(doc)
 
     // Update the resumeToken with the `_id` field from the document.
     if let resumeToken = doc["_id"] as? Document {
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -93,6 +93,8 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
     // we have to copy because libmongoc owns the pointer.
     let doc = Document(copying: pointee)
 
+    print(doc)
+
     // Update the resumeToken with the `_id` field from the document.
     if let resumeToken = doc["_id"] as? Document {
         self.resumeToken = ChangeStreamToken(resumeToken: resumeToken)
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -21,6 +21,7 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
     public private(set) var resumeToken: ChangeStreamToken
 
     /// A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
+    // TODO: SWIFT-374: remove this property.
     private let client: MongoClient
 
     /// A `Connection` stored to make sure the client connection stays valid until the change stream is destroyed.
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -1,6 +1,6 @@
 import mongoc
 
-/// A `ChangeStreamToken` that wraps `resumeToken`.
+/// A `ChangeStreamToken` used for wrapping the resume token.
 public struct ChangeStreamToken: Codable {
     private let resumeToken: Document
 
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -70,7 +70,7 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
 
   /// Returns the next `T` in the change stream or nil if there is no next value.
   /// Will block for a maximum of `maxAwaitTimeMS` milliseconds as specified in the
-  /// `ChangeStreamOptions`, or the server default timeout if omitted.
+  /// `ChangeStreamOptions`, or for the server default timeout if omitted.
   public func next() -> T? {
     let cursor = self.changeStream
     // Allocate space for a reference to a BSON pointer.
@@ -116,16 +116,16 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
    * which returns `nil` and requires manually checking for an error
    * afterward.
    * Will block for a maximum of `maxAwaitTimeMS` milliseconds as specified in the
-   * `ChangeStreamOptions`, or the server default timeout if omitted.
+   * `ChangeStreamOptions`, or for the server default timeout if omitted.
    * - Returns: the next `T` in this change stream,
    *            or `nil` if at the end of the change stream cursor.
    * - Throws:
    *   - `ServerError.commandError` if an error occurs on the
-   * 	   server while iterating the change stream cursor.
+   *     server while iterating the change stream cursor.
    *   - `UserError.logicError` if this function is called and the
    *     session associated with this change stream is inactive.
    *   - `DecodingError` if an error occurs decoding the server's
-   * 	   response.
+   *     response.
    */
   public func nextOrError() throws -> T? {
     if let next = self.next() {
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -78,7 +78,6 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
         out.deallocate()
     }
 
-    // Check that there’s another document in the stream.
     guard mongoc_change_stream_next(self.changeStream, out) else {
       return nil
     }
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -1,155 +1,155 @@
 import mongoc
-/// A token used for manually resuming a change stream. Pass this to the `resumeAfter` or `startAfter` fields of
+/// A token used for manually resuming a change stream. Pass this to the `resumeAfter` field of
 /// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
 public struct ChangeStreamToken: Codable {
     private let resumeToken: Document
 
     internal init(resumeToken: Document) {
-      self.resumeToken = resumeToken
+        self.resumeToken = resumeToken
     }
 
     public init(from decoder: Decoder) throws {
-      self.resumeToken = try Document(from: decoder)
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
+import mongoc
+
+/// A `ChangeStreamToken` used for wrapping the resume token.
```
patrickfreed: This docstring could be a little more descriptive. 

e.g. something like:
 "A token used for manually resuming a change stream. Pass this to the `resumeAfter` or `startAfter` fields of `ChangeStreamOptions` to resume or start a change stream where a previous one left off.

See \[the official documentation\](https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream) for more information on how to use this."

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    public init(from decoder: Decoder) throws {
+      self.resumeToken = try Document(from: decoder)
+    }
+}
+
+/// A MongoDB ChangeStream.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
+public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
+  /// The cached resume token.
```
patrickfreed: Likewise some of the documentation added to the definition of `ChangeStreamToken` can be reworded and used here to better describe what this is / what it's used for.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+    let cursor = self.changeStream
+    // Allocate space for a reference to a BSON pointer.
+    let out = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
+
+    // Check that there’s another document in the stream.
+    guard mongoc_change_stream_next(cursor, out) else {
+      return nil
+    }
+
+    defer {
```
patrickfreed: Let's move this defer block to immediately after the allocation. Otherwise, if the above guard fails we'll leak memory.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+        out.deinitialize(count: 1)
+        out.deallocate()
+    }
+
+    guard let pointee = out.pointee else {
+        fatalError("mongoc_change_stream_next returned true, but document is nil")
```
patrickfreed: This message can be rephrased to not mention libmongoc and be more generic. e.g. "Could not advance change stream"

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    // Update the resumeToken with the `_id` field from the document.
+    if let resumeToken = doc["_id"] as? Document {
+        self.resumeToken = ChangeStreamToken(resumeToken: resumeToken)
+    } else {
+        self.swiftError = RuntimeError.internalError(message: "_id field is missing from the change stream document.")
+        return nil
+    }
+
+    do {
```
patrickfreed: I think the indentation is off in this block

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+      return try decoder.decode(T.self, from: doc)
+    } catch {
+      self.swiftError = error
+    }
+
+    return nil
+  }
+
+  /**
+   * Returns the next `T` in this change stream or
```
patrickfreed: These docstrings can be expanded to the full width that our formatter accepts (I think its 100? maybe 120? characters)

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    if let error = self.error {
+      throw error
+    }
+    return nil
+  }
+
+  /**
+   * Initializes a `ChangeStream`.
```
patrickfreed: ditto expansion

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,174 @@
...
+                client: MongoClient,
+                session: ClientSession? = nil,
+                decoder: BSONDecoder) throws {
+    self.changeStream = changeStream
+    self.session = session
+    self.client = client
+    self.decoder = decoder
+    self.resumeToken = ChangeStreamToken(resumeToken: [])
+
+    if let err = self.error {
```
patrickfreed: ditto indentation

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,67 @@
...
+   /// that were changed, and the new value for those fields.
+   public let updatedFields: Document
+
+   /// An array of field names that were removed from the document.
+   public let removedFields: [String]
+}
+
+/// An enum representing the type of operation for this change.
+public enum OperationType: String, Codable {
+   case insert
```
patrickfreed: The linter will probably complain that these need to be documented. The docstrings can be short and basically just link to official docs (if relevant ones exist)

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// The response document type from a `ChangeStream`.
+public struct ChangeStreamDocument<T: Codable>: Codable {
+    /// Describes the type of operation for this change.
+    public let operationType: OperationType
+
+    /// An opaque token for use when resuming an interrupted change
+    /// stream.
+    public let _id: ChangeStreamToken
```
patrickfreed: could use the link I mentioned earlier about how to use these.

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    /// A document containing the database and collection names in
+    /// which this change happened.
+    public let ns: MongoNamespace
+
+    /** Only present for options of type ‘insert’, ‘update’,
```
patrickfreed: This line should be moved one down so that the `/**` is on its own line.

Also, most of the documentation in this file can be expanded to the full column limit.

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,67 @@
...
+    public let documentKey: Document?
+
+    /// An `UpdateDescription` containing updated and removed fields in
+    /// this operation. Only present for operations of type `update`.
+    public let updateDescription: UpdateDescription?
+
+   /**
+    * Always present for operations of type ‘insert’ and ‘replace’. Also
+    * present for operations of type ‘update’ if the user has specified
+    * ‘updateLookup’ in the ‘fullDocument’ arguments to the ‘$changeStream’
```
patrickfreed: instead of talking about passing stuff to the `$changeStream` stage, let's rephrase this in terms of our structures. (e.g. if the user specified `FullDocument.updateLookup` in the `ChangeStreamOptions` used to create the `ChangeStream` that emitted this document.)

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,101 @@
+/// Describes the modes for configuring the fullDocument field of a
+/// `ChangeStreamDocument`.
+public enum FullDocument: RawRepresentable, Codable {
+    /// The change stream document will include both a delta describing the
```
patrickfreed: I think I mentioned this in the design doc, but this comment should be updated to only describe what this option is doing to the updateLookup field and can omit the parts about the `UpdateDescription` field.

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,101 @@
...
+            self = .updateLookup
+        default:
+            self = .other(rawValue)
+        }
+    }
+}
+
+/// Options to use when creating a `ChangeStream`.
+public struct ChangeStreamOptions: Codable {
+    /** Indicates how the `fullDocument` field of a `ChangeStreamDocument` should
```
patrickfreed: ditto line break and expansion

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,101 @@
...
+     * be filled out by the server.
+     * By default (indicated by a nil value for this option), the fullDocument field
+     * in the change stream document will always be present in the case of 'insert'
+     * and 'replace' operations (containing the document being inserted) and will be
+     * nil for all other operations.
+     */
+    public let fullDocument: FullDocument?
+
+    /// Specifies the logical starting point for the new change stream.
+    public let resumeAfter: Document?
```
patrickfreed: the type here should be a `ChangeStreamToken?`. Also ditto the documentation stuff I said elsewhere about `ChangeStreamToken`s.

Since this type is by nature opaque (and complicated), we need to provide a lot of docs so users can figure out how to use it.

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    /// The maximum amount of time in milliseconds for the server to wait on new documents
+    // to satisfy a change stream query. Uses the server default timeout when omitted.
+    public let maxAwaitTimeMS: Int64?
+
+    /**
+     * The number of documents to return per batch.
+     * This option is sent only if the caller explicitly provides a value. The
```
patrickfreed: Here and elsewhere, we don't need to say "This option is only sent when...". That's direction to the drivers authors from the spec authors and not really meant for users to read.

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * The change stream will only provide changes that occurred at or after
+     * the specified timestamp. Any command run against the server will return
+     * an operation time that can be used here.
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.runCommand/
+     */
+    public let startAtOperationTime: Timestamp?
+
+    /**
+     * Similar to `resumeAfter`, this option takes a resume token and starts a
```
patrickfreed: This docstring is really good / informative. If you could update the `resumeAfter` one to look like this, that would be great.

One interesting thing to note: this is a 4.2 feature, so libmongoc might not have implemented it fully yet. If so, we may have to keep it internal until libmongoc does.

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -0,0 +1,139 @@
...
+
+        let changeStreamPtr: OpaquePointer = mongoc_collection_watch(coll._collection, pipeline._bson, opts?._bson)
+        var replyPtr = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
+        defer {
+            replyPtr.deinitialize(count: 1)
+            replyPtr.deallocate()
+        }
+
+        let decoder = BSONDecoder()
+       let changeStream = try ChangeStream<ChangeStreamDocument<Document>>(stealing: changeStreamPtr,
```
patrickfreed: indentation here and below is off

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -0,0 +1,139 @@
+import mongoc
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+final class ChangeStreamTest: MongoSwiftTestCase {
```
patrickfreed: these tests are a good starting point. I would say it might be worth adding some assertions on the `fullDocument` field (for updates, inserts, and some ops that are neither). You can wait to adding those tests until later PRs if you would like.

Sources/MongoSwift/ChangeStreamOptions.swift
ncarbon: @patrickfreed I see `startAfter` being used in mongoc here: https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-change-stream.c#L122

Sources/MongoSwift/ChangeStreamOptions.swift
ncarbon: @patrickfreed I don't think I'm fully understanding the issue here. By option do you mean the `ChangeStreamOptions.fullDocument`, not the `FullDocument` type? And are you referring to the docstring for the `updateLookup` case?

Tests/MongoSwiftTests/ChangeStreamTests.swift
ncarbon: I added some assertions on `fullDocument` for update and insert ops. Will add more in the next PR.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    var replyPtr = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
+    defer {
+        replyPtr.deinitialize(count: 1)
+        replyPtr.deallocate()
+    }
+
+    var error = bson_error_t()
+    guard mongoc_cursor_error_document(self.changeStream, &error, replyPtr) else {
```
kmahar: should this be `mongoc_change_stream_error_document`?

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
+import mongoc
+
+/// A token used for manually resuming a change stream. Pass this to the `resumeAfter` or `startAfter` fields of
+/// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
+public struct ChangeStreamToken: Codable {
+    private let resumeToken: Document
+
+    public init(resumeToken: Document) {
```
kmahar: does this need to be publicly initializable, given that users will be reading instances of this type from the `ChangeStream.resumeToken` property?

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    guard let pointee = out.pointee else {
+        fatalError("Could not advance the change stream.")
+    }
+
+    // we have to copy because libmongoc owns the pointer.
+    let doc = Document(copying: pointee)
+
+    // Update the resumeToken with the `_id` field from the document.
```
kmahar: Swift style nit: I think it would be more idiomatic to write this block using a `guard`, like:
```swift
guard let resumeToken = doc["_id"] as? Document else {
    // handle error 
}
self.resumeToken = ChangeStreamToken(resumeToken: resumeToken)
```

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /// Decoder for decoding documents into type `T`.
+  internal let decoder: BSONDecoder
+
+  /// Returns the next `T` in the change stream or nil if there is no next value. Will block for a maximum of
+  /// `maxAwaitTimeMS` milliseconds as specified in the `ChangeStreamOptions`, or for the server default timeout
+  /// if omitted.
+  public func next() -> T? {
+    let cursor = self.changeStream
```
kmahar: i think it's kind of confusing to name this variable `cursor` since it's not really a cursor. we only use it in one place - let's just directly pass in `self.changeStream` to `mongoc_change_stream_next`?

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+        out.deallocate()
+    }
+
+    // Check that there’s another document in the stream.
+    guard mongoc_change_stream_next(cursor, out) else {
+      return nil
+    }
+
+    guard let pointee = out.pointee else {
+        fatalError("Could not advance the change stream.")
```
kmahar: nit: let's make this a little more descriptive,  something like `mongoc_change_stream_next` returned `true`, but did not fill out the next document"

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+    } catch {
+        self.swiftError = error
+    }
+
+    return nil
+  }
+
+  /**
+   * Returns the next `T` in this change stream or `nil`, or throws an error if one occurs -- compared to `next()`,
+   * whichreturns `nil` and requires manually checking for an error afterward. Will block for a maximum of
```
kmahar:
```suggestion
   * which returns `nil` and requires manually checking for an error afterward. Will block for a maximum of
```

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+}
+
+/// A MongoDB ChangeStream.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
+public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
+  /// A `ChangeStreamToken` used for manually resuming a change stream.
+  public private(set) var resumeToken: ChangeStreamToken
+
+  /// A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
+  private var client: MongoClient?
```
kmahar: since change streams cannot be manually closed, I think we can make `client`, `session`, and `changeStream` all non-optional and immutable.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  /// Cleans up internal state.
+  deinit {
+    guard let cursor = self.changeStream else {
+        return
+    }
+    mongoc_change_stream_destroy(changeStream)
+    self.changeStream = nil
```
kmahar: per my comment above about making the properties immutable and non optional, these three lines nil-ing out properties can go away.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+      throw err
+    }
+  }
+
+  /// Cleans up internal state.
+  deinit {
+    guard let cursor = self.changeStream else {
+        return
+    }
+    mongoc_change_stream_destroy(changeStream)
```
kmahar: I think we can just directly pass in `self.changeStream` here and we don't need the guard above.

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,76 @@
...
+   public let updatedFields: Document
+
+   /// An array of field names that were removed from the document.
+   public let removedFields: [String]
+}
+
+/// An enum representing the type of operation for this change event.
+public enum OperationType: String, Codable {
+   /// Specifies an operation of type `insert`.
+   /// SeeAlso: https://docs.mongodb.com/manual/reference/change-events/index.html#insert-event
```
kmahar:
```suggestion
   /// - SeeAlso: https://docs.mongodb.com/manual/reference/change-events/index.html#insert-event
```
I think this might be needed for docs tool to pick up on them/format them correctly. same applies to the below `SeeAlso`s

Sources/MongoSwift/MongoNamespace.swift
```diff
@@ -1,11 +1,15 @@
 /// Represents a MongoDB namespace for a database or collection.
-internal struct MongoNamespace {
-    internal let db: String
-    internal let collection: String?
+public struct MongoNamespace: Codable {
+    public let db: String
+    public let collection: String?
+
+    public enum CodingKeys: String, CodingKey {
```
kmahar: I think the `CodingKeys` can be `private`.

Sources/MongoSwift/MongoNamespace.swift
```diff
@@ -1,11 +1,15 @@
 /// Represents a MongoDB namespace for a database or collection.
-internal struct MongoNamespace {
-    internal let db: String
-    internal let collection: String?
+public struct MongoNamespace: Codable {
+    public let db: String
```
kmahar: this and collection probably need docstrings now that this is a public type. I would just keep them very basic like:
"The database name."
"The collection name if this is a collection's namespace, or nil otherwise`.

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.runCommand/
+    public let startAtOperationTime: Timestamp?
+
+    /// A `ChangeStreamToken` similar to `resumeAfter` except `startAfter` will allow users to watch collections
+    /// that have been dropped and recreated or newly renamed collections without missing any notifications.
+    /// The server will report an error if `startAfter` and `resumeAfter` are both specified.
+    /// - SeeAlso: https://docs.mongodb.com/master/changeStreams/#change-stream-start-after
+    public let startAfter: ChangeStreamToken?
+
+    /// Initializes a `ChangeStreamOption`.
```
kmahar:
```suggestion
    /// Initializes a `ChangeStreamOptions`.
```

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,165 @@
...
+/// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
+public struct ChangeStreamToken: Codable {
+    private let resumeToken: Document
+
+    public init(resumeToken: Document) {
+        self.resumeToken = resumeToken
+    }
+
+    public init(from decoder: Decoder) throws {
```
kmahar: why do we need a custom implementation of this?

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// Specifies a collation.
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/command/aggregate
+    public let collation: Document?
+
+    /// The change stream will only provide changes that occurred at or after the specified timestamp.
+    /// Any command run against the server will return an operation time that can be used here.
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.runCommand/
+    public let startAtOperationTime: Timestamp?
+
+    /// A `ChangeStreamToken` similar to `resumeAfter` except `startAfter` will allow users to watch collections
```
patrickfreed: I would reword this as: 
"Similar to `resumeAfter`, this option takes a `ChangeStreamToken` which will serve as the logical starting point for the new change stream. This option differs from `resumeAfter` in that it will allow a change stream to receive notifications even after an invalidate event (e.g. it will allow watching a collection that has been dropped and recreated).

Note: the server will report an error if `startAfter` and `resumeAfter` are both specified.

\- SeeAlso: https://docs.mongodb.com/master/changeStreams/#change-stream-start-after"

Sources/MongoSwift/ChangeStream.swift
ncarbon: I think @kmahar 's [comment](https://github.com/mongodb/mongo-swift-driver/pull/298#discussion_r304034688) about this needing to be more descriptive and state that the document is nil makes sense. So maybe just not mention the libmongoc method but still say that the document is nil in this case?

Sources/MongoSwift/ChangeStream.swift
ncarbon: Oops, that's right. This was changed with SPEC-1343. Updated.

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,81 @@
+/// Describes the modes for configuring the fullDocument field of a `ChangeStreamDocument`.
+public enum FullDocument: RawRepresentable, Codable {
+    /// The change stream document will include both a delta describing the changes to the document,
+    /// as well as a copy of the entire document that was changed from some time after the change occurred.
+    case updateLookup
```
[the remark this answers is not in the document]
ncarbon: That makes sense. Updated.

Sources/MongoSwift/ChangeStream.swift
ncarbon: The `resumeToken` within `ChangeStreamToken` wasn't getting decoded (coding keys error), so @patrickfreed suggested using a decoder init.

Sources/MongoSwift/ChangeStream.swift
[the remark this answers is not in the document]
ncarbon: If that's the case, the `resumeToken` still gets updated/extracted from the `_id` field in `next`. @kmahar I think this init should not be public then.

Sources/MongoSwift/ChangeStream.swift
patrickfreed: Sounds good to me!

Sources/MongoSwift/ChangeStream.swift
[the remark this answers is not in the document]
ncarbon: @patrickfreed  If that is the case, wouldn't the resume token still get updated from the `id` field? @kmahar  brought up whether this init needed to be public or not. So maybe this init should be internal?

Sources/MongoSwift/ChangeStream.swift
ncarbon: There was a decoding error while updating `resumeToken` from the `_id` field in `next`.
`"No value associated with key CodingKeys(stringValue: \"resumeToken\", intValue: nil) (\"resumeToken\").", underlyingError: nil))`
So @patrickfreed  suggested using an decoder init so the `resumeToken` gets initialized as a document.

Sources/MongoSwift/ChangeStream.swift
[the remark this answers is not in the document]
ncarbon: @patrickfreed Actually, when I first added the `ChangeStreamToken` struct, it only had this public init. Then we added the decoder init when we encountered this decoding error: `"No value associated with key CodingKeys(stringValue: \"resumeToken\", intValue: nil) (\"resumeToken\").", underlyingError: nil))`
When the `resumeToken` gets updated from the `_id` field in the document returned from libmongoc, it always uses the decoder init. 
I think I am not seeing or understanding how this public init will be used/needed now. If the user opted for `ChangeStream<Document>` wouldn't the `ChangeStream.resumeToken` still get updated in `next` and just extract it from there?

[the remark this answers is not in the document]
patrickfreed: >I think I am not seeing or understanding how this public init will be used/needed now. If the user opted for ChangeStream<Document> wouldn't the ChangeStream.resumeToken still get updated in next and just extract it from there?

You're absolutely right, good point! I forgot that we already cache it for them. In that case, this initializer doesn't need to be public. In fact, since you said we only use the decoder initializer, it doesn't actually have to exist at all, right?

Sources/MongoSwift/ChangeStream.swift
patrickfreed: Right. The `ChangeStreamToken` type doesn't actually care about the fields present in it (and they're subject to change according to server version), so the `Codable` conformance is just deferred to the underlying document being wrapped. There isn't actually a `resumeToken` field (as per the error received when relying on the synthesized conformance) in the `_id` field of a change stream event.

Sources/MongoSwift/ChangeStream.swift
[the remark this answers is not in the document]
ncarbon: @patrickfreed Actually, both inits are needed since `ChangeStream.resumeToken` is a non-optional and needs to be initialized in `ChangStream.init`. It also used when updating the `resumeToken` in `next`.
I think this init can just be `internal`?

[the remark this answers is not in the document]
patrickfreed: Oh, right. Yep `internal` is good.

ncarbon: So far, this is only used internally, so I will make it `internal`

Sources/MongoSwift/ChangeStream.swift
kmahar: nit: use `self.changeStream`. swift doesn't require explicit `self` but we try to always use it. (there's a swiftlint rule for this but it misses most violations 🤦‍♀️

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,155 @@
+import mongoc
+/// A token used for manually resuming a change stream. Pass this to the `resumeAfter` or `startAfter` fields of
+/// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
+public struct ChangeStreamToken: Codable {
+    private let resumeToken: Document
+
+    internal init(resumeToken: Document) {
+      self.resumeToken = resumeToken
```
kmahar: indentation looks off here.
is there maybe some setting you can enable in VSCode to deal with these indentation issues?
(if you can't find one maybe ask @mbroadst since he uses it)
```suggestion
        self.resumeToken = resumeToken
```

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,155 @@
...
+    } catch {
+        self.swiftError = error
+    }
+
+    return nil
+  }
+
+  /**
+   * Returns the next `T` in this change stream or `nil`, or throws an error if one occurs -- compared to `next()`,
+   * whichreturns `nil` and requires manually checking for an error afterward. Will block for a maximum of
```
kmahar:
```suggestion
   * which returns `nil` and requires manually checking for an error afterward. Will block for a maximum of
```

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    public init(from decoder: Decoder) throws {
+      self.resumeToken = try Document(from: decoder)
+    }
+}
+
+/// A MongoDB ChangeStream.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
+public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
```
kmahar: this whole class/maybe whole file appears to be indented with 2 spaces rather than 4.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,155 @@
...
+        return extractMongoError(error: error, reply: reply)
+    }
+
+    // Otherwise, the only feasible error is that the user tried to advance a dead change stream cursor,
+    // which is a logic error. We will still parse the mongoc error to cover all cases.
+    return extractMongoError(error: error)
+  }
+
+  /// Decoder for decoding documents into type `T`.
+  internal let decoder: BSONDecoder
```
kmahar: is this used outside the class or can it be private?

kmahar: also, nit, let's put this up with all of the other stored properties above `error`

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,76 @@
+/// An `UpdateDescription` containing fields that will be present in the change stream document for
```
kmahar: indentation also seems off in this file. `UpdateDescription` and `OperationType`'s properties are only indented three spaces

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,76 @@
...
+    * the `ChangeStream` that emitted this document.
+    *
+    * For operations of type `insert’ and `replace’, this key will contain the document being inserted, or the new
+    * version of the document that is replacing the existing document, respectively.
+    *
+    * For operations of type `update’, this key will contain a copy of the full version of the document from some point
+    * after the update occurred. If the document was deleted since the updated happened, it will be null.
+    */
+    public let fullDocument: T?
+ }
```
kmahar:
```suggestion
}
```

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,91 @@
...
+        default:
+            self = .other(rawValue)
+        }
+    }
+}
+
+/// Options to use when creating a `ChangeStream`.
+public struct ChangeStreamOptions: Codable {
+    /**
+     * Indicates how the `fullDocument` field of a `ChangeStreamDocument` should be filled out by the server.
```
kmahar: nit: I'd suggest just saying "change stream document" rather than `ChangeStreamDocument` as it will also apply to the documents returned in the case where you have a `ChangeStream<Document>`

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,91 @@
...
+     * The change stream will attempt to resume notifications starting after the operation associated with
+     * the provided token.
+     * - Note: A change stream cannot be resumed after an invalidate event (e.g. a collection drop or rename).
+     *         Use the `startAfter` option in those cases instead.
+     * - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
+     */
+    public let resumeAfter: ChangeStreamToken?
+
+    /// The maximum amount of time in milliseconds for the server to wait on new documents to satisfy a
+    /// change stream query. Uses the server default timeout when omitted.
```
kmahar: nit: "Uses the server default timeout when omitted" sounds like *we* use/send the server default if they omit it.

instead let's phrase as "If omitted, the server will use its default timeout."

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,91 @@
...
+    public let maxAwaitTimeMS: Int64?
+
+    /// The number of documents to return per batch. The default is to not send a value.
```
kmahar: nit: I think "the default is to not send a value" is more a drivers internal thing. I would suggest phrasing this like the `maxAwaitTimeMS` suggestion I made above:
"If omitted, the server will use its default batch size."

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,91 @@
+/// Describes the modes for configuring the fullDocument field of a `ChangeStreamDocument`.
```
kmahar: similar to nit below, we may want to say "change stream document" rather than `ChangeStreamDocument`.

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -0,0 +1,91 @@
...
+public enum FullDocument: RawRepresentable, Codable {
+    /// Specifies that the `fullDocument` field will contain a copy of the entire document that was changed from
+    /// some time after the change occurred. If the document was deleted since the updated happened, it will be null.
```
kmahar: nit: null -> nil

Sources/MongoSwift/ChangeStreamDocument.swift
```diff
@@ -0,0 +1,76 @@
...
+   /**
+    * Always present for operations of type `insert` and `replace`. Also present for operations of type `update` if the
+    * user has specified `.updateLookup` for the `fullDocument` option in the `ChangeStreamOptions` used to create
+    * the `ChangeStream` that emitted this document.
+    *
+    * For operations of type `insert’ and `replace’, this key will contain the document being inserted, or the new
+    * version of the document that is replacing the existing document, respectively.
+    *
+    * For operations of type `update’, this key will contain a copy of the full version of the document from some point
+    * after the update occurred. If the document was deleted since the updated happened, it will be null.
```
kmahar: same null -> nil nit

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -0,0 +1,151 @@
...
+        let client = try MongoClient()
+        let db = client.db(type(of: self).testDatabase)
+        defer { try? db.drop() }
+        let coll = try db.createCollection(self.getCollectionName(suffix: "1"))
+        let session = try client.startSession()
+        let options = ChangeStreamOptions(fullDocument: .updateLookup)
+        let opts = try encodeOptions(options: options, session: session)
+        let pipeline: Document = []
+
+        try client.connectionPool.withConnection { conn in
```
kmahar: just FYI: this approach is fine for the purpose of this test, but in the next PR when you add watch methods you'll want to switch to doing 
```swift
let conn = try client.connectionPool.checkOut()
try coll.withMongocCollection(from: conn) { collPtr in
    ...
    let changeStream = try ChangeStream<ChangeStreamDocument<Document>>(stealing: changeStreamPtr,
                                                                        client: client,
                                                                        connection: conn, // changestream stores source conn
                                                                        session: session,
                                                                        decoder: decoder)
```

and then `ChangeStream.deinit` will do `self.client.connectionPool.checkIn(self.connection)`.

this will ensure that a `mongoc_client_t` is not destroyed before a `mongoc_change_stream_t` that depends on it is.
(note `mongoc_change_stream_t` stores/uses a pointer its source `mongoc_client_t` - [see here](https://github.com/mongodb/mongo-c-driver/blob/c88b84ff383386f2491ff5298f1a1fd7bae6dd7a/src/libmongoc/src/mongoc/mongoc-change-stream-private.h#L47) - but it's up to us to ensure the client isn't destroyed)

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -0,0 +1,151 @@
...
+            try coll.withMongocCollection(from: conn) { collPtr in
+                // TODO: Use MongoCollection.watch() instead `mongoc_collection_watch` of once it gets added
+                let changeStreamPtr: OpaquePointer = mongoc_collection_watch(collPtr, pipeline._bson, opts?._bson)
+                var replyPtr = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
+                defer {
+                    replyPtr.deinitialize(count: 1)
+                    replyPtr.deallocate()
+                }
+                expect(try ChangeStream<ChangeStreamDocument<Document>>(stealing: changeStreamPtr,
```
kmahar: you might want to do something like the second two tests where you trigger events being posted to the change stream.

Sources/MongoSwift/ChangeStreamOptions.swift
kmahar: that's the master branch of libmongoc, which contains unreleased code. if you look at the [r1.14 branch version of the file](https://github.com/mongodb/mongo-c-driver/blob/r1.14/src/libmongoc/src/mongoc/mongoc-change-stream.c), which is what is in the 1.14 release, it's not there.
so I don't think we can support `startAfter` yet unfortunately.

kmahar: so I think we need to make this property internal and add a comment under the docstring like: "TODO: make this public when support is added for 4.2 change stream features". and then remove the `startAfter` param from the options init and remove mention of it from elsewhere in the file

ncarbon: Swift still wants the `startAfter` property to be initialized. Should we just remove this from the code for now (or store some value/comment it all out?) ?

Sources/MongoSwift/ChangeStream.swift
ncarbon: oops, yeah 😓. I think I was trusting vscode every time I hit enter and didn't realize it wasn't 4 spaces. I just got an extension called `indent one space` and another one `indenticator`  so I think that should help now.

Tests/MongoSwiftTests/ChangeStreamTests.swift
ncarbon: I think I just wanted to test that creating a change stream did not throw an error. Would it be preferable to combine this with the rest of the tests instead?

Tests/MongoSwiftTests/ChangeStreamTests.swift
ncarbon: Updated. In the future, `MongoCollection` will no longer store a reference to `mongoc_client_t`, so `ChangeStream` would just store a `Connection` instead? I added a todo similar to the one in `MongoClient` : `// TODO: SWIFT-374: remove this property.`
Let me know if this is okay.

Sources/MongoSwift/ChangeStreamOptions.swift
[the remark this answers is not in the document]
patrickfreed: Sorry, upon reading it again I think it needs one last addition. The beginning should read something like "Specifies that the `fullDocument` field of an update event..." to clarify when different behavior will be observed.

Sources/MongoSwift/ChangeStreamOptions.swift
patrickfreed: Yeah, that should be fine. Also can you file a ticket so that we remember to do the work and update the TODO with the ticket number?

Sources/MongoSwift/ChangeStreamOptions.swift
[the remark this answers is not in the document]
ncarbon: updated

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/// A MongoDB ChangeStream.
+/// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
+public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
+    /// A `ChangeStreamToken` used for manually resuming a change stream.
+    public private(set) var resumeToken: ChangeStreamToken
+
+    /// A `MongoClient` stored to make sure the source client stays valid until the change stream is destroyed.
+    // TODO: SWIFT-374: remove this property.
```
kmahar: per my other comment, no todo needed, this will stick around forever

Tests/MongoSwiftTests/ChangeStreamTests.swift
kmahar: actually, `MongoCollection` will always need to store a `MongoClient` to ensure that the `MongoClient` and its `ConnectionPool` stay alive for as long as the `MongoCollection` does. So no todo needed.

Tests/MongoSwiftTests/ChangeStreamTests.swift
kmahar: actually let's hold off on this until you add the watch methods, and then we can just explicitly test watching on a MongoClient there

kmahar:
no further comments but will wait to see if others request any changes before I approve 👍

PR #299, opened 2019-07-22
SWIFT-374 Add pooled mode to ConnectionPool, and have ClientSessions store their source Connections

kmahar:
This is the final PR for SWIFT-374 🎉 will make some comments inline

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -55,13 +55,30 @@ private func withSessionOpts<T>(wrapping options: ClientSessionOptions?,
  */
 public final class ClientSession {
     /// Error thrown when an inactive session is used.
-    internal static var SessionInactiveError = UserError.logicError(message: "Tried to use an inactive session")
+    internal static let SessionInactiveError = UserError.logicError(message: "Tried to use an inactive session")
+    /// Error thrown when a user attempts to use a session with a client it was not created from.
+    internal static let ClientMismatchError = UserError.invalidArgumentError(
+        message: "Sessions may only be used with the MongoClient used to create them")
 
-    /// Pointer to the underlying `mongoc_client_session_t`.
-    internal fileprivate(set) var _session: OpaquePointer?
+    /// Enum for tracking the state of a session.
+    internal enum State {
+        /// Indicates that the session is active. Stores a pointer to the underlying `mongoc_client_session_t` and the
+        /// source `Connection` for this session.
+        case active(session: OpaquePointer, connection: Connection)
+        /// Indicates that the session has been ended.
+        case inactive
+    }
+
+    /// Indicates the state of this session.
+    internal private(set) var state: State
 
     /// Returns whether this session has been ended or not.
-    internal var active: Bool { return self._session != nil }
+    internal var active: Bool {
+        if case .active = self.state {
+            return true
+        }
+        return false
+    }
 
     /// The client used to start this session.
     public let client: MongoClient
@@ -69,21 +86,29 @@ public final class ClientSession {
     /// The session ID of this session.
     public let id: Document
 
-    /// The most recent cluster time seen by this session.
-    /// If no operations have been executed using this session and `advanceClusterTime` has not been called, this will
-    /// be `nil`.
+    /// The most recent cluster time seen by this session. This value will be nil if either of the following are true:
+    /// - No operations have been executed using this session and `advanceClusterTime` has not been called.
+    /// - This session has been ended.
     public var clusterTime: Document? {
-        guard let time = mongoc_client_session_get_cluster_time(self._session) else {
+        guard case let .active(session, _) = self.state,
+                    let time = mongoc_client_session_get_cluster_time(session) else {
             return nil
         }
         return Document(copying: time)
     }
 
-    /// The operation time of the most recent operation performed using this session.
+    /// The operation time of the most recent operation performed using this session. This value will be nil if either
+    /// of the following are true:
+    /// - No operations have been performed using this session.
+    /// - This session has been ended.
     public var operationTime: Timestamp? {
+        guard case let .active(session, _) = self.state else {
+            return nil
+        }
+
         var timestamp: UInt32 = 0
         var increment: UInt32 = 0
-        mongoc_client_session_get_operation_time(self._session, &timestamp, &increment)
+        mongoc_client_session_get_operation_time(session, &timestamp, &increment)
 
         guard timestamp != 0 && increment != 0 else {
             return nil
@@ -99,25 +124,42 @@ public final class ClientSession {
         self.options = options
         self.client = client
 
-        self._session = try withSessionOpts(wrapping: options) { opts in
+        let connection = try client.connectionPool.checkOut()
+        let session: OpaquePointer = try withSessionOpts(wrapping: options) { opts in
             var error = bson_error_t()
-            guard let session = mongoc_client_start_session(client._client, opts, &error) else {
+            guard let session = mongoc_client_start_session(connection.clientHandle, opts, &error) else {
+                // we won't call end(), so need to check the connection back in here manually.
+                client.connectionPool.checkIn(connection)
                 throw extractMongoError(error: error)
             }
             return session
         }
+
+        self.state = .active(session: session, connection: connection)
         // swiftlint:disable:next force_unwrapping
-        self.id = Document(copying: mongoc_client_session_get_lsid(self._session)!) // always returns a value
+        self.id = Document(copying: mongoc_client_session_get_lsid(session)!) // always returns a value
+    }
+
+    /// Retrieves this session's underlying connection. Throws an error if the provided client was not the client used
+    /// to create this session, or if this session has been ended.
+    internal func getConnection(forUseWith client: MongoClient) throws -> Connection {
+        guard case let .active(_, connection) = self.state else {
+            throw ClientSession.SessionInactiveError
+        }
+        guard self.client == client else {
+            throw ClientSession.ClientMismatchError
+        }
+        return connection
     }
 
     /// Destroy the underlying `mongoc_client_session_t` and set this session to inactive.
     /// Does nothing if this session is already inactive.
     internal func end() {
-        guard self.active else {
-            return
+        if case let .active(session, connection) = self.state {
+            mongoc_client_session_destroy(session)
+            self.client.connectionPool.checkIn(connection)
+            self.state = .inactive
         }
-        mongoc_client_session_destroy(self._session)
-        self._session = nil
     }
 
     /// Cleans up internal state.
@@ -126,38 +168,44 @@ public final class ClientSession {
     }
 
     /**
-     * Advances the clusterTime for this session to the given time, if it is greater than the current clusterTime.
-     * If the provided clusterTime is less than the current clusterTime, this method has no effect.
+     * Advances the clusterTime for this session to the given time, if it is greater than the current clusterTime. If
+     * the session has been ended, or if the provided clusterTime is less than the current clusterTime, this method has
+     * no effect.
      *
      * - Parameters:
      *   - clusterTime: The session's new cluster time, as a `Document` like `["cluster time": Timestamp(...)]`
      */
     public func advanceClusterTime(to clusterTime: Document) {
-        mongoc_client_session_advance_cluster_time(self._session, clusterTime._storage._bson)
+        if case let .active(session, _) = self.state {
+            mongoc_client_session_advance_cluster_time(session, clusterTime._bson)
+        }
     }
 
     /**
      * Advances the operationTime for this session to the given time if it is greater than the current operationTime.
-     * If the provided operationTime is less than the current operationTime, this method has no effect.
+     * If the session has been ended, or if the provided operationTime is less than the current operationTime, this
+     * method has no effect.
      *
      * - Parameters:
      *   - operationTime: The session's new operationTime
      */
     public func advanceOperationTime(to operationTime: Timestamp) {
-        mongoc_client_session_advance_operation_time(self._session, operationTime.timestamp, operationTime.increment)
+        if case let .active(session, _) = self.state {
+            mongoc_client_session_advance_operation_time(session, operationTime.timestamp, operationTime.increment)
+        }
     }
 
     /// Appends this provided session to an options document for libmongoc interoperability.
     /// - Throws:
     ///   - `UserError.logicError` if this session is inactive
     internal func append(to doc: inout Document) throws {
-        guard self.active else {
+        guard case let .active(session, _) = self.state else {
             throw ClientSession.SessionInactiveError
         }
 
         var error = bson_error_t()
         try withMutableBSONPointer(to: &doc) { docPtr in
-            guard mongoc_client_session_append(self._session, docPtr, &error) else {
+            guard mongoc_client_session_append(session, docPtr, &error) else {
                 throw extractMongoError(error: error)
             }
         }
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -37,17 +37,15 @@ internal class ConnectionPool {
 
     /// Initializes the pool in pooled mode using the provided `ConnectionString`.
     internal init(from connString: ConnectionString) throws {
-        // TODO SWIFT-374: this initializer should actually create a `mongoc_client_pool_t` and do appropriate setup
-        // on that.
-        guard let clientHandle = mongoc_client_new_from_uri(connString._uri) else {
+        guard let pool = mongoc_client_pool_new(connString._uri) else {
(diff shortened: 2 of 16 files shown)


Tests/MongoSwiftTests/SDAMMonitoringTests.swift
```diff
@@ -52,32 +52,33 @@ final class SDAMTests: MongoSwiftTestCase {
...
         // check event count and that events are of the expected types
-        expect(receivedEvents.count).to(equal(5))
+        expect(receivedEvents.count).to(equal(7))
         expect(receivedEvents[0]).to(beAnInstanceOf(TopologyOpeningEvent.self))
         expect(receivedEvents[1]).to(beAnInstanceOf(TopologyDescriptionChangedEvent.self))
         expect(receivedEvents[2]).to(beAnInstanceOf(ServerOpeningEvent.self))
         expect(receivedEvents[3]).to(beAnInstanceOf(ServerDescriptionChangedEvent.self))
         expect(receivedEvents[4]).to(beAnInstanceOf(TopologyDescriptionChangedEvent.self))
 
+        // we are ignoring the last two events here. I think they are being generated by mongoc in error.
```
kmahar: asked in #mongoc and waiting to hear back. 
this test is based on the standalone test case in the SDAM yaml tests. 

the extra events are a ServerChangedEvent and a TopologyChangedEvent but in both of them the previous and new descriptions are identical.
I think it's likely that these events are generated in error but libmongoc doesn't notice because they only look at the first 5 events that the yaml test says should be there.

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -38,3 +41,25 @@ internal func encodeOptions<T: Encodable>(options: T?, session: ClientSession?)
...
+
+/**
+ * This method takes in a client and an optional session and determines the appropriate connection the method should
+ * use to interact with mongoc.
+ * If session is non-nil: returns the session's underlying connection. Throws an error if the session is inactive or
+ * was not created from the provided client.
+ * If session is nil: returns a new connection from the client's underlying pool. **The caller is responsible for
+ * ensuring the returned connection is checked back into the pool.**
+ */
+internal func resolveConnection(client: MongoClient, session: ClientSession? = nil) throws -> Connection {
```
patrickfreed: The way this is currently written, it's impossible for the caller to know whether they should be returning the connection to the pool or not (i.e. whether or not the connection came from a session or from the pool). Above, we have to unwrap the session to figure this out (and pass it back in as an optional anyways). 

I think it could be useful to combine this function and `pool.withConnection` into a single `withResolvedConnection(client:session:func)` which would return the connection in the appropriate cases.

That way we could rewrite above as something like:
```swift
internal func execute<...>(...) {
    return try withResolvedConnection(client: client, session: session) {
        try operation.execute(using: $0, session: session)
    }
}
```
And have the guarantee that the connection is returned when it should be (and not when it shouldn't).

Sources/MongoSwift/ClientSession.swift
```diff
@@ -56,34 +56,58 @@ private func withSessionOpts<T>(wrapping options: ClientSessionOptions?,
 public final class ClientSession {
     /// Error thrown when an inactive session is used.
     internal static var SessionInactiveError = UserError.logicError(message: "Tried to use an inactive session")
+    /// Error thrown when a user attempts to use a session with a client it was not created from.
+    internal static var ClientMismatchError = UserError.invalidArgumentError(
+        message: "Sessions may only be used with the MongoClient used to create them")
+
+    /// Enum for tracking the state of a session.
+    internal enum State {
+    /// Indicates that the session is active. Stores a pointer to the underlying `mongoc_client_session_t` and the
```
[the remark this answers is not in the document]
kmahar: done

Sources/MongoSwift/ClientSession.swift
```diff
@@ -56,34 +56,58 @@ private func withSessionOpts<T>(wrapping options: ClientSessionOptions?,
...
 
-    /// Pointer to the underlying `mongoc_client_session_t`.
-    internal fileprivate(set) var _session: OpaquePointer?
+    /// Indicates the state of this session.
+    internal private(set) var state: State
 
     /// Returns whether this session has been ended or not.
-    internal var active: Bool { return self._session != nil }
+    internal var active: Bool {
+        if case .active = self.state {
```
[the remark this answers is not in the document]
kmahar: since `State` isn't `Equatable` you have to match it with the case syntax. 
I considered making it `Equatable` but it feels a little weird as we would have to ignore the associated values in the `active` case (or also somehow make `connection` `Equatable`) and this is the only place we'd use it

Tests/MongoSwiftTests/SDAMMonitoringTests.swift
kmahar: update: this is a CDRIVER bug, [CDRIVER-3256](https://jira.mongodb.org/browse/CDRIVER-3256). they weren't running the SDAM monitoring tests in pooled mode so they missed this.

the spec says the events should not be emitted when the old and new descriptions are the same, so I have updated our SDAM monitoring code so that it compares the descriptions for the changed events, and does not publish the event if they are the same.

changed this test back to expecting 5 events.

Sources/MongoSwift/Operations/Operation.swift
kmahar: ya, I felt a little weird about the fact that you can't tell unless if you inspect whether the session you passed in was `nil`.

your proposal works for the case above, and the case where we are re-doing the code from above in `bulkWrite`.

however, we are calling `resolveConnection` from a bunch of places that return a cursor, and in those cases we don't want to check the connection back in til later. (same will be true for change stream methods).

it's also kind of weird in the cursor case/change stream case, because we rely on the presence of `self.session` to determine whether we need to return the connection. maybe having `resolveConnection` be explicit and return both  a `Connection` and a boolean indicating whether the connection is now owned by the caller would be better?

we could then add `withResolvedConnection` which returns the connection to the pool if `resolveConnection`

and then maybe cursor/change streams could store an explicit boolean indicating whether they own their `Connection`.

Sources/MongoSwift/SDAM.swift
```diff
@@ -214,7 +238,12 @@ public struct TopologyDescription {
     public let type: TopologyType
 
     /// The replica set name.
-    public var setName: String? { return self.servers[0].setName }
+    public var setName: String? {
+        guard !self.servers.isEmpty else {
```
kmahar: this would crash on empty self.servers 😬

Sources/MongoSwift/ClientSession.swift
```diff
@@ -126,38 +156,44 @@ public final class ClientSession {
     }
 
     /**
-     * Advances the clusterTime for this session to the given time, if it is greater than the current clusterTime.
-     * If the provided clusterTime is less than the current clusterTime, this method has no effect.
+     * Advances the clusterTime for this session to the given time, if it is greater than the current clusterTime. If
+     * the session has been ended, or if the provided clusterTime is less than the current clusterTime, this method has
```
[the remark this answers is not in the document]
mbroadst: agreed, this is fine. You note that the method similarly doesn't throw an exception if you try to advance the time using an older time.

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -37,17 +37,15 @@ internal class ConnectionPool {
 
     /// Initializes the pool in pooled mode using the provided `ConnectionString`.
     internal init(from connString: ConnectionString) throws {
-        // TODO SWIFT-374: this initializer should actually create a `mongoc_client_pool_t` and do appropriate setup
-        // on that.
-        guard let clientHandle = mongoc_client_new_from_uri(connString._uri) else {
+        guard let pool = mongoc_client_pool_new(connString._uri) else {
```
mbroadst: :champagne:

Sources/MongoSwift/Counter.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+/// Threadsafe counter that generates an increasing sequence of integers.
+internal class Counter {
```
mbroadst: :open_mouth:  how... can it be that swift has not atomics

Sources/MongoSwift/Counter.swift
```diff
@@ -0,0 +1,16 @@
...
+    private let queue = DispatchQueue(label: "counter queue")
```
mbroadst: I think we should probably use a more unique identifier here, afaict this is registered with GCD so you might have conflicts.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -133,30 +130,26 @@ public class MongoClient {
     /// If command and/or server monitoring is enabled, indicates what event types notifications will be posted for.
     internal var monitoringEventTypes: [MongoEventType]?
 
+    /// A unique identifier for this client.
+    internal let _id = counter.next()
+
+    /// Counter for generating client _ids.
+    internal static let counter = Counter()
```
mbroadst: nit: this is more of a `ClientIdProducer` or generator

Sources/MongoSwift/Operations/Operation.swift
mbroadst: I was actually thinking about the `resolveConnection` duplication for cursors above. In node I chose to have cursors constructed with an operation, so they could reuse `executeOperation` internally. We have different requirements here, but could we perhaps accept e.g. a lambda in the `MongoCursor` constructor, so you'd have some pseudocode like:

```
 return try MongoCursor(from: cursor, client: self, connection: conn, decoder: self.decoder, session: session, initializer: { conn -> OpaquePointer in 
   // call your mongoc function here
})
```
(caveat emptor: I have no recollection what closure capture rules are in swift, this probably won't even compile)

The idea mostly would be that you could essentially treat the initializer like an operation, and localize the connection resolution to one place. This would have the added benefit of only having `resolveConnection` called in two places in the codebase, at which point the confusion Patrick points out might be more acceptable.

Sources/MongoSwift/SDAM.swift
mbroadst: :man_facepalming:

kmahar:
BTW when I rebased I had to make some updates to the DB enumeration code because it was a bit behind on operations layer changes - still used the passed-in session rather than a stored one, and still used `self._client._client`

Sources/MongoSwift/Operations/Operation.swift
patrickfreed: Since we've already verified that the session exists and we therefore know the connection we're checking out will come from it and not the pool, I think it would be clearer if, instead of invoking `resolveConnection` here, we added some of the below logic to a method on `ClientSession` itself and called that. Then we'd reserve usage of `resolveConnection` exclusively for circumstances in which it's ambiguous where the connection could come from. 

This is a bit nitpicky I know, but imo it clears up the intention of the code a little bit.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -234,7 +237,7 @@ public class MongoClient {
      *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
      *     on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
      *
-     * - Returns: A `MongoCursor` over `Document`s describing the databases matching provided criteria
+     * - Returns: A `[DatabaseSpecification]` containing the databases matching provided criteria.
```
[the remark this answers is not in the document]
kmahar: I was doing command+F to find all the methods that returned MongoCursors to update to the new initializer so I stumbled upon this

Sources/MongoSwift/Operations/Operation.swift
kmahar: so I added back the initial `getConnection(forUseWith client: MongoClient)` method and now that's used instead.
then I realized at that point the only place we were even using `resolveConnection` anymore was in the `MongoCursor` initializer, now that we have this new lambda thing for cursor init. so I've just deleted the method and moved the logic there. 
I think that is clearer because `MongoCursor.init` now mirrors `MongoCursor.deinit` - we both assign and clean up `self._connection` based on whether `self._session` is nil.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -21,22 +21,17 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
-    internal init(from cursor: OpaquePointer,
-                  client: MongoClient,
-                  connection: Connection,
+    internal init(client: MongoClient,
                   decoder: BSONDecoder,
-                  session: ClientSession?) throws {
-        self._cursor = cursor
+                  session: ClientSession?,
+                  initializer: (Connection) -> OpaquePointer) throws {
+        self._connection = try session?.getConnection(forUseWith: client) ?? client.connectionPool.checkOut()
```
patrickfreed: nice one-liner

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -21,9 +21,14 @@ internal struct DefaultOperationExecutor: OperationExecutor {
...
                                         client: MongoClient,
                                         session: ClientSession?) throws -> T.OperationResult {
-        // TODO SWIFT-374: if session is non-nil, use its underlying Connection
+        // if a session was provided, use its underlying connection
+        if let session = session {
+            let conn = try session.getConnection(forUseWith: client)
+            return try operation.execute(using: conn, session: session)
+        }
+        // otherwise use a new connection from the pool
         return try client.connectionPool.withConnection { conn in
```
patrickfreed: We may run into trouble here with change streams since `watch` both performs I/O during creation of the stream _and_ holds onto the connection for future use. Currently this isn't a problem for us because cursors aren't constructed via operations.

Sources/MongoSwift/MongoCursor.swift
kmahar: 😎

Sources/MongoSwift/Operations/Operation.swift
kmahar: yeah... I think like cursors, change stream operations are probably not going to be executable via the executor.

patrickfreed: I think we probably want to avoid a scenario where certain operations aren't run through the executor. Cursors are okay to not Operation-ify imo since they're not actually performing any I/O. For change streams we'd want them to be actual `Operation`s.

Perhaps modifying the `Operation` protocol to include some computed property that determines whether the connection is still being held or not would be useful.

something like:
```swift
let connection = try client.connectionPool.checkOut()
defer { 
    if !operation.holdsConnection {
        client.connectionPool.checkIn(connection)
    }
}
return try operation.execute(using: connection, session: nil)
```

kmahar: > I think we probably want to avoid a scenario where certain operations aren't run through the executor.

this is already true for `NextOperation`, which was done so that we avoid checking out a new connection (which would be ignored) each time we call `next`. by manually calling execute we can avoid that. 
if we want those and change stream next ops executed via executor, we might consider adding some flag on an operation that indicates the executor should pass in a dummy connection.

for the operation that calls `watch`, what you propose seems reasonable, though I wonder if we can come up with some abstraction that covers both that and the next operations 🤔

I updated [SWIFT-523](https://jira.mongodb.org/browse/SWIFT-523) with some details from this convo, feel free to edit/add 

ugh, why does mongoc make everything so hard 🤦‍♀

patrickfreed: We could possibly use the same computed property to determine if we need to check out a connection at all. So the property could be more answering the question "is this operation currently holding onto a connection?" rather than "does this type of operation hold onto its connections?". 

Then we'd have:
- Pre execution: If it is holding onto a connection, don't give it a new one. 
- Post execution: If we checked something out and the operation is still holding onto a connection, don't check anything back in.

kmahar:
apologies for squashing all the prior commits together, having like 18 commits was too much of a pain when rebasing 😛 
the latest commit contains new changes to bring the change streams code in line with the new `MongoCursor` changes that are introduced in this PR: specifically, the `ChangeStream` init now takes in a closure that constructs the `mongoc_change_stream_t` given a `Connection`. this allows us to do all the connection resolution for change streams (that decides whether we use a session's connection or a new one) in one place.
cc @ncarbon

Sources/MongoSwift/Operations/Operation.swift
mbroadst: I agree that operations should pertain to I/O, in particular with a server. A change stream is interesting because its an aggregate operation, and then it becomes a cursor. I like the second idea @patrickfreed, I don't think this is a question so much of whether we should check out a connection, but whether the "execution context" already has a bound connection bypassing the need to check one out at all. Let's defer this to SWIFT-253, but prioritize doing that work

kmahar: ok i've added these comments to the ticket so we can refer back to them later 👍

PR #301, opened 2019-07-23
SWIFT-176 Add Change Streams watch methods

ncarbon:
This is the second PR for ChangeStreams support.


Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
         return MongoDatabase(name: name, client: self, options: options)
     }
 
+    /**
+     * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+     * system collections i.g. "config", "local", and "admin" databases. By default, the type `CollectionType` is
```
patrickfreed: Since we're on a client here, there is no such `CollectionType`, so that sentence can be removed.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+      *   - `UserError.invalidArgumentError` if the options passed formed an invalid combination or the pipeline passed
+      *     is invalid.
+      *   - `UserError.invalidArgumentError` if the `_id` field is projected out of the change stream documents by the
+      *     pipeline.
+      * - SeeAlso: https://docs.mongodb.com/manual/reference/system-collections/
+      */
+     public func watch<T: Codable>(_  pipeline: [Document],
+                                   options: ChangeStreamOptions?  =  nil,
+                                   session: ClientSession? = nil,
+                                   withFullDocumentType: T.Type) throws ->
```
patrickfreed: The `T` here isn't the type of the `fullDocument` field--it's the type of the entire change stream document itself. This parameter label should instead be something like `withReturnType` (as per the design doc) or `withEventType`. I kind of prefer one of the latter two since this type affects what the change stream returns, not what this method returns.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -272,6 +272,91 @@ public class MongoDatabase {
...
+     public func watch(_ pipeline: [Document],
+                       options: ChangeStreamOptions? = nil,
+                       session: ClientSession? = nil) throws ->
+                       ChangeStream<ChangeStreamDocument<Document>> {
+        return try self.watch(pipeline, options: options, session: session, withFullDocumentType: Document.self)
+     }
+
+    /**
+     * Starts a `ChangeStream` on a database. Excludes system collections. Associates the specified `Codable` type `T`
+     * with the `fullDocument` field in the `ChangeStreamDocument`.
```
patrickfreed: I'd amend this docstring to say "...`ChangeStreamDocument`s emitted by the returned `ChangeStream`." This also applies to other docstrings for similar watch methods on client/collection.

Also the docstring alignment is off by one. The top '/' should be aligned with the 'p' in `public`.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -133,4 +133,92 @@ public class MongoCollection<T: Codable> {
...
+    /**
+     * Starts a `ChangeStream` on a collection. By default, the type `CollectionType` is associated with the
+     * `fullDocument`field in `ChangeStreamsDocument`.
+     * - Parameters:
+     *   - Pipeline: The pipeline of stages to append to an initial `ChangeStream` stage.
+     *   - Options: An optional `ChangeStreamOptions` to use on the initial `ChangeStream` stage.
+     *   - Session: An optional `ChangeStream` to use with this change stream.
+     * - Returns: A change stream on a specific collection.
+     * - Throws:
+     *   - `ServerError.commandError` if an error occurs on the server while creating the change stream.
```
patrickfreed: Since a `ServerError` can be thrown here, that implies that we're doing i/o as part of this function call. Generally we like to encapsulate all our i/o in `Operation`s, so we should probably do that here, too. 

If we're not performing I/O, then this error case should be impossible.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -133,4 +133,92 @@ public class MongoCollection<T: Codable> {
...
+                                                             connection: connection,
+                                                             session: session,
+                                                             decoder: self.decoder)
+        }
+    }
+
+    /**
+     * Starts a `ChangeStream` on a collection. Associates the specified `Codable` type `T` with the returned
+     * `ChangeStream`.
+     * - Parameters:
```
patrickfreed: Let's add here (and for all the `withXType` parameters) descriptions of what they're doing. In this case, we could say something like "The type that the change events emitted from the change stream will be decoded to". 

I also think it's worth pointing out in the "Throws" section that for methods in which a user can specify a `Codable` type, a `DecodingError` will be thrown if the decoding doesn't work. e.g. if I pass in `struct { let nonsense: Int }` as the return type, I'll get an error.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+        let opts = try encodeOptions(options: options, session: session)
+        let changeStreamPtr: OpaquePointer = mongoc_client_watch(self._client, pipeline._bson, opts?._bson)
+        return try ChangeStream<ChangeStreamDocument<T>>(stealing: changeStreamPtr,
+                                                         client: self,
+                                                         connection: connection,
+                                                         session: session,
+                                                         decoder: self.decoder)
+     }
+
+     /**
```
patrickfreed: db/client level `watch` is only supported on MongoDB 4.0, so I think we should call that out in the docstrings for the associated methods.

Sources/MongoSwift/MongoClient.swift
ncarbon: Oops, yes, the parameter it was supposed to be labeled `withReturnType`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+     * - Returns: a `ChangeStream` on all collections in all databases in a cluster.
+     * - Throws:
+     *   - `ServerError.commandError` if an error occurs on the server while creating the cursor.
+     *   - `UserError.invalidArgumentError` if the options passed formed an invalid combination or the pipeline passed
+     *     is invalid.
+     *   - `UserError.invalidArgumentError` if the `_id` field is projected out of the change stream documents by the
+     *     pipeline.
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/system-collections/
+     */
+     public func watch(_  pipeline: [Document],
```
kmahar: I think in all these methods, pipeline should have a default value of `[:]` (empty document), so that a user can omit it if they don't want to do anything fancy.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+     * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+     * system collections i.g. "config", "local", and "admin" databases. By default, the type `CollectionType` is
+     * associated with the `fullDocument` field in the `ChangeStreamsDocument`.
+     * - Parameters:
+     *   - Pipeline: The pipeline of stages to append to an initial `ChangeStream` stage.
+     *   - Options: An optional `ChangeStreamOptions` to use on the initial `ChangeStream` stage.
+     *   - Session: An optional `ChangeStream` to use with this change stream.
+     * - Returns: a `ChangeStream` on all collections in all databases in a cluster.
+     * - Throws:
+     *   - `ServerError.commandError` if an error occurs on the server while creating the cursor.
```
kmahar:
```suggestion
     *   - `ServerError.commandError` if an error occurs on the server while creating the change stream.
```

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+     *   - `UserError.invalidArgumentError` if the options passed formed an invalid combination or the pipeline passed
+     *     is invalid.
+     *   - `UserError.invalidArgumentError` if the `_id` field is projected out of the change stream documents by the
+     *     pipeline.
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/system-collections/
```
kmahar: maybe worth also adding a seealso link to all of these docstrings that points to the page on change streams:
https://docs.mongodb.com/manual/changeStreams/
and maybe another linking to allowed aggregation stages for the pipeline:
https://docs.mongodb.com/manual/meta/aggregation-quick-reference/

(applies to all the methods)

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
         return MongoDatabase(name: name, client: self, options: options)
     }
 
+    /**
+     * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+     * system collections i.g. "config", "local", and "admin" databases. By default, the type `CollectionType` is
+     * associated with the `fullDocument` field in the `ChangeStreamsDocument`.
+     * - Parameters:
+     *   - Pipeline: The pipeline of stages to append to an initial `ChangeStream` stage.
```
kmahar: 1. "Pipeline" -> "pipeline" so it matches the capitalization of the parameter
I think mentioning the "initial changestream stage" is a little confusing because users might not necessarily realize this helper is implemented via aggregate. 
it might be clearer to say something like "An array of aggregation pipeline stages to apply to the events returned by the change stream." 

same applies to other places this parameter is used

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+     *   - Options: An optional `ChangeStreamOptions` to use on the initial `ChangeStream` stage.
```
kmahar: 1. "Options" -> "options" so it matches the capitalization of the parameter
2. similarly to my other comment about mentioning the initial stage, I'd suggest phrasing this more like "An optional `ChangeStreamOptions` to use when constructing the `ChangeStream`"

same applies to other places this parameter is used

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
+     }
+
+     /**
+      * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+      * system collections i.g. "config", "local", and "admin" databases. Associates the specified `Codable` type `T`
+      * with the `fullDocument` field in the `ChangeStreamDocument`.
+      * - Parameters:
+      *   - Pipeline: The pipeline of stages to append to an initial `ChangeStream` stage.
+      *   - Options: An optional `ChangeStreamOptions` to use on the initial `ChangeStream` stage.
+      *   - Session: An optional `ChangeStream` to use with this change stream.
```
kmahar: let's also include the `withFullDocumentType` parameter in this list

Sources/MongoSwift/MongoClient.swift
```diff
@@ -278,6 +278,96 @@ public class MongoClient {
...
     }
 
+    /**
+     * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+     * system collections i.g. "config", "local", and "admin" databases. By default, the type `CollectionType` is
+     * associated with the `fullDocument` field in the `ChangeStreamsDocument`.
+     * - Parameters:
+     *   - Pipeline: The pipeline of stages to append to an initial `ChangeStream` stage.
+     *   - Options: An optional `ChangeStreamOptions` to use on the initial `ChangeStream` stage.
+     *   - Session: An optional `ChangeStream` to use with this change stream.
```
kmahar:
```suggestion
     *   - session: An optional `ClientSession` to use with this change stream.
```

Sources/MongoSwift/MongoCollection.swift
kmahar: `watch` does indeed send the initial aggregate command to the server so ya I/O is happening

I think the design of an appropriate operation type that can be used for all the commands is non-trivial (should there be separate ones for client, db, coll? should we combine them in a single op and use an enum?) so I'd be ok with adding a follow on ticket for refactoring the watch calls into an operation rather than doing it right now

Sources/MongoSwift/MongoCollection.swift
ncarbon: Yeah, unless I overlooked, `mongoc_watch` does not call the server only creates the cursor and the command to be sent later.
Removed `ServerError` from the docstrings.

kmahar: I think it does talk to the server. `mongoc_client_watch` calls `_mongoc_change_stream_new_from_client`, which calls `_change_stream_init`, which calls `_make_cursor`, which calls `mongoc_client_read_command_with_opts` and sends a read command. 
see [here](https://github.com/mongodb/mongo-c-driver/blob/2d7d5f20261eb797afed22a17c397c8c91014605/src/libmongoc/src/mongoc/mongoc-change-stream.c#L232)

docstring for `make_cursor` also says "Construct and send the aggregate command and create the resulting cursor."

(you could also verify this with command monitoring)

note this is different from the behavior of methods that return `MongoCursor`s. all of those are lazy and don't send the initial cursor command until you try to read from the cursor.

Sources/MongoSwift/MongoCollection.swift
ncarbon: Yeah, unless I overlooked, it seems like `mongoc_watch` does not call the server, only creates the change stream cursor and prepares the commands to be sent later?
I have removed the `ServerError` from the docstrings for the `watch` methods.

patrickfreed: Leaving it for a follow on ticket sounds good to me. As part of that we'll also need a `ChangeStreamNextOperation` (or modification to the existing `NextOperation`) too.

Sources/MongoSwift/MongoCollection.swift
ncarbon: Oops, I see that now. Also, when I deleted my comment about not seeing the call to the server (realized it was in the wrong place), I had not seen your response yet. 😅

Sources/MongoSwift/MongoCollection.swift
ncarbon: Just created two tix. One for the `watch` operation ([SWIFT-521](https://jira.mongodb.org/browse/SWIFT-521)), and one for `ChangeStreamNextOperation` ([SWIFT-523](https://jira.mongodb.org/browse/SWIFT-523)).

Sources/MongoSwift/MongoClient.swift
patrickfreed: One little nit: the 4.0 warning is best for a `- Note:` at the bottom of the docstring.

Sources/MongoSwift/MongoClient.swift
patrickfreed: @kmahar @mbroadst Any opinions on `withEventType` vs `withReturnType` for the parameter label?

Sources/MongoSwift/MongoClient.swift
ncarbon: Updated. I set its default value to [] though, then just wrap it in a `Document` before sending it to libmongoc, similar to how it's done [here](https://github.com/mongodb/mongo-swift-driver/blob/00fe70e9764d3d9a01a3cbd409470a7ab11db390/Sources/MongoSwift/MongoCollection%2BRead.swift#L59).

Sources/MongoSwift/MongoClient.swift
patrickfreed: Oh, I should have said only on 4.0+ (since 4.2 release is just around the corner).

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -4,38 +4,53 @@ import Nimble
...
+        let res2 = changeStream.next()
+        // expect the change stream to contain a document for changes on a different collection in the same database
+        expect(res2).toNot(beNil())
+        expect(res2?.operationType).to(equal(.insert))
+        expect(res2?.fullDocument?["x"]).to(bsonEqual(123))
+        // expect the resumeToken to be updated to the _id field of the most recently accessed document
+        expect(changeStream.resumeToken).to(equal(res2?._id))
+
+        let db2 = client.db("db2")
+        defer { try? db1.drop() }
```
patrickfreed:
```suggestion
        defer { try? db2.drop() }
```

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -272,6 +272,110 @@ public class MongoDatabase {
...
+      */
+     public func watch<T: Codable>(_ pipeline: [Document] = [],
+                                   options: ChangeStreamOptions? = nil,
+                                   session: ClientSession? = nil,
+                                   withFullDocumentType: T.Type) throws ->
+                                   ChangeStream<ChangeStreamDocument<T>> {
+        let pipeline: Document = ["pipeline": pipeline]
+        let connection = try self._client.connectionPool.checkOut()
+        let opts = try encodeOptions(options: options, session: session)
+        return try self.withMongocDatabase(from: connection) { dbPtr in
```
patrickfreed: instead of duplicating this code, we can call through to `self.watch(pipeline, options: options, session: session, withReturnType: ChangeStreamDocument<T>)`. This also applies to the other `withFullDocumentType` variants

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -4,38 +4,53 @@ import Nimble
 import XCTest
 
 final class ChangeStreamTest: MongoSwiftTestCase {
```
patrickfreed: I think a test verifying the `fullDocument` `Codable` integration works + `CollectionType` interop would be nice to have.

patrickfreed: Looks good. I think one more test verifying that the `CollectionType` is used by default (when watching collections) would also be useful.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -272,6 +272,103 @@ public class MongoDatabase {
...
+                                   options: ChangeStreamOptions? = nil,
+                                   session: ClientSession? = nil,
+                                   withReturnType: T.Type) throws ->
+                                   ChangeStream<T> {
+        let pipeline: Document = ["pipeline": pipeline]
+        let connection = try self._client.connectionPool.checkOut()
+        let opts = try encodeOptions(options: options, session: session)
+        return try self.withMongocDatabase(from: connection) { dbPtr in
+            let changeStreamPtr: OpaquePointer = mongoc_database_watch(dbPtr, pipeline._bson, opts?._bson)
+            return try ChangeStream<T>(stealing: changeStreamPtr,
```
patrickfreed: According to the [Updating the Cached Resume Token](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#updating-the-cached-resume-token) section of the spec, in the cases of `startAfter` or `resumeAfter`, we need to set the cached token to one we use to resume/start after. I think that logic can be put into the initializer of `ChangeStream` so it'll get transitively included here and all other `watch` functions.

On a somewhat related note, the `resumeToken` field of a `ChangeStream` should be optional, since it won't always be set on creation. 

And one final somewhat related note: `init(resumeToken:)` on `ResumeToken` can have the label omitted since it is redundant (i.e. `init(_ resumeToken: Document)`).

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -45,122 +60,124 @@ final class ChangeStreamTest: MongoSwiftTestCase {
...
-            // test that the resumeToken is updated
-            expect(changeStream.resumeToken).to(equal(res3?._id))
+        let changeStream = try coll.watch(options: options)
+
+        // expect the change stream to contain a change document for the `insert` operation
+        try coll.insertOne(["x": 1])
+        let res1 = changeStream.next()
+        expect(res1).toNot(beNil())
+        expect(res1?.operationType).to(equal(.insert))
+        expect(res1?.fullDocument?["x"]).to(bsonEqual(1))
```
ncarbon: @patrickfreed `testChangeStreamOnACollection` creates a watch on a collection using `CollectionType` as default. It looks at the fields and checks they're not nil. Is this what you meant by [your commen](https://github.com/mongodb/mongo-swift-driver/pull/301#discussion_r308414653)t? I can add more assertions for more change events.

Sources/MongoSwift/MongoClient.swift
kmahar: oh yes, i was thinking it's a document, but it's an array 👍

Sources/MongoSwift/MongoClient.swift
kmahar: I like `withEventType`

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -3,12 +3,17 @@ import mongoc
...
         self.resumeToken = resumeToken
     }
 
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self.resumeToken)
+    }
+
     public init(from decoder: Decoder) throws {
         self.resumeToken = try Document(from: decoder)
```
kmahar: nit: while this works, I think it would be better to write it using a `SingleValueDecodingContainer` so it's symmetric to the `encode` method:
```swift
let container = try decoder.singleValueContainer()
self.resumeToken = try container.decode(Document.self)
```

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -137,11 +142,18 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
      *   - `UserError.invalidArgumentError` if the `mongoc_change_stream_t` was created with invalid options.
      */
     internal init(stealing changeStream: OpaquePointer,
+                  options: ChangeStreamOptions? = nil,
                   client: MongoClient,
                   connection: Connection,
                   session: ClientSession? = nil,
                   decoder: BSONDecoder) throws {
-        self.resumeToken = ResumeToken(resumeToken: [])
+        if let options = options {
```
kmahar: nit: these `if` statements can be combined using optional chaining: `if let resumeAfter = options?.resumeAfter`

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -3,12 +3,17 @@ import mongoc
 /// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
 public struct ResumeToken: Codable, Equatable {
-    private let resumeToken: Document
+    public let resumeToken: Document
```
kmahar: why did this become public?

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -133,4 +133,100 @@ public class MongoCollection<T: Codable> {
 
         return try body(collection)
     }
+
+    /**
+     * Starts a `ChangeStream` on a collection. By default, the type `CollectionType` is associated with the
+     * `fullDocument` field in `ChangeStreamDocument` emitted by the returned `ChangeStream`.
```
kmahar:
```suggestion
     * `fullDocument` field in `ChangeStreamDocument`s emitted by the returned `ChangeStream`.
```

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -133,4 +133,100 @@ public class MongoCollection<T: Codable> {
...
+     * - Parameters:
+     *   - pipeline: An array of aggregation pipeline stages to apply to the events returned by the change stream.
+     *   - options: An optional `ChangeStreamOptions` to use when constructing the `ChangeStream`.
```
kmahar: nit: for the other two param descriptions you say "change stream" vs `ChangeStream`, let's be consistent and just say "change stream" for all of them?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -304,6 +304,107 @@ public class MongoClient {
         return MongoDatabase(name: name, client: self, options: options)
     }
 
+     /**
+      * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+      * system collections i.g. "config", "local", and "admin" databases. By default, the type `Document` is
```
kmahar: I think we could omit the "by default, ..." sentence. it implies that this method has some way to override that behavior, whereas really they need to use another method.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -304,6 +304,107 @@ public class MongoClient {
...
+      *   - session: An optional `ClientSession` to use with this change stream.
+      *   - withFullDocumentType: The type that the change events emitted from the change stream will be decoded to.
+      * - Returns: A `ChangeStream` on all collections in all databases in a cluster.
+      * - Throws:
+      *   - `ServerError.commandError` if an error occurs on the server while creating the change stream.
+      *   - `ServerError.commandError` if the pipeline passed is invalid.
+      *   - `UserError.invalidArgumentError` if the options passed formed an invalid combination.
+      *   - `UserError.invalidArgumentError` if the `_id` field is projected out of the change stream documents by the
+      *     pipeline.
+      *   - `DecodingError` if an error occurs while decoding user-defined `withFullDocumentType` `Codable` type.
```
kmahar: would we see the decoding errors now? it seems to me like they wouldn't occur until the user calls `next()` on the change stream.

Sources/MongoSwift/ChangeStream.swift
ncarbon: oops, I made that public while trying to debug. I've made it private now.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -304,6 +304,104 @@ public class MongoClient {
...
+
+     /**
+      * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+      * system collections i.g. "config", "local", and "admin" databases. Associates the specified `Codable` type `T`
+      * with the `fullDocument` field in the `ChangeStreamDocument` emitted by the returned `ChangeStream`.
+      * - Parameters:
+      *   - pipeline: An array of aggregation pipeline stages to apply to the events returned by the change stream.
+      *   - options: An optional `ChangeStreamOptions` to use when constructing the change stream.
+      *   - session: An optional `ClientSession` to use with this change stream.
+      *   - withFullDocumentType: The type that the change events emitted from the change stream will be decoded to.
```
kmahar: i would consider the "change event" to be the whole event. (consider `withEventType` below where `eventType` means the whole evwent)
maybe just phrase this similarly to above, like  "the type that the full document field of the emitted `ChangeStreamDocument`s will be decoded to.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -304,6 +304,104 @@ public class MongoClient {
...
+
+     /**
+      * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+      * system collections i.g. "config", "local", and "admin" databases. Associates the specified `Codable` type `T`
+      * with the returned `ChangeStream`.
+      * - Parameters:
+      *   - pipeline: An array of aggregation pipeline stages to apply to the events returned by the change stream.
+      *   - options: An optional `ChangeStreamOptions` to use when constructing the change stream.
+      *   - session: An optional `ClientSession` to use with this change stream.
+      *   - withEventType: The type that the entire change stream response will be decoded to.
```
kmahar: maybe add on to the end of this something like "and that will be returned when iterating through the change stream" to make it clear that changestream.next() returns this type

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -133,4 +133,97 @@ public class MongoCollection<T: Codable> {
 
         return try body(collection)
     }
+
+    /**
+     * Starts a `ChangeStream` on a collection. By default, the type `CollectionType` is associated with the
```
kmahar: similar to my comment in prev review maybe get rid of the "by default" phrasing and say more like
"The type `CollectionType` will be associated..."

Sources/MongoSwift/MongoClient.swift
kmahar: same applies to this method on DB and collection

Sources/MongoSwift/MongoClient.swift
```diff
@@ -304,6 +304,104 @@ public class MongoClient {
         return MongoDatabase(name: name, client: self, options: options)
     }
 
+     /**
+      * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+      * system collections i.g. "config", "local", and "admin" databases.
```
kmahar: I think this method excludes system collections *and* those databases. system collections can be in any DB and are named like `someDB.system.something`

kmahar: same applies to other docstrings in the file

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -133,4 +133,97 @@ public class MongoCollection<T: Codable> {
...
+     *   - options: An optional `ChangeStreamOptions` to use when constructing the change stream.
+     *   - session: An optional `ClientSession` to use with this change stream.
+     * - Returns: A `ChangeStream` on a specific collection.
+     * - Throws:
+     *   - `ServerError.commandError` if an error occurs on the server while creating the change stream.
+     *   - `ServerError.commandError` if the pipeline passed is invalid.
+     *   - `UserError.invalidArgumentError` if the options passed formed an invalid combination.
+     *   - `UserError.invalidArgumentError` if the `_id` field is projected out of the change stream documents by the
+     *     pipeline.
+     * - SeeAlso:
```
kmahar: maybe include your system collections see also link in all methods since it applies to all of them

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -45,122 +60,193 @@ final class ChangeStreamTest: MongoSwiftTestCase {
...
+        let coll = try db.createCollection(self.getCollectionName(suffix: "1"))
+        let options = ChangeStreamOptions(fullDocument: .updateLookup)
+        let pipeline: [Document] = [["$project": ["_id": 0] as Document]]
+        let changeStream = try coll.watch(pipeline, options: options)
+        try coll.insertOne(["a": 1])
+        expect(try changeStream.nextOrError()).to(throwError())
     }
 
-    func testChangeStreamOnACollection() throws {
+    func testChangeStreamWithWithCodableType() throws {
```
kmahar: withwith -> with

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,8 +3,98 @@ import mongoc
...
+        let db1 = client.db("db1")
+        defer { try? db1.drop() }
+        let coll1 = db1.collection("coll1")
+        let coll2 = db1.collection("coll2")
+        try coll1.insertOne(["a": 1])
+        try coll2.insertOne(["x": 123])
+
+        let res1 = changeStream.next()
+        // expect the change stream to contain a change document for the `insert` operation
+        expect(res1).toNot(beNil())
```
kmahar: there seem to be a lot of cases in these tests where you check the same few fields of a change stream document - consider introducing some kind of helper or helpers where you pass in a change stream document and expected values for a few fields and it does the comparison

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,8 +3,98 @@ import mongoc
...
+        expect(res1?.operationType).to(equal(.insert))
+        expect(res1?.fullDocument?["a"]).to(bsonEqual(1))
```
kmahar: are you just comparing the one field "a" because of `_id`?
if you want, you can choose `_id` yourself and then it will always be predictable. for ex.:
```swift
let doc1: Document = ["a": 1, "_id": 1]
try coll1.insertOne(doc1)
...
expect(res1?.fullDocument).to(equal(doc1))
```

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -14,153 +104,177 @@ final class ChangeStreamTest: MongoSwiftTestCase {
...
             return
         }
 
         let client = try MongoClient()
+        let db = client.db(type(of: self).testDatabase)
+        defer { try? db.drop() }
 
-        guard try client.serverVersion() >= ServerVersion(major: 4, minor: 0) else {
-            print("Skipping test case for server version \(try client.serverVersion())")
+        struct MyType: Codable {
```
kmahar: consider making this type `Equatable`, then below you can just check if `res1.fullDocument` equals the exact instance you insert

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -272,6 +272,103 @@ public class MongoDatabase {
...
+      * - Returns: A `ChangeStream` on all collections in a database.
+      * - Throws:
+      *   - `ServerError.commandError` if an error occurs on the server while creating the change stream.
+      *   - `ServerError.commandError` if the pipeline passed is invalid.
+      *   - `UserError.invalidArgumentError` if the options passed formed an invalid combination.
+      *   - `UserError.invalidArgumentError` if the `_id` field is projected out of the change stream documents by the
+      *     pipeline.
+      * - SeeAlso:
+      *   - https://docs.mongodb.com/manual/changeStreams/
+      *   - https://docs.mongodb.com/manual/meta/aggregation-quick-reference/
```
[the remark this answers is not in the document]
ncarbon: I combined client and database change stream tests since those have the 4.0 server version requirement.

Sources/MongoSwift/MongoClient.swift
kmahar: I'm fine with that, though I wonder if deviating from the spec there is problematic

patrickfreed: Another option (which is what rust is doing) is `ChangeStreamEventDocument`, which deviates slightly less from the spec but is pretty verbose.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -5,20 +5,26 @@ import mongoc
...
-        self.resumeToken = try Document(from: decoder)
+        let container = try decoder.singleValueContainer()
+        self.resumeToken = try container.decode(Document.self)
     }
 }
 
 /// A MongoDB ChangeStream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
 public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
     /// A `ResumeToken` used for manually resuming a change stream.
```
patrickfreed: This may be out of scope for this review, but I think it's important to note here that this resume token is associated with the most recent event seen by this change stream. (i.e. if a user  wanted to pick up where one left off, they'd use this token).

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -38,8 +44,11 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
     /// Used for storing Swift errors.
     private var swiftError: Error?
 
-    /// The error that occurred while iterating the change stream, if one exists. This should be used to check
-    /// for errors after `next()` returns `nil`.
+    /** The error that occurred while iterating the change stream, if one exists. This should be used to check
```
patrickfreed: The `/**` should be on its own line.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -304,6 +304,106 @@ public class MongoClient {
...
+      * Starts a `ChangeStream` on a `MongoClient`. Allows the client to observe all changes in a cluster - excluding
+      * system collections and the "config", "local", and "admin" databases.
+      * - Parameters:
+      *   - pipeline: An array of aggregation pipeline stages to apply to the events returned by the change stream.
+      *   - options: An optional `ChangeStreamOptions` to use when constructing the change stream.
+      *   - session: An optional `ClientSession` to use with this change stream.
+      * - Returns: a `ChangeStream` on all collections in all databases in a cluster.
+      * - Throws:
+      *   - `ServerError.commandError` if an error occurs on the server while creating the change stream.
+      *   - `ServerError.commandError` if the pipeline passed is invalid.
```
patrickfreed: I think this error case can be omitted since it's covered under the "error occurs on the server while creating the change stream".

Sources/MongoSwift/MongoCollection+ChangeStreams.swift
```diff
@@ -0,0 +1,102 @@
+import mongoc
+
+extension MongoCollection {
+    /**
+     * Starts a `ChangeStream` on a collection. The `CollectionType` will be associated with the
+     * `fullDocument` field in `ChangeStreamDocument`s emitted by the returned `ChangeStream`.
+     * Excludes system collections.
```
patrickfreed: What happens if I call `watch` on a system collection?

ncarbon: It throws a `commandError`. Should we explicitly say a `commandError` is thrown when a system collection is used or doesn't the sentence`ServerError.commandError if an error occurs on the server while creating the change stream.`  already cover it?

patrickfreed: Yeah I think it does. Perhaps just rewording "Excludes ..." to "The server will return an error if this is called on a system collection." would be more informative.

Sources/MongoSwift/MongoClient.swift
kmahar: hmm... ya i think i'd maybe just go with `ChangeStreamEvent` over that

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -14,153 +108,232 @@ final class ChangeStreamTest: MongoSwiftTestCase {
...
+        let resumeAfter = changeStream1.resumeToken
+        let changeStream2 = try coll.watch(options: ChangeStreamOptions(resumeAfter: resumeAfter))
+        // expect this change stream to have its resumeToken set to the resumeAfter
+        expect(changeStream2.resumeToken).to(equal(resumeAfter))
+        expect(changeStream2.error).to(beNil())
+    }
 
-        guard try client.serverVersion() >= ServerVersion(major: 4, minor: 0) else {
-            print("Skipping test case for server version \(try client.serverVersion())")
+    func testChangeStreamResumeTokenError() throws {
```
kmahar: this name doesn't seem to match what the test is checking.

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -14,153 +108,232 @@ final class ChangeStreamTest: MongoSwiftTestCase {
...
+        expect(clientChangeStream.error).to(beNil())
+        expect(clientChange?.operation).to(equal("insert"))
+        expect(clientChange?.fullDocument).to(equal(expectedFullDocument))
+        expect(dbChangeStream).toNot(beNil())
+        expect(dbChangeStream.error).to(beNil())
+        expect(dbChange?.operation).to(equal("insert"))
+        expect(dbChange?.fullDocument).to(equal(expectedFullDocument))
+    }
+
+    func testChangeStreamOnWithFullDocumentType() throws {
```
kmahar: is there an extra "on" in this test name?

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -14,153 +108,232 @@ final class ChangeStreamTest: MongoSwiftTestCase {
...
+
+        let doc1: Document = ["_id": 1, "x": 1, "y": 2]
+
+        let client = try MongoClient()
+        let db = client.db(type(of: self).testDatabase)
+        defer { try? db.drop() }
+
+        let coll = try db.createCollection(self.getCollectionName(suffix: "1"))
+
+        // test that the change stream works on a collection when using withFullDocumentType
```
kmahar: it doesn't look like we are actually using `withFullDocumentType` here?

Sources/MongoSwift/ChangeStream.swift
ncarbon: @patrickfreed So should `ResumeToken`'s docstring be updated or is this for its private field `resumeToken`?

Tests/MongoSwiftTests/ChangeStreamTests.swift
ncarbon: I thought this error was related to the resumeToken. I will change it to `testChangeStreamProjectOutIdError` because I see now how it will describe the test better.

Tests/MongoSwiftTests/ChangeStreamTests.swift
ncarbon: fixed typo

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -273,15 +273,14 @@ public class MongoDatabase {
     }
 
      /**
-      * Starts a `ChangeStream` on a database. Excludes system collections.
+      * Starts a `ChangeStream` on a database. The server will return an error if this is called on a system collection.
```
kmahar: I think this should go back to the original text about excluding system collections - since a change stream is opened on an entire DB it would be impossible for this method to be used to open a change stream on a system collection.
same applies to other 2 DB watch methods

Tests/MongoSwiftTests/ChangeStreamTests.swift
kmahar: ahhh, i see. you're right, it does relate to the resume token. 👍to more descriptive name though!

Sources/MongoSwift/ChangeStream.swift
patrickfreed: Just this field. I was saying that the docstring for this field should provide some context as to where it came from.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -3,22 +3,29 @@ import mongoc
 /// `ChangeStreamOptions` to resume or start a change stream where a previous one left off.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream
 public struct ResumeToken: Codable, Equatable {
+    /// A `Document` associated with the most recent event seen by this change stream.
```
patrickfreed: This docstring doesn't actually need to be added ("most recent event seen by this change stream" doesn't apply outside the context of a change stream, and this docstring is on the resume token struct which isn't tied to a stream).

Sources/MongoSwift/MongoDatabase.swift
patrickfreed: Sorry to ping-pong on this, but I think Kaitlin just meant that database `watch` methods should go back to the old wording. I think for collections it still makes sense to have the error warning.

ncarbon: oops, I shouldn't have used find+replace 😅. Will fix.

PR #302, opened 2019-08-01
SWIFT-375 Make MongoClient threadsafe, and greatly simplify monitoring API

kmahar:
putting this up as a PR against the SWIFT-374 branch so I don't have to wait for SWIFT-374 to get merged to get feedback. 

This ticket was technically to make `MongoClient`, `MongoDatabase`, and `MongoCollection` threadsafe.
The only mutable state any of them had was a couple properties related to APM on `MongoClient`, which were mutable because we allow you to toggle monitoring on and off. I don't really recall why we did it this way, but it seems unnecessary to support that. I feel like generally you just want it on or you want it off and don't need to switch between the two states.
This also required us to up-front set callbacks for every single type of event as we didn't know what event types you would later decide to enable.

So rather than adding code to synchronize reading/writing those properties across threads, I just decided to make it so you turn on monitoring at the start and it stays on.

Also, previously the API for command and SDAM monitoring required you to take two steps to enable monitoring:
1. create a `MongoClient` with `initializeMonitoring: true`
2. call `MongoClient.enableMonitoring` and pass in the type of events you want to monitor

Now, when creating the client you just pass in `commandMonitoring: true` and/or `serverMonitoring: true`, and possibly a custom `NotificationCenter` to publish to.

The code for setting callbacks now also lives in `ConnectionPool` rather than in `MongoClient`.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -1,12 +1,8 @@
 import Foundation
 import mongoc
 
-/// A protocol for monitoring events to implement, specifying their type and name.
+/// A protocol for monitoring events to implement, specifying their name.
 public protocol MongoEvent {
-    /// The MongoEventType corresponding to the event. This event will be posted if
-    /// monitoring is enabled for its eventType.
-    static var eventType: MongoEventType { get }
-
     /// The name this event will be posted under.
     static var eventName: Notification.Name { get }
 }
@@ -21,9 +17,6 @@ private protocol InitializableFromOpaquePointer {
 /// An event published when a command starts. The event is stored under the key `event`
 /// in the `userInfo` property of `Notification`s posted under the name .commandStarted.
 public struct CommandStartedEvent: MongoEvent, InitializableFromOpaquePointer {
-    /// The type of this event.
-    public static var eventType: MongoEventType { return .commandMonitoring }
-
     /// The name this event will be posted under.
     public static var eventName: Notification.Name { return .commandStarted }
 
@@ -61,9 +54,6 @@ public struct CommandStartedEvent: MongoEvent, InitializableFromOpaquePointer {
 /// An event published when a command succeeds. The event is stored under the key `event`
 /// in the `userInfo` property of `Notification`s posted under the name .commandSucceeded.
 public struct CommandSucceededEvent: MongoEvent, InitializableFromOpaquePointer {
-    /// The type of this event.
-    public static var eventType: MongoEventType { return .commandMonitoring }
-
     /// The name this event will be posted under.
     public static var eventName: Notification.Name { return .commandSucceeded }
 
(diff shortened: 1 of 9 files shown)


Sources/MongoSwift/APM.swift
```diff
@@ -529,56 +525,42 @@ public enum MongoEventType {
 }
```
patrickfreed: The above enum can be removed now, right?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -6,9 +6,15 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
...
     public var retryWrites: Bool?
 
-    /// Indicates whether this client should be set up to enable monitoring command and server discovery and monitoring
-    /// events.
-    public var eventMonitoring: Bool
+    /// Indicates whether this client should publish command monitoring events.
+    public var commandMonitoring: Bool = false
+
+    /// Indicates whether this client should publish server monitoring events.
+    public var serverMonitoring: Bool = false
```
patrickfreed: I think it would be useful to document which events are covered by which flag, similar to how it's documented on the `MongoEventType` enum.

kmahar: done

Sources/MongoSwift/APM.swift
kmahar: good point, removed

Sources/MongoSwift/MongoClient.swift
```diff
@@ -173,8 +198,11 @@ public class MongoClient {
         self.readPreference = options.readPreference ?? ReadPreference(.primary)
         self.encoder = BSONEncoder(options: options)
         self.decoder = BSONDecoder(options: options)
+        self.notificationCenter = options.notificationCenter ?? NotificationCenter.default
```
mbroadst: I wonder, if we're making more dramatic changes, whether we should get rid of storing a notification center at all. If we instead supported passing in lambdas, or some sort of protocol for callbacks, then the user could decide to use whatever means of notification they wanted (a notification center, some sort of sswg log facility, writing to a file, etc).

kmahar: yeah, we made a ticket about this a while back: [SWIFT-360](https://jira.mongodb.org/browse/SWIFT-360)

I definitely agree we should change this API to give the user a bit more flexibility about what happens. technically now they still have it, because the observer they register on the notification center can do anything they want it to. but we may as well just let them register a callback with us directly.
I think this should end up being kind of symmetric to the logging API SSWG introduced, and we should think about bringing in autoclosures so that if e.g. a user only cares about `CommandStartedEvent`s we don't waste time filling out structs for the other command-related events which will just get thrown away. (this may be tricky with the mongoc event object lifetimes though...)

I think this kind of needs a mini-design... it would be nice if we can get it into the same release as these changes, but I think we could switch to the callbacks in a non-breaking way in the future if we want - we could just default to still publishing to the notification center but allow you to override

kmahar: per conversation in triage today with @patrickfreed and @rathisekaran I'm going to merge this PR in as-is. I am marking SWIFT-360 with the next fix version so we remember to revisit it before releasing these new changes.

PR #303, opened 2019-08-05
SWIFT-403 Add an Example project using Perfect

kmahar:
This adds a new very basic example project that uses Perfect, and updates the Vapor and Kitura examples to better match up with it.

Only kinda weird thing about this is that I could see someone potentially coming to our repo between when this gets merged and when we release pooling, and not having access to the threadsafe types yet 🤔 so maybe I'll wait to actually merge this until just before the release.

### DIFF
--- a/Examples/Kitura/Package.swift
+++ b/Examples/Kitura/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:4.2
 import PackageDescription
 
 let package = Package(
--- a/Examples/Vapor/Package.swift
+++ b/Examples/Vapor/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:4.2
 import PackageDescription
 
 let package = Package(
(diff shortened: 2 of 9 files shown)


Examples/Kitura/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:4.2
```
patrickfreed: is this downgrade necessary?

Examples/Vapor/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:4.2
```
patrickfreed: ditto q on downgrade

Examples/Kitura/Package.swift
kmahar: not necessary, but I thought it better communicated that we support 4.2+. I imagined that someone might copy these `Package.swift` files and think they had to use Swift 5.0 based on this

PR #306, opened 2019-08-07
Swift-452 Test runner for Change Streams

ncarbon:
This is the test runner for change streams. This does not include the APM assertions.


Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,7 +3,155 @@ import mongoc
...
+
+                guard version >= test.minServerVersion else {
+                    print("Skipping tests from file, minimum required server version not met.")
+                    continue
+                }
+                print("Executing test: \(test.description)")
+
+                var changeStream: ChangeStream<ChangeStreamTestEventDocument>?
+                do {
+                    changeStream = try createChangeStream(client: client, testFile: testFile, test: test)
```
patrickfreed: I think the creation of the change stream can be deferred to the `run` method on `ChangeStreamTest` since it really is part of running the individual test.

I'd update the signature of `ChangeStreamTest::run` to `func run(globalClient: MongoClient, dbName: String, collName: String) throws ...`, which is similar to `SpecTest::run` and should provide the necessary fields to run the test.

Tests/MongoSwiftTests/SpecTestRunner/ChangeStreamSpecTest.swift
```diff
@@ -0,0 +1,136 @@
+import mongoc
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+internal struct ChangeStreamAnyTestOperation: Decodable {
+    let anyTestOperation: AnyTestOperation
+
+    let database: String?
```
patrickfreed: I believe these can be non-optional

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,7 +3,155 @@ import mongoc
...
+                    seenError = error
+                }
+
+                if let failPoint = test.failPoint {
+                    try self.activateFailPoint(failPoint)
+                }
+
+                try test.run(client: globalClient, seenError: seenError, changeStream: changeStream)
+
+                // TODO: add APM assertions
```
patrickfreed: this TODO can be moved into the body of `run`. Also, can you file a ticket for the follow on work and add it to the TODO?

Tests/MongoSwiftTests/SpecTestRunner/ChangeStreamSpecTest.swift
```diff
@@ -0,0 +1,136 @@
...
+            guard let collName = operation.collection else {
+                print("The collection name for running the operation is missing.")
+                return
+            }
+
+            let db = client.db(dbName)
+            let coll = db.collection(collName)
+
+            do {
+                try operation.anyTestOperation.op.execute(client: client,
```
patrickfreed: probably a little cleaner to define an `execute` method on `ChangeStreamOperation` that calls through to the underlying operation. Then you can make the fields of `ChangeStreamOperation` private.

Tests/MongoSwiftTests/SpecTestRunner/ChangeStreamSpecTest.swift
```diff
@@ -0,0 +1,136 @@
...
+            do {
+                try changeStream?.nextOrError()
+            } catch {
+                seenError = error
+            }
+        }
+
+        if seenError != nil {
+            // assert errors match expected errors
+            assertError(seenError: seenError)
```
patrickfreed: this conditional can be made into an `if let` and the signature of `assertError` updated to take in a non-optional.
for the success branch, you can just pass in `changeStream!` since you know it will be non-nil if the error is non-nil.

Tests/MongoSwiftTests/SpecTestRunner/ChangeStreamSpecTest.swift
```diff
@@ -0,0 +1,136 @@
...
+        // TODO: assert that the error matches expected error
+    }
+
+    private func assertSuccess(changeStream: ChangeStream<ChangeStreamTestEventDocument>?) {
+        // assert that no error was expected for the test
+        expect(self.result.error).to(beNil())
+        // assert that the changes received from changeStream MATCH the results in result.success
+        if let changeStream = changeStream, let expectedChanges = self.result.success {
+            for expectedChange in expectedChanges {
+                if let change = changeStream.next() {
```
patrickfreed: we want to fail the test if this `if let` doesn't succeed, so maybe rewrite it as:
```swift
let change = changeStream.next()
expect(changeStream.error).to(beNil(), description: self.description)
expect(change).toNot(beNil(), description: self.description)
expect(change!).to(equal(expectedChange), description: self.description)
```

As another note: all the `expect` calls in this test should have `self.description` as the description so that we know which individual test failed when looking at failed assertions.

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,7 +3,155 @@ import mongoc
...
+/// Struct representing a single test within a spec test JSON file.
+internal struct ChangeStreamTest: Decodable, ChangeStreamSpecTest {
+    let description: String
+    let minServerVersion: ServerVersion
+
+    /// The configureFailPoint command document to run to configure a fail point on the primary server.
+    let failPoint: FailPoint?
+
+    /// The entity on which to run the change stream.
+    let target: String
```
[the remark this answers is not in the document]
ncarbon: It'll still need the `ChangeStreamTest` to grab the database and collection names,  `changeStreamPipeline`, and  `changeStreamOptions`.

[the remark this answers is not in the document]
ncarbon: I'll do that

patrickfreed:
@kmahar this should be ready for a look now

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,6 +3,274 @@ import mongoc
...
+            let db1 = globalClient.db(testFile.databaseName)
+            let db2 = globalClient.db(testFile.database2Name)
+            print("\n------------\nExecuting tests from file \(testFilesPath)/\(testFile.name)...\n")
+            for test in testFile.tests {
+                try db1.drop()
+                try db2.drop()
+                _ = try db1.createCollection(testFile.collectionName)
+                _ = try db2.createCollection(testFile.collection2Name)
+
+                let testTopologies = test.topology.map { TopologyDescription.TopologyType(from: $0) }
```
kmahar: should we do the topology/version checks before we bother dropping DBs and creating collections?

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,6 +3,274 @@ import mongoc
...
+            return testFile
+        }
+
+        let globalClient = try MongoClient(MongoSwiftTestCase.connStr)
+
+        let version = try globalClient.serverVersion()
+        let topology = MongoSwiftTestCase.topologyType
+
+        for testFile in tests {
+            let db1 = globalClient.db(testFile.databaseName)
```
kmahar: will we ever drop these DBs after the last test case completes?

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,6 +3,274 @@ import mongoc
...
+
+    /// An array of tests that are to be run independently of each other.
+    let tests: [ChangeStreamTest]
+}
+
+final class ChangeStreamSpecTests: MongoSwiftTestCase, FailPointConfigured {
+    var activeFailPoint: FailPoint?
+
+    override func tearDown() {
+        self.disableActiveFailPoint()
```
kmahar: it might be helpful to clarify why we need this on top of the deferred one in a comment (I think the answer is that if the test fails the defer won't be reached...?)

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -3,6 +3,274 @@ import mongoc
...
+
+internal struct ChangeStreamTestEvent: Codable, Equatable {
+    let operationType: String
+
+    let ns: MongoNamespace?
+
+    let fullDocument: Document?
+
+    public static func == (lhs: ChangeStreamTestEvent, rhs: ChangeStreamTestEvent) -> Bool {
+        let lhsFullDoc = lhs.fullDocument?.filter { $0.key != "_id" }
```
kmahar: do we care about making sure _id is present or no?

Tests/LinuxMain.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.15.0 — https://github.com/krzysztofzablocki/Sourcery
```
kmahar: think you need to update your sourcery so this doesn't break travis

PR #310, opened 2019-08-19
SWIFT-531 Test against SSL variants on Evergreen

kmahar:
This PR adds SSL versions of most of our test variants and does some internal refactoring in the tests to correctly set up SSL when running against those variants.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -416,7 +416,7 @@ buildvariants:
      - ".3.6"
   rules:
   # pre 4.0 we are just using legacy linux mongoDB since ubuntu 18.04 didn't
-  # exist when 3.6 came out. it legacy doesn't link to OpenSSL so we can't run
+  # exist when 3.6 came out. legacy doesn't link to OpenSSL so we can't run
   # SSL tests there.
   - if:
       os-fully-featured: "ubuntu-18.04"
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -103,6 +103,7 @@ functions:
           ${PREPARE_SHELL}
           MONGODB_VERSION=${MONGODB_VERSION} \
             TOPOLOGY=${TOPOLOGY} \
+            SSL=${SSL} \
             sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
     # run-orchestration generates expansion file with the MONGODB_URI for the cluster
     - command: expansions.update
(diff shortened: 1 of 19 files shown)


.evergreen/install-dependencies.sh
```diff
@@ -13,7 +13,7 @@ export SWIFTENV_ROOT="${INSTALL_DIR}/swiftenv"
 export PATH=/opt/cmake/bin:${SWIFTENV_ROOT}/bin:$PATH
 
 # should be set by EVG eventuallty
-LIBMONGOC_VERSION="r1.13"
+LIBMONGOC_VERSION="r1.14"
```
kmahar: 1.13 does not work with OpenSSL 1.1.1 which is what is on the Ubuntu 18.04 evg hosts. I guess it's maybe not ideal to not be testing against our min required version, but we do still use 1.13 on Travis anyway, and are going to drop 1.13 support pretty soon, so I'm not too concerned.

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -19,6 +19,12 @@ final class MongoClientTests: MongoSwiftTestCase {
             _ = try client.db($0).createCollection("c")
         }
 
+        defer {
```
kmahar: these were not getting cleaned up properly

Tests/MongoSwiftTests/MongoCollection+IndexTests.swift
```diff
@@ -2,6 +2,8 @@
 import Nimble
 import XCTest
 
+private var _client: MongoClient?
```
kmahar: we were inadvertently using the shared client defined in `MongoCollectionTests.swift` which seemed a little weird

.evergreen/config.yml
```diff
@@ -424,6 +424,14 @@ buildvariants:
       ssl: "ssl"
     then:
       remove_tasks: ".3.6"
+  # workaround for CDRIVER-3318 preventing us from doing any version checks
```
kmahar: both `buildInfo` and `serverStatus` seem to hang indefinitely on SSL sharded clusters on MacOS so we have no way to do a version check. this makes running a lot of the tests impossible so I've just removed these variants for now.

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -245,6 +245,12 @@ final class ChangeStreamSpecTests: MongoSwiftTestCase, FailPointConfigured {
     }
 
     func testChangeStreamSpec() throws {
+        // TODO SWIFT-539: unskip
```
kmahar: in the situations where we did actually manage to get the server version and run these (on repl sets), this test just ended up hanging and timing out later on. so also skipping for now.

PR #312, opened 2019-08-21
SWIFT-173 Add initial DNS seedlist discovery test runner

kmahar:
This adds the test runner for the DNS seedlist discovery tests.

### DIFF
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -98,7 +98,7 @@ internal class ConnectionPool {
     }
 
     /// Temporary API for testing purposes. This will likely change in SWIFT-471.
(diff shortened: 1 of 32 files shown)


Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -98,7 +98,7 @@ internal class ConnectionPool {
     }
 
     /// Temporary API for testing purposes. This will likely change in SWIFT-471.
-    internal func setSSLOpts(caFile: String?, pemFile: String?) throws {
+    internal func setSSLOpts(caFile: String?, pemFile: String?, allowInvalidHostnames: Bool? = nil) throws {
```
kmahar: adding the capability to set this was necessary because the hosts in the test URIs do not match up with the MO replica sets.

PR #317, opened 2019-08-29
SWIFT-607 Implement collection enumeration spec

kaywux:
Implement the collection enumeration spec using listCollections, listCollectionNames, and ListMongoCollections and add relevant tests.

JIRA ticket: https://jira.mongodb.org/browse/SWIFT-607

Evergreen: https://evergreen.mongodb.com/version/5d6843727742ae4163f0cdc2

### DIFF
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -2,6 +2,8 @@ import Foundation
 
 /// `BSONDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
 public class BSONDecoder {
+    // swiftlint:disable explicit_acl
+    // some kind of swiftlint bug here.
     @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
     internal static var iso8601Formatter: ISO8601DateFormatter = {
         let formatter = ISO8601DateFormatter()
--- a/Sources/MongoSwift/BSON/BSONDecoder.swift
+++ b/Sources/MongoSwift/BSON/BSONDecoder.swift
@@ -2,8 +2,6 @@ import Foundation
 
 /// `BSONDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
 public class BSONDecoder {
-    // swiftlint:disable explicit_acl
-    // some kind of swiftlint bug here.
     @available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)
     internal static var iso8601Formatter: ISO8601DateFormatter = {
         let formatter = ISO8601DateFormatter()
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -1,16 +1,5 @@
 import mongoc
 
-/// Options to use when executing a `listCollections` command on a `MongoDatabase`.
-public struct ListCollectionsOptions: Encodable {
-    /// The batchSize for the returned cursor.
-    public var batchSize: Int?
-
-    /// Convenience initializer allowing any/all parameters to be omitted or optional
-    public init(batchSize: Int? = nil) {
-        self.batchSize = batchSize
-    }
-}
-
 /// Options to set on a retrieved `MongoCollection`.
 public struct CollectionOptions: CodingStrategyProvider {
     /// A read concern to set on the returned collection. If one is not specified, the collection will inherit the
@@ -223,7 +212,7 @@ public struct MongoDatabase {
      *   - options: Optional `ListCollectionsOptions` to use when executing this command
      *   - session: Optional `ClientSession` to use when executing this command
      *
-     * - Returns: a `MongoCursor` over an array of collections
+     * - Returns: a `MongoCursor` over an array of `CollectionSpecification`s
      *
      * - Throws:
      *   - `userError.invalidArgumentError` if the options passed are an invalid combination.
@@ -232,14 +221,7 @@ public struct MongoDatabase {
     public func listCollections(_ filter: Document? = nil,
                                 options: ListCollectionsOptions? = nil,
                                 session: ClientSession? = nil) throws -> MongoCursor<CollectionSpecification> {
-        var opts = try encodeOptions(options: options, session: session)
-        if let filterDoc = filter {
-            opts = opts ?? Document()
-            // swiftlint:disable:next force_unwrapping
-            opts!["filter"] = filterDoc // guaranteed safe because of nil coalescing default.
-        }
-
-        let operation = ListCollectionsOperation(database: self, options: opts, nameOnly: false)
+        let operation = ListCollectionsOperation(database: self, filter: filter, options: options, nameOnly: false)
         guard case let .specs(result) = try self._client.executeOperation(operation, session: session) else {
             throw RuntimeError.internalError(message: "Invalid result")
         }
@@ -247,7 +229,7 @@ public struct MongoDatabase {
     }
 
     /**
-     * Gets a list of `MongoCollection`s.
+     * Gets a list of `MongoCollection`s in this database.
      *
      * - Parameters:
      *   - filter: a `Document`, optional criteria to filter results by
@@ -267,7 +249,7 @@ public struct MongoDatabase {
     }
 
     /**
-     * Gets a list of names of collections.
+     * Gets a list of names of collections in this database.
      *
      * - Parameters:
      *   - filter: a `Document`, optional criteria to filter results by
@@ -283,14 +265,7 @@ public struct MongoDatabase {
     public func listCollectionNames(_ filter: Document? = nil,
                                     options: ListCollectionsOptions? = nil,
                                     session: ClientSession? = nil) throws -> [String] {
-        var opts = try encodeOptions(options: options, session: session)
-        if let filterDoc = filter {
-            opts = opts ?? Document()
-            // swiftlint:disable:next force_unwrapping
-            opts!["filter"] = filterDoc // guaranteed safe because of nil coalescing default.
-        }
-
-        let operation = ListCollectionsOperation(database: self, options: opts, nameOnly: true)
+        let operation = ListCollectionsOperation(database: self, filter: filter, options: options, nameOnly: true)
         guard case let .names(result) = try self._client.executeOperation(operation, session: session) else {
             throw RuntimeError.internalError(message: "Invalid result")
         }
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -252,8 +252,7 @@ public struct MongoDatabase {
      * Gets a list of names of collections in this database.
      *
      * - Parameters:
-     *   - filter: a `Document`, optional criteria to filter results by. For `listConnectionNames` the filter expression
-     *             can only filter based on a collection’s name.
+     *   - filter: a `Document`, optional criteria to filter results by
      *   - options: Optional `ListCollectionsOptions` to use when executing this command
      *   - session: Optional `ClientSession` to use when executing this command
      *
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -56,15 +56,6 @@ public struct CollectionOptions: CodingStrategyProvider {
     }
 }
 
-/// Specifications of a collection returned when executing `listCollections`.
-public struct CollectionSpecification: Codable {
-    /// The name of the collection.
-    public var name: String
-
-    /// Options of the collection.
-    public var collectionOptions: CreateCollectionOptions?
-}
-
 /// Options to use when executing dropDatabase command.
 public struct DropDatabaseOptions: Codable {
     /// An optional `WriteConcern` to use for the command.
@@ -292,7 +283,12 @@ public struct MongoDatabase {
     public func listCollectionNames(_ filter: Document? = nil,
                                     options: ListCollectionsOptions? = nil,
                                     session: ClientSession? = nil) throws -> [String] {
-        let opts = try encodeOptions(options: options, session: session)
+        var opts = try encodeOptions(options: options, session: session)
+        if let filterDoc = filter {
+            opts = opts ?? Document()
+            // swiftlint:disable:next force_unwrapping
+            opts!["filter"] = filterDoc // guaranteed safe because of nil coalescing default.
+        }
 
         let operation = ListCollectionsOperation(database: self, options: opts, nameOnly: true)
         guard case let .names(result) = try self._client.executeOperation(operation, session: session) else {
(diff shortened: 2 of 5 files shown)


Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,56 @@
+import mongoc
+
+/// Specifications of a collection returned when executing `listCollections`.
+public struct CollectionSpecification: Codable {
+    /// The name of the collection.
+    public var name: String
+
+    /// Type returned.
```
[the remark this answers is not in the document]
patrickfreed: Yeah, I think using an enum here is a great idea, and including the `other(String)` case as we've done elsewhere seems appropriate.

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,110 @@
+import mongoc
+
+/// Describes the type of data store returned when executing `listCollections`
```
kmahar:
```suggestion
/// Describes the type of data store returned when executing `listCollections`.
```

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/// An operation corresponding to a "listCollections" command on a database.
+internal struct ListCollectionsOperation: Operation {
+    private let database: MongoDatabase
+    private let filter: Document?
+    private let options: ListCollectionsOptions?
+    private let nameOnly: Bool?
+
+    internal init(database: MongoDatabase, filter: Document? = nil, options: ListCollectionsOptions?, nameOnly: Bool?) {
```
kmahar: nit: does `filter` need a default value? I think we are always passing it in whenever we call this initializer

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -156,3 +182,11 @@ extension CreateCollectionOptions: Equatable {
                // locale matches.
     }
 }
+
+extension CollectionSpecification: Equatable {
+    public static func == (lhs: CollectionSpecification, rhs: CollectionSpecification) -> Bool {
```
kmahar: if all of the properties stored in a type are `Equatable`, then the compiler will synthesize conformance to`Equatable ` for the type - basically it will provide a default implementation that just compares all the properties for equality. 

Since what you have written here is equivalent to the synthesized implementation, you should just be able to do
`extension CollectionSpecification: Equatable {}` and leave out the definition of `==`. 

[Here](https://swiftunboxed.com/internals/synthesized-equatable-conformance/) is a blog post about the synthesized conformance. The compiler also generates `Codable` conformance for you automatically in a lot of cases, too, like for your new `CollectionSpecification` type.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -112,26 +112,52 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
-        var collectionInfo = try db.listCollections().map { try decoder.decode(CollectionInfo.self, from: $0) }
+
+        var collectionInfo = try Array(db.listCollections()) as [CollectionSpecification]
         collectionInfo.sort { $0.name < $1.name }
 
         expect(collectionInfo).to(haveCount(3))
 
-        let expectedFoo = CollectionInfo(name: "foo", type: "collection", options: fooOptions)
+        let expectedFoo = CollectionSpecification(name: "foo",
+                                                  type: CollectionType(rawValue: "collection"),
```
kmahar: you should be able to initialize a new `CollectionType` here by just saying `.collection`. based on the context the  compiler knows the type is `CollectionType` so it will look for a matching enum case there. same applies a couple lines down with `.view`.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
kaywux: When I remove the == definition, I get the following error 

> type 'CollectionSpecification' does not conform to protocol 'Equatable'

kmahar: we talked about this offline. it turns out you only get the synthesized implementation of the method if you make the type `Equatable` in the same module that it is defined in. we cannot do that, since we need our custom impl of `CreateCollectionOptions` to live in the tests only. therefore we need to hand-write the `Equatable` implementation here.

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,110 @@
...
+                }
+                return collections
+            }
+        }
+        if self.nameOnly ?? false {
+            let cursor: MongoCursor<Document> = try MongoCursor(client: self.database._client,
+                                                                decoder: self.database.decoder,
+                                                                session: session,
+                                                                initializer: initializer)
+            return .names(cursor.map { $0["name"] as? String ?? "" })
```
kmahar: we should maybe throw an internal error here if the server response is missing a `name` field. that would be pretty unexpected. (note that in the non name only case, we would also throw if it were missing, because the decoder would look for a name in each doc and not find it.)

kmahar: [here](https://github.com/mongodb/mongo-swift-driver/blob/5850eafc2d008c5bff0508af2506ec59f01cfd04/Sources/MongoSwift/Operations/ListDatabasesOperation.swift#L71) is an example of such an error

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -99,7 +99,11 @@ internal struct ListCollectionsOperation: Operation {
                                                                 decoder: self.database.decoder,
                                                                 session: session,
                                                                 initializer: initializer)
-            return .names(cursor.map { $0["name"] as? String ?? "" })
+            return try .names(cursor.map { guard let name = $0["name"] as? String else {
+                throw RuntimeError.internalError(message: "Invalid server response: collection has no name")
```
kmahar: nit: the indentation is a little weird here in that the `throw` is lined up with the `}` in the line below. I think maybe we should indent this `throw...` line in one more.

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,114 @@
...
+        case "view":
+            self = .view
+        default:
+            self = .other(rawValue)
+        }
+    }
+}
+
+/// Specifications of a collection returned when executing `listCollections`.
+public struct CollectionSpecification: Codable {
```
patrickfreed: This struct is missing a few fields that the server returns back, namely `info` and `idIndex`. See the [official docs](https://docs.mongodb.com/manual/reference/command/listCollections/#listCollections.cursor) for more info on them. On a related note, we can add a `seeAlso` with that link here in the docstring for this struct. You can take a look at examples in [`ChangeStream.swift`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/MongoCollection%2BChangeStreams.swift) for what the formatting of that is.

For `info`, instead of using `Document`, we can define another struct (e.g. `CollectionSpecificationInfo`) that has the required fields and conforms to `Codable`.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -223,30 +212,64 @@ public class MongoDatabase {
...
+     * - Returns: An array of `MongoCollection`s that match the provided filter.
+     *
+     * - Throws:
+     *   - `userError.invalidArgumentError` if the options passed are an invalid combination.
+     *   - `UserError.logicError` if the provided session is inactive.
+     */
+    public func listMongoCollections(_ filter: Document? = nil,
+                                     options: ListCollectionsOptions? = nil,
+                                     session: ClientSession? = nil) throws -> [MongoCollection<Document>] {
+        return try self.listCollectionNames(filter, options: options, session: session).map { self.collection($0) }
```
patrickfreed: From the [section on "returning a list of collection objects"](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#returning-a-list-of-collection-objects): 
>Drivers MUST specify the nameOnly option when executing the listCollections command for this method, except drivers MUST NOT set nameOnly if a filter specifies any keys other than name.

So it looks like we need to inspect the `filter` (if there is one) to see what we should set `nameOnly` to. The reason being that there are big performance differences to using `nameOnly: true` vs `nameOnly: false`, and we want to take advantage of them where possible.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -223,30 +212,64 @@ public class MongoDatabase {
...
+     *
+     * - Returns: An array of `MongoCollection`s that match the provided filter.
+     *
+     * - Throws:
+     *   - `userError.invalidArgumentError` if the options passed are an invalid combination.
+     *   - `UserError.logicError` if the provided session is inactive.
+     */
+    public func listMongoCollections(_ filter: Document? = nil,
+                                     options: ListCollectionsOptions? = nil,
+                                     session: ClientSession? = nil) throws -> [MongoCollection<Document>] {
```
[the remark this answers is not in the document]
kmahar: hmm, maybe, though unless if all your collections are structured the same way it might not be that useful. 
we could maybe add another variant that takes in a type parameter but I think we should leave this one as it is, for symmetry with the `collection` methods on DB.

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,146 @@
...
+ * Info about the collection that is returned with a `listCollections` call.
+ *
+ * - SeeAlso:
+ *   - https://docs.mongodb.com/manual/reference/command/listCollections/#listCollections.cursor
+ */
+public struct CollectionSpecificationInfo: Codable {
+    /// Tells whether or not the data store is read only.
+    public let readOnly: Bool
+
+    /// The collection's UUID - once established, this does not change. The UUID remains the same
```
kmahar: under what conditions is is this value `nil`? let's document it.

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Info about the collection that is returned with a `listCollections` call.
+ *
+ * - SeeAlso:
+ *   - https://docs.mongodb.com/manual/reference/command/listCollections/#listCollections.cursor
+ */
+public struct CollectionSpecificationInfo: Codable {
+    /// Tells whether or not the data store is read only.
```
kmahar: nit: `Tells` -> `Indicates`

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,146 @@
...
+            // swiftlint:disable:next force_unwrapping
+            opts!["filter"] = filterDoc // Guaranteed safe because of nil coalescing default.
+
+            let keys = filterDoc.keys.filter { $0 != "name" }
+            if self.nameOnly && !keys.isEmpty {
+                // swiftlint:disable:next force_unwrapping
+                opts!["nameOnly"] = false // Same as above.
+            } else {
+                // swiftlint:disable:next force_unwrapping
+                opts!["nameOnly"] = self.nameOnly // Same as above.
```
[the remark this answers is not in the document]
kaywux: Spoke offline about `listCollectionNames` - according to spec:
> Drivers MUST always specify the nameOnly option when they only intend to access collection names from the listCollections command result, except drivers MUST NOT set nameOnly if a filter specifies any keys other than name

Thus we won't throw an error when a non-name filter key is passed in to `listCollectionNames` and keep the filter-checking logic in the operations layer.

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
patrickfreed: This struct still needs the `seeAlso` link

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -112,26 +113,59 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
     }
-}
 
-struct CollectionInfo: Decodable, Equatable {
-    let name: String
-    let type: String
-    let options: CreateCollectionOptions
+    func testListCollections() throws {
+        let client = try MongoClient.makeTestClient()
+        let db = client.db(type(of: self).testDatabase)
```
patrickfreed: Probably want to drop this database before running the test just in case there were some collections added by other tests.

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -112,26 +113,59 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
...
 
         expect(collectionInfo[2].name).to(equal("system.views"))
     }
-}
 
-struct CollectionInfo: Decodable, Equatable {
-    let name: String
-    let type: String
-    let options: CreateCollectionOptions
+    func testListCollections() throws {
```
patrickfreed: I think it would be useful to add a test using command monitoring here to verify that `nameOnly: false` is sent in `listCollectionNames` and `listMongoCollections` with nil / name only filters.

To make that easier, you can use the `withCommandMonitoring` function. Check out the [change stream tests](https://github.com/mongodb/mongo-swift-driver/blob/master/Tests/MongoSwiftTests/ChangeStreamTests.swift#L548) for an example of how to use it.

patrickfreed: Test looks great!

Tests/MongoSwiftTests/MongoDatabaseTests.swift
```diff
@@ -112,26 +113,74 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
         )
 
         expect(try db.createCollection("fooView", options: viewOptions)).toNot(throwError())
-        let decoder = BSONDecoder()
-        var collectionInfo = try db.listCollections().map { try decoder.decode(CollectionInfo.self, from: $0) }
+
+        var collectionInfo = try Array(db.listCollections()) as [CollectionSpecification]
```
kmahar: nit: do we need the type annotation? I'd think it would be inferred by the return type of `listCollections`

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -2,8 +2,6 @@ import Foundation
 
 /// `BSONDecoder` facilitates the decoding of BSON into semantic `Decodable` types.
 public class BSONDecoder {
-    // swiftlint:disable explicit_acl
```
kmahar: I think the version of swiftlint on travis is still broken for this rule so let's leave it here for now

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -0,0 +1,145 @@
...
+    /// The type of data store returned.
+    public let type: CollectionType
+
+    /// Options that were used when creating this collection.
+    public let options: CreateCollectionOptions?
+
+    /// Contains info pertaining to the collection.
+    public let info: CollectionSpecificationInfo
+
+    /// Provides info on the _id index of the collection, is nil when this data store is of type view.
```
kmahar:
```suggestion
    /// Provides info on the _id index of the collection. `nil` when this data store is of type view.
```

PR #318, opened 2019-09-05
SWIFT-517 Return Connection to pool as soon as MongoCursor is exhausted

kmahar:
This PR implements [SWIFT-517](https://jira.mongodb.org/browse/SWIFT-517) and adds some tests around tailable and non-tailable cursor behavior.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -22,8 +22,8 @@ extension MongoCollection {
                      session: ClientSession? = nil) throws -> MongoCursor<CollectionType> {
         let opts = try encodeOptions(options: options, session: session)
         let rp = options?.readPreference?._readPreference
-
-        return try MongoCursor(client: self._client, decoder: self.decoder, session: session) { conn in
+        let isTailable = [.tailable, .tailableAwait].contains(options?.cursorType)
+        return try MongoCursor(client: self._client, decoder: self.decoder, session: session, isTailable: isTailable) { conn in
             self.withMongocCollection(from: conn) { collPtr in
                 guard let cursor = mongoc_collection_find_with_opts(collPtr, filter._bson, opts?._bson, rp) else {
                     fatalError(failedToRetrieveCursorMessage)
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -22,11 +22,10 @@ extension MongoCollection {
                      session: ClientSession? = nil) throws -> MongoCursor<CollectionType> {
         let opts = try encodeOptions(options: options, session: session)
         let rp = options?.readPreference?._readPreference
-        let isTailable = [.tailable, .tailableAwait].contains(options?.cursorType)
         return try MongoCursor(client: self._client,
                                decoder: self.decoder,
                                session: session,
-                               isTailable: isTailable) { conn in
+                               cursorType: options?.cursorType) { conn in
             self.withMongocCollection(from: conn) { collPtr in
                 guard let cursor = mongoc_collection_find_with_opts(collPtr, filter._bson, opts?._bson, rp) else {
                     fatalError(failedToRetrieveCursorMessage)
@@ -289,6 +288,10 @@ public enum CursorType {
      * - SeeAlso: https://docs.mongodb.com/meta-driver/latest/legacy/mongodb-wire-protocol/#op-query
      */
     case tailableAwait
+
+    internal var isTailable: Bool {
+        return self != .nonTailable
+    }
 }
 
 /// Options to use when executing a `find` command on a `MongoCollection`.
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -99,21 +139,40 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
+        guard case let .open(cursor, conn, _, session) = self.state else {
+            self.error = ClosedCursorError
+            return nil
+        }
+
         do {
             let operation = NextOperation(cursor: self)
-            let out = try operation.execute(using: self._connection, session: self._session)
-            self.swiftError = nil
+            guard let out = try operation.execute(using: conn, session: session) else {
```
kmahar: something to note about this design: for a (non-tailable) cursor with `x` documents, we won't close the cursor until the `x + 1`th call to `next()`, where `mongoc_cursor_next` returns `nil`.
we could consider calling `mongoc_cursor_next` and preloading the next document each time, but I wasn't sure if that was worth doing...  we generally expect users to iterate via for loop, right? in which case they will always make all `x + 1` calls to `next()` since that calls `next()` until it returns `nil`.
(note that for non-tailable `mongoc_cursor_more` won't help, as it will always return `true` until a call to `mongoc_cursor_next` has returned `false`.)

in the tailable case,  we *could* preemptively call `mongoc_cursor_more` after each non-nil `out` value and see if there is potentially anything else or if we just retrieved the last doc. (right now we only do this after a nil `out`), and if we are at the end immediately close the cursor but allow the user to call `next()` one more time without triggering a `ClosedCursorError`.

all of this seemed a bit overly complicated to me but if anyone feels like either of those changes would be worth making let me know.

Tests/MongoSwiftTests/MongoCursorTests.swift
```diff
@@ -0,0 +1,112 @@
...
+import XCTest
+
+private let doc1: Document = ["_id": 1, "x": 1]
+private let doc2: Document = ["_id": 2, "x": 2]
+private let doc3: Document = ["_id": 3, "x": 3]
+private let logicError = UserError.logicError(message: "")
+
+final class MongoCursorTests: MongoSwiftTestCase {
+    func testNonTailableCursor() throws {
+        try self.withTestNamespace { _, db, coll in
```
kmahar: I love this `withTestNamespace` thing, ty @patrickfreed

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -22,8 +22,11 @@ extension MongoCollection {
                      session: ClientSession? = nil) throws -> MongoCursor<CollectionType> {
         let opts = try encodeOptions(options: options, session: session)
         let rp = options?.readPreference?._readPreference
-
-        return try MongoCursor(client: self._client, decoder: self.decoder, session: session) { conn in
+        let isTailable = [.tailable, .tailableAwait].contains(options?.cursorType)
+        return try MongoCursor(client: self._client,
+                               decoder: self.decoder,
+                               session: session,
+                               isTailable: isTailable) { conn in
```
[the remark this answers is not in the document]
kmahar: good point. done

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -1,19 +1,53 @@
...
+    }
+
+    /// Returns the ID used by the server to track the cursor. `nil` until mongoc actually talks to the server by
+    /// iterating the cursor, and `nil` after mongoc has fetched all the results from the server.
+    internal var id: Int64? {
+        guard case let .open(cursor, _, _, _) = self.state else {
+            return nil
+        }
+        let id = mongoc_cursor_get_id(cursor)
+        return id == 0 ? nil : id
```
mbroadst: nitpick: I think there's a meaningful difference between 0 and `nil` in this case. 0 means the cursor is closed server side, but nil might indicate the server was never actually initialized or even run.

Sources/MongoSwift/MongoCursor.swift
mbroadst: Would this code trigger the issue:
```
let singleDocFromCursor = try cursor.next()
.. do something with the single document from the response, but don't keep iterating ..
```
?

Even in that case it might not be an issue because the cursor will likely fall out of scope (where we will clean it up implicitly). I think this behavior is fine. If we had more of a java iterator approach, then the `hasNext` would need to cache a document like one of your proposals above:
```
while (cursor.hasNext()) {
   doSomethingWith(cursor.next())
}
```

Sources/MongoSwift/MongoCursor.swift
kmahar: we are kind of limited here by the API mongoc provides -- a `0` return value from `mongoc_cursor_get_id` can indicate either of those cases:
> This number is zero until the driver actually uses a server when executing the query, and after it has fetched all results from the server.

we could consider adding some kind of boolean flag in `MongoCursor` that tracks if we've iterated at all, and if we haven't, return `nil` here, and otherwise directly return the 0. do you think that's worth it?

Sources/MongoSwift/MongoCursor.swift
kmahar: yes, it would trigger it unless if no documents were actually returned in the cursor. if the cursor has one or more documents, then it will stay open until it goes out of scope.

Sources/MongoSwift/MongoCursor.swift
mbroadst: nah, let's just leave this as-is, and consider it in a future where we're not using libmongoc

PR #319, opened 2019-09-06
SWIFT-608 Implement index enumeration spec

kaywux:
Implement the index enumeration spec using listIndexes and listIndexNames. Note that some aspects of the spec weren't followed for the following reasons:
-  Naming of the fields `v`, `key`, `default_language`, and `language_override`, as they conflict with the existing index management spec (https://github.com/mongodb/specifications/blob/master/source/index-management.rst#standard-api) . Instead used `version`, `keys`, `defaultLanguage`, and `languageOverride` for the respective fields.
-  Naming of the field `2dsphereIndexVersion` because of Swift constraints on naming variables with numbers.
- Optional `ns` field - the latest version of the server has omitted this field, so `testIndexOptions` couldn't depend on and test the presence of the field.

JIRA ticket: https://jira.mongodb.org/browse/SWIFT-608

Evergreen: https://evergreen.mongodb.com/version/5d769864e3c33168706697cb

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -1,7 +1,7 @@
 import mongoc
 
 /// A struct representing an index on a `MongoCollection`.
-public struct IndexModel: Encodable {
+public struct IndexModel: Codable {
     /// Contains the required keys for the index.
     public let keys: Document
 
@@ -335,20 +335,33 @@ extension MongoCollection {
      * - Parameters:
      *   - session: Optional `ClientSession` to use when executing this command
      *
+     * - Returns: A `MongoCursor` over the `IndexModel`s.
+     *
+     * - Throws: `UserError.logicError` if the provided session is inactive.
+     */
+    public func listIndexes(session: ClientSession? = nil) throws -> MongoCursor<IndexModel> {
+        let operation = ListIndexesOperation(collection: self, nameOnly: false)
+        guard case let .specs(result) = try self._client.executeOperation(operation, session: session) else {
+            throw RuntimeError.internalError(message: "Invalid result")
+        }
+        return result
+    }
+
+    /**
+     * Retrieves a list of names of the indexes currently on this collection.
+     *
+     * - Parameters:
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
      * - Returns: A `MongoCursor` over the index names.
      *
      * - Throws: `UserError.logicError` if the provided session is inactive.
      */
-    public func listIndexes(session: ClientSession? = nil) throws -> MongoCursor<Document> {
-        let opts = try encodeOptions(options: Document(), session: session)
-
-        return try MongoCursor(client: self._client, decoder: self.decoder, session: session) { conn in
-            self.withMongocCollection(from: conn) { collPtr in
-                guard let cursor = mongoc_collection_find_indexes_with_opts(collPtr, opts?._bson) else {
-                    fatalError(failedToRetrieveCursorMessage)
-                }
-                return cursor
-            }
+    public func listIndexNames(session: ClientSession? = nil) throws -> [String] {
+        let operation = ListIndexesOperation(collection: self, nameOnly: true)
+        guard case let .specs(result) = try self._client.executeOperation(operation, session: session) else {
+            throw RuntimeError.internalError(message: "Invalid result")
         }
+        return result
     }
 }
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -1,55 +1,5 @@
 import mongoc
 
-// /// A struct representing an index on a `MongoCollection`.
-// public struct IndexModel: Codable {
-//     /// Contains the required keys for the index.
-//     public let keys: Document
-
-//     /// Contains the options for the index.
-//     public let options: IndexOptions?
-
-//     // /// Contains the version of the index.
-//     // public let v: Int32?
-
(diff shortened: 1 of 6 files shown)


Tests/MongoSwiftTests/MongoCollection+IndexTests.swift
```diff
@@ -78,13 +79,22 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
...
 
+        let sv = try _client?.serverVersion() ?? ServerVersion("8000")
+        guard try sv < ServerVersion("4.3.0") else {
+            print("Skipping test, fails with latest server version because of missing ns field")
+            return
+        }
+
+        let nsString = MongoSwiftTestCase.isMacOS ? "test.MongoCollection_IndexTests_testIndexOptions" :
+                                                    "test.MongoCollection_IndexTests.testIndexOptions"
+
```
kaywux: The `ns` field seems to be different based on whether it is running on MacOS ("_") or Ubuntu ("."). This was the only way I could get the evergreen tests to pass, but please let me know if there is a better way to do this.

patrickfreed: So, upon further reflection, I've realized that setting the `ns` field on `IndexOptions` doesn't actually make a lot of sense. When calling `listIndexes` on a particular `MongoCollection`, you know that all indexes returned from that query will have the same namespace (that of the collection you called the method on). Likewise for `createIndex`, we wouldn't want to allow the user to create an index on a different namespace from the collection they were invoking the command on. Also, as proved by the tests that were failing on "latest", the server isn't going to even include `ns` in future responses from the server. For all those reasons, I think it's okay to just omit the field altogether. Worst case, we can add it back later for spec compliance. 

Sorry for sending the wrong signals on what to do on this one--I hadn't read the spec in its entirety and made some incorrect assumptions.

Sources/MongoSwift/Operations/ListIndexesOperation.swift
```diff
@@ -0,0 +1,51 @@
...
+
+        let initializer = { (conn: Connection) -> OpaquePointer in
+            self.collection.withMongocCollection(from: conn) { collPtr in
+                guard let indexes = mongoc_collection_find_indexes_with_opts(collPtr, opts?._bson) else {
+                    fatalError(failedToRetrieveCursorMessage)
+                }
+                return indexes
+            }
+        }
+        if self.nameOnly {
```
patrickfreed: because there isn't a server-side `nameOnly` parameter, I think we can just have this operation always return a cursor over the specifications, and we can map that cursor to a `[String]` in the body of `listCollectionNames`.

The separation in `ListCollectionOperation` and `ListDatabaseOperation` is mainly due to the fact that `nameOnly` affects both the format of the server's replies as well as the semantics of the command itself. In this case, the server returns the same thing regardless of whether we only want the names or not, so I think the operation should do the same.

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -139,11 +145,12 @@ public struct IndexOptions: Codable {
         self.collation = collation
     }
 
-    // Encode everything besides the name, as we will handle that when encoding the `IndexModel`
+    // The `name` key is encoded in `IndexModel` but needed here for decoding
     private enum CodingKeys: String, CodingKey {
-        case background, expireAfterSeconds, sparse, storageEngine, unique, indexVersion = "v",
+        case background, expireAfterSeconds, name, sparse, storageEngine, unique, version = "v",
```
patrickfreed: Because `name` is now part of the options encoding/decoding, we should stop doing the weird encoding of `name` in `IndexModel`. Instead, the logic of checking if the user specified the `name` and possibly setting the default should live in the body of `CreateIndexesOperation` (where it probably should have been originally).

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -335,20 +342,28 @@ extension MongoCollection {
...
-                guard let cursor = mongoc_collection_find_indexes_with_opts(collPtr, opts?._bson) else {
-                    fatalError(failedToRetrieveCursorMessage)
-                }
-                return cursor
-            }
-        }
+    public func listIndexNames(session: ClientSession? = nil) throws -> [String] {
+        let operation = ListIndexesOperation(collection: self)
+        let models = try self._client.executeOperation(operation, session: session)
+        return models.map { $0.options?.name ?? $0.defaultName }
```
patrickfreed: the server should always return a name field, so lets throw an error here instead of supplying a name on the fly. If a name doesn't show up here, it's likely to due a bug in the driver somewhere, and we don't want the behavior caused by that bug to be hidden.

An interesting concept in swift that is related to this is `rethrows`: Similar to how you can mark a function as `throws`, meaning it may return errors and must be invoked with `try`, you can mark a method as `rethrows`, meaning it will only be `throws` if a closure passed to it is `throws`.

The `map` method on `Array` is `rethrows`, so we can still use it here with a `throws` closure:
```swift
return try models.map { model in 
    guard let name = model.options?.name else {
        throw RuntimeError.internalError("Server response missing a 'name' field")
    }
    return name
}
```

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -139,11 +145,12 @@ public struct IndexOptions: Codable {
         self.collation = collation
     }
 
-    // Encode everything besides the name, as we will handle that when encoding the `IndexModel`
+    // The `name` key is encoded in `IndexModel` but needed here for decoding
```
patrickfreed: I think this comment can be removed.

PR #322, opened 2019-09-12
SWIFT-523 Change stream operations for creation and iteration

kaywux:
Create a `ChangeStreamNextOperation` that is called by the `next` function of a change stream. The `watch` functionality for `MongoClient`, `MongoDatabase` and `MongoCollection` is encapsulated in the new `WatchOperation` class.

JIRA ticket: https://jira.mongodb.org/browse/SWIFT-523

Evergreen: https://evergreen.mongodb.com/version/5d8a6c6f5623430d610e3606


Sources/MongoSwift/ChangeStream.swift
```diff
@@ -84,38 +84,11 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
...
-                    RuntimeError.internalError(message: "_id field is missing from the change stream document.")
-            return nil
-        }
-        self.resumeToken = ResumeToken(resumeToken)
-
         do {
-            return try decoder.decode(T.self, from: doc)
+            let operation = ChangeStreamNextOperation(changeStream: self)
+            let out = try operation.execute(using: self.connection, session: self.session)
+            self.swiftError = nil
```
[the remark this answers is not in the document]
patrickfreed: Yeah, I can't think of a situation in which we would want to clear an error that we've encountered. Also, I think the `guard` that was moved into the operation should just remain in `next` at the change stream level, so that no operation will even be created in the event that an error has been seen. Agreed on automatically closing.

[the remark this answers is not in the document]
kaywux: I closed the `ChangeStream` if an error was encountered, but should I also close in the case that there are no documents returned and an error occurred / the cursor can't possibly return any more data (unsure of the situations where this happens)? I am assuming that a `ChangeStream` operates like a tailable cursor (which might be wrong) so am looking at cases 2 and 3 for when the `next` operation of `MongoCursor` doesn't return a doc.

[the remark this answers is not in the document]
kmahar: yup, you are right that change streams behave like tailable cursors.

so case 2 where we encountered some error iterating a change stream is definitely applicable here.
case 3 for tailable cursors is something that we can encounter in cases like [this](https://github.com/mongodb/mongo-swift-driver/blob/master/Tests/MongoSwiftTests/MongoCursorTests.swift#L61). if nothing matches the initial tailable cursor query (in the test that occurs because the collection is empty) then the server will not keep the cursor alive and `mongoc_cursor_more` would return false.

I don't think case 3 is relevant for change streams as the server won't close a change stream even if it is initially empty. and mongoc provides us no way to check something like `mongoc_cursor_more` anyway.

if we're going to make this closing change now, I think we should switch `ChangeStream` to store all of its properties in a state enum much like `MongoCursor` does, so that we can nil everything out at once when we close it. and we probably need to/should save off errors similarly to how we do there rather than recomputing each time.
right now we have a couple problems in that we will potentially call `close()` (and therefore `mongoc_change_stream_destroy`) twice, once via manual `close` and once via `deinit`, and in that after we call `close` we can't access any mongoc error that was encountered after calling `close()`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -420,11 +420,8 @@ public class MongoClient {
                                   options: ChangeStreamOptions?  =  nil,
                                   session: ClientSession? = nil,
                                   withEventType: T.Type) throws -> ChangeStream<T> {
-        let pipeline: Document = ["pipeline": pipeline]
-        let opts = try encodeOptions(options: options, session: session)
-        return try ChangeStream<T>(options: options, client: self, decoder: self.decoder, session: session) { conn in
-            mongoc_client_watch(conn.clientHandle, pipeline._bson, opts?._bson)
-        }
+        let operation = try WatchOperation<T, T>(target: .client(self), pipeline: pipeline, options: options)
```
patrickfreed: Let's update the generic argument to `watch` to disambiguate from the collection's parameter, maybe `EventType` would be better? Then we can use `WatchOperation<CollectionType, EventType>`. This also applies to other `watch` functions.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -84,40 +91,27 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
...
-
         do {
-            return try decoder.decode(T.self, from: doc)
+            // If an error exists, refuse iterating the change stream to avoid overwriting the original error.
+            guard self.error == nil else {
+                return nil
+            }
+            let operation = ChangeStreamNextOperation(changeStream: self)
+            guard let out = try operation.execute(using: connection, session: session) else {
+                if self.getChangeStreamError() != nil {
```
patrickfreed: we probably want to be assigning this error (if non nil) to `self.error` before closing.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -84,40 +91,28 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
...
-
         do {
-            return try decoder.decode(T.self, from: doc)
+            // If an error exists, refuse iterating the change stream to avoid overwriting the original error.
+            guard self.error == nil else {
+                return nil
+            }
+            let operation = ChangeStreamNextOperation(changeStream: self)
+            guard let out = try operation.execute(using: connection, session: session) else {
+                self.error = self.getChangeStreamError
```
kmahar: I think this needs `()` to call the method

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -24,35 +27,39 @@ public struct ResumeToken: Codable, Equatable {
...
      *  - Errors:
      *    - `DecodingError` if an error occurs while decoding the server's response.
      */
-    public var error: Error? {
-        if let err = self.swiftError {
+    private func getChangeStreamError() -> Error? {
+        guard case let .open(changeStream, _, _, _) = self.state else {
+            return nil
+        }
+        if let err = self.error {
```
kmahar: I think we can skip this check because this could return any error, not just an error from mongoc (e.g. some decoding error).
I also don't think there's currently a code path where we call this method and this property would have already been set to non-nil

Sources/MongoSwift/Operations/ChangeStreamNextOperation.swift
```diff
@@ -0,0 +1,48 @@
...
+internal struct ChangeStreamNextOperation<T: Codable>: Operation {
+    private let changeStream: ChangeStream<T>
+
+    internal init(changeStream: ChangeStream<T>) {
+        self.changeStream = changeStream
+    }
+
+    internal func execute(using connection: Connection, session: ClientSession?) throws -> T? {
+        // If an error exists, refuse iterating the change stream to avoid overwriting the original error.
+        guard self.changeStream.error == nil else {
```
kmahar: we already check this in `next`, just before constructing the operation, so can probably skip it here.

Sources/MongoSwift/ChangeStream.swift
kaywux: Sorry about this, I accidentally deleted the `()` during a previous change and didn't double check my code

Sources/MongoSwift/Operations/ChangeStreamNextOperation.swift
```diff
@@ -0,0 +1,44 @@
...
+        self.changeStream = changeStream
+    }
+
+    internal func execute(using connection: Connection, session: ClientSession?) throws -> T? {
+        // We already check this in `ChangeStream.next()` in order to extract the relevant connection and session,
+        // but error again here just in case.
+        guard case let .open(changeStreamPtr, _, _, _) = changeStream.state else {
+            throw ClosedChangeStreamError
+        }
+
```
kmahar: comparing to the `NextOperation` for cursors, I noticed we are doing a check there to see if the session is still active and throwing an error if not. I think we probably need to check that here as well, what do you think @patrickfreed ?

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -84,40 +88,28 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
...
-            self.swiftError =
-                    RuntimeError.internalError(message: "_id field is missing from the change stream document.")
-            return nil
-        }
-        self.resumeToken = ResumeToken(resumeToken)
-
         do {
-            return try decoder.decode(T.self, from: doc)
+            // If an error exists, refuse iterating the change stream to avoid overwriting the original error.
+            guard self.error == nil else {
```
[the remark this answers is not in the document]
kaywux: That makes sense, I'll remove this guard statement

Sources/MongoSwift/Operations/NextOperation.swift
```diff
@@ -1,11 +1,20 @@
...
-    private let cursor: MongoCursor<T>
+    private let target: NextOperationTarget<T>
 
-    internal init(cursor: MongoCursor<T>) {
-        self.cursor = cursor
+    internal init(target: NextOperationTarget<T>) {
+        self.target = target
     }
 
     internal func execute(using connection: Connection, session: ClientSession?) throws -> T? {
```
kaywux: I decided to combine `ChangeStreamNextOperation` and `NextOperation` because there was a lot of repeated code, but I am now running into a problem with the linter's Cyclomatic Complexity Violation. Is there another way to structure this code that is less complex than what I am doing?

kmahar: combining seems like a good idea to me. 

I'm not sure of a better way to structure this, I'd be fine with just turning off the cyclomatic complexity rule here. 

you might consider trying to identify all of the shared properties of these types and putting them on a protocol, something like:
```swift
protocol CursorLike {
    associatedtype OutputType: Decodable
    var state: CursorLikeState { get } // you'd have to start using same state type for both cursor and changestream
    static var closedError: Error { get }
    var nextFunc: (OpaquePointer, UnsafeMutablePointer<BSONPointer?>) -> (Bool) { get }
    var decoder: BSONDecoder { get }
}
```

(poorly named but you get the idea)

then you could make `NextOperation` generic on this type and you could do stuff like
```swift
guard case let .open(ptr, _, _, _) = target.state else {
    throw target.closedError
}

guard nextFunc(ptr, out) else {
    return nil
}

return target.decoder.decode(CursorType.OutputType.self, from: doc)
```

that said, it might not be worth it to do all that, but interesting to consider!

kaywux: Got it, I'll turn off the swiftlint rule in this case, but will keep this protocol method in mind for the future!

PR #323, opened 2019-09-23
SWIFT-619: Add logic for executing operations with different connection behavior

kmahar:
This PR adds a way for `Operation`s to specify how they interact with `Connection`s passed to their execute method (SWIFT-619). 
`NextOperation` now uses the `.ignores` case which allows us to execute it via executor and skip checking out a new connection rather than doing it manually.
I've also added `StartSessionOperation` and `EndSessionOperation` which utilize `.steals` and `.ignores` respectively. 

cc @kaywux as the `WatchOperation` and `NextOperation` she is adding via #322 will end up using these as well.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -122,7 +122,7 @@ public final class ClientSession {
     public let options: ClientSessionOptions?
 
     /// Initializes a new client session.
-    internal init(client: MongoClient, options: ClientSessionOptions?) throws {
+    internal init(client: MongoClient, options: ClientSessionOptions? = nil) throws {
         self.options = options
         self.client = client
 
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -122,7 +122,7 @@ public final class ClientSession {
     public let options: ClientSessionOptions?
 
     /// Initializes a new client session.
-    internal init(client: MongoClient, options: ClientSessionOptions? = nil) throws {
+    internal init(client: MongoClient, options: ClientSessionOptions?) throws {
         self.options = options
         self.client = client
 
--- a/Sources/MongoSwift/Operations/Operation.swift
+++ b/Sources/MongoSwift/Operations/Operation.swift
@@ -4,7 +4,7 @@ internal protocol Operation {
     /// The result type this operation returns.
     associatedtype OperationResult
     /// Indicates how this operation interactions with `Connection`s.
-    static var connectionUsage: ConnectionUsage { get }
+    var connectionUsage: ConnectionUsage { get }
 
     /// Executes this operation using the provided connection and optional session, and returns its corresponding
     /// result type.
@@ -13,7 +13,7 @@ internal protocol Operation {
 
 extension Operation {
     /// This is the behavior of most operations so default to this.
-    internal static var connectionUsage: ConnectionUsage { return .uses }
+    internal var connectionUsage: ConnectionUsage { return .uses }
 }
 
 /// Uses to indicate how an `Operation` type uses `Connection`s passed to its execute method.
@@ -22,12 +22,12 @@ internal enum ConnectionUsage {
     /// over responsibility for later returning it to the pool. This applies to e.g. `WatchOperation` where the
     /// resulting `ChangeStream` will hold onto its source `Connection` until deinitialization.
     case steals
-    /// This operation will ignore the connection passed to its execute method, as it already is holding onto a
-    /// connection. This applies to e.g. `NextOperation` where the operation must use its parent cursor's source
-    /// connection rather than an arbitrary one from the pool.
-    case ignores
-    /// This operation will use the connection provided to its execute method to execute itself. It will not save it
-    /// or pass it off for later usage. This applies to the majority of operations.
+    /// This operation is already holding onto the provided connection, which should be used to execute it. This
+    /// applies to e.g. `NextOperation` where the operation must use its parent cursor's source connection rather than
+    /// an arbitrary one from the pool.
+    case owns(Connection)
+    /// This operation will use the connection provided to its execute method to execute itself. It will not save it or
+    /// pass it off for later usage. This applies to the majority of operations.
     case uses
 }
 
@@ -44,16 +44,14 @@ internal struct DefaultOperationExecutor: OperationExecutor {
     internal func execute<T: Operation>(_ operation: T,
                                         client: MongoClient,
                                         session: ClientSession?) throws -> T.OperationResult {
-        switch T.connectionUsage {
+        switch operation.connectionUsage {
         case .steals:
             // don't return the connection to the pool, as the operation will handle it
             let conn = try session?.getConnection(forUseWith: client) ?? client.connectionPool.checkOut()
             return try operation.execute(using: conn, session: session)
-        case .ignores:
-            // we won't use the connection so pass in a dummy. initialization always succeeds with this bit pattern.
-            // swiftlint:disable:next force_unwrapping
-            let dummy = Connection(OpaquePointer(bitPattern: 1)!)
-            return try operation.execute(using: dummy, session: session)
+        case let .owns(conn):
+            // pass in the connection this operation already owns
+            return try operation.execute(using: conn, session: session)
         case .uses:
             // if a session was provided, use its underlying connection
             if let session = session {
--- a/Sources/MongoSwift/Operations/Operation.swift
+++ b/Sources/MongoSwift/Operations/Operation.swift
@@ -3,9 +3,32 @@
 internal protocol Operation {
     /// The result type this operation returns.
     associatedtype OperationResult
+    /// Indicates how this operation interactions with `Connection`s.
+    static var connectionUsage: ConnectionUsage { get }
+
     /// Executes this operation using the provided connection and optional session, and returns its corresponding
     /// result type.
     func execute(using connection: Connection, session: ClientSession?) throws -> OperationResult
+}
+
+extension Operation {
+    /// This is the behavior of most operations so default to this.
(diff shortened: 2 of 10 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -157,11 +122,14 @@ public final class ClientSession {
...
-            self.client.connectionPool.checkIn(connection)
-            self.state = .inactive
+        guard self.active else {
+            return
         }
+        let operation = EndSessionOperation(session: self)
+        // since `end()` is called from `deinit` and a `defer` block in `withSession`, we can't throw here. however,
+        // in practice this will never throw anyway. the branch taken in `executeOperation` doesn't throw, and neither
+        // does `EndSessionOperation`'s execute method.
+        try? self.client.executeOperation(operation)
```
kmahar: exact details are to be worked out in async design, but idea is that we can execute this asynchronously if the client used to create this session was an async client

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -21,14 +44,26 @@ internal struct DefaultOperationExecutor: OperationExecutor {
...
+            let conn = try session?.getConnection(forUseWith: client) ?? client.connectionPool.checkOut()
             return try operation.execute(using: conn, session: session)
-        }
-        // otherwise use a new connection from the pool
-        return try client.connectionPool.withConnection { conn in
-            try operation.execute(using: conn, session: nil)
+        case .ignores:
+            // we won't use the connection so pass in a dummy. initialization always succeeds with this bit pattern.
+            // swiftlint:disable:next force_unwrapping
+            let dummy = Connection(OpaquePointer(bitPattern: 1)!)
```
patrickfreed: I think it would be better if we modified the `Operation` protocol method to take in a `Connection?` rather than just passing in a null pointer.

kmahar: we'd have to unwrap the `Connection` in all of the operations that actually use it then, do you think that's preferable?

patrickfreed: That's also not great, but it feels a little off to side-step the compiler like we're currently doing here. I don't feel super strongly about this though, so if you think it's fine, then it's fine with me.

kmahar:
Update: I've removed the code for handling sessions creation/destruction-related I/O. that work will be done as part of the async PRs.
this now just includes the new way for connections to indicate their behavior, and updates `WatchOperation` to use this new strategy

Sources/MongoSwift/Operations/Operation.swift
kmahar: one alternative I thought of:
we make the `connectionUsage` property non-static, and make `.ignores` something like `.owns(Connection)` instead.

then whenever it's that case, we just pass in the already-owned `Connection` to the `execute` method instead.

so e.g. `NextOperation` would be `.owns` and use the cursor's source connection in that place.

thoughts?

patrickfreed: Yeah I think that sounds good. That's actually how I currently have it designed in the Rust driver, too (though we went that path mostly for Rust specific reasons).

kmahar: updated to that approach

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -3,11 +3,34 @@
...
     func execute(using connection: Connection, session: ClientSession?) throws -> OperationResult
 }
 
+extension Operation {
+    /// This is the behavior of most operations so default to this.
+    internal var connectionUsage: ConnectionUsage { return .uses }
+}
+
+/// Uses to indicate how an `Operation` type uses `Connection`s passed to its execute method.
+internal enum ConnectionUsage {
```
mbroadst: nitpick: I think it might be more consistent to call this a "strategy".

What is the difference between an operation which retains a connection (owns), and one which "steals" a connection? It seems like in both cases we're saying "the connection passed into this operation should become the responsibility of the operation, it should close it/return it eventually" It seems to me the distinction lies in the initialization of the operation. 

Rather, would it make sense to have only two strategies: `retained(conn)` and `unretained`, and the stealing would be done though `init(stealing: conn, session)`?

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -21,14 +44,24 @@ internal struct DefaultOperationExecutor: OperationExecutor {
     internal func execute<T: Operation>(_ operation: T,
                                         client: MongoClient,
                                         session: ClientSession?) throws -> T.OperationResult {
-        // if a session was provided, use its underlying connection
-        if let session = session {
-            let conn = try session.getConnection(forUseWith: client)
+        switch operation.connectionUsage {
+        case .steals:
+            // don't return the connection to the pool, as the operation will handle it
```
mbroadst: the meat of this case looks like a condensed version of the `.uses` case below, which is where I'm getting the smell for the suggestion above about reducing to two cases.

Sources/MongoSwift/Operations/WatchOperation.swift
```diff
@@ -31,32 +33,34 @@ internal struct WatchOperation<CollectionType: Codable, ChangeStreamType: Codabl
...
-            return try ChangeStream<ChangeStreamType>(options: self.options,
-                                                      client: client,
-                                                      decoder: client.decoder,
-                                                      session: session) { conn in
-                mongoc_client_watch(conn.clientHandle, pipeline._bson, opts?._bson)
-            }
+        case let .client(c):
+            client = c
+            decoder = c.decoder
+            changeStream = mongoc_client_watch(connection.clientHandle, pipeline._bson, opts?._bson)
```
mbroadst: why is this one different than collection and database?

Sources/MongoSwift/Operations/Operation.swift
kmahar: strategy definitely sounds better, i didn't really like this name.

I agree that these cases are very similar. I was mainly trying to avoid calling the
`try session?.getConnection(forUseWith: client) ?? client.connectionPool.checkOut()` line every time we construct an operation like this. but at the moment anyway that is only 3 places (once for each of collection/db/client watch) so it seems acceptable.

that said I feel like we can do better on naming the cases, because it's still not quite true that e.g. `NextOperation` owns the connection or takes responsibility for it.... Rather, the cursor / change stream that created the operation owns it. in the `WatchOperation` case, the operation temporarily owns it, but quickly passes off ownership to the `ChangeStream`.

maybe `.bound(to: Connection)` and `.unbound` better conveys this? i.e. based on the context that the operation was created it in we've already bound it to a particular connection.

mbroadst: that sgtm, “retained” immediately came to mine,  but I think bound similarly conveys intent. Perhaps a helper function is in order for the repetitive code issue, I thought at some point we discussed `resolveConnection(session, pool)`?

Sources/MongoSwift/Operations/WatchOperation.swift
mbroadst: sorry, I misread this to be assigning a changestream type rather than a pointer!

mbroadst: sorry, I misread this to be assigning a changestream type rather than a pointer!

Sources/MongoSwift/Operations/Operation.swift
kmahar: yeah, i feel like we had that in a previous PR but ended up not needing it? seems useful now though and we can also use it in `MongoCursor`'s init which does the same thing.

I also added a `releaseConnection` method that handles the reverse logic of releasing a connection that was obtained via this method.

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -21,15 +40,36 @@ internal struct DefaultOperationExecutor: OperationExecutor {
...
                                         client: MongoClient,
                                         session: ClientSession?) throws -> T.OperationResult {
-        // if a session was provided, use its underlying connection
-        if let session = session {
-            let conn = try session.getConnection(forUseWith: client)
+        switch operation.connectionStrategy {
+        case let .bound(conn):
+            // pass in the connection this operation is already bound to
             return try operation.execute(using: conn, session: session)
+        case .unbound:
```
patrickfreed: We could consolidate all the checking in / checking out of connections to this spot by adding some more checks. 
- For instance, after executing an `.unbound` operation, we could check to see if it has become `.bound`. If so, then don't check the connection back in. Otherwise, do check it in. This would essentially be the "stealing" behavior that was once here. 
- Similarly, after executing a `.bound` connection, we could check if it has become `.unbound`, and if so, check the connection back in. This case would be if a `NextOperation` failed on a cursor or something like that.

What do you think about this approach? I slightly prefer it since it would restrict a lot of the connection lifetime management to just this method, rather than having calls to `resolveConnection` and `checkIn` in more places. Some cases will need to remain no matter what though, like if a cursor goes out of scope before it's exhausted.

kmahar: consolidating the logic seems nice. I tried to test this out but quickly ran into some problems surrounding mutability - the `Operation.execute` method would need to become mutating for this to work (so we could change the `connectionStrategy` there), and also `OperationExecutor.execute` would need to take the operation as an `inout` parameter which feels kind of weird.
I think, for now, maybe we should just stick with how things are and open a ticket about exploring this more at a future point?

patrickfreed: For what it's worth, the Rust Operation layer's equivalents are mutating / take in mutable references. That said, I think it's fine if we defer further work on this since we need to get through some other higher priority things pre 1.0. I filed [SWIFT-633](https://jira.mongodb.org/browse/SWIFT-633) for that.

kmahar: 👍 ty!

PR #326, opened 2019-10-04
SWIFT-628 Introduce the BSON enum

patrickfreed:
This PR introduces the `BSON` enum as defined in the BSON API v2 Design. It also updates a few tests to use it.


Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -0,0 +1,413 @@
...
+    case decimal128(Decimal128)
+
+    /// A BSON minKey.
+    case minKey
+
+    /// A BSON maxKey.
+    case maxKey
+
+    /// If this `BSON` is an `.int32`, return it as an `Int32`. Otherwise, return nil.
+    public var int32Value: Int32? {
```
patrickfreed: these are the helpers that simply extract a value directly if possible with no coercion. There are helpers like this for types that have associated values (e.g. `.int32(Int32)`). For types like `.null` or `.maxKey`, a user can just do `if bson == .null { ... }`, so theres no real point in adding helpers.

Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -0,0 +1,413 @@
...
+    public var timestampValue: Timestamp? {
+        guard case let .timestamp(t) = self else {
+            return nil
+        }
+        return t
+    }
+
+    /// Return this BSON as an `Int` if possible.
+    /// This will coerce non-integer numeric cases (e.g. `.double`) into an `Int` if such coercion would be lossless.
+    public func asInt() -> Int? {
```
patrickfreed: these are the helpers that coerce different `BSON` cases into Swift types. For now, I've just added ones for the numeric types and `.array`, but it can always be expanded. For instance, maybe an `asString()` could be useful for coercing `.string`, `.symbol`, and even `.objectId`.

Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -0,0 +1,413 @@
...
+        case let .double(d):
+            return Decimal128(String(d))
+        default:
+            return nil
+        }
+    }
+
+    /// Return this BSON as an `[T]` if possible, transforming each element via the provided closure.
+    /// If the closure returns nil, the value will be discarded from the resultant array.
+    public func asArray<T>(_ transform: (BSON) -> T?) -> [T]? {
```
patrickfreed: this could be named `map` since that's basically what it's doing, but I didn't want to imply that cases besides `.array` could be mapped. It could have been named `mapArray` to solve that concern, but I went with `asArray` to be consistent with other coercion methods.

Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -0,0 +1,413 @@
...
+        return self.arrayValue?.compactMap(transform)
+    }
+}
+
+/// Extension providing the internal API of `BSON`
+extension BSON {
+    /// List of all BSONValue types. Can be used to exhaustively check each one at runtime.
+    internal static var allBSONTypes: [BSONValue.Type] = [
```
patrickfreed: this will be used by the `Decodable` conformance to try a bunch of different types.

Sources/MongoSwift/BSON/BSON.swift
kmahar: did you have a particular use case in mind for this method / is this something we expect users to do a lot or to do a lot ourselves?
just wondering why have this vs. just letting users do `myBSON.arrayValue?.map` or `myBSON.arrayValue?.compactMap` themselves.

Tests/MongoSwiftTests/BSONValueTests.swift
```diff
@@ -23,77 +23,76 @@ final class BSONValueTests: MongoSwiftTestCase {
...
-            alternate: [
-                "foo": 1.414,
-                "bar": "swift",
-                "nested": [ "a": 1, "b": "different" ] as Document
-                ] as Document
-        )
-        // Check that when an array contains non-BSONValues, we return false
-        let arr = [[String: Int]()]
-        expect(arr.bsonEquals(arr)).to(beFalse())
+        // TODO SWIFT-230: unskip
```
kmahar:
```suggestion
        // TODO SWIFT-630: unskip
```

kmahar: going to that ticket did take me back to maybe the stupidest bug I ever introduced in the driver though 🙂

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -750,6 +750,45 @@ extension Int64: BSONNumber {
     }
 }
 
+/// A struct to represent the BSON Code type.
+public struct Code: BSONValue, Equatable, Codable, Hashable {
```
kmahar: should `CodeWithScope` be updated to have non-optional scope and always have BSON type `.javascriptWithScope`? or is that coming in future PR?

Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -0,0 +1,413 @@
...
+
+    /// A BSON undefined.
+    case undefined
+
+    /// A BSON ObjectId.
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/bson-types/#objectid
+    case objectId(ObjectId)
+
+    /// A BSON boolean.
+    case bool(Bool)
```
kmahar: I'm noticing we have some naming mismatches between this enum and the `BSONType` enum:
- `bool` vs `boolean`
- `date` vs `dateTime`
- `regex` vs `regularExpression`
- `javascript` vs `code`
- `javascriptWithScope` vs `codeWithScope`

I don't feel very strongly about either option in any of these cases but it does seem to me like they should be consistent.

Sources/MongoSwift/BSON/BSON.swift
kmahar: i never thought about this in the design, but what happens to the types like null, undefined, minkey, maxkey now, since they aren't actually used in the enum?

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -750,6 +750,45 @@ extension Int64: BSONNumber {
...
+    /// A string containing Javascript code.
+    public let code: String
+
+    /// Initializes a `CodeWithScope` with an optional scope value.
+    public init(code: String) {
+        self.code = code
+    }
+
+    public init(from decoder: Decoder) throws {
+        throw getDecodingError(type: CodeWithScope.self, decoder: decoder)
```
kmahar: `CodeWithScope` -> `Code`?

Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+extension BSON: ExpressibleByFloatLiteral {
+    public init(floatLiteral value: Double) {
+        self = .double(value)
+    }
+}
+
+extension BSON: ExpressibleByIntegerLiteral {
```
kmahar: so if I have an `Int` saved in a variable and I want to turn it into a BSON do I have to cast it?

e.g. like
```swift
let x = 1
let bson = BSON.int64(Int64(x))
```

Sources/MongoSwift/BSON/BSON.swift
kmahar: I think it's `if case .null = bson`, but in any case (lol) I agree helper isn't really valuable

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: Good point. I've decided on the following:
- `.bool`: expresses the same idea as "boolean" but is slightly briefer
- `.datetime`: since dates and datetimes are different things, I felt it was important to be precise here. I did away with the camel case since a datetime is thing itself rather than a combination of two words.
- `.regex`: communicates the same idea as "regularExpression" but is much briefer
- `.code` and `.codeWithScope`: "code" is much briefer than "javascript" but still stays true to the spec

Let me know if this scheme makes sense. I also updated some of the docstrings to be more descriptive.

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: `==` also works here due to the `Equatable` conformance, but yeah that's another option thats pretty much equally simple. The benefit of using `Equatable` though is that you get the negation `!=`, and you can use it outside of control flow situations, e.g. `let x = BSON.null == BSON.string("not equal")`

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: Honestly, its only benefit is that it's shorter. The reason it exists at all is because I originally was using the following:
```swift
public func asArray<T>(of: T.Type) -> [T]? {
    self.arrayValue?.compactMap { $0.bsonValue as? T }
}
```
I quickly realized that this method is not a good idea. I was doing things like `a.getArray(of: Int.self)` or `a.getArray(of: [Int32.self])` and getting `[]` back unknowingly. To fix that, I decided on the user passing a lambda instead, but I never really evaluated whether the method was still worth using. 

I use it quite frequently in the driver and the tests, but that doesn't really mean it's _useful_. I'm not sure if the slight increase in brevity is worth the loss in explicitness, anymore. What do you think?

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: I was going to expand on this more in the next PR, but I think we can at the least internalize those types (and then remove them after that if need be). If I remember right, we were still using them in the pure swift bson implementation to encode to bytes and what not.

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: Good catch. In the tests I was using `BSON(integerLiteral: x)` but meant to add a better initializer for this (e.g. `BSON(x)` or `BSON(from int: Int)`. For now I've just gone with the no-label option.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: Yep. That'll come in the next PR.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: fixed.

Tests/MongoSwiftTests/BSONValueTests.swift
patrickfreed: oh wow, lol

Sources/MongoSwift/BSON/BSON.swift
kmahar: oooo right, good point

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: Okay, for now I'll just remove it. We can always add new helpers in the future as needed.

PR #327, opened 2019-10-07
SWIFT-501 Conflicting client options from struct and URI

kaywux:
Currently passing client options in through both the ClientOptions struct and the connection string URI has some unintended consequences. It has been decided that values provided in the ClientOptions struct will have precedence over values provided in corresponding URI fields. Documentation has been added to inform users of this. SWIFT-515 has also been affected by this decision and the necessary changes are included in this branch.

JIRA tickets: https://jira.mongodb.org/browse/SWIFT-501, https://jira.mongodb.org/browse/SWIFT-515

Design doc: https://docs.google.com/document/d/1ArrzgIDwPL2ppW1zbhiA9Gki-HGzoNiEeT5gX6rDAeU/edit?usp=sharing

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -199,7 +199,7 @@ public class MongoClient {
     /**
      * Create a new client connection to a MongoDB server. For options that included in both the connection string URI
      * and the ClientOptions struct, the final value is set in descending order of priority: the value specified in
-     * ClientOptions (if non-nil), the value specified in the URI, or the server’s default value if both are unset.
+     * ClientOptions (if non-nil), the value specified in the URI, or the default value if both are unset.
      *
      * - Parameters:
      *   - connectionString: the connection string to connect to.
@@ -363,7 +363,7 @@ public class MongoClient {
 
     /**
      * Gets a `MongoDatabase` instance for the given database name. If an option is not specified in the optional
-     * `DatabaseOptions` param, the database will inherit the value from the parent client or the server default if
+     * `DatabaseOptions` param, the database will inherit the value from the parent client or the default if
      * the client’s option is not set. To override an option inherited from the client (e.g. a read concern) with the
      * default value, it must be explicitly specified in the options param (e.g. ReadConcern(), not nil).
      *
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -40,18 +40,18 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     /// is not specified, the application's default `NotificationCenter` will be used.
     public var notificationCenter: NotificationCenter?
 
-    /// Specifies a ReadConcern to use for the client. To override a read concern set in the URI with the server's
-    /// default, specify an empty ReadConcern here.
+    /// Specifies a ReadConcern to use for the client. Explicitly setting the read concern to `nil` here is the same as
+    /// not specifying one at all.
     public var readConcern: ReadConcern?
 
-    /// Specifies a WriteConcern to use for the client. To override a write concern set in the URI with the server's
-    /// default, specify an empty WriteConcern here.
+    /// Specifies a WriteConcern to use for the client. Explicitly setting the write concern to `nil` here is the same
+    /// as not specifying one at all.
     public var writeConcern: WriteConcern?
 
     // swiftlint:disable redundant_optional_initialization
 
-    /// Specifies a ReadPreference to use for the client. To override a read preference set in the URI with the server's
-    /// default, specify an empty ReadPreference here.
+    /// Specifies a ReadPreference to use for the client. Explicitly setting the read preference to `nil` here is the
+    /// same as not specifying one at all.
     public var readPreference: ReadPreference? = nil
 
     /// Specifies the `DateCodingStrategy` to use for BSON encoding/decoding operations performed by this client and any
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -51,7 +51,7 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     // swiftlint:disable redundant_optional_initialization
 
     /// Specifies a ReadPreference to use for the client. To override a read preference set in the URI with the server's
-    /// default, specify a ReadPreference of mode .primary here.
+    /// default, specify an empty ReadPreference here.
     public var readPreference: ReadPreference? = nil
 
     /// Specifies the `DateCodingStrategy` to use for BSON encoding/decoding operations performed by this client and any
@@ -104,15 +104,15 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
 /// Options to use when retrieving a `MongoDatabase` from a `MongoClient`.
 public struct DatabaseOptions: CodingStrategyProvider {
     /// A read concern to set on the retrieved database. If one is not specified, the database will inherit the
-    /// client's read concern.
+    /// client's read concern. To set the read concern to the server's default, specify an empty ReadConcern.
     public var readConcern: ReadConcern?
 
     /// A read preference to set on the retrieved database. If one is not specified, the database will inherit the
-    /// client's read preference.
+    /// client's read preference. To set the read preference to the server's default, specify an empty ReadPreference.
     public var readPreference: ReadPreference?
 
     /// A write concern to set on the retrieved database. If one is not specified, the database will inherit the
-    /// client's write concern.
+    /// client's write concern. To set the write concern to the server's default, specify an empty WriteConcern.
     public var writeConcern: WriteConcern?
 
     /// Specifies the `DateCodingStrategy` to use for BSON encoding/decoding operations performed by this database and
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -1,9 +1,12 @@
 import Foundation
 import mongoc
 
-/// Options to use when creating a `MongoClient`.
+/// Options to use when creating a `MongoClient`. For fields that are also included in the connection string URI
+/// (ie. readConcern, writeConcern, readPreference, retryWrites), the values specified here will override the ones
```
patrickfreed: We should omit the examples here since the number of overlaps will grow once we have options/uri parity, and we don't want to have to update this string once we do that. Alternatively, can rewrite "ie." to "e.g." to signify that the list isn't comprehensive.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -1,9 +1,12 @@
...
+/// specified in the URI. If they are not specified here or in the URI, the server's default read concern will be used.
```
patrickfreed: "read concern" -> "values"

Sources/MongoSwift/MongoClient.swift
```diff
@@ -1,9 +1,12 @@
...
 public struct ClientOptions: CodingStrategyProvider, Decodable {
-    /// Determines whether the client should retry supported write operations.
+    /// Determines whether the client should retry supported write operations. Explicitly setting retry writes to `nil`
+    /// is the same as not specifying one at all.
```
patrickfreed: instead of including "Explicitly setting ..." in every option's docstring, it would be easier to just include it (maybe in combination with something like the updated string I mentioned below) in the docstring for the options struct itself.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -92,15 +96,15 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
 /// Options to use when retrieving a `MongoDatabase` from a `MongoClient`.
 public struct DatabaseOptions: CodingStrategyProvider {
     /// A read concern to set on the retrieved database. If one is not specified, the database will inherit the
-    /// client's read concern.
+    /// client's read concern. To set the read concern to the server's default, specify an empty ReadConcern.
```
patrickfreed: Similar to my other comment, I think we can consolidate these various repeated sentences to the options structs' docstrings themselves. Also, we want to be emphasizing that a user would only need to do this if a non-default option was set somewhere up the chain, so we could say "To override a option inherited from the parent client (e.g. a read preference) with the default value for that option, it must be explicitly specified (e.g. `ReadConcern()`, not `nil`).", or something along those lines. Ditto for `CollectionOptions`.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -96,6 +96,11 @@ public final class ReadPreference {
         self._readPreference = mongoc_read_prefs_new(mode.readMode)
     }
 
+    /// Initialize a new empty `ReadPreference` with the server's default.
```
patrickfreed: Read preferences are actually a client side feature (we do send them to the server for mongos routing and to determine if secondary reads are okay, though), so there is no "server default". Instead, we can just say "Initializes a new `ReadPreference` with the default mode (primary)".

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -96,6 +96,11 @@ public final class ReadPreference {
...
+    public init() {
+        self._readPreference = mongoc_read_prefs_new(MONGOC_READ_PRIMARY)
```
patrickfreed: to reduce our libmongoc exposure, it might be better if we routed this through an existing initiailzer.
e.g.
```swift
self.init(.primary)
```

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -96,6 +96,11 @@ public final class ReadPreference {
         self._readPreference = mongoc_read_prefs_new(mode.readMode)
     }
 
+    /// Initializes a new ReadPreference with the default mode (primary).
+    public init() {
```
patrickfreed: because this is a class and this initializer now delegates to a different one, you need to add the `convenience` keyword. [This page](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html) gives a good overview of the rules around `init`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -1,7 +1,9 @@
 import Foundation
 import mongoc
 
-/// Options to use when creating a `MongoClient`.
+/// Options to use when creating a `MongoClient`. For options that are also included in the connection string URI, the
```
kmahar: IMO, the second sentence on maybe makes more sense to include in the docstring for the `MongoClient` initializer, so it's clear what the connection string you're referring to is.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -89,18 +89,18 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     }
 }
 
-/// Options to use when retrieving a `MongoDatabase` from a `MongoClient`.
+/// Options to use when retrieving a `MongoDatabase` from a `MongoClient`. If an option is not specified here, the
```
kmahar: similar to my comment above, I think the new text might make more sense to include in the docstring for `MongoClient.db`.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -1,17 +1,17 @@
 import mongoc
 
-/// Options to set on a retrieved `MongoCollection`.
+/// Options to set on a retrieved `MongoCollection`. If an option is not specified here, the collection will inherit the
```
kmahar: similar to my comment above, this might make more sense if it were in the docstring for `MongoDatabase.collection` (both the one that gives you back a `MongoCollection<Document>` as well as the `withType` variant.)

Sources/MongoSwift/ReadPreference.swift
kmahar: there are a few places in the driver we explicitly create a `.primary` read preference to serve as a default. I think it would be good to update those places to call this new init

Sources/MongoSwift/MongoClient.swift
```diff
@@ -202,7 +197,9 @@ public class MongoClient {
     public let writeConcern: WriteConcern?
 
     /**
-     * Create a new client connection to a MongoDB server.
+     * Create a new client connection to a MongoDB server. For options that included in both the connection string URI
+     * and the ClientOptions struct, the final value is set in descending order of priority: the value specified in
+     * ClientOptions (if non-nil), the value specified in the URI, or the server’s default value if both are unset.
```
kmahar: nit: maybe just say "default" instead of "server's default" since some options .e.g `ReadPreference` are actually a driver default 
same applies to other usages

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -41,7 +41,7 @@ internal struct ListDatabasesOperation: Operation {
 
     internal func execute(using connection: Connection, session: ClientSession?) throws -> ListDatabasesResults {
         // spec requires that this command be run against the primary.
-        let readPref = ReadPreference(.primary)
```
patrickfreed: this command must be explicitly run against the primary, so I think we can leave this as-is.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -771,7 +771,7 @@ final class CodecTests: MongoSwiftTestCase {
 
         let rc = ReadConcern(.majority)
         let wc = try WriteConcern(wtimeoutMS: 123)
-        let rp = ReadPreference(.primary)
+        let rp = ReadPreference()
```
patrickfreed: because we're testing encoding here and not the default read preference behavior, it's probably best to explicitly specify which case we expect in the output. for that reason, I think we can leave this one as-is.

Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -8,7 +8,7 @@ final class ReadPreferenceTests: MongoSwiftTestCase {
     }
 
     func testMode() {
-        let primary = ReadPreference(.primary)
```
patrickfreed: I think instead of replacing this case, we should just add a new one verifying the default behavior.

Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -154,18 +154,18 @@ final class ReadPreferenceTests: MongoSwiftTestCase {
     }
 
     func testDatabaseReadPreference() throws {
-        let primary = ReadPreference(.primary)
+        let primary = ReadPreference()
```
patrickfreed: this change and the one on 185 can be undone for similar reasons to the others.

Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -67,11 +67,11 @@ final class ReadPreferenceTests: MongoSwiftTestCase {
     }
 
     func testEquatable() throws {
-        expect(ReadPreference(.primary)).to(equal(ReadPreference(.primary)))
```
patrickfreed: similar to the other cases, we're testing the `Equatable` conformance here, not the default `ReadPreference` behavior, so I think this test can be left as-is.

PR #328, opened 2019-10-07
SWIFT-364, SWIFT-395, SWIFT-609, SWIFT-600 Bump mongoc version to 1.15.1

kmahar:
Bumps min libmongoc version to 1.15.1 ([SWIFT-364](https://jira.mongodb.org/browse/SWIFT-600)) and fixes various new errors introduced by the version bump and switch to having retryable reads on by default.

Also implements a bunch of quick fixes we had depending on that:
* [SWIFT-369](https://jira.mongodb.org/browse/SWIFT-369) `wtimeoutMS` is truncated if set to a value > `Int32.max`
* [SWIFT-609](https://jira.mongodb.org/browse/SWIFT-609) Enable change stream prose test that verifies no resume is attempted for specific error labels
* [SWIFT-600](https://jira.mongodb.org/browse/SWIFT-600) `Decimal128.writeToCurrentPosition` should pass a non-mutable pointer to `bson_iter_overwrite_decimal128`

### DIFF
--- a/Tests/MongoSwiftTests/ChangeStreamTests.swift
+++ b/Tests/MongoSwiftTests/ChangeStreamTests.swift
@@ -527,7 +527,7 @@ final class ChangeStreamTests: MongoSwiftTestCase {
 
         // the next set of assertions relies on the presence of the NonResumableChangeStreamError label, which was
         // introduced in 4.1.1 via SERVER-40446.
-        guard MongoClient.makeTestClient().serverVersion >= ServerVersion(major: 4, minor: 1, patch: 1) else {
+        guard try MongoClient.makeTestClient().serverVersion() >= ServerVersion(major: 4, minor: 1, patch: 1) else {
             return
         }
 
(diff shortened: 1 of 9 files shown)


Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -422,7 +422,7 @@ final class ChangeStreamTests: MongoSwiftTestCase {
             let stage = pipeline[0]
             let streamDoc = stage["$changeStream"] as? Document
             expect(streamDoc).toNot(beNil())
-            return streamDoc!.filter { $0.key != "startAtOperationTime" }
+            return streamDoc!.filter { $0.key != "resumeAfter" }
```
kmahar: this change is due to a bug in libmongoc's resume logic that was fixed in 1.15.1: https://github.com/mongodb/mongo-c-driver/commit/dc2f9c1ea27699727a7e2349f1310178e7630c09

kmahar: actually, I don't think this is the cause, will keep investigating

kmahar: ok, I see why this happened now. as of libmongoc 1.15.0, libmongoc will cache the `postBatchResumeToken` from the initial aggregate command and use that for resuming rather than the operation time, per the spec.
`postBatchResumeToken` exists starting on 4.0.7, and we only run this test on our tested versions that support `failCommand`, which as of now are all greater than 4.0.7.

PR #330, opened 2019-10-08
SWIFT-632 Drop Swift 4.2 support

kmahar:
Drops support for Swift 4.2 and fixes various deprecation warnings that surfaced now that we are no longer using the Swift 5 compiler in 4.2 compatibility mode.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -1377,14 +1377,15 @@ internal func getDecodingError<T: BSONValue>(type: T.Type, decoder: Decoder) -> 
 
 extension Data {
     /// Gets access to the start of the data buffer in the form of an UnsafeMutablePointer<CChar>. Useful for calling C
-    /// API methods that expect a location for a string. **You must only call this method on Data instances with 
+    /// API methods that expect a location for a string. **You must only call this method on Data instances with
     /// count > 0 so that the base address will exist.**
     /// Based on https://mjtsai.com/blog/2019/03/27/swift-5-released/
     fileprivate mutating func withUnsafeMutableCStringPointer<T>(body: (UnsafeMutablePointer<CChar>) throws -> T)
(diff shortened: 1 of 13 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -1374,3 +1374,19 @@ internal func getDecodingError<T: BSONValue>(type: T.Type, decoder: Decoder) ->
     // Non-BSONDecoders are currently unsupported
     return bsonDecodingUnsupportedError(type: T.self, at: decoder.codingPath)
 }
+
+extension Data {
+    /// Gets access to the start of the data buffer in the form of an UnsafeMutablePointer<CChar>. Useful for calling C
+    /// API methods that expect a location for a string. **You must only call this method on Data instances with 
+    /// count > 0 so that the base address will exist.**
+    /// Based on https://mjtsai.com/blog/2019/03/27/swift-5-released/
+    fileprivate mutating func withUnsafeMutableCStringPointer<T>(body: (UnsafeMutablePointer<CChar>) throws -> T)
```
kmahar: the old way we were doing this was deprecated

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -547,3 +556,17 @@ extension Document: Hashable {
         hasher.combine(self.rawBSON)
     }
 }
+
+extension Data {
+    /// Gets access to the start of the data buffer in the form of an UnsafePointer<UInt8>?. Useful for calling C API
+    /// methods that expect the location of an uint8_t. The pointer provided to `body` will be null if the `Data` has
+    /// count == 0.
+    /// Based on https://mjtsai.com/blog/2019/03/27/swift-5-released/
+    fileprivate func withUnsafeBytePointer<T>(body: (UnsafePointer<UInt8>?) throws -> T) rethrows -> T {
```
kmahar: the old way we were doing this was deprecated

PR #331, opened 2019-10-11
SWIFT-133 Implement new Count API

kaywux:
The `count()` function for `MongoCollection` has been deprecated and replaced with `countDocuments()` and `estimatedDocumentCount()`. The new API has been added, along with changes to CRUD and other tests.

JIRA ticket: https://jira.mongodb.org/browse/SWIFT-133

Evergreen: https://evergreen.mongodb.com/version/5da0c38d32f41762b94647cc

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -79,33 +79,10 @@ extension SyncMongoCollection {
         }
     }
 
-    // TODO: SWIFT-133: mark this method deprecated https://jira.mongodb.org/browse/SWIFT-133
     /**
-     * Counts the number of documents in this collection matching the provided filter.
-     *
-     * - Parameters:
-     *   - filter: a `Document`, the filter that documents must match in order to be counted
-     *   - options: Optional `CountOptions` to use when executing the command
-     *   - session: Optional `SyncClientSession` to use when executing this command
-     *
-     * - Returns: The count of the documents that matched the filter
-     *
-     * - Throws:
-     *   - `ServerError.commandError` if an error occurs that prevents the command from performing the write.
-     *   - `UserError.invalidArgumentError` if the options passed in form an invalid combination.
-     *   - `EncodingError` if an error occurs while encoding the options to BSON.
-     */
-    public func count(
-        _ filter: Document = [:],
-        options: CountOptions? = nil,
-        session: SyncClientSession? = nil
-    ) throws -> Int {
-        let operation = CountOperation(collection: self, filter: filter, options: options)
-        return try self._client.executeOperation(operation, session: session)
-    }
-
-    /**
-     * Counts the number of documents in this collection matching the provided filter.
+     * Counts the number of documents in this collection matching the provided filter. Note that an empty filter will
+     * force a scan of the entire collection. For a fast count of the total documents in a collection see
+     * `estimatedDocumentCount`.
      *
      * - Parameters:
      *   - filter: a `Document`, the filter that documents must match in order to be counted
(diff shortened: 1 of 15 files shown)


Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -71,22 +71,8 @@ extension MongoCollection {
...
-     *
-     * - Returns: The count of the documents that matched the filter
-     *
-     * - Throws:
-     *   - `ServerError.commandError` if an error occurs that prevents the command from performing the write.
-     *   - `UserError.invalidArgumentError` if the options passed in form an invalid combination.
-     *   - `EncodingError` if an error occurs while encoding the options to BSON.
-     */
+     /// :nodoc:
+    @available(*, deprecated, message: "Use countDocuments or estimatedDucmentCount instead.")
```
patrickfreed: We plan on removing this for 1.0 right?

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -26,7 +26,8 @@ final class ClientSessionTests: MongoSwiftTestCase {
         (name: "find", body: { _ = try $0.find([:], session: $1).nextOrError() }),
         (name: "aggregate", body: { _ = try $0.aggregate([], session: $1).nextOrError() }),
         (name: "distinct", body: { _ = try $0.distinct(fieldName: "x", session: $1) }),
-        (name: "count", body: { _ = try $0.count(session: $1) })
+        (name: "count", body: { _ = try $0.countDocuments(session: $1) }),
```
patrickfreed: would probably be best to use the full names here (i.e. "countDocuments" and "estimatedDocumentCount")

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -284,10 +294,26 @@ private class BulkWriteTest: CrudTest {
 
 /// A class for executing `count` tests
 private class CountTest: CrudTest {
```
patrickfreed: do we still need this class?

Sources/MongoSwift/MongoCollection+Read.swift
kaywux: Yes, should I just remove the original `count` function altogether?

patrickfreed: We're going to be breaking tons of API next release, so might as well just remove this altogether as part of that.

Tests/MongoSwiftTests/CrudTests.swift
```diff
@@ -79,7 +87,8 @@ final class CrudTests: MongoSwiftTestCase {
 private struct CrudTestFile: Decodable {
     let data: [Document]
     let testDocs: [Document]
-    var tests: [CrudTest] { return try! testDocs.map { try makeCrudTest($0) } }
+    var tests: [CrudTest] { return try!
+        testDocs.filter { ($0["operation"] as? Document)?["name"] as? String != "count" }.map { try makeCrudTest($0) } }
```
patrickfreed: this can probably be left as-is, and instead we just skip the test above in `doTests`. That way all the test skipping is centralized.

kmahar: doing that would require that we keep around the `CountTest` type as a placeholder to parse into. the files mix count and countDocuments so we can't skip a whole file, either.

patrickfreed: Oh, good point. I actually personally prefer the placeholder approach. Imo we should be able to deserialize all the tests, regardless of which ones we're going to run, and the testing logic should decide which ones get skipped. It feels a little weird here to be inspecting the documents directly right before we deserialize them. 

This isn't a huge deal of course, but the placeholder approach is the one I'm taking for the retryable reads tests, so I figured it was worth mentioning here.

kmahar: that's fine with me, switched to using a placeholder.

PR #332, opened 2019-10-14
SWIFT-599 add helper for reading in JSON files

kmahar:
I did this in the process of implementing the auth tests rather than copy-pasting the code another time... putting up in a separate PR to cut down on noise in the auth one.
I also created a new `SpecTestUtils.swift` file where this lives along with some other helpers that are only relevant in the context of spec tests.

### DIFF
--- a/Tests/MongoSwiftTests/ReadWriteConcernTests.swift
+++ b/Tests/MongoSwiftTests/ReadWriteConcernTests.swift
@@ -527,7 +527,7 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
     }
 
     func testConnectionStrings() throws {
-        let testFiles = try retrieveSpecTestFiles(specName: "read-write-concern", 
+        let testFiles = try retrieveSpecTestFiles(specName: "read-write-concern",
                                                   subdirectory: "connection-string",
                                                   asType: Document.self)
         for (_, asDocument) in testFiles {
@@ -564,9 +564,9 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
 
     func testDocuments() throws {
         let encoder = BSONEncoder()
-        let testFiles = try retrieveSpecTestFiles(specName: "read-write-concern", 
-                                          subdirectory: "document",
-                                          asType: Document.self)
+        let testFiles = try retrieveSpecTestFiles(specName: "read-write-concern",
+                                                  subdirectory: "document",
+                                                  asType: Document.self)
 
         for (_, asDocument) in testFiles {
             let tests: [Document] = try asDocument.get("tests")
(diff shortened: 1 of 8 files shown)


Tests/MongoSwiftTests/SpecTestRunner/SpecTestUtils.swift
```diff
@@ -0,0 +1,69 @@
...
+    init(fromJSONFile file: URL) throws {
+        let jsonString = try String(contentsOf: file, encoding: .utf8)
+        try self.init(fromJSON: jsonString)
+    }
+}
+
+/// Given a spec folder name (e.g. "crud") and optionally a subdirectory name for a folder (e.g. "read") retrieves an
+/// array of [(filename, file decoded to type T)].
+internal func retrieveSpecTestFiles<T: Decodable>(specName: String, 
+                                                subdirectory: String? = nil,
```
patrickfreed: probably need an indent here and in the line below

Tests/MongoSwiftTests/SpecTestRunner/SpecTestUtils.swift
```diff
@@ -0,0 +1,69 @@
...
+                                                asType: T.Type) throws -> [(String, T)] {
```
patrickfreed: this tuple approach is much better than what I was doing before with setting the name after initialize the file...

Tests/MongoSwiftTests/SpecTestRunner/SpecTestUtils.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
```
patrickfreed: good idea to introduce one of these. the single `TestUtils` was bound to get bloated with spec test stuff eventually (if it wasn't already).

Tests/MongoSwiftTests/ReadWriteConcernTests.swift
```diff
@@ -565,11 +564,11 @@ final class ReadWriteConcernTests: MongoSwiftTestCase {
 
     func testDocuments() throws {
         let encoder = BSONEncoder()
-        let docsPath = "\(MongoSwiftTestCase.specsPath)/read-write-concern/tests/document"
-        let testFiles = try FileManager.default.contentsOfDirectory(atPath: docsPath).filter { $0.hasSuffix(".json") }
-        for filename in testFiles {
-            let testFilePath = URL(fileURLWithPath: "\(docsPath)/\(filename)")
-            let asDocument = try Document(fromJSONFile: testFilePath)
+        let testFiles = try retrieveSpecTestFiles(specName: "read-write-concern", 
+                                          subdirectory: "document",
```
patrickfreed: ditto indent

Tests/MongoSwiftTests/SpecTestRunner/SpecTestUtils.swift
kmahar: lol good catch i totally forgot to lint. i usually ignore the indicators sublime shows me in the tests/ directory because 95% of them are for `explicit_acl` (plugin is not smart enough to handle nested configurations... yet)

PR #333, opened 2019-10-14
SWIFT-429 Implement auth spec tests

kmahar:
This implements the auth JSON and prose tests as best as we can right now.

I am putting this up as a PR against the spec file helper branch for now so it can be reviewed separately. I'll switch the base branch to master once that gets merged.

FYI this fails on Travis Swift 4.2 right now. I didn't bother special-casing the call to `mongoc_uri_get_mechanism_properties` as I expect that #330 dropping 4.2 will be merged before this is.

I also made a somewhat related change to `ConnectionString` as part of this, will explain more in inline comment.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -5,45 +5,41 @@ internal class ConnectionString {
     /// Pointer to the underlying `mongoc_uri_t`.
     internal let _uri: OpaquePointer
 
-    /// Initializes a new `ConnectionString` with the provided options. Updates `options` to correctly reflect the
-    /// final options set on the `ConnectionString`.
-    internal init(_ connectionString: String, options: inout ClientOptions) throws {
+    /// Initializes a new `ConnectionString` with the provided options.
+    internal init(_ connectionString: String, options: ClientOptions? = nil) throws {
         var error = bson_error_t()
         guard let uri = mongoc_uri_new_with_error(connectionString, &error) else {
             throw extractMongoError(error: error)
         }
         self._uri = uri
 
-        if let rc = options.readConcern {
+        if let rc = options?.readConcern {
             self.readConcern = rc
         }
-        options.readConcern = self.readConcern
 
-        if let wc = options.writeConcern {
+        if let wc = options?.writeConcern {
             self.writeConcern = wc
         }
-        options.writeConcern = self.writeConcern
 
-        if let rp = options.readPreference {
+        if let rp = options?.readPreference {
             self.readPreference = rp
         }
-        options.readPreference = self.readPreference
 
-        if let rw = options.retryWrites {
+        if let rw = options?.retryWrites {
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -5,45 +5,33 @@ internal class ConnectionString {
     /// Pointer to the underlying `mongoc_uri_t`.
     internal let _uri: OpaquePointer
 
-    /// Initializes a new `ConnectionString` with the provided options. Updates `options` to correctly reflect the
-    /// final options set on the `ConnectionString`.
-    internal init(_ connectionString: String, options: inout ClientOptions) throws {
+    /// Initializes a new `ConnectionString` with the provided options.
+    internal init(_ connectionString: String, options: ClientOptions? = nil) throws {
```
kmahar: for some reason I wrote this stupid `inout` API where the connection string would update the client options struct to reflect the final values on the URI so that the `MongoClient` could have correct values for read concern, write concern, and read preference.
however, it's way simpler to just have `MongoClient` read back those values by calling the appropriate getters on the connection string after it creates it (see `MongoClient.init`.)

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -5,45 +5,33 @@ internal class ConnectionString {
...
 
-        if let rw = options.retryWrites {
+        if let rw = options?.retryWrites {
             mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYWRITES, rw)
         }
-        if let rr = options.retryReads {
-            mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYREADS, rr)
-        }
 
-        // we can't get values for retryReads and retryWrites individually, so we will read
```
kmahar: the client doesn't get the options struct back anymore, and doesn't actually care what the values of retryReads and retryWrites are, so we don't need to worry about retrieving those values anymore to send back.

Tests/MongoSwiftTests/AuthTests.swift
```diff
@@ -0,0 +1,300 @@
...
+                    return false
+                default:
+                    return str
+                }
+            }
+        }
+    }
+
+    /// Returns the credential configured on this URI. Will be empty if no options are set.
+    fileprivate var credential: Credential {
```
patrickfreed: I imagine that eventually we would want to incorporate this into `ClientOptions`, as well as all other options supported by the URI. I filed SWIFT-640 for the specific auth options case, but perhaps we would want a "URI/ClientOptions Parity" epic for the entire project.

Perhaps add a TODO here or somewhere else in this file for updating the tests to go over the `ClientOptions` path too?

kmahar: I added a TODO by the prose tests, since those are what the spec says should be tested via both URI and options struct.

kmahar: also opened [SWIFT-641](https://jira.mongodb.org/browse/SWIFT-641) with your proposed epic

PR #336, opened 2019-10-15
SWIFT-630 Integrate the BSON enum into the public API

patrickfreed:
[SWIFT-630](https://jira.mongodb.org/browse/SWIFT-630)

This PR updates the public API and tests to use the new `BSON` enum. As part of this, the following changes have been made:
- The `BSONValue` protocol is now internal
- `Array` conditionally conforms to `BSONValue` when `Element == BSON`
- `AnyBSONValue` has been removed
- `bsonEquals` and its associated Nimble matcher have been removed

Most of the implementation of the BSON library has remained the same with the exception of the encoder/decoder, which required some changes.


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -4,56 +4,59 @@ import Foundation
...
-    maxKey = 0x7f
+    case maxKey = 0x7f
 }
 
 /// A protocol all types representing `BSONType`s must implement.
-public protocol BSONValue {
+internal protocol BSONValue: Codable {
     /// The `BSONType` of this value.
-    var bsonType: BSONType { get }
+    static var bsonType: BSONType { get }
```
patrickfreed: each `BSONValue` corresponds to exactly one BSON type now, so this property can be static.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -69,119 +72,36 @@ public protocol BSONValue {
...
-        return self == otherAsSelf
+extension BSONValue {
+    internal var bsonType: BSONType {
+        return type(of: self).bsonType
     }
 }
 
-/// A protocol that numeric `BSONValue`s should conform to. It provides functionality for converting to BSON's native
-/// number types.
-public protocol BSONNumber: BSONValue {
```
patrickfreed: this is all implemented on `BSON` now

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -197,50 +117,33 @@ extension Array: BSONValue {
...
-            return false
-        }
-        return self.count == otherArr.count && zip(selfArr, otherArr).allSatisfy { lhs, rhs in lhs.bsonEquals(rhs) }
-    }
 }
 
 /// A struct to represent the BSON null type.
-public struct BSONNull: BSONValue, Codable, Equatable {
-    public var bsonType: BSONType { return .null }
+internal struct BSONNull: BSONValue, Codable, Equatable {
```
patrickfreed: users have no need of this struct anymore unless they wan't to specifically decode `BSONNull` as a field in their structs, but that doesn't seem like a common or useful case. So now all the `BSONValue`s that aren't included in `BSON` are internal (`BSONNull`, `BSONUndefined`, `MinKey`, `MaxKey`).

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -586,223 +499,90 @@ extension Decimal128: Hashable {
...
-        return iter.withBSONIterPointer { iterPtr in
+        return .double(iter.withBSONIterPointer { iterPtr in
             self.init(bson_iter_double(iterPtr))
-        }
-    }
-}
-
-/// An extension of `Int` to represent the BSON Int32 or Int64 type.
-/// On 64-bit systems, `Int` corresponds to a BSON Int64. On 32-bit systems, it corresponds to a BSON Int32.
-extension Int: BSONNumber {
```
patrickfreed: since `Int` doesn't correspond to a `BSON` case, it no longer is a `BSONValue`.

`Int` can still be used with our BSON library in the following ways:
- Encoding/Decoding (e.g. having an `Int` property of a struct)
- Integer literals (e.g. `let x: Document = ["x": 1]`)
- through an initializer on `BSON` (e.g. `let x: Document = ["x": BSON(someInt)]`)
- through a conversion accessor on `BSON` (e.g. `x["x"]?.asInt()`)

Users also always have the option of explicit conversion (e.g. `x["x"] = .int64(Int64(someInt))`)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -853,17 +622,56 @@ public struct CodeWithScope: BSONValue, Equatable, Codable, Hashable {
             let scopeDoc = Document(stealing: scopeData)
 
             return self.init(code: code, scope: scopeDoc)
+        })
+    }
+}
+
+/// A struct to represent the BSON Code type.
+public struct Code: BSONValue, Equatable, Codable, Hashable {
```
patrickfreed: Keeping with the 1:1 mapping between `BSONValue` and `BSON`, we now have separate `Code` and `CodeWithScope` structs.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -323,6 +324,22 @@ extension Document {
         self._storage = DocumentStorage()
     }
 
+    internal init(keyValuePairs: [(String, BSON)]) {
```
patrickfreed: this was previously just contained in `Document`'s `ExpressibleByDictionaryLiteral` conformance, but I needed to move it here so that `BSON` could use it in its own `ExpressibleByDictionaryLiteral` conformance.

Tests/MongoSwiftTests/SpecTestRunner/Match.swift
```diff
@@ -10,34 +10,58 @@ internal protocol Matchable {
     /// This assumes `expected` is NOT a placeholder value (i.e. 42/"42"). Use `matches` if `expected` may be a
     /// placeholder.
     /// https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests#spec-test-match-function
-    func contentMatches(expected: Any) -> Bool
```
patrickfreed: the new API allowed me to rewrite most of the `Matchable` stuff in a much better way, which was cool.

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -240,7 +240,7 @@ internal class _BSONDecoder: Decoder {
     }
 
     /// Initializes `self` with the given top-level container and options.
-    fileprivate init(referencing container: BSONValue,
+    fileprivate init(referencing container: BSON,
```
kmahar: I realize this is just preserving existing behavior, but given that `decode` only takes in a `Document` is there a reason why we shouldn't just take a `Document` here?

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -314,39 +314,54 @@ internal struct _BSONDecodingStorage {
...
+                                          debugDescription: "Expected a non-null type."))
+        }
+
+        guard let typed = f(value) else {
+            throw DecodingError._typeMismatch(at: self.codingPath, expectation: T.self, reality: value.bsonValue)
+        }
+        return typed
+    }
+
+    fileprivate func unboxBSONValue<T: BSONValue>(_ value: BSON, as type: T.Type) throws -> T {
```
kmahar: can you explain what the difference is between `unboxBSON` and`unboxBSONValue` now / why we need both? (possibly should go into docstrings.)

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -511,14 +535,6 @@ private struct _BSONKeyedDecodingContainer<K: CodingKey>: KeyedDecodingContainer
         return entry
     }
 
-    /// Decode a BSONValue type from this container for the given key.
-    private func decodeBSONType<T: BSONValue>(_ type: T.Type, forKey key: Key) throws -> T {
-        let entry = try getValue(forKey: key)
-        return try self.decoder.with(pushedKey: key) {
```
kmahar: is still `with(pushedKey:)` thing still called on the new code paths that used to use this?

Tests/MongoSwiftTests/SpecTestRunner/Match.swift
```diff
@@ -47,38 +71,38 @@ extension Array: Matchable {
...
 
-/// Extension that adds MATCHES functionality to `BSONValue`.
-extension BSONValue {
-    internal func bsonMatches(expected: BSONValue) -> Bool {
-        if let matchable = self as? Matchable {
-            return matchable.matches(expected: expected)
+/// Extension that adds MATCHES functionality to `BSON`.
+extension BSON: Matchable {
+    internal func isPlaceholder() -> Bool {
+        return self.asInt() == 42 || self.stringValue == "42"
```
kmahar: rather than hard-coding `42` and `"42"` in twice, maybe it would be better to just check whether `self.asInt().isPlaceholder() || self.stringValue?.isPlaceholder()`?

Tests/MongoSwiftTests/SpecTestRunner/Match.swift
```diff
@@ -10,34 +10,58 @@ internal protocol Matchable {
...
-extension Array: Matchable {
-    internal func contentMatches(expected: Any) -> Bool {
-        guard let expected = expected as? [Any], expected.count <= self.count else {
+extension Array: Matchable where Element: Matchable {
+    internal func contentMatches(expected: [Element]) -> Bool {
+        guard expected.count <= self.count else {
             return false
         }
 
         for (aV, eV) in zip(self, expected) {
```
kmahar: if you feel like it you could one line this with `allSatisfy`

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -68,12 +68,12 @@ internal struct ListDatabasesOperation: Operation {
...
         }
 
-        guard let databases = reply["databases"] as? [Document] else {
+        guard let databases = reply["databases"]?.arrayValue?.compactMap({ $0.documentValue }) else {
             throw RuntimeError.internalError(message: "Invalid server response: \(reply)")
         }
 
         if self.nameOnly ?? false {
-            return .names(databases.map { $0["name"] as? String ?? "" })
+            return .names(databases.map { $0["name"]?.stringValue ?? "" })
```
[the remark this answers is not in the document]
patrickfreed: fixed

Sources/MongoSwift/Operations/WatchOperation.swift
```diff
@@ -24,7 +24,7 @@ internal struct WatchOperation<CollectionType: Codable, ChangeStreamType: Codabl
                   options: ChangeStreamOptions?,
                   stealing connection: Connection) throws {
         self.target = target
-        self.pipeline = pipeline
+        self.pipeline = .array(pipeline.map { .document($0) })
```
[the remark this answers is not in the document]
patrickfreed: I think what you said generally holds. If we need to use a value for its `BSONValue`, don't put it in a `BSON` unless we need to (e.g. for putting into a document).

In this case, we don't need the pipeline except to put it into a document, so I figured it could just be mapped to `BSON` in the constructor. If we needed to inspect the pipeline in `execute`, we would want to keep it as `[Document]`.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -85,7 +85,7 @@ public struct ReadConcern: Codable {
 
     /// Initializes a new `ReadConcern` from a `Document`.
     public init(_ doc: Document) {
-        if let level = doc["level"] as? String {
+        if let level = doc["level"]?.stringValue {
```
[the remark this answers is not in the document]
patrickfreed: Makes me wonder if we need this initializer at all really given the `Codable` conformance.

Tests/MongoSwiftTests/ClientSessionTests.swift
```diff
@@ -321,9 +321,11 @@ final class ClientSessionTests: MongoSwiftTestCase {
         try client.withSession { session in
             let date = Date()
             expect(session.clusterTime).to(beNil())
-            let newTime: Document = ["clusterTime": Timestamp(timestamp: Int(date.timeIntervalSince1970), inc: 100)]
+            let newTime: Document = [
+                "clusterTime": .timestamp(Timestamp(timestamp: Int(date.timeIntervalSince1970), inc: 100))
```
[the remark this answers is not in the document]
patrickfreed: Yeah, it's especially unfortunate that a field of a `Timestamp` is "timestamp". That said, I feel like you only ever read this type like you say, so it shouldn't be a huge problem.

Tests/MongoSwiftTests/MongoError+Equatable.swift
```diff
@@ -54,11 +54,8 @@ extension ServerError: Equatable {
 
 extension BulkWriteResult: Equatable {
     public static func == (lhs: BulkWriteResult, rhs: BulkWriteResult) -> Bool {
-        let iidsEqual = lhs.insertedIds.mapValues { AnyBSONValue($0) } == rhs.insertedIds.mapValues { AnyBSONValue($0) }
-        let uidsEqual = lhs.upsertedIds.mapValues { AnyBSONValue($0) } == rhs.upsertedIds.mapValues { AnyBSONValue($0) }
-
-        return iidsEqual
-                && uidsEqual
+        return lhs.insertedIds == rhs.insertedIds
```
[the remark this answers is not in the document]
patrickfreed: The compiler doesn't synthesize it here for some reason. It must have to do with it being in an extension outside of the originating module. We could just have `BulkWriteResult` conform to `Equatable`, but that would be a public API change that we'd have to commit to.

Tests/MongoSwiftTests/SpecTestRunner/Match.swift
patrickfreed: yep good point. updated.

Tests/MongoSwiftTests/SpecTestRunner/Match.swift
patrickfreed: good idea, updated.

Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -487,254 +496,223 @@ final class CodecTests: MongoSwiftTestCase {
...
+        let x: BSON
 
-        init(_ x: BSONValue) {
-            self.x = AnyBSONValue(x)
+        init(_ x: BSON) {
+            self.x = x
         }
     }
 
     // test encoding/decoding AnyBSONValues with BSONEncoder and Decoder
```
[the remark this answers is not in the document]
patrickfreed: done

Sources/MongoSwift/ReadConcern.swift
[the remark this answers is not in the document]
kmahar: this is only used in tests and I believe was added before we had codable conformance (back when this wrapped a `mongoc_read_concern_t`. 
I think we should remove this from the public API pre-1.0 and update our tests to use a decoder instead. opened [SWIFT-645](https://jira.mongodb.org/browse/SWIFT-645)

Sources/MongoSwift/BSON/BSONDecoder.swift
kmahar: opened [SWIFT-646](https://jira.mongodb.org/browse/SWIFT-646)

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -592,8 +614,10 @@ private struct _BSONKeyedDecodingContainer<K: CodingKey>: KeyedDecodingContainer
         return try self.decoder.with(pushedKey: key) {
             let value = try getValue(forKey: key)
 
-            guard let array = value as? [BSONValue] else {
-                throw DecodingError._typeMismatch(at: self.codingPath, expectation: [BSONValue].self, reality: value)
+            guard let array = value.arrayValue else {
+                throw DecodingError._typeMismatch(at: self.codingPath,
+                                                  expectation: [BSONValue].self,
```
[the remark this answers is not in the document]
patrickfreed: whoops, fixed.

Sources/MongoSwift/BSON/BSONDecoder.swift
kmahar: what if you are decoding e.g. a `Bool` or a `String` (line 582, 597)? it seems to me like those can get called directly now without adding the key that is being decoded to the key path

patrickfreed: Ohh, I misunderstood what you were saying. Yeah that's a very good point; I shouldn't be using an "unboxing" method to do "decoding". I've gone ahead and reverted removing this method and made the two examples you mentioned call it again.

PR #339, opened 2019-10-18
Introduce SwiftFormat

patrickfreed:
This PR runs [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) on the driver and adds  it as part of our linting task.

In general, I tried to emulate our swiftlint config as much as possible so that this formatter would help us enforce existing rules rather than making up new ones. The one notable exception to that is the [wrapArguments](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapArguments) rule, which I've standardized to "before-first". I don't think there is a right answer here as to which option is better, but I do think that selecting _any_ option for consistency is the right idea (as @Utagai [said long ago](https://docs.google.com/a/10gen.com/document/d/1hAE9PyPq0zb0eNHTBwFJ8mfdfk1FZQo0ZSYucWCMsX4/edit?disco=AAAACYUBAVY)). The majority of the diff from this PR is due to this rule. 

e.g.
```swift
public func hello(a: Int, 
                  b: Int) throws -> Int { ... }

let a = try hello(a: 1,
                  b: 2)
```
becomes
```swift
public init(
    a: Int,
    b: Int
) throws { ... }

let a = try hello(
    a: 1,
    b: 2
)
```
In general, I found that this approach to wrapping produces far more readable and condensed code than only doing the "after-first" option, so I went with it.

Another thing to note is that this rule only applies to _how_ we format multiline arguments; it doesn't decide _when_ we do it. 

e.g. the following will be untouched by the formatter:
```swift
public func a(a: Int, b: Int) throws -> Int { ... }

let b = a(a: 1, b: 2)
```
This means that not all output produced by the formatter will pass `swiftlint` due to line length issues, so manual line breakage may be necessary. The formatter will make sure that the way the lines are broken is fixed up, though.

**Why not `swift-format`?**
There is an official formatter being worked on by Apple called [`swift-format`](https://github.com/apple/swift-format). I decided not to use it for this because it supports a very small number of rules compared to SwiftFormat. I think that long term, it may make sense to start using the official formatter once it becomes more mature.


Tests/MongoSwiftTests/CodecTests.swift
```diff
@@ -159,23 +161,25 @@ final class CodecTests: MongoSwiftTestCase {
...
 
         // check that a UInt32 too large for an Int32 gets converted to Int64
-        expect(try encoder.encode(Numbers(uint32: 4294967295))).to(equal(["uint32": .int64(4294967295)]))
+        expect(try encoder.encode(Numbers(uint32: 4_294_967_295))).to(equal(["uint32": .int64(4_294_967_295)]))
 
         // check that UInt, UInt64 too large for an Int32 gets converted to Int64
-        expect(try encoder.encode(Numbers(uint64: 4294967295))).to(equal(["uint64": .int64(4294967295)]))
-        expect(try encoder.encode(Numbers(uint: 4294967295))).to(equal(["uint": .int64(4294967295)]))
+        expect(try encoder.encode(Numbers(uint64: 4_294_967_295))).to(equal(["uint64": .int64(4_294_967_295)]))
+        expect(try encoder.encode(Numbers(uint: 4_294_967_295))).to(equal(["uint": .int64(4_294_967_295)]))
```
patrickfreed: adding underscores in large number literals is another interesting change that's on by default. Let me know what you guys think about it.

kmahar: I like it 👍

Guides/Development.md
```diff
@@ -78,6 +78,10 @@ To regenerate the files, run `make documentation` from the project's root direct
 ## Linting and Style
 We use [SwiftLint](https://github.com/realm/SwiftLint#using-homebrew) for linting. You can see our configuration in the `.swiftlint.yml` file in the project's root directory.  Run `swiftlint` in the `/Sources` directory to lint all of our files. Running `swiftlint autocorrect` will correct some types of violations.
 
+We use [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) for formatting the code. You can see our configuration in the `.swiftformat` file in the project's root directory. Our linter config contains a superset of the rules that our formatter does, so some manual tweaking may be necessary to satisy both once the formatter is run (e.g. line length enforcement). Most of the time, the formatter should put the code into a format that passes the linter.
```
kmahar: satisy -> satisfy

.swiftformat
```diff
@@ -0,0 +1,36 @@
...
+--disable trailingCommas
+
+# first element of collection on newline after opening brace/bracket
+--wrapcollections "before-first"
+
+# put first argument on its own line
+--wraparguments "before-first"
+
+# if x == 1 vs if 1 == x
+--disable yodaConditions
```
kmahar: this one seems reasonable to leave on to me, 1 == x looks kinda weird

patrickfreed: yodaConditions actually enforces the 1 == x, so it's being disabled here; I agree, look weird it does. It's called a "yoda" condition because if read it sounds like you're speaking like Yoda from star wars...lol

.swiftlint.yml
```diff
@@ -53,6 +53,7 @@ excluded:
   - Package.swift
   - Examples/*/Package.swift
   - Tests/LinuxMain.swift
+  - SwiftFormat
```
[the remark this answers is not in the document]
patrickfreed: done

Guides/Development.md
patrickfreed: good catch, fixed

.swiftformat
kmahar: lmao who would want that.....

PR #340, opened 2019-10-22
SWIFT-642 Update examples to use the new BSON API

patrickfreed:
This PR updates the examples and the readme to use the new API.

### DIFF
--- a/Examples/Docs/Sources/DocsExamples/main.swift
+++ b/Examples/Docs/Sources/DocsExamples/main.swift
@@ -24,7 +24,7 @@ private func causalConsistency() throws {
     try items.insertOne(["sku": "nuts-111", "name": "Pecans", "start": .datetime(currentDate)], session: s1)
     // End Causal Consistency Example 1
 
-    let client2 = try MongoClient()
+    let client2 = try SyncMongoClient()
 
     // Start Causal Consistency Example 2
     try client2.withSession(options: ClientSessionOptions(causalConsistency: true)) { s2 in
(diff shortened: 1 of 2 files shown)


Examples/Docs/Sources/DocsExamples/main.swift
```diff
@@ -17,11 +17,11 @@ private func causalConsistency() throws {
...
-        update: ["$set": ["end": currentDate] as Document],
+        filter: ["sku": "111", "end": .null],
+        update: ["$set": ["end": .datetime(currentDate)]],
         session: s1
     )
-    try items.insertOne(["sku": "nuts-111", "name": "Pecans", "start": currentDate], session: s1)
+    try items.insertOne(["sku": "nuts-111", "name": "Pecans", "start": .datetime(currentDate)], session: s1)
     // End Causal Consistency Example 1
 
     let client2 = try MongoClient()
```
kmahar: does this need to be updated too to make these compile?

kmahar: we have [this ticket](https://jira.mongodb.org/browse/SWIFT-536) open about making sure examples still compile when we do a release, but I wonder if it would be better to make that an evergreen task instead so we catch these things earlier? what do you think?

patrickfreed: Fixed it to compile. 

I do think we should incorporate compiling the examples into our evg matrix at least by 1.0, but we do have that circularity issue that I don't really know how to get around. Pre 1.0, we do have the problem that the API in the examples might be vastly different than the API in the latest release, but that will go away real soon so I don't think it is a big issue. I think the API difference is why we were only planning to automate it on releases.

kmahar: I added a comment on the ticket about adding this to evg

PR #343, opened 2019-10-24
Introduce base classes for client and sessions

kmahar:
This PR introduces base classes for clients and sessions and updates the synchronous implementations to inherit from them. It also updates the operation layer to use these base classes rather than the sync variants.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -97,7 +97,7 @@ public class ClientSession {
     public let options: ClientSessionOptions?
 
     /// This type is not meant to be instantiated directly. Should only be instantiated via subclasses.
-    private init(client: MongoClient, options: ClientSessionOptions?) throws {
+    fileprivate init(client: MongoClient, options: ClientSessionOptions?) throws {
         self._client = client
         self.options = options
         let connection = try client.connectionPool.checkOut()
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -97,6 +97,7 @@ public class ClientSession {
     public let options: ClientSessionOptions?
 
     /// This type is not meant to be instantiated directly. Should only be instantiated via subclasses.
+    // swiftformat:disable:next redundantFileprivate
     fileprivate init(client: MongoClient, options: ClientSessionOptions?) throws {
         self._client = client
         self.options = options
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -97,7 +97,7 @@ public class ClientSession {
     public let options: ClientSessionOptions?
 
     /// This type is not meant to be instantiated directly. Should only be instantiated via subclasses.
-    fileprivate init(client: MongoClient, options: ClientSessionOptions?) throws {
+    private init(client: MongoClient, options: ClientSessionOptions?) throws {
         self._client = client
         self.options = options
         let connection = try client.connectionPool.checkOut()
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -31,7 +31,7 @@ private func withSessionOpts<T>(
 
 /// A base class for `SyncClientSession` and `AsyncClientSession`.
 public class ClientSession {
-    internal let _client: MongoClient
+    private let _client: MongoClient
 
     /// Error thrown when an inactive session is used.
     internal static let SessionInactiveError = UserError.logicError(message: "Tried to use an inactive session")
@@ -203,6 +203,8 @@ public class ClientSession {
  *   - https://docs.mongodb.com/manual/core/causal-consistency-read-write-concerns/
  */
 public final class SyncClientSession: ClientSession {
+    // we store this in addition to the private `MongoClient` stored in the base class so that we can expose a sync
+    // client in the public API rather than the base type.
     /// The client used to start this session.
     public let client: SyncMongoClient
 
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -29,7 +29,37 @@ private func withSessionOpts<T>(
     return try body(opts)
 }
 
-/// A base class for `SyncClientSession` and `AsyncClientSession`.
+/**
+ * A MongoDB client session for use with clients, databases, and sessions.
+ * This class represents a logical session used for ordering sequential operations.
+ *
+ * This class serves as a base class for both `SyncClientSession` and `AsyncClientSession`. The only difference between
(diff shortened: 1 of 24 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -29,35 +29,10 @@ private func withSessionOpts<T>(
...
- * `withSession` or `startSession`.
- *
- * - SeeAlso:
- *   - https://docs.mongodb.com/manual/core/read-isolation-consistency-recency/#sessions
- *   - https://docs.mongodb.com/manual/core/causal-consistency-read-write-concerns/
- */
-public final class SyncClientSession {
+/// A base class for `SyncClientSession` and `AsyncClientSession`.
+public class ClientSession {
+    internal let _client: MongoClient
```
kmahar: we now store the client both in the base class and here. the reason for this is that exposing the client is (for whatever reason) part of the public API for sessions, and it seems odd for a `SyncClientSession` to just return a `MongoClient` rather than a `SyncMongoClient`.
at the same time, the base class needs access to the parent client to implement e.g. `getConnection`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -252,12 +235,49 @@ public class SyncMongoClient {
...
+     *   - options: optional `ClientOptions` to use for this client
+     *
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/
+     *
+     * - Throws:
+     *   - A `UserError.invalidArgumentError` if the connection string passed in is improperly formatted.
+     *   - A `UserError.invalidArgumentError` if the connection string specifies the use of TLS but libmongoc was not
+     *     built with TLS support.
      */
-    @available(*, deprecated, message: "Use SyncMongoClient(stealing:) instead.")
```
kmahar: I deleted this deprecated init while I was here since it needed to go anyway.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -272,14 +292,8 @@ public class SyncMongoClient {
...
      */
-    public init(stealing pointer: OpaquePointer) {
-        self.connectionPool = ConnectionPool(stealing: pointer)
-        self.encoder = BSONEncoder()
-        self.decoder = BSONDecoder()
-        self.readConcern = nil
-        self.readPreference = ReadPreference()
-        self.writeConcern = nil
-        self.notificationCenter = NotificationCenter.default
+    override public init(stealing pointer: OpaquePointer) {
```
kmahar: I kind of wonder if we should just remove this from the public API altogether. we definitely don't need to add a corresponding async init but I'm not sure we even need this one (or "single" mode in the pool) anymore.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -124,11 +96,11 @@ public final class SyncClientSession {
     /// The options used to start this session.
     public let options: ClientSessionOptions?
 
-    /// Initializes a new client session.
-    internal init(client: SyncMongoClient, options: ClientSessionOptions? = nil) throws {
+    /// This type is not meant to be instantiated directly. Should only be instantiated via subclasses.
+    // swiftformat:disable:next redundantFileprivate
```
kmahar: this seems to be a swiftformat bug, opened https://github.com/nicklockwood/SwiftFormat/issues/493

Sources/MongoSwift/ClientSession.swift
```diff
@@ -29,35 +29,10 @@ private func withSessionOpts<T>(
...
- *
- * To disable causal consistency, set `causalConsistency` to `false` in the `ClientSessionOptions` passed in to either
- * `withSession` or `startSession`.
- *
- * - SeeAlso:
- *   - https://docs.mongodb.com/manual/core/read-isolation-consistency-recency/#sessions
- *   - https://docs.mongodb.com/manual/core/causal-consistency-read-write-concerns/
- */
-public final class SyncClientSession {
+/// A base class for `SyncClientSession` and `AsyncClientSession`.
```
patrickfreed: This base class is going to be the type that is accepted by all of our public methods, so it will probably be a common occurrence for a user viewing the docs to click on the session parameter of one of those methods and then be redirected to this docstring. For that reason, I think we should probably keep the long docstring on this class in addition to the subclasses, with this one perhaps having a section about how the sync/async variants derive from this class.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -216,3 +173,57 @@ public final class SyncClientSession {
...
+
+/**
+ * A MongoDB client session for use with synchronous clients, databases, and sessions.
+ * This class represents a logical session used for ordering sequential operations.
+ *
+ * To create a client session, use `startSession` or `withSession` on a `SyncMongoClient`.
+ *
+ * If `causalConsistency` is not set to `false` when starting a session, read and write operations that use the session
+ * will be provided causal consistency guarantees depending on the read and write concerns used. Using "majority"
+ * read and write preferences will provide the full set of guarantees. See
```
patrickfreed: this diff has made me notice an existing typo here: this should be "concerns" not "preferences"

Sources/MongoSwift/ClientSession.swift
```diff
@@ -216,3 +173,57 @@ public final class SyncClientSession {
...
+ * https://docs.mongodb.com/manual/core/read-isolation-consistency-recency/#sessions for more details.
+ *
+ * e.g.
+ *   ```
```
patrickfreed: I wonder if jazzy supports the `swift` word here for syntax highlighting, or if it does it automatically.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -252,12 +236,50 @@ public class SyncMongoClient {
...
+}
+
+/// A synchronous MongoDB Client.
+public class SyncMongoClient: MongoClient {
+    private let operationExecutor: SyncOperationExecutor = DefaultSyncOperationExecutor()
+
     /**
-     * :nodoc:
+     * Create a new client connection to a MongoDB server. For options that included in both the connection string URI
+     * and the ClientOptions struct, the final value is set in descending order of priority: the value specified in
```
patrickfreed: Since these docstrings are markdown, we can write \`ClientOptions\`. Jazzy might also automatically link it to the docstring for `ClientOptions` if the backticks are used.

Sources/MongoSwift/MongoClient.swift
patrickfreed: I think given the current mobile situation, this can probably be removed.

Sources/MongoSwift/ClientSession.swift
kmahar: fixed.

Sources/MongoSwift/ClientSession.swift
kmahar: good point. updated

kmahar:
closing as I think this is obsoleted by recent changes to the async design.

PR #350, opened 2019-11-20
SWIFT-660 Use reference counting to check in Connections

kmahar:
This PR does the following:
- Make `Connection` a class, which automatically checks the `mongoc_client_t` back into the pool upon `deinit`
    - This enables us to remove `releaseConnection` and just ensure that any place we were calling it we now drop our reference to the `Connection` being released
- Add the override option to pass a particular `Connection` into `OperationExecutor.execute`, and use this for executing `NextOperation`s
- Remove `ConnectionStrategy`, as the combination of the previous bullet point plus the ability to simply "steal" a `Connection` via maintaining a reference to it makes the strategies obsolete
- I've also updated the `MongoCursor` init to look more like the one for `ChangeStream` and no longer take in a closure that constructs a `mongoc_cursor_t`. Once `find`, `aggregate`, and `startSession` are operationized in upcoming PRs, this will make it so the only place we ever check out a `Connection` is directly within the executor.

### DIFF
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -1,12 +1,24 @@
 import mongoc
 
 /// A connection to the database.
-internal struct Connection {
+internal class Connection {
     /// Pointer to the underlying `mongoc_client_t`.
     internal let clientHandle: OpaquePointer
+    /// The pool this connection belongs to.
+    private let pool: ConnectionPool
 
-    internal init(_ clientHandle: OpaquePointer) {
+    internal init(clientHandle: OpaquePointer, pool: ConnectionPool) {
         self.clientHandle = clientHandle
+        self.pool = pool
+    }
+
+    deinit {
+        switch self.pool.state {
+        case let .open(pool):
+            mongoc_client_pool_push(pool, self.clientHandle)
+        case .closed:
+            assertionFailure("ConnectionPool was already closed")
+        }
     }
 }
 
@@ -55,30 +67,19 @@ internal class ConnectionPool {
         self.state = .closed
     }
 
-    /// Checks out a connection. This connection must be returned to the pool via `checkIn`.
+    /// Checks out a connection. This connection will return itself to the pool when its reference count drops to 0.
     internal func checkOut() throws -> Connection {
         switch self.state {
         case let .open(pool):
-            return Connection(mongoc_client_pool_pop(pool))
+            return Connection(clientHandle: mongoc_client_pool_pop(pool), pool: self)
         case .closed:
             throw RuntimeError.internalError(message: "ConnectionPool was already closed")
-        }
-    }
-
-    /// Returns a connection to the pool.
-    internal func checkIn(_ connection: Connection) {
-        switch self.state {
-        case let .open(pool):
-            mongoc_client_pool_push(pool, connection.clientHandle)
-        case .closed:
-            fatalError("ConnectionPool was already closed")
         }
     }
 
     /// Executes the given closure using a connection from the pool.
     internal func withConnection<T>(body: (Connection) throws -> T) throws -> T {
         let connection = try self.checkOut()
-        defer { self.checkIn(connection) }
         return try body(connection)
     }
 
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -24,19 +24,23 @@ extension MongoCollection {
     ) throws -> MongoCursor<CollectionType> {
         let opts = try encodeOptions(options: options, session: session)
         let rp = options?.readPreference?._readPreference
+
+        let connection = try resolveConnection(client: self._client, session: session)
+        let cursor: OpaquePointer = self.withMongocCollection(from: connection) { collPtr in
+            guard let cursor = mongoc_collection_find_with_opts(collPtr, filter._bson, opts?._bson, rp) else {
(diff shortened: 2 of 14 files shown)


Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -1,12 +1,24 @@
...
         self.clientHandle = clientHandle
+        self.pool = pool
+    }
+
+    deinit {
+        switch self.pool.state {
+        case let .open(pool):
+            mongoc_client_pool_push(pool, self.clientHandle)
+        case .closed:
+            assertionFailure("ConnectionPool was already closed")
```
kmahar: this was previously a `fatalError` but I think it is maybe better to not crash in release mode because of this... similar to our choice in the design use use `assert`s to make sure `close` has been called on e.g. `MongoClient`. what do you think?

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -55,30 +67,19 @@ internal class ConnectionPool {
...
-            mongoc_client_pool_push(pool, connection.clientHandle)
-        case .closed:
-            fatalError("ConnectionPool was already closed")
-        }
-    }
-
     /// Executes the given closure using a connection from the pool.
     internal func withConnection<T>(body: (Connection) throws -> T) throws -> T {
         let connection = try self.checkOut()
-        defer { self.checkIn(connection) }
```
kmahar: I debated removing this method altogether, but I figured it might still be useful to us if there is some future situation where you want to be able to quickly release a connection without manually nil-ing out your reference to it.

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -24,19 +24,23 @@ extension MongoCollection {
     ) throws -> MongoCursor<CollectionType> {
         let opts = try encodeOptions(options: options, session: session)
         let rp = options?.readPreference?._readPreference
+
+        let connection = try resolveConnection(client: self._client, session: session)
+        let cursor: OpaquePointer = self.withMongocCollection(from: connection) { collPtr in
```
kmahar: in the next PR all of this will be removed into an operation and we won't need to call `resolveConnection` anymore as it will happen in the executor. I just left it like this for now to cut down on the size of the diff in this PR.

ditto for aggregate below.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -56,34 +56,30 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         self.state = .open(cursor: cursor, connection: connection, client: client, session: session)
         self.cursorType = cursorType ?? .nonTailable
         self.decoder = decoder
         self.error = nil
 
-        if let err = self.getMongocError() {
-            // Errors in creation of the cursor are limited to invalid argument errors, but some errors are reported
-            // by libmongoc as invalid cursor errors. These would be parsed to .logicErrors, so we need to rethrow them
-            // as the correct case.
-            throw UserError.invalidArgumentError(message: err.errorDescription ?? "")
```
kmahar: I messaged about this on Slack before but I'm not sure we ever came to a resolution. but because of the fact that some mongoc methods returning cursors (like the list indexes one) do immediate I/O, this isn't actually true and we can see multiple types of errors show up here, for example, a server selection error. 

something we *could* do is in cases like `find` and `aggregate` where the method doesn't do I/O immediately, we could first see if there is an error and if so convert to invalidArgumentError, *then* trigger I/O, and if that errors then throw it as normal. 

but, in whatever the specific cases you are referring to in this comment are, that might lead to inconsistency in error reporting across cursor methods, because in the immediate I/O cases I think we would still throw those errors as logicErrors even if they should be converted as we can't tell where they came from?

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -117,20 +117,20 @@ internal struct ListCollectionsOperation: Operation {
...
             let cursor: MongoCursor<Document> = try MongoCursor(
+                stealing: indexes,
+                connection: connection,
                 client: self.database._client,
                 decoder: self.database.decoder,
-                session: session,
-                initializer: initializer
+                session: session
             )
             return try .names(cursor.map {
```
kmahar: in future PR I will update this so we aren't iterating the cursor and generating new operations/tieing up an extra thread from within this operation. and also update it so we are making sure this cursor didn't encounter an error while iterating.

Sources/MongoSwift/MongoCursor.swift
patrickfreed: If you look at [`_mongoc_cursor_new_with_opts`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-cursor.c#L230), pretty much every error there is an invalid argument error, but they're all returned as `MONGOC_ERROR_CURSOR_INVALID_CURSOR` errors. libmongoc also happens to use this domain/code for other things, such the error returned after the user iterates a dead cursor. For that reason, our parser in MongoError.swift returns a `logicError` when seeing that domain/code combo, and for the cursor creation specific cases, the original line just took the message out and put it in an argument error. 

I think what we can do is update `parseMongocError` to parse that combo as an argument error instead of a logic error, due to the fact that we now close cursors in Swift land and don't rely on mongoc to report errors for the iteration of dead cursors.

kmahar: that makes sense to me. changed.

PR #351, opened 2019-11-22
SWIFT-661 Introduce find and aggregate operations, and make find/aggregate methods do immediate I/O

kmahar:
Apologies for the size of the diff - most of this is just moving code around as I moved the options plus implementations of find and aggregate into operations files.

I put `Hint` in its own file because it is used by both `FindOptions` and `AggregateOptions` and doesn't have an obvious home otherwise.

I think mostly important to review here are changes to `MongoCursor` and `NextOperation`.


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -67,11 +78,16 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         self.error = nil
-
+        self.cached = .none
+        // If there was an error constructing the cursor, throw it.
         if let error = self.getMongocError() {
             self.close()
             throw error
         }
+
+        // Cache the first document.
```
kmahar: so initially I had planned to put `cacheDocument` in its own method. however, the only time we ever needed to call it was from right here, and there is no harm in always caching the first document. as I alluded to in our conversation earlier I think having it as a standalone method complicates the different states the driver can end up in.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -139,36 +176,47 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
-                    self.close()
-                }
-                return nil
-            }
-            return out
+            return try client.executeOperation(operation, using: connection, session: session)
         } catch {
-            // This indicates that an error occurred executing the `NextOperation`. Currently the only possible error
-            // is a `DecodingError` when decoding a `Document` to this cursor's type.
+            // This indicates that an error occurred executing the `NextOperation`.
```
kmahar: since the logic above checking `self.getMongocError()` now lives in `getNextDocumentFromMongocCursor`, all sorts of errors, not just decoding errors, can be thrown by `executeOperation` and end up being thrown here.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -139,36 +176,47 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
             self.error = error
-            // Since we encountered an error, close the cursor.
-            self.close()
```
kmahar: this is now handled in `getNextDocumentFromMongocCursor`.

Sources/MongoSwift/Operations/NextOperation.swift
```diff
@@ -17,52 +17,39 @@ internal struct NextOperation<T: Codable>: Operation {
...
         case let .cursor(cursor):
-            // We already check this in `MongoCursor.next()` in order to extract the relevant connection and
-            // session, but error again here just in case.
-            guard case let .open(cursorPtr, _, _, _) = cursor.state else {
-                throw ClosedCursorError
-            }
-            guard mongoc_cursor_next(cursorPtr, out) else {
-                return nil
-            }
+            return try cursor.getNextDocumentFromMongocCursor()
```
kmahar: at this point it feels kind of weird to have this be an operation, but given that we need the logic for calling `mongoc_cursor_next` to be reusable for document caching and for the future `ForEachOperation` and `AllOperation`, I'm not sure how to really improve this.

we could consider just directly calling ` try cursor.getNextDocumentFromMongocCursor()` by using the soon-to-be-added threadpool for method for executing an arbitrary non-operation block instead.

Sources/MongoSwift/Operations/FindOperation.swift
```diff
@@ -0,0 +1,229 @@
...
+        self.collection = collection
+        self.filter = filter
+        self.options = options
+    }
+
+    internal func execute(
+        using connection: Connection,
+        session: ClientSession?
+    )
+        throws -> MongoCursor<CollectionType> {
```
patrickfreed: I feel like this line could be on the same as the closing paren, does SwiftFormat not put it there? (If not, does it complain if you put it there?)

Sources/MongoSwift/Operations/NextOperation.swift
patrickfreed: I think using the arbitrary non-operation blocks sounds like a decent idea. Once we go pure swift, this weirdness will go away since we'll be able to access the first batch from the `find` command, then only use `getMore` after that.

Sources/MongoSwift/Operations/FindOperation.swift
kmahar: it does not complain. I maybe did this myself before I ran the formatter.... updated to be on one line.

Sources/MongoSwift/Operations/NextOperation.swift
kmahar: I added a note about this on [SWIFT-668](https://jira.mongodb.org/browse/SWIFT-668) (the ticket where the new `execute` method will be added).

PR #354, opened 2019-11-26
SWIFT-662, SWIFT-652: Decompose operations, and check cursor errors after iterating

kmahar:
The only operation that appears to create other operations from within it right now is `ListCollectionsOperation`.
I've updated that to directly use the new `getNextDocumentFromMongocCursor` method instead of iterating the cursor and generating new ops.

This did make me realize a problem with always caching the first document though: for situations like this, there is no way to retrieve the cached one without calling `next()`. So I added a parameter `cacheFirstDocument` to the `MongoCursor` init which defaults to false.

I also audited the driver for places we iterate cursors and only found the one in `ListCollectionsOperation` plus one in `listIndexNames` so I have fixed both of those.

I noticed that `listIndexNames` is doing a lot of unnecessary decoding work so I have opened [SWIFT-677](https://jira.mongodb.org/browse/SWIFT-677) about optimizing that.

### DIFF
--- a/Sources/MongoSwift/Operations/ListCollectionsOperation.swift
+++ b/Sources/MongoSwift/Operations/ListCollectionsOperation.swift
@@ -132,12 +132,19 @@ internal struct ListCollectionsOperation: Operation {
                 decoder: self.database.decoder,
                 session: session
             )
-            return try .names(cursor.map {
-                guard let name = $0["name"]?.stringValue else {
+
+            var names = [String]()
+            // call the cursor method for getting the next document directly rather than just iterating through the
+            // cursor so that we avoid generating more concurrent operations and tying up more of the executor's
+            // threads.
+            while let nextDoc = try cursor.getNextDocumentFromMongocCursor() {
+                guard let name = nextDoc["name"]?.stringValue else {
                     throw RuntimeError.internalError(message: "Invalid server response: collection has no name")
                 }
-                return name
-            })
+                names.append(name)
+            }
+
+            return .names(names)
         }
         let cursor: MongoCursor<CollectionSpecification> = try MongoCursor(
             stealing: indexes,
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -383,11 +383,16 @@ extension MongoCollection {
     public func listIndexNames(session: ClientSession? = nil) throws -> [String] {
         let operation = ListIndexesOperation(collection: self)
         let models = try self._client.executeOperation(operation, session: session)
-        return try models.map { model in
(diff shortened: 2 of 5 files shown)


Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -132,12 +132,19 @@ internal struct ListCollectionsOperation: Operation {
...
                 session: session
             )
-            return try .names(cursor.map {
-                guard let name = $0["name"]?.stringValue else {
+
+            var names = [String]()
+            // call the cursor method for getting the next document directly rather than just iterating through the
+            // cursor so that we avoid generating more concurrent operations and tying up more of the executor's
+            // threads.
+            while let nextDoc = try cursor.getNextDocumentFromMongocCursor() {
```
patrickfreed: what if instead we just rename this method to something more generic (e.g. `fetchNextDocument`) that will check the cached one first before going to libmongoc? I think that would be a more straightforward approach than introducing a new parameter which is possible to forget. Also, I think that whether the cursor caches a document is really more of an implementation detail of `MongoCursor` that shouldn't necessarily be configurable via the initializer. What do you think?

kmahar: well, the idea is that most of the the time we don't need to remember to use this parameter at all as it defaults to false - I don't imagine it being relevant in any new uses going forward (or imagine us actually adding any more uses.)

that said, I'm not opposed to moving the logic to check for a cached document into this method, but it does introduce a new problem with `next`. presumably with this change the implementation of `next` would then just create and execute a `NextOperation` which would call `fetchNextDocument`, and we would in that way receive either a cached document or a new document (or nothing if we are at the end of the cursor).

however, in the case where a cursor is initially empty or only has one doc, it is possible for us to end up in a state where we have a cached document and the cursor is closed. 
we can't execute a `NextOperation` for a closed cursor though, since we need a `Connection` around to pass into the `execute` method. so how do we get the cached document? 

two options (open to any other ideas you might have too) are:
a) duplicate the cached document check in both `next` and `fetchNextDocument` (seems messy)
b) as we previously discussed, stop using `NextOperation` altogether and just directly call `fetchNextDocument`in `next`, and later on run that call in the thread pool. less messy but moves away from our goal of operationizing all I/O

patrickfreed: Good points all around. I think keeping the approach you have in this PR makes sense in light of all of that. Maybe rename the parameter to something that indicates that it determines whether I/O happens, like `open` or something?

kmahar: renamed to `forceIO`

PR #357, opened 2019-12-02
SWIFT-666 Create MongoSwiftSync module and add sketch of sync API

kmahar:
This introduces a new `MongoSwiftSync` module containing an unimplemented synchronous API. 

This was mostly a copy-paste job where I deleted the source of each method body and replaced it with a fatal error, so I don't think a super thorough review is needed, but you can see whether this looks reasonable anyway.


.swiftformat
```diff
@@ -1,4 +1,4 @@
---exclude Sources/MongoSwift/MongoSwiftVersion.swift
+--exclude Sources/MongoSwift/MongoSwiftVersion.swift,Sources/MongoSwiftSync/*
```
kmahar: I had to disable this because otherwise the formatter automatically replaces each parameter name with an underscore as they are all unused. I opened [SWIFT-680](https://jira.mongodb.org/browse/SWIFT-680) about reenabling it so we don't forget.

kmahar:
Oh yeah, I forgot about that. I ended up writing a sourcery template to generate the exports list for us.
In order to prevent re-exporting the types defined separately in the sync module, I added a `skipSyncExport` annotation to the relevant async-only types in `MongoSwift`. 
I tried to figure out a way to automatically generate a list of overlapping type names to exclude, but Sourcery just works by taking in a bunch of input files and as far as I can tell there is no way to filter by which file or module a type was defined in.

PR #360, opened 2019-12-04
SWIFT-667 Split up test targets

kmahar:
This PR updates `Tests/` to use 4 test targets as follows:

1. `TestsCommon`: This contains code that is shared across multiple test targets, and does not differ across the sync and async APIs. 
2. `BSONTests`: This contains BSON tests. It depends on `TestCommon` and `MongoSwift`, but eventually when we split out the BSON library should just depend on that rather than `MongoSwift`.
3. `MongoSwiftTests`: This currently contains all of the non-BSON tests, along with various sync-specific test helpers. Some of those helpers may eventually get asynchronous versions, depending on what we need when implementing async-specific tests. In the next PR, this target will be updated to contain only tests for types that are shared across the sync and async modules, e.g. `ReadConcern`.
4. `MongoSwiftSyncTests`: This currently is empty. In the next PR, it will be updated to contain tests that are specific to the sync API. This will also include moving over the `SpecTestRunner` as that runs spec tests in a synchronous fashion.


Tests/.swiftlint.yml
```diff
@@ -4,3 +4,4 @@ disabled_rules:
   - force_unwrapping
   - nesting
   - explicit_acl
+  - missing_docs
```
kmahar: now that `TestsCommon` is in use and has public methods, we were getting missing docs errors for a few undocumented things there.

Tests/MongoSwiftTests/SyncTestUtils.swift
```diff
@@ -0,0 +1,161 @@
+import Foundation
```
kmahar: everything in this file used to be in `TestUtils.swift`.

Tests/TestsCommon/CommonTestUtils.swift
```diff
@@ -0,0 +1,172 @@
+import Foundation
```
kmahar: Everything here used to be in `TestUtils.swift`. All that has changed is marking things open and public as needed so they can be accessed from the other test targets.

Tests/TestsCommon/SpecTestUtils.swift
```diff
@@ -1,5 +1,5 @@
 import Foundation
-@testable import MongoSwift
+import MongoSwift
```
kmahar: I moved this file into common because some of these utils are used by the BSON corpus tests.

kmahar:
the placeholder file is necessary so that the `MongoSwiftSyncTests` folder exists (git seems to ignore empty folders). without that folder, SwiftPM errors.

Tests/TestsCommon/CommonTestUtils.swift
```diff
@@ -0,0 +1,172 @@
...
+import mongoc
+@testable import MongoSwift
+import Nimble
+import XCTest
+
+// sourcery: disableTests
```
patrickfreed: does this mean none of the tests are currently running?

Tests/MongoSwiftSyncTests/Placeholder.swift
```diff
@@ -0,0 +1 @@
+
```
patrickfreed: this file probably needs a comment or something to pass the linter

Tests/TestsCommon/CommonTestUtils.swift
kmahar: nah, this is so that sourcery doesn't include the base class in the list of generated tests in LinuxMain, see [here](https://github.com/mongodb/mongo-swift-driver/blob/1cd83228be80c121c6eb0670f77eab8477de7182/Tests/LinuxMain.stencil#L5)

Tests/MongoSwiftSyncTests/Placeholder.swift
kmahar: added one.

PR #362, opened 2019-12-06
SWIFT-643 Add a findOne method

ali-mir:
Added a findOne method and tests.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -47,11 +47,8 @@ extension MongoCollection {
         options: FindOneOptions? = nil,
         session: ClientSession? = nil
     ) throws -> T? {
-        var findOneOptions: FindOptions?
-        if let options = options {
-            findOneOptions = FindOptions(findOneOptions: options)
-        }
-        let cursor = try self.find(filter, options: findOneOptions, session: session)
+        let options = options.map { FindOptions(from: $0) }
+        let cursor = try self.find(filter, options: options, session: session)
         return try cursor.nextOrError()
     }
 
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -34,7 +34,7 @@ extension MongoCollection {
      *   - options: Optional `FindOneOptions` to use when executing the command
      *   - session: Optional `ClientSession` to use when executing this command
      *
-     * - Returns:  the resulting `Document`
+     * - Returns:  the resulting `Document`, or nil if there is no match
      *
      * - Throws:
      *   - `UserError.invalidArgumentError` if the options passed are an invalid combination.
@@ -51,7 +51,7 @@ extension MongoCollection {
         if let options = options {
             findOneOptions = FindOptions(findOneOptions: options)
         }
-        let cursor = try self.find(_: filter, options: findOneOptions, session: session)
+        let cursor = try self.find(filter, options: findOneOptions, session: session)
         return try cursor.nextOrError()
     }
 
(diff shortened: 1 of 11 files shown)


Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -26,6 +26,35 @@ extension MongoCollection {
...
 
+    /**
+     * Finds a single document in this collection that matches the provided filter.
+     *
+     * - Parameters:
+     *   - filter: A `Document` that should match the query
+     *   - options: Optional `FindOneOptions` to use when executing the command
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Returns:  the resulting `Document`
```
kmahar: what happens if there is no match? let's clarify.

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -26,6 +26,35 @@ extension MongoCollection {
...
+    public func findOne(
+        _ filter: Document = [:],
+        options: FindOneOptions? = nil,
+        session: ClientSession? = nil
+    ) throws -> T? {
+        var findOneOptions: FindOptions?
+        if let options = options {
+            findOneOptions = FindOptions(findOneOptions: options)
+        }
+        let cursor = try self.find(_: filter, options: findOneOptions, session: session)
```
kmahar: nit: I think you can just say `filter` instead of `_: filter`. the use of that in the `find` signature is so users don't have to provide a parameter label

Sources/MongoSwift/Operations/FindOperation.swift
```diff
@@ -182,6 +182,23 @@ public struct FindOptions: Codable {
         self.sort = sort
     }
 
+    internal init(findOneOptions: FindOneOptions) {
```
kmahar: I think this initializer also needs to copy over the read preference. I think we should probably also test using a read preference with this method doesn't error in `ReadPreferenceTests.swift`. 

this makes me realize we don't really have a good way right now of ensuring specified read preferences are actually sent to / respected by libmongoc. one way we might be able to test that is by using an unsatisfiable read preference (e.g. containing tags that match no members in the replica set) and ensuring commands fail.... what do you think @patrickfreed ?

I think that's out of scope for this ticket but something we should think about. I've opened [SWIFT-685](https://jira.mongodb.org/browse/SWIFT-685) for discussion.

patrickfreed: That sounds like a reasonable approach 👍 

With regards to the naming of this, our conversion initializerss generally follow the form of `init(from:)`, so this could be updated to match that (e.g. see [`UUID.init(from)`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/BSON/BSONValue.swift#L219))

Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -26,6 +26,35 @@ extension MongoCollection {
...
+     *   - `UserError.logicError` if the provided session is inactive.
+     *   - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+
+    public func findOne(
+        _ filter: Document = [:],
+        options: FindOneOptions? = nil,
+        session: ClientSession? = nil
+    ) throws -> T? {
+        var findOneOptions: FindOptions?
```
patrickfreed: this could be condensed into a [`map`](https://developer.apple.com/documentation/swift/optional/1539476-map):
e.g.
```swift
let options = options.map { FindOptions(findOneOptions: $0) }
// or
let options = options.map(FindOptions.init(findOneOptions:))
```
I personally like using these functional thingies, but it's up to you which you think look better or prefer. I just figured I'd point out that something like this is possible.

Issue #366, opened 2019-12-09
crashes with concurrent connection on OS X: mongoc-socket.c:592 mongoc_socket_errno(): precondition failed: sock


gurugeek:
### Swift version

What does this command give you?
```
Apple Swift version 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9)
Target: x86_64-apple-darwin19.0.0
```

### Operating system

What does this command give you?
```
Darwin Davids-MacBook-Pro.local 19.0.0 Darwin Kernel Version 19.0.0: Wed Oct 23 18:29:05 PDT 2019; root:xnu-6153.41.3~44/RELEASE_X86_64 x86_6
```

### Driver version

What does this command give you?
```
   "version": "0.1.3"
```

### What is the version(s) of `mongod` that you are running with the driver?
Try running:
```
db version v4.2.1
git version: edf6d45851c0b9ee15548f0f847df141764a317e
allocator: system
modules: none
build environment:
    distarch: x86_64
    target_arch: x86_64
```

### How did you install `libmongoc` and `libbson` on your system

brew

### Version of `libmongoc` and `libbson`

What does this command give you?
```
mongo-c-driver 1.15.1
```

## What is the problem? 

code, a simple Route with Kitura framework:

```
router.get("/:name") {request, response, next
        
        in guard let name = request.parameters["name"]
        else {
            _ = response.send(status: .badRequest)
            return next()
        }
        let query: Document = ["slug": name]
        let documents = try collection.find(query)
        // check if an error occurred while iterating the cursor
        if let error = documents.error { throw error
        }
        print("documents")
        print(documents)

        var page: [String: String] = [:]
        for d in documents {
            print(d)
            page = [
                "title": d.title,
                "slug": d.slug,
                "body": d.body,
                "date": d.date,
            ]
            break
        }

        if page.isEmpty {
            // response.send ("<h1>&#127934  not found </h1>")

            let title = ["title": "Page not Found - 404 "]
            try response.render("404.stencil", context: title)
            next()
        } else {
            try response.render("page.stencil", with: page)
            response.status(.OK)
        }
    }

```

runs fine without concurrent connections. A simple test with apache benchmark
ab -n 100 -c 10 http://127.0.0.1:8080/index

triggers the crash 

## Reproducing the bug

a) Kitura route as above (select a document from a database)
b) swift run  ( all will work fine here )

c) ab -n 100 -c 10 http://127.0.0.1:8080/index

[...]

PR #367, opened 2019-12-10
Update example projects to not share collections across threads

kmahar:
Currently we show them doing this.

However we still have not released client pooling which makes collections threadsafe, and this is confusing for users (see: #366).

We should change these back once we have actually tagged our next release. I opened [SWIFT-686](https://jira.mongodb.org/browse/SWIFT-686) with fixVersion 1.0.0-rc0 so we remember to do this.


kmahar:
Oh also I renamed the example project directories to get rid of that circular dependency error.

PR #369, opened 2019-12-11
SWIFT-667, SWIFT-342 hodgepodge of small test changes

kmahar:
This contains a lot of small and I think non-controversial changes I've identified we should make as part of or as a precursor to splitting up the tests.

in particular:
- moving a couple of methods that are only used in tests and are defined internally in `MongoSwift` into `TestsCommon`
- Adding a couple missing components of the sync API
- removing various unused imports to make it easier to see which files can be moved to the sync test suite with no changes

### DIFF
--- a/Sources/MongoSwiftSync/MongoCollection+Read.swift
+++ b/Sources/MongoSwiftSync/MongoCollection+Read.swift
@@ -39,7 +39,6 @@ extension MongoCollection {
      *   - `UserError.logicError` if the provided session is inactive.
      *   - `EncodingError` if an error occurs while encoding the options to BSON.
      */
-
     public func findOne(
         _ filter: Document = [:],
         options: FindOneOptions? = nil,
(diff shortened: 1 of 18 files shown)


Sources/MongoSwiftSync/MongoCollection+Read.swift
```diff
@@ -24,6 +24,30 @@ extension MongoCollection {
...
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Returns:  the resulting `Document`, or nil if there is no match
+     *
+     * - Throws:
+     *   - `UserError.invalidArgumentError` if the options passed are an invalid combination.
+     *   - `UserError.logicError` if the provided session is inactive.
+     *   - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+
```
patrickfreed: extra newline here

kmahar: good catch. this existed in the existing implementation too; fixed both

PR #370, opened 2019-12-11
SWIFT-667 Move majority of tests to MongoSwiftSyncTests

kmahar:
This PR moves over a large number of tests to `MongoSwiftSyncTests`. I've also made some minor changes to support sharing stuff across the test modules and cut down on the number of times we need to `@testable` import in the sync tests module. 

For now, I've just split up one test file, `AuthTests`, into components testing internal and public API that can live in the async and sync test modules, respectively. I will do another PR after this to further split up some of our tests for types like `ReadConcern`.

Also, for now I've left all of these as still using the functioning sync API, since we basically wouldn't be able to run any tests if we switched to the unimplemented one.

As the current API becomes async and the sync one is reimplemented I will switch the imports over.


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -39,9 +39,8 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
         return false
     }
 
-    /// Returns the ID used by the server to track the cursor. `nil` until mongoc actually talks to the server by
-    /// iterating the cursor, and `nil` after mongoc has fetched all the results from the server.
-    internal var id: Int64? {
+    /// Returns the ID used by the server to track the cursor. `nil` once all results have been fetched from the server.
+    public var id: Int64? {
```
kmahar: I think we decided we wanted this to be public anyway, and I made it public in the new sync sketch. I only noticed it now because `MongoCursorTests` was using it and I removed the `@testable` from the import there.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -208,3 +208,22 @@ extension TopologyDescription.TopologyType {
         }
     }
 }
+
+public struct TestError: LocalizedError {
```
kmahar: this was being used across a few different test files and it seems general enough that it should just live here.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -208,3 +208,22 @@ extension TopologyDescription.TopologyType {
...
+    public var message: String
+    public var errorDescription: String { return self.message }
+
+    public init(message: String) {
+        self.message = message
+    }
+}
+
+/// Possible authentication mechanisms.
+public enum AuthMechanism: String, Decodable {
```
kmahar: this is used by both the sync and async auth tests so I put it here.
eventually this will be moved out into the driver itself as part of the public API.

Sources/TestsCommon/ServerVersion.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+
+/// A struct representing a server version.
+public struct ServerVersion: Comparable, Decodable, CustomStringConvertible {
```
kmahar: similarly, this seems general enough that it can just be part of the common module.

Tests/MongoSwiftSyncTests/ChangeStreamTests.swift
```diff
@@ -162,7 +161,7 @@ internal struct ChangeStreamTest: Decodable {
     internal func run(globalClient: MongoClient, database: String, collection: String) throws {
         let client = try MongoClient.makeTestClient(options: ClientOptions(commandMonitoring: true))
 
-        let center = client.notificationCenter
+        let center = NotificationCenter.default
```
kmahar: this was to get rid of the `@testable`. since we aren't ever passing in a notification center when we create the client, it will always be the default. also, this API is going away soon anyway.

Tests/MongoSwiftSyncTests/SyncAuthTests.swift
```diff
@@ -0,0 +1,139 @@
...
+        // assume the URI does not end with a / and also has no params, e.g. mongodb://localhost:27017
+        return "\(joined)/?authMechanism=\(mech.rawValue)"
+    }
+}
+
+final class SyncAuthTests: MongoSwiftTestCase {
+    // TODO: SWIFT-640: spec says "Drivers that allow specifying auth parameters in code as well as via connection
+    // string should test both for the test cases described below". Once we support setting auth options via options
+    // struct we should test that here too.
+    func testAuthProseTests() throws {
```
kmahar: the prose tests only use the public API, so I could move them over. 

the JSON tests use the `ConnectionString` type so they need to be in the async test module.

PR #371, opened 2019-12-13
SWIFT-675 Iterate over Result in MongoCursor next()

ali-mir:
Updated the `next()` function in `MongoCursor.swift` to iterate over the [Result](https://developer.apple.com/documentation/swift/result) type. Updated tests to reflect this change.


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +197,58 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
-            return doc
+            if let doc = doc {
+                return .success(doc)
+            }
+            return nil
         }
 
         // We already closed the mongoc cursor, either because we reached the end or encountered an error.
         guard case let .open(_, connection, client, session) = self.state else {
             self.error = ClosedCursorError
```
kmahar: I believe your changes here mean we can get rid of error caching altogether. We were doing it before because we didn't have a way to propagate any errors encountered when calling `next()`. Now we are returning those in the `Result`, so I think you can just remove the `error` property altogether and any places we are setting it.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +197,58 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
+            switch $0 {
+            case let .success(t):
+                return t
+            case let .failure(error):
+                throw error
+            }
+        }
+    }
+
     /// Returns the next `Document` in this cursor, or `nil`. After this function returns `nil`, the caller should use
```
kmahar: this docstring needs to be updated.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +197,58 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
     /// the `.error` property to check for errors. For tailable cursors, users should also check `isAlive` after this
     /// method returns `nil`, to determine if the cursor has the potential to return any more data in the future.
-    public func next() -> T? {
+    public func next() -> Result<T, Error>? {
         if case let .cached(doc) = self.cached {
             self.cached = .none
-            return doc
+            if let doc = doc {
```
kmahar: style nit: guard vs if usage is somewhat subjective, but I think this is a case where `guard let` is appropriate, like:
```swift
guard let doc = doc else {
    return nil
}
return .success(doc)
```
I would consider the doc being nil the sort of "edge case" here so I think it's slightly more appropriate to have that be nested in a conditional.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +197,58 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         guard case let .open(_, connection, client, session) = self.state else {
             self.error = ClosedCursorError
-            return nil
+            return .failure(ClosedCursorError)
         }
 
         let operation = NextOperation(target: .cursor(self))
         do {
-            return try client.executeOperation(operation, using: connection, session: session)
+            let result = try client.executeOperation(operation, using: connection, session: session)
```
kmahar: similarly to my comment above, I think you could use a `guard` here to check that result is non-nil.
you can also combine the assignment + check in one line, like `guard let result = try client.executeOperation...`

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +197,58 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
+            case let .success(doc):
+                return doc
+            case let .failure(error):
+                throw error
+            }
         }
         return nil
     }
 
+    public func all() throws -> [T] {
```
kmahar: this method needs a docstring.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +197,58 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
      *   - `DecodingError` if an error occurs decoding the server's response.
      */
     public func nextOrError() throws -> T? {
-        if let next = self.next() {
```
kmahar: (posting here as I can't comment on the actual line) this method's docstring references the old behavior of `next()`. we should update it to reflect your changes.

Tests/MongoSwiftTests/MongoCollection+BulkWriteTests.swift
```diff
@@ -203,13 +203,13 @@ final class MongoCollection_BulkWriteTests: MongoSwiftTestCase {
...
-        expect(cursor.next()).to(equal(["_id": 4, "x": 44]))
+        expect(try cursor.nextOrError()).to(equal(["_id": 2, "x": 24]))
+        expect(try cursor.nextOrError()).to(equal(["_id": 3, "x": 34]))
+        expect(try cursor.nextOrError()).to(equal(["_id": 4, "x": 44]))
         expect(cursor.next()).to(beNil()) // cursor ends
         expect(cursor.error).to(beNil())
-        expect(cursor.next()).to(beNil()) // iterate after cursor ends
-        expect(cursor.error as? UserError).to(equal(UserError.logicError(message: "")))
+        expect(try cursor.nextOrError()).to(throwError(UserError.logicError(message: ""))) // iterate after cursor ends
+//        expect(cursor.error as? UserError).to(equal(UserError.logicError(message: "")))
```
kmahar: this line can maybe be deleted now?

Tests/MongoSwiftTests/MongoCollectionTests.swift
```diff
@@ -100,7 +100,7 @@ final class MongoCollectionTests: MongoSwiftTestCase {
     }
 
     func testAggregate() throws {
-        expect(Array(try self.coll.aggregate([["$project": ["_id": 0, "cat": 1]]])))
+        expect(Array(try self.coll.aggregate([["$project": ["_id": 0, "cat": 1]]]).all()))
```
kmahar: do you still need to wrap this in `Array(...)`, given that `all` returns an array?

Tests/MongoSwiftTests/MongoCursorTests.swift
```diff
@@ -23,7 +23,7 @@ final class MongoCursorTests: MongoSwiftTestCase {
             // insert and read out one document
             try coll.insertOne(doc1)
             cursor = try coll.find()
-            var results = Array(cursor)
+            var results = try Array(cursor.all())
```
kmahar: we are implicitly testing it now via the other methods we call which depend on it, but I think it would be good to add a test somewhere in this file that explicitly tests out the updated `next()` method and consumes the returned `Result`s.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +192,65 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
-    /// Returns the next `Document` in this cursor, or `nil`. After this function returns `nil`, the caller should use
-    /// the `.error` property to check for errors. For tailable cursors, users should also check `isAlive` after this
-    /// method returns `nil`, to determine if the cursor has the potential to return any more data in the future.
-    public func next() -> T? {
+    /**
+     * Returns an array of type  `Document` from this cursor, or throws
+     * an error if any of the `Result` objects contain an error.
+     * - Returns: an array of type `Document`
+     * - Throws:
+     *   - `ServerError.commandError` if an error occurs on the server while iterating the cursor.
```
kmahar:
```suggestion
     *   - `ServerError.commandError` if an error occurs on the server while creating or iterating the cursor.
```

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +192,65 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         }
         return nil
     }
 
-    /// Returns the next `Document` in this cursor, or `nil`. After this function returns `nil`, the caller should use
-    /// the `.error` property to check for errors. For tailable cursors, users should also check `isAlive` after this
-    /// method returns `nil`, to determine if the cursor has the potential to return any more data in the future.
-    public func next() -> T? {
+    /**
+     * Returns an array of type  `Document` from this cursor, or throws
```
kmahar: since this is a cursor over a generic `Codable` type `T`, this may not actually be an array of documents.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +192,65 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
+            switch $0 {
+            case let .success(t):
+                return t
+            case let .failure(error):
+                throw error
+            }
+        }
+    }
+
+    /// Returns a `Result` containing the next `Document` in this cursor, `nil`
```
kmahar: 1) similar to my comment above, this can be any `Codable` type `T`

2) I think it would be good if this docstring also listed the possible errors that can occur calling `next`, as you have above for `nextOrError`.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -197,37 +192,65 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
      *   - `DecodingError` if an error occurs decoding the server's response.
      */
     public func nextOrError() throws -> T? {
```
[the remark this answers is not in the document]
patrickfreed: `Result` does provide a [`get`](https://developer.apple.com/documentation/swift/result/3139397-get) method, which can achieve a similar result:
```swift
let document = try cursor.next()?.get()
```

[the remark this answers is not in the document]
kmahar: hm maybe we should just be encouraging users to use that then (and using it ourselves in tests?). so looping would be like
```swift
for elt in cursor {
    print(try elt.get())
}
```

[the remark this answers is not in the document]
ali-mir: Kaitlin and I spoke offline, and agreed that we can replace `nextOrError` with the `get` method.

ali-mir:
image
Here is what the docstring for `next()` would look like with the changes proposed by this PR. I've removed the `Throws` section as this PR introduces the `Result` API to `next()`.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -187,47 +182,60 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     }
 
     /**
-     * Returns the next `Document` in this cursor or `nil`, or throws an error if one occurs -- compared to `next()`,
-     * which returns `nil` and requires manually checking for an error afterward.
-     * - Returns: the next `Document` in this cursor, or `nil` if at the end of the cursor
+     * Returns an array of type  `T` from this cursor, or throws
```
kmahar: nit: the fact that this uses `Result`s and gets errors from them is an implementation detail that doesn't really matter for the user. I think you could probably just omit everything after "this cursor" .

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -187,47 +182,60 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
+     * an error if any of the `Result` objects contain an error.
+     * - Returns: an array of type `T`
      * - Throws:
-     *   - `ServerError.commandError` if an error occurs on the server while iterating the cursor.
+     *   - `ServerError.commandError` if an error occurs on the server while fetching more results from the server.
```
kmahar:
```suggestion
     *   - `ServerError.commandError` if an error occurs while fetching more results from the server.
```

Tests/MongoSwiftTests/MongoCursorTests.swift
```diff
@@ -64,35 +64,34 @@ final class MongoCursorTests: MongoSwiftTestCase {
...
             // insert a doc so something matches initial query
             try coll.insertOne(doc1)
             cursor = try coll.find(options: cursorOpts)
 
             // for each doc we insert, check that it arrives in the cursor next,
             // and that the cursor is still alive afterward
-            let checkNextResult: (Document) -> Void = { doc in
-                let results = Array(cursor)
+            let checkNextResult: (Document) throws -> Void = { doc in
+                let results = try Array(cursor.all())
```
kmahar: nit: the `Array` isn't needed, since `cursor.all()` already returns an array.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -187,47 +182,60 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
         }
-        return nil
     }
 
-    /// Returns the next `Document` in this cursor, or `nil`. After this function returns `nil`, the caller should use
-    /// the `.error` property to check for errors. For tailable cursors, users should also check `isAlive` after this
-    /// method returns `nil`, to determine if the cursor has the potential to return any more data in the future.
-    public func next() -> T? {
+    /**
+     * Returns a `Result` containing the next `T` in this cursor, `nil`
```
kmahar: I think the ordering here of talking about a result, then a nil value, then a result again, and then a nil value/tailable cursor, is a little confusing.

I think you could reorder it something like

Returns a `Result` containing the next `T` in this cursor, or an error if one occurred. Returns `nil` if the cursor is exhausted. For tailable cursors, users should check `isAlive` after this method returns `nil`, to determine if the cursor has the potential to return any more data in the future.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -187,47 +181,59 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
...
-    /// method returns `nil`, to determine if the cursor has the potential to return any more data in the future.
-    public func next() -> T? {
+    /**
+     * Returns a `Result` containing the next `T` in this cursor,  or an error if one occurred.
+     * Returns `nil` if the cursor is exhausted. For tailable cursors,
+     * users should also check `isAlive` after this method returns `nil`,
+     * to determine if the cursor has the potential to return any more data in the future.
+     * - Returns: `nil` if the end of the cursor has been reached, or a `Result`. On success, the
+     *   `Result` contains the next `T`, and on failure contains:
+     *   - `ServerError.commandError` if an error occurs while fetching more results from the server.
```
patrickfreed: After SWIFT-502, these enum based errors were replaced with structs, so the prefixes can be dropped.
e.g. "ServerError.commandError" becomes just "CommandError"

kmahar: good point. @ali-mir, were these a copy-paste from somewhere else in the driver we didn't update as part of SWIFT-502?

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -25,8 +25,7 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     /**
      * Indicates whether this cursor has the potential to return more data. This property is mainly useful for
      * tailable cursors, where the cursor may be empty but contain more results later on. For non-tailable cursors,
-     * the cursor will always be dead as soon as `next()` returns `nil`, or as soon as `next().get()` returns `nil` or
-     * throws an error.
+     * the cursor will always be dead  as soon as `next()` returns `nil` or a failed `Result`.
```
kmahar: there's an extra space b/w "dead" and "as"

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -182,11 +181,10 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     }
 
     /**
-     * Returns an array of type  `T` from this cursor, or throws
-     * an error if any of the `Result` objects contain an error.
+     * Returns an array of type  `T` from this cursor.
```
kmahar: there is an extra space between "type" and "`T`"

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -203,13 +201,13 @@ public class MongoCursor<T: Codable>: Sequence, IteratorProtocol {
     }
 
     /**
-     * Returns a `Result` containing the next `T` in this cursor, `nil`
-     * if the cursor is exhausted, or an error if one occurs. For tailable cursors,
+     * Returns a `Result` containing the next `T` in this cursor,  or an error if one occurred.
```
kmahar: extra space b/w "," and "or"

Sources/MongoSwift/MongoCursor.swift
ali-mir: good catch! @kmahar I copy pasted it from the `nextOrError()` docstring before this PR was merged. Updating the docstrings now.

kmahar:
@patrickfreed do you have any more comments for this?

PR #373, opened 2019-12-13
SWIFT-667 Split up additional tests

kmahar:
previously, we had one test file for each of `ReadConcern`, `WriteConcern`, and `ReadPreference`. 

`ReadWriteConcernTests.swift` contained tests for both types, in the following formats:
1) Tests for constructing various instances of the type
2) Tests for proper inheritance of these between clients, DBs, and collection
3) Tests that the libmongoc equivalents of clients, DBs, and collections have these options set on them properly
4) Tests for using them to call various methods
5) Spec tests which check read and write concern parsing from URIs and documents

These have now been split into:
- `MongoSwiftTests/ReadWriteConcernSpecTests.swift`: contains the spec tests (5)
- `MongoSwiftSyncTests/ReadWriteConcernOperationTests.swift`: test for using read and write concerns to call various methods (4)
- `MongoSwiftTests/ReadConcernTests.swift` and `MongoSwiftTests/WriteConcernTests.swift`: contain tests for the types themselves (1), (2), and (3)
    * Note these will need to change very slightly with the async API as they call `createCollection`

`ReadPreferenceTests.swift` had tests in the following formats:
1) Tests for constructing various instances of the type
2) Tests for proper inheritance of these between clients, DBs, and collection
3) Tests for using them to call various methods

This has now been split into:
- `MongoSwiftSyncTests/ReadPreferenceOperationTests.swift` (contains 3)
- `MongoSwiftTests/ReadPreferenceTests.swift` (contains 1 and 2)


Tests/MongoSwiftSyncTests/ReadPreferenceOperationTests.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        // expect running other commands to not throw errors when passing in a valid read preference
+        expect(try coll.find(options: FindOptions(readPreference: ReadPreference(.primary)))).toNot(throwError())
+        expect(try coll.findOne(options: FindOneOptions(readPreference: ReadPreference(.primary)))).toNot(throwError())
+
+        expect(try coll.aggregate(
+            [["$project": ["a": 1]]],
+            options: AggregateOptions(readPreference: ReadPreference(.secondaryPreferred))
+        ))
+            .toNot(throwError())
```
patrickfreed: these `toNot`s seem misformatted, does swiftformat not move them? I feel like they look better like the second one.

kmahar: I agree, it looks weird. swiftformat doesn't move them but I've put them on the same line as the )).

PR #376, opened 2019-12-18
Share clients in examples

kmahar:
Undoing the changes from before since we are releasing this now.


patrickfreed:
looks like swiftformat needs to be run, but besides that lgtm

kmahar:
oops, fixed

PR #377, opened 2019-12-20
SWIFT-668, SWIFT-669: Make MongoSwift.MongoClient async and implement MongoSwiftSync.MongoClient

kmahar:
This PR:
- makes the operation executor async
- makes `MongoClient` and its (non cursor/change stream) API methods async
- Implements (non-cursor/change stream) sync `MongoClient` API methods
- moves a lot of now non-functional tests into a `DisabledTests` folder. this seemed preferable to an enormous diff from commenting out all of the non-functional ones.

BTW the diff is significantly smaller if you hide whitespace changes because a lot of the changes in e.g. `ReadConcernTests.swift` involved indenting code


.evergreen/config.yml
```diff
@@ -426,10 +426,11 @@ buildvariants:
...
       remove_tasks: ".3.6"
 
-- matrix_name: "atlas-connect"
-  matrix_spec: 
-    os-fully-featured: ubuntu-18.04
-    swift-version: "*"
-  display_name: "Atlas Connectivity ${swift-version} ${os-fully-featured}"
-  tasks:
-    - ".atlas-connect"
+# TODO SWIFT-703: reenable
```
kmahar: currently the atlas tests will crash, since I switched them to use the sync API and sync DB + collection are unimplemented.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -491,7 +521,16 @@ public class MongoClient {
         using connection: Connection? = nil,
         session: ClientSession? = nil
     ) throws -> T.OperationResult {
-        return try self.operationExecutor.execute(operation, using: connection, client: self, session: session)
+        return try self.operationExecutor.execute(operation, using: connection, client: self, session: session).wait()
+    }
+
+    /// Executes an `Operation` asynchronously using this `MongoClient` and an optionally provided session.
+    internal func executeOperationAsync<T: Operation>(
```
kmahar: eventually `executeOperation` above this will go away and this method will drop the `async`, but for now I've left the above so the types that are not yet async still compile.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -253,6 +267,23 @@ public class MongoClient {
...
+        assert(self.isClosed, "MongoClient was not closed before deinitialization")
+    }
+
+    /// Closes this `MongoClient`. Call this method exactly once when you are finished using the client. You must
+    /// ensure that all operations using the client have completed before calling this. The returned future must be
+    /// fulfilled before the client goes out of scope.
+    public func close() -> EventLoopFuture<Void> {
+        return self.operationExecutor.execute {
+            self.connectionPool.close()
+            self.isClosed = true
```
kmahar: this is technically kind of racey since you could call this from one thread while still using the client in another, but I think the requirements I've documented (call it exactly once, when everything else is done) make it acceptable.

I'm not sure how often people will even *really* need to call this method since I imagine people will often just have one client that for the whole application and only clean it up when their app is shutting down...

Sources/MongoSwiftSync/MongoClient.swift
```diff
@@ -34,7 +41,28 @@ public class MongoClient {
...
+            self.asyncClient = try MongoSwift.MongoClient(connectionString, using: eventLoopGroup, options: options)
+            self.eventLoopGroup = eventLoopGroup
+        } catch {
+            try eventLoopGroup.syncShutdownGracefully()
+            throw error
+        }
+    }
+
+    deinit {
+        do {
```
kmahar: I'm not sure if there is really a better way to handle these.

I wrote it this way in separate try/catches so that even if closing the client fails we will still shut down the event loop group.

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
```diff
@@ -158,18 +158,18 @@ internal func captureCommandEvents(
...
-            if let event = self.next() {
-                return event
-            } else if let error = self.error {
-                throw error
-            }
-        }
-        return nil
-    }
-}
+// extension ChangeStream {
```
kmahar: doesn't compile right now because `ChangeStreamTests` have been moved to the disabled folder

Tests/MongoSwiftTests/AsyncTestUtils.swift
```diff
@@ -14,6 +16,34 @@ extension MongoClient {
...
+        do {
+            try self.close().wait()
+        } catch {
+            print("Error closing test client: \(error)")
+        }
+    }
+}
+
+extension MongoSwiftTestCase {
+    internal func withTestClient<T>(options: ClientOptions? = nil, f: (MongoClient) throws -> T) throws -> T {
```
kmahar: eventually this may look more like `withTestNamespace`, but for now we only have a working async client so I left it like this.

Sources/MongoSwiftSync/MongoClient.swift
patrickfreed: Once we have swift-log integration, I think this would be a good place to do some error logging. Besides that, there's not really much else we can do from deinit in terms of error reporting, so until then this lgtm.

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -11,31 +13,69 @@ internal protocol Operation {
...
-internal struct DefaultOperationExecutor: OperationExecutor {
+internal class DefaultOperationExecutor: OperationExecutor {
+    /// A group of event loops to use for running operations in the thread pool.
+    private let eventLoopGroup: EventLoopGroup
+    /// The thread pool to execute operations in.
+    private let threadPool: NIOThreadPool
+
+    internal init(eventLoopGroup: EventLoopGroup) {
+        self.eventLoopGroup = eventLoopGroup
+        self.threadPool = NIOThreadPool(numberOfThreads: 5)
```
patrickfreed: might want to store this "5" in a constant somewhere for easy for easier discoverability, also maybe a comment on how we arrived at this number

kmahar: good catch, this reminded me that I meant to make it an option. I've done that, and made the default value a static constant on `MongoClient`. 
5 doesn't really have a great justification yet so I've opened [SWIFT-705](https://jira.mongodb.org/browse/SWIFT-705) about doing some basic testing to arrive at a good number and documenting why at that point.

.travis.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     - stage: post-tests
       name: code coverage
       os: osx
-      osx_image: xcode10.2
+      osx_image: xcode11.3
```
kmahar: it seems `llvm-cov` does not have the lcov output format as of LLVM/Xcode 10. I bumped this and went ahead and bumped all of our other Xcode versions on Travis too.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -53,6 +53,15 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
...
     public var serverMonitoring: Bool = false
 
+    /**
+     * `MongoSwift.MongoClient` provides an asynchronous API by running all blocking operations off of their
+     * originating threads in a thread pool. `MongoSwiftSync.MongoClient` is implemented as a wrapper of the async
+     * client which waits for each corresponding asynchronous operation to complete and then returns the result.
+     * This option specifies the size of the thread pool used by the asynchronous client, and determines the max
+     * number of concurrent operations that may be performed using a single client.
+     */
+    public var threadPoolSize: Int? = MongoClient.defaultThreadPoolSize
```
patrickfreed: One thing to consider when making this configurable is that if we ever remove it, it will be a breaking change. I'm pretty sure we were saying that version 2 would be pure swift (i.e. when this change would happen), but I don't know if we've settled on that for sure.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -191,8 +191,7 @@ public class MongoClient {
 
     private let operationExecutor: OperationExecutor
 
-    /// Default size for a client's NIOThreadPool.
-    // TODO: SWIFT-705 document size justification.
+    /// Default size for a client's NIOThreadPool. TODO: SWIFT-705 document size justification.
```
patrickfreed: we probably want to move the todo to the line above the docstring, otherwise it'll be rendered by jazzy.

Sources/MongoSwift/MongoClient.swift
kmahar: yeah, I thought about that. I think that should indeed be version 2.0. but worst case if it's not for some reason, we can deprecate this and add a compiler warning that it no longer has any effect

Sources/MongoSwift/MongoClient.swift
kmahar: I think we will remove this before the next release occurs, but I've moved it anyway in case not 👍

PR #379, opened 2020-01-02
SDAM monitoring updates

patrickfreed:
This PR introduces a number of updates to the SDAM monitoring API:

**Misc. updates:**
- `ConnectionId` refactored to `Address`. This more accurately reflects what is actually stored within the id, and the field name / object returned are flexibly specified in the SDAM monitoring spec:
    - >The connection id is the unique identifier of the driver’s Connection object that wraps the socket. For languages that do not have this object, this MUST a string of “hostname:port” or an object that that contains the hostname and port as attributes. The name of this field is flexible to match the object that is returned from the driver. Examples are, but not limited to, ‘address’, ‘serverAddress’, ‘connectionId’,
- Fields that would always be `nil` on `TopologyDescription` and `ServerDescription` have just been omitted. We can add them back once we have a pure Swift implementation of SDAM if necessary.
- Removed `opTime` from `ServerDescription` since it [only needs to be recorded by mongos](https://jira.mongodb.org/browse/SPEC-1525).
- Introduce the `compatible` property on `TopoogyDescription` as described in the [SDAM spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription).
- Convert integer types on `ServerDescription` and `TopologyDescription` to just be `Int`s. Rationale: 
    - these would overflow 32 bits (which is only a concern if Swift even ran on 32-bit systems, which it doesn't)
    - the SDAM spec is not very prescriptive of how the data is actually stored, it just gives a suggested outline

**[SWIFT-601](https://jira.mongodb.org/browse/SWIFT-601) changes:**
- Introduced an umbrella `MongoSDAMEvent` protocol that is empty at the moment. 
    - Inherits from `MongoEvent`
    - not directly conformed to.
- Introduced `MongoTopologyUpdateEvent` protocol: 
    - inherits from `MongoSDAMEvent`
    - conformed to by `TopologyDescriptionChangedEvent`, `TopologyOpeningEvent`, and `TopologyClosedEvent`
- Introduced `MongoServerUpdateEvent` protocol:
    - inherits from `MongoSDAMEvent`
    - conformed to by `ServerDescriptionChangedEvent`, `ServerOpeningEvent`, and `ServerClosingEvent`
- Introduced `MongoServerHeartbeatEvent` protocol
    - inherits from `MongoSDAMEvent`
    - conformed to by `ServerHeartbeatStartedEvent`, `ServerHeartbeatSucceededEvent`, and `ServerHeartbeatFailedEvent`

[SWIFT-288](https://jira.mongodb.org/browse/SWIFT-288) was also completed as part of this work.


.swiftformat
```diff
@@ -1,4 +1,5 @@
 --exclude Sources/MongoSwift/MongoSwiftVersion.swift,Sources/MongoSwiftSync/*
+--exclude .build/*
```
patrickfreed: not sure why this wasn't a problem for me before, maybe a linux only thing?

Tests/DisabledTests/ClientSessionTests.swift
```diff
@@ -80,7 +80,7 @@ final class ClientSessionTests: MongoSwiftTestCase {
     ]
 
 // This function causes the compiler to crash on older versions of swift due to a bug in the compiler.
-#if swift(>=5.1)
+#if os(OSX) && swift(>=5.1)
```
patrickfreed: this crash still happens on linux 5.1, so I just made this a macOS only test until we can be sure linux is working.

Sources/MongoSwift/SDAM.swift
```diff
@@ -57,40 +65,35 @@ public struct ServerDescription {
...
     /// address the client uses.
-    public let connectionId: ConnectionId
-
-    /// The last error related to this server.
-    public let error: MongoError? = nil // currently we will never set this
+    public let address: Address
 
-    /// The duration of the server's last ismaster call.
+    /// The duration in milliseconds of the server's last ismaster call.
     public var roundTripTime: Int64?
```
kmahar: are all of these just `var`s so we can set them outside of the `init` in `parseIsMaster`? or is it possible I did this so the values could change over time in a true SDAM implementation? 
(I realize I wrote all of this code but I have no recollection of what I was thinking...)

Sources/MongoSwift/SDAM.swift
```diff
@@ -261,7 +250,6 @@ public struct TopologyDescription {
 
     /// Returns `true` if the topology has a readable server available, and `false` otherwise.
     public func hasReadableServer() -> Bool {
-        // (this function should take in an optional ReadPreference, but we have yet to implement that type.)
```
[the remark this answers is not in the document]
patrickfreed: it's still open, we just waaay down in the backlog. Once we implement SDAM in Swift we can pick it up again. Added the todo comment

Sources/MongoSwift/SDAM.swift
patrickfreed: I remember thinking about this before break but I don't remember what conclusion I came to. `var` -> `let` is definitely a breaking change, but the other way I think should be fine. Do you think it's worth making these constants here? In the currently implementation they definitely shouldn't be mutated by the user or us.

Sources/MongoSwift/SDAM.swift
```diff
@@ -156,38 +153,36 @@ public struct ServerDescription {
...
+        self.address = Address(mongoc_server_description_host(description))
         self.roundTripTime = mongoc_server_description_round_trip_time(description)
-        // we have to copy because libmongoc owns the pointer.
-        let isMaster = Document(copying: mongoc_server_description_ismaster(description))
-        self.parseIsMaster(isMaster)
-
+        self.lastUpdateTime = Date(msSinceEpoch: mongoc_server_description_last_update_time(description))
         let serverType = String(cString: mongoc_server_description_type(description))
         // swiftlint:disable:next force_unwrapping
         self.type = ServerType(rawValue: serverType)! // libmongoc will always give us a valid raw value.
```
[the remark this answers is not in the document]
patrickfreed: A new server type would be a mega breaking change for all drivers I think. The only case I can think of that might introduce one is multi-master, but that might be handled entirely by mongos or something. @mbroadst probably has a better perspective on this.

Sources/MongoSwift/SDAM.swift
patrickfreed: Updated them all to be `let`.

Sources/MongoSwift/SDAM.swift
```diff
@@ -1,14 +1,15 @@
 import Foundation
 import mongoc
 
-/// A struct representing a server connection, consisting of a host and port.
-public struct ConnectionId: Equatable {
-    /// A string representing the host for this connection.
+/// A struct representing a server address, consisting of a host and port.
+public struct Address: Equatable {
```
mbroadst: nit: let's either call it `ServerAddress` or correct the comment to explain this is a generic address. Also, now that we have NIO as a dependency, is there a good reason we aren't just using their [SocketAddress](https://github.com/apple/swift-nio/blob/master/Sources/NIO/SocketAddresses.swift)? It seems to serve a similar purpose.

Sources/MongoSwift/SDAM.swift
```diff
@@ -57,137 +65,105 @@ public struct ServerDescription {
...
 
 extension ServerDescription: Equatable {
     public static func == (lhs: ServerDescription, rhs: ServerDescription) -> Bool {
-        // Compare everything except `error` it is always nil and `MongoError`s are not `Equatable`
-        return lhs.connectionId == rhs.connectionId &&
-            lhs.roundTripTime == rhs.roundTripTime &&
-            lhs.lastWriteDate == rhs.lastWriteDate &&
-            lhs.opTime == rhs.opTime &&
+        // As per the SDAM spec, only some fields are necessary to compare for equality.
+        return lhs.address == rhs.address &&
```
mbroadst: This will at some point need to account for a stored `error` on the `ServerDescription` as well.

Sources/MongoSwift/SDAM.swift
```diff
@@ -233,25 +208,20 @@ public struct TopologyDescription {
...
-    public let stale: Bool = false // currently, this will never be set
+    /// The servers comprising this topology.
+    public let servers: [ServerDescription]
 
-    /// Exists if any server's wire protocol version range is incompatible with the client's.
-    public let compatibilityError: MongoError? = nil // currently, this will never be set
+    /// Whether every server's wire protocol version range is compatible with MongoSwift's.
+    public var compatible: Bool {
+        return self.servers.allSatisfy {
+            $0.maxWireVersion >= MIN_SUPPORTED_WIRE_VERSION
```
mbroadst: This logic is incorrect. From the spec:
```
A ServerDescription which is not Unknown is incompatible if:

minWireVersion > clientMaxWireVersion, or
maxWireVersion < clientMinWireVersion
```

Here we are not ignoring `Unknown` servers, nor are we checking the `minWireVersion` requirements. Additionally, the reason the error itself is stored is because it will likely change based on what part of this is incompatible, its unclear you'd have that information where you eventually report the error to the user.

Sources/MongoSwift/SDAM.swift
```diff
@@ -281,22 +251,12 @@ public struct TopologyDescription {
...
-extension TopologyDescription: Equatable {
-    public static func == (lhs: TopologyDescription, rhs: TopologyDescription) -> Bool {
-        // Compare everything except `error` it is always nil and `MongoError`s are not `Equatable`
-        return lhs.type == rhs.type &&
-            lhs.setName == rhs.setName &&
-            lhs.maxSetVersion == rhs.maxSetVersion &&
-            lhs.maxElectionId == rhs.maxElectionId &&
-            lhs.servers == rhs.servers &&
-            lhs.stale == rhs.stale
+        self.servers = servers ?? []
```
mbroadst: nit: can't you reduce this to one line?

```
self.servers = size > 0 ? Array(buffer).map { ServerDescription($0!) } : []
```

Sources/MongoSwift/SDAM.swift
patrickfreed: done. I added a `try!` when converting from a `mongoc_host_list_t` to a `SocketAddress`, since I figured libmongoc would only contain well-formatted addresses in its server descriptions. Is that a safe assumption to make?

patrickfreed: `SocketAddress` only works on resolved IP addresses, so things like "localhost" or a domain name won't be able to be stored in it, which I think doesn't work for us. As such, I've reverted the commits that enabled its usage in the driver.

patrickfreed: I've updated the comment to say a "network address" instead of "server address"

Sources/MongoSwift/SDAM.swift
patrickfreed: For now I've just removed it.

PR #380, opened 2020-01-02
SWIFT-690 Store mongod logs after evergreen runs

patrickfreed:
[SWIFT-690](https://jira.mongodb.org/browse/SWIFT-690)

Taken from the [go driver's config](https://github.com/mongodb/mongo-go-driver/blob/master/.evergreen/config.yml#L177).

[example patch](https://evergreen.mongodb.com/task/mongo_swift_driver_tests_all__os_fully_featured~ubuntu_18.04_swift_version~5.0_ssl~nossl_test_3.6_replica_set_patch_0d211072d7b37efa64a9fef0ecbdb66b6e35d5d5_5e0e4ed60ae606081514989d_20_01_02_20_14_50)

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -194,7 +194,6 @@ pre:
   - func: "install dependencies"
 
 post:
-  - func: "upload-mo-artifacts"
   - func: "cleanup"
 
 tasks:
@@ -209,6 +208,7 @@ tasks:
             TOPOLOGY: "server"
         - func: "run tests"
         - func: "stop mongo-orchestration"
+        - func: "upload-mo-artifacts"
 
     - name: "test-3.6-replica_set"
       tags: ["3.6", "replica_set"]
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -172,11 +172,29 @@ functions:
...
+        permissions: public-read
+        content_type: ${content_type|application/x-gzip}
+        display_name: "mongodb-logs.tar.gz"
+
 pre:
   - func: "fetch source"
   - func: "install dependencies"
 
 post:
+  - func: "upload-mo-artifacts"
```
kmahar: post is run after *every* task, which may present problems for our tasks that do not actually use mongo-orchestration (e.g. the Atlas tests may fail because this directory won't exist).
the duplication is annoying but it might be better to add this manually to each task which calls "stop mongo-orchestration"

patrickfreed: done

PR #381, opened 2020-01-02
SWIFT-670 Make MongoSwift.MongoDatabase async and implement MongoSwiftSync.MongoDatabase

kmahar:
This PR updates `MongoSwift.MongoDatabase` to be async by using the client's new `executeOperationAsync` method, and implements `MongoSwiftSync.MongoDatabase` as a wrapper of the newly async DB.

This allowed me to reenable some tests and also required minor changes to already-enabled read/write concern tests to `wait()` on results.

### DIFF
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -172,7 +172,7 @@ public struct MongoDatabase {
      *   - options: Optional `CreateCollectionOptions` to use for the collection
      *   - session: Optional `ClientSession` to use when executing this command
      *
-     * - Returns: the newly created `MongoCollection<Document>`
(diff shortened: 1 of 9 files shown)


Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -250,7 +251,8 @@ public struct MongoDatabase {
      *   - options: Optional `ListCollectionsOptions` to use when executing this command
      *   - session: Optional `ClientSession` to use when executing this command
      *
-     * - Returns: An array of `MongoCollection`s that match the provided filter.
+     * - Returns: An `EventLoopFuture<[MongoCollection<Document>]>`s containing collections that match the provided
```
patrickfreed: can remove the "s" after `EventLoopFuture<...>`

kmahar: fixed.

PR #382, opened 2020-01-06
SWIFT-709 ignore vendored files for language statistics

mbroadst:
Adding a specially crafted line to our `.gitattributes` file will indicate that GitHub should skip certain directories when doing language statistics on the project.


kmahar:
@mbroadst can we merge?

PR #383, opened 2020-01-07
SWIFT-690 Don't archive logs until cluster has finished shutting down

patrickfreed:
This is a follow-up to SWIFT-690. Right now, we're seeing some purple failures on the waterfall due to the log upload function beginning before the mongods are all finished shutting down. [The logs are still changing while we attempt to archive them, and it results in the task failing.](https://evergreen.mongodb.com/task_log_raw/mongo_swift_driver_tests_all__os_fully_featured~ubuntu_18.04_swift_version~5.0_ssl~ssl_test_4.0_replica_set_865b5660a707b2445279d70f695321bae534252c_20_01_06_16_34_45/0?type=T#L1641) I think this is due "upload-mo-artifacts" being called directly after "stop-mongo-orchestration", rather than in post.

To alleviate this, I simply added a 1 second sleep to "upload mo artifacts", and it seems to be enough breathing room to prevent this issue. [example patch](https://evergreen.mongodb.com/version/5e1388b4e3c3315b3af19253)

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -92,7 +92,7 @@ functions:
             # If this was a patch build, doing a fresh clone would not actually test the patch
             cp -R ${PROJECT_DIRECTORY}/ $DRIVERS_TOOLS
           else
-            git clone git://github.com/mongodb-labs/drivers-evergreen-tools.git $DRIVERS_TOOLS
+            git clone -b blocking-stop git://github.com/patrickfreed/drivers-evergreen-tools.git $DRIVERS_TOOLS
           fi
           echo "{ \"releases\": { \"default\": \"$MONGODB_BINARIES\" }}" > $MONGO_ORCHESTRATION_HOME/orchestration.config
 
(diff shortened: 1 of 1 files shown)


patrickfreed:
It doesn't look like such a thing is possible unfortunately.

Also, I updated it to sleep for 5 seconds (up from 1) since every now and then 1 second would still be too short.

patrickfreed:
@mbroadst any feelings on this?

patrickfreed:
The best path forward here actually seemed to be running a new script that issues a request to mongo-orchestration to shut down the cluster explicitly. The nice thing about this is that the script will block until mongo-orchestration returns, indicating the cluster has completely finished shutting down and that we can safely tar the logs.

I figured that such a script might be useful for other drivers, so I added it to a fork of drivers-evergreen-tools. For the sake of testing, I've updated our `config.yml` to point to that fork and run [an example patch](https://evergreen.mongodb.com/version/5e20d5322fbabe19ba7e323e). If y'all think this is the right approach, I'll make a PR against drivers-evergreen-tools with the script, and once that is merged, I can update our `config.yml` to point back to the mongodb-labs version instead of my fork.

patrickfreed:
Here's the branch of my drivers-evergreen-tools fork containing the relevant changes: https://github.com/patrickfreed/drivers-evergreen-tools/tree/blocking-stop

patrickfreed:
With the changes in https://github.com/10gen/mongo-orchestration/pull/276, this is no longer necessary.

PR #384, opened 2020-01-08
SWIFT-671 Make MongoCollection read/write methods async, and implement synchronous equivalents

kmahar:
This PR makes the non-cursor read/write methods on `MongoCollection` async, and implements their synchronous equivalents.

There were a few test files that *mostly* can run but needed cursors so I enabled them and commented out the cursor portions of the files.

### DIFF
--- a/Sources/MongoSwiftSync/MongoClient.swift
+++ b/Sources/MongoSwiftSync/MongoClient.swift
@@ -22,7 +22,7 @@ public class MongoClient {
     private let eventLoopGroup: MultiThreadedEventLoopGroup
 
     /// The underlying async client.
-    internal let asyncClient: MongoSwift.MongoClient
+    private let asyncClient: MongoSwift.MongoClient
 
     /**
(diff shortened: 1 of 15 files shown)


Sources/MongoSwiftSync/MongoClient.swift
```diff
@@ -160,7 +160,7 @@ public class MongoClient {
      * - Returns: a `MongoDatabase` corresponding to the provided database name
      */
     public func db(_ name: String, options: DatabaseOptions? = nil) -> MongoDatabase {
-        return MongoDatabase(name: name, client: self, options: options)
+        return MongoDatabase(asyncDB: self.asyncClient.db(name, options: options))
```
kmahar: the change to this initializer is for consistency with how I ended up doing it for `MongoCollection`. I did it differently there because we can end up with a sync `MongoCollection` via both `createCollection` and `collection` and so I wanted to just be able to pass an async collection into the init.

I also realized there is no need for the sync DBs and collections to store their parent clients so that eliminated a need to pass that in

PR #386, opened 2020-01-09
SWIFT-702 Make index management and find-and-modify methods async, implement sync versions

kmahar:
This PR implements non-cursor index management and find-and-modify methods on `MongoCollection`.

Per our discussion on Slack I also added a new `DropIndexesResult` type.

Unfortunately basically all of the index tests rely on `MongoCursor` being implemented (they use `listIndexes` to verify results) so I can't really enable that file without commenting out most of it yet. 

Also made some other small changes I'll explain in inline comments

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -548,7 +548,25 @@ public class MongoClient {
         using connection: Connection? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<T.OperationResult> {
+        guard !self.isClosed else {
+            return self.makeFailedFuture(MongoClient.ClosedClientError)
+        }
+
+        if let session = session {
+            if case .ended = session.state {
+                return self.makeFailedFuture(ClientSession.SessionInactiveError)
+            }
+            guard session.client == self else {
+                return self.makeFailedFuture(ClientSession.ClientMismatchError)
+            }
+        }
+
         return self.operationExecutor.execute(operation, using: connection, client: self, session: session)
+    }
+
+    /// Creates an `EventLoopFuture<T>` that fails with the provided error.
+    internal func makeFailedFuture<T>(_ error: Error) -> EventLoopFuture<T> {
+        return self.operationExecutor.makeFailedFuture(error)
     }
 }
 
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -189,7 +189,7 @@ public struct TLSOptions {
 public class MongoClient {
     internal let connectionPool: ConnectionPool
 
-    private let operationExecutor: OperationExecutor
+    internal let operationExecutor: OperationExecutor
 
     // TODO: SWIFT-705 document size justification.
     /// Default size for a client's NIOThreadPool.
@@ -548,25 +548,7 @@ public class MongoClient {
         using connection: Connection? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<T.OperationResult> {
-        guard !self.isClosed else {
-            return self.makeFailedFuture(MongoClient.ClosedClientError)
-        }
-
-        if let session = session {
-            if case .ended = session.state {
-                return self.makeFailedFuture(ClientSession.SessionInactiveError)
-            }
-            guard session.client == self else {
-                return self.makeFailedFuture(ClientSession.ClientMismatchError)
-            }
-        }
-
         return self.operationExecutor.execute(operation, using: connection, client: self, session: session)
-    }
-
-    /// Creates an `EventLoopFuture<T>` that fails with the provided error.
-    internal func makeFailedFuture<T>(_ error: Error) -> EventLoopFuture<T> {
-        return self.operationExecutor.makeFailedFuture(error)
     }
 }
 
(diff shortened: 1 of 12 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -548,8 +548,19 @@ public class MongoClient {
...
+            return self.makeFailedFuture(MongoClient.ClosedClientError)
+        }
+        if case .ended = session?.state {
+            return self.makeFailedFuture(ClientSession.SessionInactiveError)
+        }
         return self.operationExecutor.execute(operation, using: connection, client: self, session: session)
     }
+
+    /// Creates an `EventLoopFuture<T>` that fails with the provided error.
+    internal func makeFailedFuture<T>(_ error: Error) -> EventLoopFuture<T> {
```
kmahar: so last time I fixed the bulk write error validation problem by just moving it into the operation's execute method, but I ran into a few more validation-related cases in this PR that were tricky to move over. I also realized that waiting to fail until we are in the thread pool is less optimal performance-wise if we can detect a problem earlier.

For that reason, I've added a `makeFailedFuture` method on `OperationExecutor` (along with this helper, for convenience) that allows us to generate failed futures without entering the thread pool.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -548,8 +548,26 @@ public class MongoClient {
         using connection: Connection? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<T.OperationResult> {
+        guard !self.isClosed else {
```
kmahar: (see comment below this one first)

previously this check happened in the executor, but now that I've added the makeFailedFuture helper I think it makes more sense to have the client do the check.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -548,8 +548,26 @@ public class MongoClient {
...
+            return self.makeFailedFuture(MongoClient.ClosedClientError)
+        }
+
+        if let session = session {
```
kmahar: I decided to add these here as it seems better to detect these cases early without having to tie up a thread in the thread pool.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -26,6 +26,9 @@ extension MongoCollection {
         options: BulkWriteOptions? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<BulkWriteResult?> {
+        guard !requests.isEmpty else {
```
kmahar: following the "fail before the thread pool if possible" logic, I moved this back into `bulkWrite`

Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -59,7 +59,6 @@ extension MongoCollection {
     ) -> EventLoopFuture<InsertManyResult?> {
         return self.bulkWrite(values.map { .insertOne($0) }, options: options, session: session)
             .flatMapThrowing { InsertManyResult(from: $0) }
-            .flatMapErrorThrowing { throw convertBulkWriteError($0) }
```
kmahar: a newly enabled test in `MongoCollectionTests` caught this. since `insertMany` throws a bulk write error we don't need to do the conversion.

Sources/MongoSwift/Operations/DropIndexesOperation.swift
```diff
@@ -15,6 +15,12 @@ public struct DropIndexOptions: Encodable {
     }
 }
 
+/// The result of performing a "dropIndexes" command.
+public struct DropIndexesResult: Decodable {
+    /// The number of indexes that were present in the collection prior to performing the drop.
+    public let nIndexesWas: Int
```
kmahar: let me know if you have naming opinions on this... I considered `previousIndexCount` or `indexCountBeforeDrop` or something but those felt kinda verbose

Sources/MongoSwift/MongoClient.swift
patrickfreed: In my change streams work, I similarly need to access the underlying `execute` on the operation executor, meaning we'd have a helper for every method on `OperationExecutor` (except for `close`). This kinda makes the helpers feel redundant when we can just expose the executor itself as `internal` instead. What do you think?

Sources/MongoSwift/Operations/DropIndexesOperation.swift
patrickfreed: I think it's okay to use the name as it is in the actual response.

Sources/MongoSwift/MongoClient.swift
kmahar: that's a good point. it also could lead to bugs if you forget the helper exists and just go directly to the executor since there are checks here that don't happen there. I'm fine with getting rid of the helper and just directly accessing the executor.

kmahar: I'll probably just do this in a separate PR right now since it is going to touch a lot of places unrelated to this PR

kmahar:
@patrickfreed for now I got rid of the new `makeFailedFuture` helper and I moved the error checking logic from the client `executeOperation` helper into the executor itself

PR #388, opened 2020-01-09
SWIFT-673 Make MongoSwift.ChangeStream async, and implement MongoSwiftSync.ChangeStream

patrickfreed:
[SWIFT-673](https://jira.mongodb.org/browse/SWIFT-673)

This PR implements the async and sync change streams. As part of this, SWIFT-688 will also be completed.

A large portion of this diff is just copying the spec test runner over from DisabledTests, so you can ignore a lot of that.


.swiftlint.yml
```diff
@@ -25,7 +25,6 @@ opt_in_rules:
   - missing_docs
   - modifier_order
   - multiline_arguments
-  - multiline_function_chains
```
patrickfreed: I removed this linter rule because it led to unnecessary (imo) line breaks when chaining or waiting on futures.

For example, the following would not be allowed:
```swift
coll.insertOne([:]).flatMap { _ in
     coll.insertOne([:])
}.flatMap { _ in
     coll.insertOne([:])    
}.flatMap { _ in
     coll.insertOne([:])
}

try coll.watch(
    pipeline,
    options: options
).wait()
```
They would have to be written as:
```swift
coll.insertOne([:]).flatMap { _ in
     coll.insertOne([:])
}
.flatMap { _ in
     coll.insertOne([:])    
}
.flatMap { _ in
     coll.insertOne([:])
}

try coll.watch(
    pipeline,
    options: options
)
.wait()
```
Which I think is too spaced out. What do you think?

Package.swift
```diff
@@ -14,7 +14,7 @@ let package = Package(
         .target(name: "MongoSwift", dependencies: ["CLibMongoC", "NIO"]),
         .target(name: "MongoSwiftSync", dependencies: ["MongoSwift"]),
         .target(name: "AtlasConnectivity", dependencies: ["MongoSwiftSync"]),
-        .target(name: "TestsCommon", dependencies: ["MongoSwift", "Nimble", "CLibMongoC"]),
+        .target(name: "TestsCommon", dependencies: ["MongoSwift", "MongoSwiftSync", "Nimble", "CLibMongoC"]),
```
patrickfreed: The "Common" in "TestsCommon" means Sync stuff can be in there too, right?

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -90,60 +105,101 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
         return extractMongoError(error: error)
     }
 
+    /// Poll the underlying mongoc_change_stream_t for the next change event.
+    internal func fetchNextDocument() throws -> T? {
```
patrickfreed: this is similar to `getNextDocumentFromMongocCursor`, but I went with a shorter and more generic name.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -90,60 +105,101 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
...
+                body(.failure(error))
+                return
             }
-            return out
-        } catch {
-            self.error = error
-            self.close()
-            return nil
+            self.pollingSemaphore.signal()
+            self.forEach(body: body)
```
patrickfreed: This method will likely be very long-lived, which means these recursion calls can add up in terms of memory cost. I'm not really sure how to get around this, since any long term polling without async/await will need recursion, though.

Sources/MongoSwiftSync/ChangeStream.swift
```diff
@@ -4,47 +4,64 @@ import MongoSwift
...
+        return self.asyncChangeStream.isAlive
+    }
 
-    /// The error that occurred while iterating the change stream, if one exists. This should be used to check
-    /// for errors after `next()` returns `nil`.
-    public var error: Error? { fatalError("unimplemented") }
+    private let asyncChangeStream: MongoSwift.ChangeStream<T>
+
+    /// The client this change strem descended from.
+    private let client: MongoClient
```
patrickfreed: The sync change stream needs a reference to the sync client, otherwise it will go out of scope and transitively close the async client before this sync change stream goes out of scope. As a result, I had to go back and add the references to the sync client to the sync db and collection.

Tests/DisabledTests/SpecTestRunner/SpecTest.swift
```diff
@@ -54,132 +54,6 @@ internal struct TestCommandStartedEvent: Decodable, Matchable {
     }
 }
 
-/// Protocol that test cases which configure fail points during their execution conform to.
```
patrickfreed: moved all this to its own file

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -90,60 +105,101 @@ public class ChangeStream<T: Codable>: Sequence, IteratorProtocol {
...
+    public func all() -> EventLoopFuture<[T]> {
+        let operation = AllOperation(target: .changeStream(self))
+        return self.client.executeOperationAsync(operation)
+    }
+
+    /// Executes the provided closure against each event seen by the change stream.
+    /// This will cause the stream to continuously poll for new events in the background. Once this method has been
+    /// called, other methods that check for events (e.g. all, next, forEach) should not be used.
+    public func forEach(body: @escaping (Result<T, Error>) -> Void) {
+        self.pollingSemaphore.wait()
```
patrickfreed: usage of this semaphore allows the user to close the change stream even if `forEach` is running in the background.

kmahar:
closing as I believe this is obsoleted by #395

PR #389, opened 2020-01-15
SWIFT-634 add Swift 5.1 variants to evg

kmahar:
Evergreen patch: https://evergreen.mongodb.com/version/5e1e4518562343482e160895 (failures are all due to timeout issue you've fixed in #383)

### DIFF
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -4,7 +4,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 
 # variables
 PROJECT_DIRECTORY=${PROJECT_DIRECTORY:-$PWD}
-SWIFT_VERSION=${SWIFT_VERSION:-4.2.4}
+SWIFT_VERSION=${SWIFT_VERSION:-5.0}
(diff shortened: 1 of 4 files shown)


.evergreen/install-dependencies.sh
```diff
@@ -4,7 +4,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 
 # variables
 PROJECT_DIRECTORY=${PROJECT_DIRECTORY:-$PWD}
-SWIFT_VERSION=${SWIFT_VERSION:-4.2.4}
+SWIFT_VERSION=${SWIFT_VERSION:-5.0}
```
patrickfreed: Do we need to still keep testing Swift 5.0.x? If so, this should probably be Swift 5.0.3.

kmahar: yes, we are still testing it. (we actually test 5.0.3, these are just fallback values in case we don't specify a SWIFT_VERSION). updated this to 5.0.3 here and in a few other scripts.

it also occurred to me we should actually be testing 5.1.3, not 5.1.0, so I've updated that in the evg config and running a new patch here: https://evergreen.mongodb.com/version/5e1f4b142a60ed140a378f37

PR #391, opened 2020-01-21
SWIFT-674 Make MongoSwift.ClientSession async, and implement MongoSwiftSync.ClientSession

kmahar:
Various comments inline

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -1,5 +1,6 @@
 import CLibMongoC
 import Foundation
+import NIO
 
 /**
  * A MongoDB client session.
@@ -111,7 +112,7 @@ public final class ClientSession {
     public let options: ClientSessionOptions?
 
     /// Initializes a new client session.
-    internal init(client: MongoClient, options: ClientSessionOptions? = nil) throws {
+    internal init(client: MongoClient, options: ClientSessionOptions? = nil) {
         self.options = options
         self.client = client
         self.state = .notStarted(opTime: nil, clusterTime: nil)
@@ -119,15 +120,15 @@ public final class ClientSession {
 
     /// Starts this session's corresponding libmongoc session, if it has not been started already. Throws an error if
     /// this session has already been ended.
-    internal func startIfNeeded() throws {
+    internal func startIfNeeded() -> EventLoopFuture<Void> {
         switch self.state {
         case .notStarted:
             let operation = StartSessionOperation(session: self)
-            try self.client.executeOperation(operation)
+            return self.client.operationExecutor.execute(operation, client: self.client, session: nil)
         case .started:
-            return
+            return self.client.operationExecutor.makeSucceededFuture(Void())
         case .ended:
-            throw ClientSession.SessionInactiveError
+            return self.client.operationExecutor.makeFailedFuture(ClientSession.SessionInactiveError)
         }
     }
 
@@ -143,18 +144,28 @@ public final class ClientSession {
         return connection
     }
 
-    /// Destroy the underlying `mongoc_client_session_t` and ends this session. Has no effect if this session is
-    /// already ended.
-    internal func end() {
-        if case let .started(session, _) = self.state {
-            mongoc_client_session_destroy(session)
+    /// Ends this `ClientSession`. Call this method when you are finished using the session. You must ensure that all
+    /// operations using this session have completed before calling this. The returned future must be fulfilled before
+    /// the session goes out of scope.
+    public func end() -> EventLoopFuture<Void> {
+        switch self.state {
+        case .notStarted, .ended:
+            self.state = .ended
+            return self.client.operationExecutor.makeSucceededFuture(Void())
+        case let .started(session, _):
+            return self.client.operationExecutor.execute {
+                mongoc_client_session_destroy(session)
+                self.state = .ended
+            }
         }
-        self.state = .ended
     }
 
     /// Cleans up internal state.
     deinit {
-        self.end()
+        guard case .ended = self.state else {
+            assertionFailure("ClientSession was not ended before going out of scope; please call ClientSession.end()")
+            return
+        }
     }
 
     /**
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -146,7 +146,7 @@ public final class ClientSession {
 
     /// Ends this `ClientSession`. Call this method when you are finished using the session. You must ensure that all
     /// operations using this session have completed before calling this. The returned future must be fulfilled before
-    /// the session goes out of scope.
+    /// this session's parent `MongoClient` is closed.
     public func end() -> EventLoopFuture<Void> {
         switch self.state {
         case .notStarted, .ended:
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -292,7 +292,7 @@ public class MongoClient {
 
     /// Closes this `MongoClient`. Call this method exactly once when you are finished using the client. You must
     /// ensure that all operations using the client have completed before calling this. The returned future must be
-    /// fulfilled before the client goes out of scope.
+    /// fulfilled before the `EventLoopGroup` provided to this client's constructor is shut down.
     public func close() -> EventLoopFuture<Void> {
         return self.operationExecutor.execute {
             self.connectionPool.close()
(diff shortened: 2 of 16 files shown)


Sources/MongoSwiftSync/MongoCollection.swift
```diff
@@ -38,8 +38,13 @@ public struct MongoCollection<T: Codable> {
     /// The underlying asynchronous collection.
     internal let asyncColl: MongoSwift.MongoCollection<T>
 
+    /// The client this collection was derived from. We store this to ensure it remains open for as long as this object
+    /// is in scope.
+    private let client: MongoClient
```
kmahar: i realized that my previous claim that we didn't need to do this was incorrect. for reference counting purposes we do indeed need to store references to sync clients from sync DBs/colls to ensure that their sync clients (and therefore their async clients) are kept alive automatically.

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
```diff
@@ -213,4 +213,39 @@ final class SyncMongoClientTests: MongoSwiftTestCase {
 
         try defaultDb.drop()
     }
+
+    // Ensure that sync clients stay alive as long as their child objects are still in scope.
+    func testClientLifetimeManagement() throws {
```
kmahar: this is basically a regression test for what I stated above about keeping clients alive for the appropriate amount of time.

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -1,8 +1,19 @@
 import Foundation
 @testable import MongoSwift
+@testable import MongoSwiftSync
```
kmahar: we need to import both modules here in order to access internal properties of async sessions e.g `id`. as a result a lot of the changes in this file are to help the compiler disambiguate between sync and async types.

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -51,13 +63,14 @@ final class ClientSessionTests: MongoSwiftTestCase {
         (name: "deleteOne", body: { _ = try $0.deleteOne([:], session: $1) }),
         (name: "deleteMany", body: { _ = try $0.deleteMany([:], session: $1) }),
         (name: "createIndex", body: { _ = try $0.createIndex([:], session: $1) }),
-        (name: "createIndex1", body: { _ = try $0.createIndex(IndexModel(keys: ["x": 1] as Document), session: $1) }),
-        (name: "createIndexes", body: { _ = try $0.createIndexes([], session: $1) }),
+        (name: "createIndex1", body: { _ = try $0.createIndex(IndexModel(keys: ["x": 1]), session: $1) }),
+        (name: "createIndexes", body: { _ = try $0.createIndexes([IndexModel(keys: ["x": 1])], session: $1) }),
```
kmahar: this was failing because I added a check in `createIndexes` for `models` being non-empty

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -384,74 +400,76 @@ final class ClientSessionTests: MongoSwiftTestCase {
             }
             defer { center.removeObserver(replyObserver) }
 
-            _ = try collection.find(session: session).next()
+            _ = try collection.countDocuments(session: session)
```
kmahar: there are a lot of places in these tests where we need to do a read and were doing a find. I switched a lot of them to `countDocuments`for the purpose of enabling the tests as they don't seem to specify which read we do. do you remember if there was a particular reason you did `find`?

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -47,25 +47,29 @@ internal class OperationExecutor {
...
-            // if a session was provided, start it if it hasn't been started already.
-            try session?.startIfNeeded()
-            // select a connection in following order of priority:
-            // 1. connection specifically provided for use with this operation
-            // 2. if a session was provided, use its underlying connection
-            // 3. a new connection from the pool
-            let connection = try connection ?? resolveConnection(client: client, session: session)
-            return try operation.execute(using: connection, session: session)
+            // start the session if needed (which generates a new operation itself), and then execute the operation.
+            return session.startIfNeeded().flatMap { self.execute { try doOperation() } }
```
patrickfreed: could these be shortened to `self.execute(doOperation)`?

Sources/MongoSwift/ClientSession.swift
```diff
@@ -143,18 +144,28 @@ public final class ClientSession {
...
     }
 
-    /// Destroy the underlying `mongoc_client_session_t` and ends this session. Has no effect if this session is
-    /// already ended.
-    internal func end() {
-        if case let .started(session, _) = self.state {
-            mongoc_client_session_destroy(session)
+    /// Ends this `ClientSession`. Call this method when you are finished using the session. You must ensure that all
+    /// operations using this session have completed before calling this. The returned future must be fulfilled before
+    /// the session goes out of scope.
```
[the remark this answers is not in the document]
kmahar: good point. I think users do need to ensure the future is fulfilled before the parent client is closed though, otherwise the session's connection might not be returned to the pool before its destroyed. I updated the docstring to reflect this.
I also updated `MongoClient.close()` with some similar language

Sources/MongoSwift/MongoClient.swift
```diff
@@ -322,11 +317,33 @@ public class MongoClient {
...
+    ) -> EventLoopFuture<T> {
+        let promise = self.operationExecutor.makePromise(of: T.self)
+        let session = self.startSession(options: options)
+        do {
+            let bodyFuture = try sessionBody(session)
+            // regardless of whether body's returned future succeeds we want to call session.end() once its complete.
+            // only once session.end() finishes can we fulfill the returned promise. otherwise the user can't tell if
+            // it is safe to close the parent client of this session, and they could inadvertently close it before the
+            // session is actually ended and its parent `mongoc_client_t` is returned to the pool.
+            bodyFuture
```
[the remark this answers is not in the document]
kmahar: I think we do need to wait for the session to finish ending before we give the user a result, per my comment above this line - we need a way for the user to know it's safe to close the client they are calling this method on.

I think what we have here where we automatically do the conversion for them by catching the error and failing the promise is probably easier so they don't really have to think about it. they might not need it that often and it would probably just be for code that could have been called outside of the closure anyway, like encoding something, but for convenience now they can do it if they want to.

kmahar:
new evg patch with test fixes all pass on ubuntu, waiting on MacOS: https://evergreen.mongodb.com/version/5e28a6c161837d46a30fcf99

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: Sounds good.

One thing to note, if `session.end()` could fail (which it can't), then the `flatMapError` would retry the same thing that just failed. Just something to keep in mind if we continue this pattern further. I guess a solution could be to `recover` the first one, though we'd lose the error from `close`.

If we plan to repeat this pattern elsewhere, we could add an internal extension to `EventLoopFuture` to encapsulate this chain in a single callback (I did something simimlar in my cursor PR):
```swift
extension EventLoopFuture {
    internal func completeAfter(promise: EventLoopPromise<Value>, f: @escaping () -> EventLoopFuture<Void>) -> EventLoopFuture<Value> {
        self.flatMap { _ in 
            f().recover { _ in () }
        }.flatMapError { _ in
            return f()
        }.whenComplete { promise.completeWith(self) }
    }
}
```
Then we could use it here like:
```swift
try sessionBody(session).completeAfter(promise: promise) { session.close() }
```
`completeAfter` isn't the best name in the world, but I couldn't really come up with anything better at the moment. It's also possible that this functionality would be better as an extension of `EventLoopPromise` and named `completeWithAfter` or something, e.g.
```swift
promise.completeWithAfter(try sessionBody(session)) { session.close() }
```
Again, this extension part probably isn't necessary unless we plan to reproduce this pattern in the future.

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
patrickfreed: I don't recall anything specific. If the spec doesn't say anything than `countDocuments` should be fine. Its just an aggregate under the hood anyways.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
kmahar: off the top of my head I can't really think of any places we need to use this, but will keep in mind in case any more arise in the rest of the implementation or tests 👍

Sources/MongoSwift/MongoClient.swift
```diff
@@ -322,11 +317,33 @@ public class MongoClient {
...
+                .flatMap { _ in
+                    session.end()
+                }
+                .flatMapError { _ in
```
patrickfreed: If you remove the `multiline_function_chains` rule from the swiftlint config (as I have done in the cursors PR), then these chains can be on the same line

kmahar: done

PR #394, opened 2020-01-27
SWIFT-704 Update async docstrings to reflect error propagation

kmahar:
Most of this PR is just slightly reformatting docstrings to clarify that futures fail with particular errors now, rather than the errors being thrown.

This excludes the docstrings for cursors and change streams, and methods that return cursors and change streams, as I think you started on those already, but if you think I should pick up any of that while you work on implementing those types let me know.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -242,8 +242,6 @@ public class MongoClient {
      *
      * - Throws:
      *   - A `InvalidArgumentError` if the connection string passed in is improperly formatted.
-     *   - A `InvalidArgumentError` if the connection string specifies the use of TLS but libmongoc was not
-     *     built with TLS support.
      */
     public init(
         _ connectionString: String = "mongodb://localhost:27017",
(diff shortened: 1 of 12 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -344,18 +348,17 @@ public class MongoClient {
...
      *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
      *     on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
      *   - session: Optional `ClientSession` to use when executing this command.
      *
-     * - Returns: An `EventLoopFuture<[DatabaseSpecification]>` containing the databases matching provided criteria.
-     *
-     * - Throws:
-     *   - `LogicError` if the provided session is inactive.
-     *   - `EncodingError` if an error is encountered while encoding the options to BSON.
+     * - Returns: An `EventLoopFuture<[DatabaseSpecification]>`. On success, the future contains an array of the
```
patrickfreed: According to the "Multiline Elements" section of Apple's [markup docs](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/MarkupSyntax.html), these blocks can begin on the next line down from their delimiters (i.e. `- Returns` in this case), and they also can be split into multiple paragraphs by inserting single empty lines between sections of text. I personally think its a little clearer if the return content starts on its own line below `- Returns`, and that the error information is separated into its own paragraph.

e.g.
https://github.com/mongodb/mongo-swift-driver/blob/763de6675ab45a7a0a5a7d3b39c99fcd0e12bb6c/Sources/MongoSwift/MongoCursor.swift#L241-L250

The way this is actually written likely could be improved, but I think the overall format is a bit clearer than a single paragraph. What do you think?

kmahar: sgtm, updated all of them

Sources/MongoSwift/MongoClient.swift
```diff
@@ -312,8 +311,16 @@ public class MongoClient {
...
      * The session is only valid within the body of the closure and will be ended after the body completes.
      *
-     * - Throws:
-     *   - `RuntimeError.compatibilityError` if the deployment does not support sessions.
+     * - Parameters:
+     *   - options: Options to use when creating the session.
+     *   - sessionBody: A closure which takes in a `ClientSession` and returns an `EventLoopFuture<T>`.
+     *
+     * - Returns:
+     *    An `EventLoopFuture<T>`, the return value of the user-provided closure.
```
patrickfreed: Could say "On success, the future contains ..." similar to the other docstrings.

kmahar: I opted to not do that here because we actually (from the user's perspective) directly return the return value of their closure (which happens to be a future because of how we've specified the type of `sessionBody`). regardless of whether that succeeds or fails we will directly return it

PR #395, opened 2020-01-31
SWIFT-673 Make MongoSwift.ChangeStream async, and implement MongoSwiftSync.ChangeStream

patrickfreed:
[SWIFT-673](https://jira.mongodb.org/browse/SWIFT-673)

This PR implements the sync and async change streams. It also refactors much of the cursor code to reduce duplicated code. [SWIFT-688](https://jira.mongodb.org/browse/SWIFT-688) is also completed as part of this work.

When reviewing this, I'd recommend starting with `CursorCommon.swift` in the async driver, since that's where the common code was moved to.


Sources/MongoSwift/CursorCommon.swift
```diff
@@ -1,7 +1,10 @@
+import CLibMongoC
```
patrickfreed: The goal of this refactor was to separate concerns and consolidate common code. To that end, the following internal types were introduced/tweaked:
- `protocol CursorProtocol`: this is the existing protocol (formerly named just `Cursor`) that defines shared behavior between our public cursor types. This now defines solely the common _public API_ between cursor like things.
- ` protocol MongocCursorWrapper`: new protocol defining behavior for direct swift wrappers of mongoc types. The two types that conform to this are `MongocCursor` and `MongocChangeStream`. This protocol allows us to use functions like `mongoc_change_stream_next` abstractly.
- `class Cursor<CursorKind: MongocCursor>`: This type contains all the logic that is currently duplicated between the existing `ChangeStream` and `MongoCursor` implementations. It's generic over the type of the mongoc wrapper.
    - Each of `MongoCursor` and `ChangeStream` now wrap one of these internal `Cursor<T>`s, and implement their public API methods by calling different methods on it. We can think of them as consumers of our internal cursor API.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -41,6 +44,17 @@ public class MongoCursor<T: Codable>: Cursor {
         case cached(T?)
         /// Indicates that there is no value cached.
         case none
+
+        /// Get the contents of the cache and clear it.
+        fileprivate mutating func clear() -> T?? {
```
patrickfreed: The outer optional refers to whether or not the cache had anything it in, the inner one refers to whether the result returned from the server was nil or not.

Sources/MongoSwiftSync/MongoCursor.swift
```diff
@@ -9,7 +9,7 @@ import MongoSwift
  *
  * It is safe to `kill(...)` a `MongoCursor` from another thread while it is blocked waiting on results, however.
  */
-public class MongoCursor<T: Codable>: LazySequenceProtocol, IteratorProtocol, Cursor {
+public class MongoCursor<T: Codable>: CursorProtocol {
```
patrickfreed: these protocols are now inherited from by `CursorProtocol`.

Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
```diff
@@ -0,0 +1,127 @@
+import MongoSwiftSync
```
patrickfreed: While I was moving test files over, it seemed like this was worthy of its own file.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -1037,9 +1031,37 @@ final class ChangeStreamTests: MongoSwiftTestCase {
         try coll.insertOne(myType)
 
         let change = try changeStream.nextWithTimeout()
-        expect(changeStream.error).to(beNil())
         expect(change).toNot(beNil())
         expect(change?.operationType).to(equal(.insert))
         expect(change?.fullDocument).to(equal(myType))
     }
+
+    func testChangeStreamLazySequence() throws {
```
patrickfreed: similar test to the existing cursor one

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -117,38 +117,36 @@ internal struct ListCollectionsOperation: Operation {
...
             guard let collections = mongoc_database_find_collections_with_opts(dbPtr, opts._bson) else {
                 fatalError(failedToRetrieveCursorMessage)
             }
             return collections
         }
 
         if self.nameOnly {
-            let cursor: MongoCursor<Document> = try MongoCursor(
-                stealing: indexes,
+            let cursor = try Cursor(
```
patrickfreed: with this new "internal cursor API", we can now operate on the internal cursor type, rather than calling internal methods on our public `MongoCursor` type.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -41,6 +44,17 @@ public class MongoCursor<T: Codable>: Cursor {
...
+            case let .cached(result):
+                self = .none
+                return result
+            case .none:
+                return nil
+            }
+        }
     }
 
     /// Tracks the caching status of this cursor.
```
patrickfreed: I decided to leave the cache as part of `MongoCursor` since it applied only to cursors and not change streams, but we could alternatively put the caching logic into `Cursor<CursorKind>` if we wanted to. It might simplify the code in `MongoCursor` a little bit.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -297,10 +297,10 @@ final class ChangeStreamSpecTests: MongoSwiftTestCase, FailPointConfigured {
 }
 
 /// Class for spec prose tests and other integration tests associated with change streams.
-final class ChangeStreamTests: MongoSwiftTestCase {
+final class SyncChangeStreamTests: MongoSwiftTestCase {
     /// How long in total a change stream should poll for an event or error before returning.
     /// Used as a default value for `ChangeStream.nextWithTimeout`
-    public static let TIMEOUT: TimeInterval = 10
+    public static let TIMEOUT: TimeInterval = 15
```
patrickfreed: I bumped this because I got a spurious error testing against latest sharded again. I'm not sure if it was caused by a performance degradation introduced by the changes in this PR or by the same server side change stream performance problems we've seen already. Based on [SERVER-44356](https://jira.mongodb.org/browse/SERVER-44356), it seems  that events may take around 10s to fire in certain circumstances, so I pushed it past 10s just to be safe.

Sources/MongoSwift/CursorCommon.swift
```diff
@@ -58,3 +61,217 @@ extension EventLoopFuture {
...
+    /// Attempt to retrieve a single document from the server, returning nil if there are no results.
+    internal func tryNext() throws -> Document? {
+        return try self.lock.withLock {
+            try self.getNextDocument()
+        }
+    }
+
+    /// Retreive all the currently available documents in the result set.
+    /// This will not exhaust the cursor.
+    internal func all() throws -> [Document] {
```
patrickfreed: internal implementation only, subject to change in future work

Tests/MongoSwiftSyncTests/MongoCursorTests.swift
```diff
@@ -158,15 +158,16 @@ final class MongoCursorTests: MongoSwiftTestCase {
             expect(cursor.isAlive).to(beTrue())
 
             let queue = DispatchQueue(label: "tailable close")
-            let allDocsLock = DispatchSemaphore(value: 1)
+            let allDocsLock = DispatchSemaphore(value: 0)
```
patrickfreed: this fixes the existing flaky test

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -3,35 +3,38 @@ import Foundation
 import NIO
 import NIOConcurrencyHelpers
 
-internal let ClosedCursorError = LogicError(message: "Cannot advance a completed or failed cursor.")
+internal struct MongocCursor: MongocCursorWrapper {
```
kmahar: similarly to what you did on the change stream wrapper, probably helpful to note here that this is a wrapper of a `mongoc_cursor_t`

Sources/MongoSwift/MongoCursor.swift
kmahar: the API feels non-intuitive to me. did you consider just returning the previous value of `CachedDocument` and switching on that where this method is called?
I think this docstring should at least clarify what you've said in this comment.

Tests/MongoSwiftSyncTests/MongoCursorTests.swift
kmahar: can you explain what the issue was?

Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -117,38 +117,36 @@ internal struct ListCollectionsOperation: Operation {
...
-                decoder: self.database.decoder,
-                session: session
+                session: session,
+                type: .nonTailable
             )
-            defer { cursor.blockingKill() }
+            defer { cursor.kill() }
 
             var names = [String]()
             // call the cursor method for getting the next document directly rather than just iterating through the
```
kmahar: this comment seems a little out of date since there's no `NextOperation` anymore, maybe update it to explain what you mentioned in your comment above about using the internal cursor type

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -27,158 +51,130 @@ public struct ResumeToken: Codable, Equatable {
...
+     *   exhausted, or an error if one occurred. The returned future will not resolve until one of those conditions is
+     *   met, potentially after multiple requests to the server.
+     *
+     *   If the future evaluates to an error, it is likely one of the following:
+     *     - `CommandError` if an error occurs while fetching more results from the server.
+     *     - `LogicError` if this function is called after the change stream has died.
+     *     - `LogicError` if this function is called and the session associated with this change stream is inactive.
+     *     - `DecodingError` if an error occurs decoding the server's response.
+     */
+    public func next() -> EventLoopFuture<T?> {
```
kmahar: so, something users may need to think about if they are using this method is that it is pretty resource-intensive on the thread pool since it continually polls. I'm not sure where the right place to document that consideration is - somewhere on this type? on the `threadPoolSize` parameter? 🤔

PR #396, opened 2020-02-01
SWIFT-703/Reenable Atlas tests

kmahar:
These were disabled for a while because they needed sync cursors to be implemented.

Also, I noticed a while back that I had only implemented tests for half of the URIs. There are both SRV and non-SRV versions in the document. I have no idea which I used the first time around since Evergreen hides private variables in the UI, but anyway, now we are running against both sets.

Also, when I initially implemented these, I was running into some weird issues getting them to work on macOS and we just decided to figure it out later on. I decided to add back macOS and it seems to work fine now so I added that too.

Evergreen patch: https://evergreen.mongodb.com/version/5e34ce3f0ae60662563a2cf0


Sources/AtlasConnectivity/main.swift
```diff
@@ -18,8 +19,7 @@ for config in configs {
         _ = try db.runCommand(["isMaster": 1])
         // findOne
         let coll = db.collection("test")
-        let res = try coll.find(options: FindOptions(limit: 1))
-        for _ in res {} // iterate cursor so we actually talk to server
+        _ = try coll.findOne()
```
mbroadst: nice

PR #398, opened 2020-02-04
SWIFT-639 Update Vapor example

kmahar:
Putting up a draft PR. We probably don't want to merge this for a while because:

1) It depends on master, and not a tagged release
2) It's `@testable` importing in order to use the `all` method on `MongoCursor`
3) It's depending on a beta release of Vapor 4
4) If we do add a `syncClose` method to `MongoClient` as proposed in SWIFT-717 we'll want to use that here

### DIFF
--- a/Examples/VaporExample/Package.swift
+++ b/Examples/VaporExample/Package.swift
@@ -1,11 +1,14 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 import PackageDescription
 
 let package = Package(
     name: "VaporExample",
+    platforms: [
+       .macOS(.v10_14)
+    ],
     dependencies: [
-        .package(url: "https://github.com/vapor/vapor", .upToNextMajor(from: "3.3.0")),
-        .package(url: "https://github.com/mongodb/mongo-swift-driver", .upToNextMajor(from: "0.1.0"))
+        .package(url: "https://github.com/vapor/vapor", .exact("4.0.0-beta.3.24")),
+        .package(url: "https://github.com/mongodb/mongo-swift-driver", .branch("master"))
     ],
     targets: [
         .target(name: "VaporExample", dependencies: ["Vapor", "MongoSwift"])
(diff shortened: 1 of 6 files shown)


Examples/VaporExample/Package.swift
```diff
@@ -1,11 +1,14 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 import PackageDescription
 
 let package = Package(
     name: "VaporExample",
+    platforms: [
+       .macOS(.v10_14)
```
kmahar: requirement of Vapor 4

Examples/VaporExample/Package.swift
```diff
@@ -1,11 +1,14 @@
...
 
 let package = Package(
     name: "VaporExample",
+    platforms: [
+        .macOS(.v10_14)
+    ],
     dependencies: [
-        .package(url: "https://github.com/vapor/vapor", .upToNextMajor(from: "3.3.0")),
-        .package(url: "https://github.com/mongodb/mongo-swift-driver", .upToNextMajor(from: "0.1.0"))
+        .package(url: "https://github.com/vapor/vapor", .exact("4.0.0-beta.3.24")),
```
patrickfreed: We may want to indicate in the somewhere (e.g. the source of this example, this example's readme, or the project's readme itself) that we only support Vaptor 4 due to Swift NIO dependencies. Seeing as Vapor might win out as the de facto swift web server, we probably want to treat compatibility with it with special care.

Examples/VaporExample/Package.swift
patrickfreed: Will this go away once it gets out of beta?

kmahar: actually, [they are planning to drop 10.14](https://forums.swift.org/t/announcement-vapor-4-will-no-longer-support-macos-10-14/33511/6) at some point before the release to use SwiftCrypto, so we'll probably have to update this to 10.15 once that happens.

Examples/VaporExample/Package.swift
kmahar: I've added mention of this to the example's README and to the `Package.swift` file. I'm currently working on a separate PR to update the driver's README and I will mention it there as well.

PR #399, opened 2020-02-04
SWIFT-639 Update Kitura example

kmahar:
Another draft PR for similar reasons to those described in #398 (excluding the concerns about using a beta version of the framework.)


kmahar:
Ready for review now that `toArray()` has been merged into master.

PR #401, opened 2020-02-12
SWIFT-593 Add auth variants to Evergreen, enable auth DNS seedlist tests

kmahar:
Various comments inline. A lot of this is just moving stuff around. Example patch: https://evergreen.mongodb.com/version/5e4382b0d1fe0769574738a3


Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -62,21 +62,19 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
...
+                    weakCertValidation: true
                 )
-
-                // mongoc connects lazily so we need to send a command.
-                let db = client.db("test")
-                _ = try db.runCommand(["isMaster": 1])
+                let opts = ClientOptions(serverMonitoring: true, tlsOptions: tlsOpts)
+                return try self.withTestClient(testCase.uri, options: opts) { client in
+                    // try selecting a server to trigger SDAM
+                    _ = try client.connectionPool.selectServer(forWrites: false)
```
kmahar: so previously we ran an `isMaster` to trigger SDAM starting and get the `TopologyDescriptionChangedEvent`s we make assertions on.

however, this approach doesn't work for the auth tests, because if you provide a connection string with a username:password in it and attempt to run an initial command (even if the command doesn't require auth), libmongoc will both start SDAM and attempt to authenticate in an initial handshake. the clusters `mongo-orchestration` starts up do not have existing users on them that match the users used in the DNS test connection strings, so we would always get an authentication error.
we *could* just ignore that error, but the failed auth attempt means the server we couldn't authenticate to gets removed from the topology description, which makes it difficult to perform the assertions about discovered hosts below.

Kevin suggested switching to using `mongoc_client_select_server` instead, as this will start up the SDAM monitoring threads and resolve TXT records, giving us the information we need below, without performing the handshake or auth.

Tests/MongoSwiftTests/AsyncTestUtils.swift
```diff
@@ -50,10 +51,14 @@ extension MongoCollection {
 }
 
 extension MongoSwiftTestCase {
-    internal func withTestClient<T>(options: ClientOptions? = nil, f: (MongoClient) throws -> T) throws -> T {
+    internal func withTestClient<T>(
+        _ uri: String = MongoSwiftTestCase.getConnectionString(),
```
kmahar: this is used by the DNS seedlist tests now

Tests/MongoSwiftSyncTests/SyncAuthTests.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+@testable import MongoSwift
```
patrickfreed: I think this is one of those imports that happens to work because `MongoSwift` is a dependency of `MongoSwiftSync`. We should probably update `Package.swift` to have `MongoSwift` as a dependency of `MongoSwiftSyncTests` to explicitly call this out, at least until we publicize the connection string stuff and include it in `MongoSwiftSync`.

As an alternative, could we `@_export` the type in `TestsCommon`? Or does that not work with `internal` types?

kmahar: good point. the` @_exported import` doesn't seem to work for an internal type unfortunately, even if I try doing `@_exported @testable import`.

I added it as an explicit dependency. it was already being imported in a few other files as well.

I also noticed that the one dependency this module had on libmongoc could now be removed because I added the `hosts` property on `ConnectionString`, so I went ahead and did that.

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -85,17 +83,42 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
                 continue
             }
 
-            expect(try doTest()).toNot(throwError(), description: testCase.comment ?? "")
+            let connStr: ConnectionString
+            do {
+                connStr = try doTest()
+            } catch {
+                expect(error).to(beNil(), description: testCase.comment ?? "")
```
patrickfreed: This expectation will always fail right? It's probably better to just use `XCTFail` or `fail` directly. Also, I think the `continue` can be omitted.

kmahar: I switched to `XCTFail`. however, the `continue` is needed for the compiler to let me use `connStr` below; otherwise, it doesn't know the variable has definitely been defined when I try to use it below.

I considered not wrapping this in a do/catch at all but then we wouldn't get the name of the test case in the failure message.

PR #403, opened 2020-02-12
SWIFT-314 Use BSONDecoder to decode isMaster responses in SDAM monitoring

patrickfreed:
[SWIFT-314](https://jira.mongodb.org/browse/SWIFT-314)

This PR updates the `ServerDescription` initializer to rely on `BSONDecoder` to decode ismaster responses instead of directly accessing the command response document.

This ticket originally sought to add graceful error handling to SDAM monitoring event construction, but without access to a logger, there really isn't anything more we can do right now. To that end, I added a number of TODO comments indicating that, once we add support for a logger, we can log when encountering errors.

### DIFF
--- a/Sources/MongoSwift/SDAM.swift
+++ b/Sources/MongoSwift/SDAM.swift
@@ -38,6 +38,26 @@ public struct Address: Equatable {
     }
 }
 
+private struct IsMasterResponse: Decodable {
+    fileprivate struct LastWrite: Decodable {
+        public let lastWriteDate: Date?
+    }
+
+    fileprivate let lastWrite: LastWrite?
+    fileprivate let minWireVersion: Int?
+    fileprivate let maxWireVersion: Int?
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/SDAM.swift
```diff
@@ -126,37 +146,29 @@ public struct ServerDescription {
 
         // initialize the rest of the values from the isMaster response.
         // we have to copy because libmongoc owns the pointer.
-        let isMaster = Document(copying: mongoc_server_description_ismaster(description))
-
-        self.lastWriteDate = isMaster["lasWrite"]?.documentValue?["lastWriteDate"]?.dateValue
```
patrickfreed: >"lasWrite"

whoops

kmahar: compelling argument for using `Codable` 🙂

PR #404, opened 2020-02-13
SWIFT-536 Automate building examples in release script

patrickfreed:
[SWIFT-536](https://jira.mongodb.org/browse/SWIFT-536)

### DIFF
--- a/etc/build-examples.sh
+++ b/etc/build-examples.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+examples=("BugReport" "Docs" "KituraExample" "PerfectExample" "VaporExample")
+
+for example_project in ${examples[@]}; do
+    echo "Building $example_project"
+    pushd "Examples/$example_project"
+    swift build
+    build_success=$?
+
+    rm -rf ./.build
+    rm Package.resolved
+    popd
+
+    if [ ${build_success} -eq 0 ]; then
+        echo "================= Building $example_project succeeded ================="
+    else
+        echo "================= Building $example_project failed ================="
+        exit 1
+    fi
+done
--- a/etc/build-examples.sh
+++ b/etc/build-examples.sh
@@ -1,21 +1,35 @@
-#!/bin/sh
+#!/bin/bash
+
+exit_code=0
 
 examples=("BugReport" "Docs" "KituraExample" "PerfectExample" "VaporExample")
 
 for example_project in ${examples[@]}; do
     echo "Building $example_project"
-    pushd "Examples/$example_project"
+    example_dir="Examples/${example_project}"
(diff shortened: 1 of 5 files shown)


patrickfreed:
One problem with this is that it won't ever try to build the changes being released. It'll just verify that the existing examples build against the latest version.

A couple of solutions are possible:
1. We modify the examples to depend on the repository they're included in (i.e. don't go to GitHub, instead depend on "../.."). Then, update the script to make tags locally, test building the examples, then publish the tags to GitHub if building succeeds. 
    - The main problem with this is that the `Package.swift`s included in our examples aren't really representative of how a user would actually include our project. This may not really be a problem given the example `Package.swift` we include in the readme.
2. We search/replace the version string in the examples' `Package.swift`s and verify they build. Seems messy but it gets the job done.
3. Assuming we'll have a branch for each minor version, we could point the examples to the latest version _branch_, verify they build, then publish the version tags afterwards.
    - The main problem here is the same problem here is the same as the first in that it will have a nonstandard `Package.swift`.
    - As a side note, we'll probably want branches for at least every major version to allow for backporting. We probably want one for each minor version too.

I'm not really sure which of these would be the best approach, or really if _any_ of these is the right one. What do you think?

kmahar:
hmmmmm...... I think we should avoid having non-standard `Package.swift`s. option 2 seems reasonable.
option 2 seems better, but agree it's rather messy.
we could also just make running this script a required step of the release process that occurs right after we publish the tags.

etc/vendor-libmongoc.sh
```diff
@@ -28,21 +28,7 @@ if [[ ! "$WORK_DIR" || ! -d "$WORK_DIR" ]]; then
...
-        ;;
-esac
-
-if ! hash ${sed} 2>/dev/null; then
-    echo "You need sed \"${sed}\" to run this script ..."
-    echo
-    echo "On macOS: brew install gnu-sed"
-    exit 43
-fi
+sed="$ETC_DIR/sed.sh"
```
patrickfreed: this was moved to a separate script so that I could use it in `build-examples.sh`.

etc/build-examples.sh
```diff
@@ -0,0 +1,35 @@
...
+    rm -rf ./.build
+    rm Package.resolved
+    git checkout Package.swift
+    popd
+
+    if [ ${build_success} -eq 0 ]; then
+        echo "================= Building $example_project succeeded ================="
+    else
+        echo "================= Building $example_project failed ================="
+        exit_code=1
```
patrickfreed: this was updated so that the script attempts to build all examples now, even if prior builds failed.

PR #406, opened 2020-02-14
SWIFT-718 Add a toArray method to async cursor and change stream

kmahar:
This adds `toArray` to async `MongoCursor` and async `ChangeStream`.

### DIFF
--- a/Sources/MongoSwift/MongoCursor.swift
+++ b/Sources/MongoSwift/MongoCursor.swift
@@ -39,24 +39,6 @@ public class MongoCursor<T: Codable>: CursorProtocol {
     /// Decoder from the client, database, or collection that created this cursor.
     internal let decoder: BSONDecoder
 
-    /// Used to store a cached next value to return, if one exists.
-    private enum CachedDocument {
-        /// Indicates that the associated value is the next value to return. This value may be nil.
-        case cached(T?)
-        /// Indicates that there is no value cached.
-        case none
-
-        /// Get the contents of the cache and clear it.
-        fileprivate mutating func clear() -> CachedDocument {
-            let copy = self
-            self = .none
-            return copy
-        }
-    }
-
-    /// Tracks the caching status of this cursor.
-    private var cached: CachedDocument
-
     /**
      * Initializes a new `MongoCursor` instance. Not meant to be instantiated directly by a user. When `forceIO` is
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/MongoCursor.swift
```diff
@@ -152,27 +152,31 @@ public class MongoCursor<T: Codable>: CursorProtocol {
...
      *
      *    If the future evaluates to an error, that error is likely one of the following:
      *      - `CommandError` if an error occurs while fetching more results from the server.
      *      - `LogicError` if this function is called after the cursor has died.
      *      - `LogicError` if this function is called and the session associated with this cursor is inactive.
      *      - `DecodingError` if an error occurs decoding the server's responses.
      */
-    internal func all() -> EventLoopFuture<[T]> {
+    public func toArray() -> EventLoopFuture<[T]> {
+        guard self.isAlive else {
```
kmahar: without this, we would never error even if you call this on a closed cursor and would always return an empty array, because `self.cached.clear()` would just return nil if the cursor is closed.

Tests/MongoSwiftTests/MongoCursorTests.swift
```diff
@@ -27,25 +27,14 @@ final class AsyncMongoCursorTests: MongoSwiftTestCase {
...
-            var results = try cursor.all().wait()
+            let results = try cursor.toArray().wait()
             expect(results).to(haveCount(1))
             expect(results[0]).to(equal(doc1))
             // cursor should be closed now that its exhausted
             expect(cursor.isAlive).to(beFalse())
             // iterating a dead cursor should error
             expect(try cursor.next().wait()).to(throwError())
 
-            // iterating after calling all should error.
```
kmahar: this was moved down into the `toArray` test below

Sources/MongoSwift/MongoCursor.swift
patrickfreed: Good catch. One thing to note is that calls to `isAlive` and `cached.clear` should probably both only occur while the cursor lock is held, which is making me think we may have to move the cache logic into the internal cursor after all.

kmahar: I've moved the caching logic into the common cursor type, and added a static property on `MongocCursorWrapper` as we discussed to indicate whether we need to cache the first document or not

PR #408, opened 2020-02-19
SWIFT-719 Add a forEach method to async cursors and change streams

kmahar:
Adds `forEach` to both types and does some internal refactoring, will explain in inline comments

### DIFF
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -193,6 +193,18 @@ public class ChangeStream<T: Codable>: CursorProtocol {
         }
     }
 
+    /**
+     * Calls the provided closure with each event in the change stream as it arrives.
+     *
+     * - Returns:
+     *     An `EventLoopFuture<Void>` which will succeed once the change stream is killed via `kill`.
+     *
+     *     If the future evaluates to an error, that error is likely one of the following:
+     *     - `CommandError` if an error occurs while fetching more results from the server.
+     *     - `LogicError` if this function is called after the change stream has died.
+     *     - `LogicError` if this function is called and the session associated with this change stream is inactive.
+     *     - `DecodingError` if an error occurs decoding the server's responses.
+     */
     public func forEach(_ body: @escaping (T) throws -> Void) -> EventLoopFuture<Void> {
         return self.client.operationExecutor.execute {
             while let next = try self.processEvent(self.wrappedCursor.next()) {
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -126,10 +126,13 @@ public class ChangeStream<T: Codable>: CursorProtocol {
      * (specified on the `ChangeStreamOptions` passed  to the method that created this change stream) before trying
      * again.
      *
-     * A thread from the pool will be occupied until the returned future is completed, so performance degradation
-     * is possible if the number of polling change streams is too close to the total number of threads in the thread
-     * pool. To configure the total number of threads in the pool, set the `ClientOptions.threadPoolSize` option
-     * during client creation.
+     * A thread from the driver's internal thread pool will be occupied until the returned future is completed, so
+     * performance degradation is possible if the number of polling change streams is too close to the total number of
+     * threads in the thread pool. To configure the total number of threads in the pool, set the
+     * `ClientOptions.threadPoolSize` option during client creation.
+     *
+     * Note: You *must not* call any change stream methods besides `kill` and `isAlive` while the future returned from
+     * this method is unresolved. Doing so will result in undefined behavior.
      *
      * - Returns:
      *   An `EventLoopFuture<T?>` evaluating to the next `T` in this change stream, `nil` if the change stream is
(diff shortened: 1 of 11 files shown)


Sources/MongoSwift/ChangeStream.swift
```diff
@@ -108,13 +108,11 @@ public class ChangeStream<T: Codable>: CursorProtocol {
         }
     }
 
-    deinit {
```
kmahar: these `deinit`s are now handled by the `deinit` in the internal `Cursor` type so we don't have to call an async method from `deinit`.

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -195,6 +193,26 @@ public class ChangeStream<T: Codable>: CursorProtocol {
         }
     }
 
+    /**
+     * Calls the provided closure with each event in the change stream as it arrives.
+     *
+     * - Returns:
+     *     An `EventLoopFuture<Void>` which will succeed once the change stream is killed via `kill`.
```
kmahar: this feels like a weird way to put this as at first read it seems unrelated to what this method does, but I'm not sure how else to say this, any ideas?

Sources/MongoSwiftSync/ChangeStream.swift
```diff
@@ -29,8 +29,8 @@ public class ChangeStream<T: Codable>: CursorProtocol {
      * This change stream will be dead if `next` returns `nil` or an error. It will also be dead if `tryNext` returns
      * an error, but will still be alive if `tryNext` returns `nil`.
      */
-    public var isAlive: Bool {
-        return self.asyncChangeStream.isAlive
+    public func isAlive() throws -> Bool {
```
kmahar: this throws now, but the only case I can think of that it would throw in is if you closed the client and then called this method.
which is actually true for basically all our methods but we don't document that anywhere else, so I wasn't sure if it was worth mentioning.

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -109,24 +109,58 @@ final class ChangeStreamTests: MongoSwiftTestCase {
...
+            let coll = try db.createCollection(self.getCollectionName()).wait()
+
+            let stream = try coll.watch().wait()
+            expect(try stream.isAlive().wait()).to(beTrue())
+
+            // future won't resolve until we close the stream later
+            let future = stream.forEach(increment)
+
+            _ = try coll.insertOne(["x": 1]).wait()
+            expect(count).toEventually(equal(1), timeout: 10)
```
kmahar: long timeout due to poor sharded change stream performance

Sources/MongoSwift/ChangeStream.swift
patrickfreed: The change stream can be closed via "invalidate" events too ([docs](https://docs.mongodb.com/manual/changeStreams/#open-a-change-stream)). Might be better to just say "... which will resolve/complete once the change stream is closed."

Sources/MongoSwiftSync/ChangeStream.swift
patrickfreed: What about just catching the error and returning false?

Tests/MongoSwiftTests/ChangeStreamTests.swift
```diff
@@ -109,24 +109,58 @@ final class ChangeStreamTests: MongoSwiftTestCase {
...
+
+            _ = try coll.insertOne(["x": 2]).wait()
+            expect(count).toEventually(equal(2), timeout: 10)
+
+            try stream.kill().wait()
+            expect(try future.wait()).toNot(throwError())
+
+            // calling forEach on dead stream should error
+            expect(try stream.forEach(increment).wait()).to(throwError(errorType: LogicError.self))
```
patrickfreed: This test made me realize we might want to give some guidance in the documentation that calling `forEach` multiple times will have undefined behavior, or really calling any methods besides `isAlive` and `kill` while the cursor/change stream has outstanding futures will result in undefined behavior.

kmahar: good point. I updated all the doc strings on relevant cursor and change stream methods to say this.

Sources/MongoSwift/MongoCursor.swift
```diff
@@ -41,6 +41,10 @@ public class MongoCursor<T: Codable>: CursorProtocol {
     /// Decoder from the client, database, or collection that created this cursor.
     internal let decoder: BSONDecoder
 
+    /// The ID used by the server to track the cursor over time. If all of the cursor's results were returnable in a
+    /// batch, or if the cursor contained no results, this value will be nil.
+    public let id: Int64?
```
kmahar: as we discussed yesterday, this is now retrieved initially and stored rather than computing it

patrickfreed: One little nit: maybe say "a _single_ batch", or "_the first_ batch"? Seems just a tad clearer to me.

kmahar: woops, I meant to say that, this phrasing is definitely weird. fixed

PR #409, opened 2020-02-20
SWIFT-360 Allow specification of an event handler instead of NotificationCenter for APM

patrickfreed:
[SWIFT-360](https://jira.mongodb.org/browse/SWIFT-360)

This PR replaces the `NotificationCenter` based APM event reporting API with a protocol oriented one ("callback based"). Users can now implement `CommandEventHandler` and `SDAMEventHandler` protocols on their own types and pass them in via the `ClientOptions` to receive events, rather than attaching observers to clients' `NotificationCenter`s.

Another key change that occurred as part of this PR is that the event protocols were replaced with enums. Enums seemed like nice fits because all the protocol implementation cases were known and that users would likely have been downcasting the protocols quite a bit.

The mapping from old protocol to enum is as follows:
- `MongoEvent` -> _removed_: It didn't seem useful for users to be able to bucket different categories of events together.
- `MongoCommandEvent` -> `CommandEvent`: I removed the `Mongo` prefix since it seemed redundant given that users can always do `MongoSwift.CommandEvent` to disambiguate, though we do still have a number of `Mongo`-prefixed types in the driver, so maybe this wasn't prudent...
- `MongoSDAMEvent` -> _removed_: "SDAM event" was too broad, so I decided to just get rid of it.
- `MongoTopologyUpdateEvent` -> `TopologyEvent`: dropped the "Update" part, since a `TopologyOpeningEvent` isn't really an "update" to the topology per se.
- `MongoServerUpdateEvent` ->  _removed_: Seemed too specific. These events were just included with the `TopologyEvent` enum. They are grouped this way in the SDAM Monitoring spec as well.
- `MongoServerHeartbeatEvent` -> `ServerHeartbeatEvent`


Sources/MongoSwift/APM.swift
```diff
@@ -1,14 +1,109 @@
...
+    func handleTopologyEvent(_ event: TopologyEvent)
+
+    /// Handle a `ServerHeartbeatEvent`.
+    /// Implementing this method is optional. If no implementation is provided, `ServerHeartbeatEvent`s will be ignored.
+    func handleServerHeartbeatEvent(_ event: ServerHeartbeatEvent)
+}
+
+public extension SDAMEventHandler {
+    /// Handle a `TopologyEvent`.
+    /// Implementing this method is optional. If no implementation is provided, `TopologyEvent`s will be ignored.
```
patrickfreed: swiftlint was complaining about these not being documented, so I just duplicated the docstrings here. Not sure what jazzy will actually pick up though.

Sources/MongoSwift/APM.swift
```diff
@@ -1,14 +1,109 @@
...
+    func handleTopologyEvent(_: TopologyEvent) {}
+
+    /// Handle a `ServerHeartbeatEvent`.
+    /// Implementing this method is optional. If no implementation is provided, `ServerHeartbeatEvent`s will be ignored.
+    func handleServerHeartbeatEvent(_: ServerHeartbeatEvent) {}
+}
+
+/// A protocol for events that are directly consumable by users to implement.
+private protocol Publishable {
```
patrickfreed: The following three private protocols help maintain the previous approach of all callbacks funneling into a single function. I also reused some of the approaches from the internal cursor implementation for dealing with the mongoc events.

Sources/MongoSwift/APM.swift
```diff
@@ -1,14 +1,109 @@
...
+
+    /// The address of the server the command was run against.
+    public var serverAddress: Address {
+        return self.event.serverAddress
+    }
 }
 
 /// A protocol for command monitoring events to implement, specifying the command name and other shared fields.
-public protocol MongoCommandEvent: MongoEvent {
+private protocol CommandEventProtocol {
```
patrickfreed: This protocol was retained privately to reduce the boilerplate in `CommandEvent`.

Sources/MongoSwift/APM.swift
```diff
@@ -55,23 +152,35 @@ public struct CommandStartedEvent: MongoCommandEvent, InitializableFromOpaquePoi
...
-    fileprivate init(_ event: OpaquePointer) {
+    fileprivate init(mongocEvent: MongocCommandStartedEvent) {
         // we have to copy because libmongoc owns the pointer.
-        self.command = Document(copying: mongoc_apm_command_started_get_command(event))
-        self.databaseName = String(cString: mongoc_apm_command_started_get_database_name(event))
-        self.commandName = String(cString: mongoc_apm_command_started_get_command_name(event))
-        self.requestId = mongoc_apm_command_started_get_request_id(event)
-        self.operationId = mongoc_apm_command_started_get_operation_id(event)
-        self.serverAddress = Address(mongoc_apm_command_started_get_host(event))
+        self.command = Document(copying: mongoc_apm_command_started_get_command(mongocEvent.ptr))
```
patrickfreed: Because the context function needed to be called elsewhere and virtually, I abstracted it to the protocol. The rest of these are only called here, so I figured it would be overkill to implement them all on the `MongocCommandStartedEvent` type (and the same thing goes for all the other mongoc event type wrappers)

Sources/MongoSwift/MongoClient.swift
```diff
@@ -4,17 +4,12 @@ import NIOConcurrencyHelpers
...
-     * - `CommandSucceededEvent`: `.commandSucceeded`
-     * - `CommandFailedEvent`: `.commandFailed`
-     */
-    public var commandMonitoring: Bool = false
-
     // swiftlint:disable redundant_optional_initialization
 
+    /// Indicates whether this client should publish command monitoring events. If non-nil, `CommandEvent`s will be
+    /// published to the provided handler.
+    public var commandEventHandler: CommandEventHandler? = nil
```
patrickfreed: Instead of a boolean flag, monitoring will simply be enabled when a monitor is present.

Sources/TestsCommon/APMUtils.swift
```diff
@@ -0,0 +1,125 @@
+import Foundation
+import MongoSwift
+
+public class TestCommandEventHandler: CommandEventHandler {
```
patrickfreed: A useful event handler for testing purposes.

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
```diff
@@ -115,49 +115,47 @@ extension MongoClient {
 /// Captures any command monitoring events filtered by type and name that are emitted during the execution of the
 /// provided closure. Only events emitted by the provided client will be captured.
 internal func captureCommandEvents(
```
patrickfreed: These methods aren't as useful anymore, but I kept them in to try to reduce the diff on this change as much as possible.

Tests/MongoSwiftTests/AsyncTestUtils.swift
```diff
@@ -120,54 +120,6 @@ extension MongoSwiftTestCase {
             }
         }
     }
-
-    /// Captures any command monitoring events filtered by type and name that are emitted during the execution of the
-    /// provided closure. Only events emitted by the provided client will be captured.
-    internal func captureCommandEvents(
```
patrickfreed: These weren't being used so I just removed them instead of fixing them.

Sources/MongoSwift/APM.swift
```diff
@@ -129,41 +250,80 @@ public struct CommandFailedEvent: MongoCommandEvent, InitializableFromOpaquePoin
...
+    }
 }
 
-/// A protocol for topology update events that relate to specific servers to implement, specifying the address of the
-/// server being updated.
-public protocol MongoServerUpdateEvent: MongoTopologyUpdateEvent {
-    /// The address of the server being updated.
-    var serverAddress: Address { get }
+/// An SDAM monitoring event related to topology updates.
+public enum TopologyEvent: Publishable {
```
kmahar: I don't have a great sense of how people use SDAM monitoring so maybe it wouldn't be that useful, but should we consider not grouping together the topology + server events into one type?

Sources/TestsCommon/APMUtils.swift
kmahar: I think some brief documentation about the types in this file would be useful to future developers

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -281,55 +272,37 @@ final class SyncClientSessionTests: MongoSwiftTestCase {
...
-        seenFind = false
-        seenGetMore = false
+        monitor.beginMonitoring()
         let cursor = try collection.find(options: FindOptions(batchSize: 2), session: session)
         expect(cursor.next()).toNot(beNil())
         expect(cursor.next()).toNot(beNil())
         expect(cursor.next()).toNot(beNil())
-        expect(seenFind).to(beTrue())
-        expect(seenGetMore).to(beTrue())
+        monitor.stopMonitoring()
```
kmahar: seems like we turn monitoring on/off and clear out events a lot, do you think something like a `withMonitoring` method that allows you to run a closure would be useful?

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
kmahar: should we open a ticket about converting to the new way of doing things at some point in the future?

Sources/MongoSwift/APM.swift
kmahar: yea, hmmm. the one plus I can think of for splitting them up is that it would make things marginally easier if you only care about topology or server but not both.
 
I believe the server events are redundant with the topology events since any server change also results in a topology change, and the topology descriptions contain server descriptions. so it seems plausible to me that you would only want topology events.

kmahar:
lgtm mod the question of splitting up server and topology events, will wait to see what @mbroadst has to say on that. 
this is way nicer than our old approach IMO and I'm thrilled to get rid of all the `NotificationCenter` junk in the tests, great job 👏

Sources/MongoSwift/APM.swift
```diff
@@ -1,14 +1,93 @@
 import CLibMongoC
 import Foundation
 
-/// A protocol for monitoring events to implement, specifying their name.
-public protocol MongoEvent {
-    /// The name this event will be posted under.
-    static var eventName: Notification.Name { get }
+/// A protocol for `CommandEvent` handlers to implement.
+public protocol CommandEventHandler: AnyObject {
```
patrickfreed: These now have class requirements so they can be stored `weak`. I think it probably makes sense that the handlers are always reference types anyways.

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
```diff
@@ -245,4 +245,24 @@ final class SyncMongoClientTests: MongoSwiftTestCase {
         coll = nil
         expect(weakClientRef).to(beNil())
     }
+
+    func testAPMCallbacks() throws {
```
patrickfreed: this is a new test

patrickfreed:
Filed [SWIFT-732](https://jira.mongodb.org/browse/SWIFT-732). Once my [evergreen patch](https://evergreen.mongodb.com/version/5e56e2f01e2d176b46b88423) passes will merge.

PR #410, opened 2020-02-20
SWIFT-639 Update Perfect example

kmahar:
Updates the Perfect example to use the async API.

### DIFF
--- a/Examples/PerfectExample/Package.swift
+++ b/Examples/PerfectExample/Package.swift
@@ -5,7 +5,7 @@ let package = Package(
     name: "PerfectExample",
     dependencies: [
         .package(url: "https://github.com/PerfectlySoft/Perfect-HTTPServer.git", from: "3.0.0"),
-        .package(url: "https://github.com/mongodb/mongo-swift-driver", .branch("master")),
+        .package(url: "https://github.com/mongodb/mongo-swift-driver", .upToNextMajor(from: "1.0.0")),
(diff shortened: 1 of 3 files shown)


Examples/PerfectExample/Package.swift
```diff
@@ -1,11 +1,12 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 import PackageDescription
 
 let package = Package(
     name: "PerfectExample",
     dependencies: [
         .package(url: "https://github.com/PerfectlySoft/Perfect-HTTPServer.git", from: "3.0.0"),
-        .package(url: "https://github.com/mongodb/mongo-swift-driver", .upToNextMajor(from: "0.1.0")),
+        .package(url: "https://github.com/mongodb/mongo-swift-driver", .branch("master")),
```
patrickfreed: before merging this, we should update this version string to be `upToNextMajor(from: "1.0")`. Alternatively, before we merge the 1.0 branch into master we can just update it then.

patrickfreed: I guess we already have SWIFT-537 for updating the versions. I guess its whether we want to do it now per example or all at once later for all examples.

kmahar: just updated this one, will do the others once they've been reviewed and there are no further changes to make (since I can't actually build with the dependency on a not-yet-existing version)

PR #411, opened 2020-02-20
SWIFT-727 Re-export cleanupMongoSwift from sync API

kmahar:
Re-exports the cleanup method so it can be called by importers of `MongoSwiftSync`.

### DIFF
--- a/Sources/MongoSwift/MongoSwift.swift
+++ b/Sources/MongoSwift/MongoSwift.swift
@@ -12,18 +12,6 @@ private final class MongocInitializer {
     }
 }
 
-/// :nodoc:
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoSwift.swift
```diff
@@ -12,18 +12,6 @@ private final class MongocInitializer {
     }
 }
 
-/// :nodoc:
-@available(*, deprecated, message: "Calling this method no longer has any effect.")
```
kmahar: fortunately happened to catch that these were still in the codebase. I searched and didn't find anything else deprecated still around.

PR #414, opened 2020-02-24
SWIFT-639 Update README

kmahar:
Updated README to talk about both modules and add async versions of the examples.


kmahar:
whoops, good catch. I added a couple sentences about that.

PR #416, opened 2020-02-26
Fix documentation generation

kmahar:
This adds a new script which handles generating documentation for both of our modules along with a pretty bare bones index page containing links to both.

It is very likely that this script does not work on Linux. A lot of the extra stuff in the SwiftNIO script is for that. But I don't want to hold up the release on getting the script to work there so I've left it out for now.

FYI: A follow up commit will be needed that cleans up everything in the `docs/` directory besides the new `README.md` and the edited `index.html`, but I figured I would just merge that in directly after this, rather than cluttering up the PR.

If you want to check out how all this looks locally I would suggest doing that deletion manually and then running the script.


etc/generate-docs.sh
```diff
@@ -0,0 +1,31 @@
...
+version=${1}
+
+jazzy_args=(--clean
+            --author 'Matt Broadstone, Kaitlin Mahar, and Patrick Freed' 
+            --readme "docs/README.md" 
+            --author_url https://github.com/mongodb/mongo-swift-driver 
+            --github_url https://github.com/mongodb/mongo-swift-driver 
+            --theme fullwidth 
+            --documentation "Guides/*.md" 
+            --github-file-prefix https://github.com/mongodb/mongo-swift-driver/tree/v${version}
```
kmahar: this makes it so, if someone clicks one of the "view on github" links in the docs, it will take them to the place the code was at the time of the tag, rather than on master.

docs/index.html
```diff
@@ -1,619 +1 @@
...
-        <section id="footer">
-          <p>&copy; 2019 <a class="link" href="" target="_blank" rel="external">Matt Broadstone, Kaitlin Mahar, and Patrick Freed</a>. All rights reserved. (Last updated: 2019-12-18)</p>
-          <p>Generated by <a class="link" href="https://github.com/realm/jazzy" target="_blank" rel="external">jazzy ♪♫ v0.13.0</a>, a <a class="link" href="https://realm.io" target="_blank" rel="external">Realm</a> project.</p>
-        </section>
-      </article>
-    </div>
-  </body>
-</div>
-</html>
+<html><head><meta http-equiv="refresh" content="0; url=mongoswift/index.html" /></head></html>
```
kmahar: copied from NIO. this makes it so if you go to the main documentation URL it forwards you to the MongoSwift docs main page (which has the README listing the modules).

Guides/Development.md
```diff
@@ -70,7 +70,7 @@ Documentation comments should generally be complete sentences and should end wit
 
 Our documentation site is automatically generated from the source code using [Jazzy](https://github.com/realm/jazzy#installation). We regenerate it via our release script each time we release a new version of the driver.
 
-If you'd like to preview how new documentation you've written will look when published, you can regenerate it by running `make documentation` and then inspecting the generated HTML files in `/docs`.
+If you'd like to preview how new documentation you've written will look when published, you can regenerate it by running `.etc/generate-docs.sh` and then inspecting the generated HTML files in `/docs`.
```
patrickfreed: typo: should be `./etc/generate-docs.sh`

kmahar: good catch, fixed

patrickfreed: I think you forgot to push the commit

kmahar:
pushed the commit from before, and also updated the path in `index.html` to have the correct case.

docs/README.md
```diff
@@ -0,0 +1,6 @@
+# mongo-swift-driver Docs
+
+The MongoDB Swift driver contains two modules:
+
+- [MongoSwift](../mongoswift/index.html), containing an asynchronous API and a BSON library
```
patrickfreed: Oh I just realized these may need to be updated too. Otherwise lgtm

kmahar: fixed

PR #420, opened 2020-03-10
SWIFT-439: Deprecate maxScan query option

JamesHeppenstall:

### DIFF
--- a/Sources/MongoSwift/Operations/FindOperation.swift
+++ b/Sources/MongoSwift/Operations/FindOperation.swift
@@ -90,6 +90,7 @@ public struct FindOptions: Codable {
     public var maxAwaitTimeMS: Int64?
 
     /// Maximum number of documents or index keys to scan when executing the query.
+    @available(*, deprecated, message: "maxScan option has been deprecated in MonogDB 4.0")
     public var maxScan: Int64?
 
     /// The maximum amount of time to allow the query to run.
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/Operations/FindOperation.swift
```diff
@@ -90,6 +90,7 @@ public struct FindOptions: Codable {
     public var maxAwaitTimeMS: Int64?
 
     /// Maximum number of documents or index keys to scan when executing the query.
+    @available(*, deprecated, message: "maxScan option has been deprecated in MonogDB 4.0")
```
kmahar: `MonogDB` -> `MongoDB` (same below)

patrickfreed:
Given that we're pre 1.0, I think I'm more in favor of just removing the option altogether rather than having a deprecation. That seems to be [what we did in Rust as well](https://jira.mongodb.org/browse/RUST-85), for reference. I think the DRIVERS ticket says to deprecate it working under the assumption that drivers can't remove it due to backwards compatibility concerns, but in this case we don't have to worry about that.

kmahar:
Ok, I assumed we should leave it for 3.6 compatibility concerns, but if that was deemed acceptable for Rust I'm good with just deleting it altogether. Sorry for the redirection @JamesHeppenstall!

JamesHeppenstall:
No problem 👍  I'll just remove it altogether - ignore my most recent commit!

PR #422, opened 2020-03-10
SWIFT-677: Reduce overhead of listIndexNames

JamesHeppenstall:

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -389,8 +389,14 @@ extension MongoCollection {
      *    - `LogicError` if this collection's parent client has already been closed.
      */
     public func listIndexes(session: ClientSession? = nil) -> EventLoopFuture<MongoCursor<IndexModel>> {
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -406,16 +412,14 @@ extension MongoCollection {
...
-            cursor.toArray()
-        }.flatMapThrowing { models in
-            try models.map { model in
-                guard let name = model.options?.name else {
+    public func listIndexNames(session: ClientSession? = nil) -> EventLoopFuture<[String]> {
+        let operation = ListIndexesOperation(collection: self, nameOnly: true)
+        return self._client.operationExecutor.execute(operation, client: self._client, session: session)
+            .flatMapThrowing { result in
+                guard case let .names(names) = result else {
                     throw InternalError(message: "Server response missing a 'name' field")
```
kmahar: I think this error case now corresponds to us getting back an unexpected result type (i.e. `.models` rather than `.names`), so we could update the error message here to match the "invalid result" one you used above.

PR #423, opened 2020-03-10
SWIFT-735 Support using ObjectId with JSONEncoder and JSONDecoder

kmahar:
As part of this I also made it so you can't use `Document` directly with `JSONEncoder` and `JSONDecoder`, and I updated some outdated docstrings.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -1165,7 +1165,7 @@ extension BSONUndefined: Hashable {
 }
 
 /// Error thrown when a BSONValue type introduced by the driver (e.g. ObjectId) is encoded not using BSONEncoder
-private func bsonEncodingUnsupportedError<T: BSONValue>(value: T, at codingPath: [CodingKey]) -> EncodingError {
+internal func bsonEncodingUnsupportedError<T: BSONValue>(value: T, at codingPath: [CodingKey]) -> EncodingError {
     let description = "Encoding \(T.self) BSONValue type with a non-BSONEncoder is currently unsupported"
 
     return EncodingError.invalidValue(
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -794,11 +794,25 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
     }
 
     public init(from decoder: Decoder) throws {
-        throw getDecodingError(type: ObjectId.self, decoder: decoder)
+        // assumes that the ObjectId is stored as a valid hex string.
+        let container = try decoder.singleValueContainer()
+        let hex = try container.decode(String.self)
+        guard let oid = ObjectId(hex) else {
+            throw DecodingError.dataCorrupted(
+                DecodingError.Context(
+                    codingPath: decoder.codingPath,
+                    debugDescription: "Invalid ObjectId hex string. Got: \(hex)"
+                )
+            )
+        }
+        self = oid
     }
 
-    public func encode(to: Encoder) throws {
-        throw bsonEncodingUnsupportedError(value: self, at: to.codingPath)
+    public func encode(to encoder: Encoder) throws {
+        // encodes the hex string for the `ObjectId`. this method is only ever reached by non-BSON encoders.
+        // BSONEncoder bypasses the method and inserts the ObjectId into a document, which converts it to BSON.
+        var container = encoder.singleValueContainer()
+        try container.encode(self.hex)
     }
 
     internal func encode(to storage: DocumentStorage, forKey key: String) throws {
--- a/Sources/MongoSwift/BSON/Document+Codable.swift
+++ b/Sources/MongoSwift/BSON/Document+Codable.swift
@@ -3,26 +3,12 @@ import Foundation
 /// An extension of `Document` to implement the `Codable` protocol.
 extension Document: Codable {
     public func encode(to encoder: Encoder) throws {
-        // if we're using a `BSONEncoder`, we can just short-circuit
-        // and directly add the `Document` to its storage.
-        if let bsonEncoder = encoder as? _BSONEncoder {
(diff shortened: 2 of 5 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -794,11 +794,25 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
...
+        self = oid
     }
 
-    public func encode(to: Encoder) throws {
-        throw bsonEncodingUnsupportedError(value: self, at: to.codingPath)
+    public func encode(to encoder: Encoder) throws {
+        // encodes the hex string for the `ObjectId`. this method is only ever reached by non-BSON encoders.
+        // BSONEncoder bypasses the method and inserts the ObjectId into a document, which converts it to BSON.
+        var container = encoder.singleValueContainer()
+        try container.encode(self.hex)
```
patrickfreed: I know we discussed this already, but the more I think about this, the more uncomfortable I become with it. A hex string being our encoded representation of `ObjectId` just seems to go so directly against the spirit and purpose of the extended JSON spec. I'm completely fine with us supporting decoding from a hex string, since we can always add support from decoding from extjson too, but we can't ensure that other drivers have the same decoding capability (hence the purpose of the extJSON spec!). 

The main rationale for this approach is that it's clunky for users to do something like `doc._id.$oid` in their frontend applications, right? To me that seems like not enough of a trouble for us to diverge so harshly from the specifications, especially given that extjson libraries can be used in the browser to avoid this.

I think we should probably discuss this a little bit more. cc @mbroadst

kmahar: I'm down to have a meeting about this, but just to provide some additional rationale and justification for the approach proposed here before that:

> The main rationale for this approach is that it's clunky for users to do something like doc._id.$oid in their frontend applications, right? 

as I've mentioned before I continue to be mainly troubled by the thought of `JSONEncoder` producing a mix of normal and extended JSON, and other encoders producing a bunch of strangely formatted stuff with $s, more than anything else. though the decreased verbosity is a plus.

In my view, we should long-term end up with the following:
- Using `JSONEncoder` with our BSON types produces normal JSON.
    - Any future, unknown encoder someone wanted to use e.g. a `YAMLEncoder` would produce a similarly readable output with no $s, extra nesting etc.
- Using a yet-to-be-written `ExtendedJSONEncoder` produces extended JSON and has both relaxed and canonical modes. This would be like `JSONEncoder` but it special cases all `BSONValue` types when they are encountered to produce their extJSON representations instead. I think now this could actually just be written very trivially as a thin wrapper around `BSONEncoder` that returns the resulting document's extJSON representation.

I view this as the best approach because:
a) It doesn't leak concepts relevant to a single encoding format into other encoders and decoders. the encoders and decoders themselves, rather than the types, handle the special cases.
b) behavior is consistent across types we own and types we do not. `JSONEncoder` produces normal JSON only, regardless of whether we own a type being encoded; `ExtendedJSONEncoder` produces extended JSON only, regardless of whether we own a type being encoded.
c) it optimizes for the more common case of people wanting to use normal JSON that is as simple/readable as possible, while providing a way to get extJSON for those who need it for type preservation purposes.

> we can't ensure that other drivers have the same decoding capability (hence the purpose of the extJSON spec!).

I think the main use case of `JSONEncoder` / `JSONDecoder` here is not interop with other drivers, but with APIs, your frontend, whatever. the plain JSON isn't getting sent to the DB, which is how drivers interact with shared data, so I don't know how much it's really a problem to use that alternative representation elsewhere. 
If you are explicitly trying to share data between drivers using JSON via files or something, we have extJSON for that, but I don't think that's the common case we should optimize for here.

patrickfreed: I definitely agree on most points here. For sure, the inconsistencies with builtin types like `Date` are a problem, and it necessitates something like an `ExtendedJSONEncoder` to solve (perhaps using the [`userInfo`](https://developer.apple.com/documentation/swift/encoder/2894907-userinfo) of the `Encoder/Decoder` may be more in line with the `Codable` "ethos" of implementing this stuff on the types themselves). I think the main point of disagreement lies in the definition of "normal JSON". To me, that seems to be exactly why "relaxed extended JSON" was defined.

From the [extended JSON spec rationale section](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#of-relaxed-and-canonical-formats):
>[Design rationale] of Relaxed and Canonical Formats:
>
> There are various use cases for expressing BSON documents in a text rather that binary format. They broadly fall into two categories:
>
> - Type preserving: for things like testing, where one has to describe the expected form of a BSON document, it's helpful to be able to precisely specify expected types. In particular, numeric types need to differentiate between Int32, Int64 and Double forms.
> - JSON-like: for things like a web API, where one is sending a document (or a projection of a document) that only uses ordinary JSON type primitives, it's desirable to represent numbers in the native JSON format. This output is also the most human readable and is useful for debugging and documentation.
>
>The two formats in this specification address these two categories of use cases.

The "JSON-like" part seems similar to the use cases of "normal JSON", albeit restricted to numbers. There is an argument to be made that "relaxed" isn't relaxed enough, but I think consistency with other drivers is more important than refining the format for just this driver.

>If you are explicitly trying to share data between drivers using JSON via files or something, we have extJSON for that, but I don't think that's the common case we should optimize for here.

I was thinking more of e.g. a front end that talks to multiple rest APIs that serve JSON serialized from different mongo deployments. If one backend is written with a driver that serves extJSON, relaxed or not, and another is written with a different driver that serves its own flavor of "normal JSON", it would require the front end to have a translation layer to reconcile the two. 

The caveat in this example is that, as mentioned, we'll never get true extended JSON support from `JSONEncoder / JSONDecoder` for builtin types like `Date`, but I don't see that as a reason for introducing a new JSON format for types that we do have control over.

As a side note: in Vapor 4 (maybe in previous versions too), you can [configure what encoder/decoder is used in the routing](https://docs.vapor.codes/4.0/content/#override-defaults). Once we introduce a way for handling relaxed/canonical, we should probably update the Vapor example to include that configuration.

PR #425, opened 2020-03-11
SWIFT-224 Optimize removing elements from documents

JamesHeppenstall:


Sources/MongoSwift/BSON/Document.swift
```diff
@@ -395,8 +422,12 @@ extension Document {
                 if let newValue = newValue {
                     try self.setValue(for: key, to: newValue)
                 } else {
-                    // TODO: SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
-                    self = self.filter { $0.key != key }
+                    let bsonOpt = bson_new()
+                    withVaList([]) { bson_copy_to_excluding_noinit_va(self._bson, bsonOpt, key, $0) }
+                    guard let bson = bsonOpt else {
```
kmahar: I think it would probably be better to check that the return value of `bson_new` is non-null before passing it into `bson_copy_to_excluding_noinit_va`. if it is null, we would likely crash in that method before getting to the fatal error.

that said, we will probably never even hit the earlier fatal error, either.

`bson_new()` returning an optional value here is sort of funny - Swift translates the signature this way because there's nothing stopping you from just returning `NULL` in place of a `bson_t *` in C, but in practice, the method will never actually do so. It may fail to allocate memory, but in that case libmongoc will send SIGABRT anyway and the method won't return: https://github.com/mongodb/mongo-swift-driver/blob/33aed8834adb1c07d811b3431ceceee1355784a4/Sources/CLibMongoC/bson/bson-memory.c#L72

so, even if we move the error up we probably still won't ever encounter it, but I think it's a bit more future-proof in case `bson_new` ever did start returning NULL without a SIGABRT being sent first.

(same applies to a couple other places you do this)

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +210,19 @@ public class DocumentIterator: IteratorProtocol {
 #endif
     }
 
+    /// Internal helper function for passing an array of strings from Swift to C (i.e. char **)
+    internal static func withArrayOfCStrings<Result>(
+        _ args: [String], _ body: ([UnsafeMutablePointer<CChar>])
```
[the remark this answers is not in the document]
JamesHeppenstall: Yeah, it seems that `strdup` automatically returns an `UnsafeMutablePointer`. More so, `free` expects the input to be mutable.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +210,19 @@ public class DocumentIterator: IteratorProtocol {
...
+            -> Result
+    ) -> Result {
+        // use strdup from C standard library to copy each input string
+        var cStrings = args.compactMap { strdup($0) }
```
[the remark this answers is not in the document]
JamesHeppenstall: `strdup` returns `nil` whenever an error occurs according to the docs. What do you suggest doing in this case?

JamesHeppenstall:
Just a note - I had to add a null pointer as the last element in `va_list` so the function `should_ignore` in `bson.c` knows when the end of the list is reached. According to the docs, calls to `va_arg` do not know the size of the corresponding `va_list`, which caused seg faults in some of the tests.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +157,44 @@ extension Document {
...
-                        try newSelf.setValue(for: pair.key, to: newValue)
                     } else {
-                        try newSelf.setValue(for: pair.key, to: pair.value)
+                        seen ? keysAfter.append(next.key) : keysBefore.append(next.key)
+                    }
+                }
+
+                // To preserve order, we construct a Document excluding keys before the key and a Document excluding
+                // keys after the key. We then merge both Documents and appropriately set the new value for the key.
+                let bsonBeforeOpt = bson_new()
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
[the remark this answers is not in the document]
kmahar: Hmm. So it looks like the only documented error case is failing to allocate memory, not unlike the libmongoc failure discussed elsewhere. probably the safest thing to do would be to make `withArrayofCStrings` a throwing method, and throw an internal error if `strdup` returns `nil`.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +157,51 @@ extension Document {
...
                     } else {
-                        try newSelf.setValue(for: pair.key, to: pair.value)
+                        seen ? keysAfter.append(next.key) : keysBefore.append(next.key)
+                    }
+                }
+
+                // To preserve order, we construct a Document excluding keys before the key and a Document excluding
+                // keys after the key. We then merge both Documents and appropriately set the new value for the key.
+                guard let bson = bson_new() else {
+                    throw InternalError(message: "Invalid BSON data")
```
kmahar: we might consider making this error a static constant so we can reuse it any place we do this. here's an example of another one of those in the driver: https://github.com/mongodb/mongo-swift-driver/blob/33aed8834adb1c07d811b3431ceceee1355784a4/Sources/MongoSwift/ClientSession.swift#L36

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +157,51 @@ extension Document {
 
                 // otherwise, we just create a new document and replace this key
             } else {
-                // TODO: SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
-                var newSelf = Document()
+                guard let iter = DocumentIterator(forDocument: self) else {
+                    throw InternalError(message: "BSON buffer is unexpectedly too small (< 5 bytes)")
```
kmahar: similarly to my comment below, if this error came from somewhere else, consider introducing a reusable error with this message.

Sources/MongoSwift/BSON/DocumentIterator.swift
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -127,23 +127,49 @@ public class DocumentIterator: IteratorProtocol {
...
+        case 0:
+            return [:]
+        default:
+            let firstExclude = excludedKeys.removeFirst()
+            guard let bson = bson_new() else {
+                fatalError("Invalid BSON data")
+            }
+            self.withArrayOfCStrings(excludedKeys) {
+                // We must append a null pointer to the array of CVarArgs so that we know when CVaList terminates.
+                let cVarArgs = $0.map { $0 as CVarArg }
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +212,31 @@ public class DocumentIterator: IteratorProtocol {
 #endif
     }
 
+    /// Internal helper function for passing an array of strings from Swift to C (i.e. char **)
+    internal static func withVaListOfCStrings(
```
kmahar: some API design related nits:

1) when I suggested the name, I hadn't considered that we could move the call to copy in here too, but it definitely makes sense to do so. given that this does more than just create a va list of c strings now, I think we could name this better to reflect what is happening -- maybe even just `copy` or `copyElements` or something like that.

2) in methods where what the parameters actually mean are ambiguous, parameter labels are very useful for someone reading usages of the method later on. here we are currently not requiring them by putting a `_` before each one. if you haven't yet, I would recommend reading over the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/) for some guidance on this. it's a Swift-y API pattern to make function calls read like phrases/sentences, so we should strive for something that reads along the lines of "copy from x to y excluding z".

3) Do we need to be able to specify `firstExclude` separately, or could we just take in all the ones to exclude in `args` and automatically use the first one as `firstExclude`?

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -127,23 +127,45 @@ public class DocumentIterator: IteratorProtocol {
...
+            return [:]
+        default:
+            let firstExclude = excludedKeys.removeFirst()
+            guard let bson = bson_new() else {
+                fatalError("Invalid BSON data")
+            }
+            do {
+                try self.withVaListOfCStrings(excludedKeys, doc._bson, bson, firstExclude)
+            } catch {
+                fatalError("Failed to copy strings")
```
kmahar: nit: rather than duplicating the error message here, which could end up getting out of sync, we could do something like `fatalError("Error creating document subsequence: \(error)")` which will provide a little more context about the call site and will insert the string description of the error that was caught.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -395,8 +417,11 @@ extension Document {
                 if let newValue = newValue {
                     try self.setValue(for: key, to: newValue)
                 } else {
-                    // TODO: SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
-                    self = self.filter { $0.key != key }
+                    guard let bson = bson_new() else {
+                        fatalError("Invalid BSON data")
+                    }
+                    withVaList([]) { bson_copy_to_excluding_noinit_va(self._bson, bson, key, $0) }
```
kmahar: with my proposed change below, we could probably route this through the helper as well.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -395,8 +417,11 @@ extension Document {
                 if let newValue = newValue {
                     try self.setValue(for: key, to: newValue)
                 } else {
-                    // TODO: SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
-                    self = self.filter { $0.key != key }
+                    guard let bson = bson_new() else {
+                        fatalError("Invalid BSON data")
```
kmahar: nit: maybe just use `InvalidBsonError.message`? same below where this is duplicated

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -43,6 +43,13 @@ public class DocumentStorage {
 /// A struct representing the BSON document type.
 @dynamicMemberLookup
 public struct Document {
+    /// Error thrown when calls to bson_new return a null pointer.
+    internal static let InvalidBsonError = InternalError(message: "Invalid BSON data")
```
kmahar: nit: we usually prefer all caps `BSON`, even in variable names

Sources/MongoSwift/BSON/DocumentIterator.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
JamesHeppenstall: 1. I changed the name to `copyElements` as requested.
2. Thanks for this - I wasn't completely sure what the parameter labels were doing in other functions so this is very helpful!
3. I've removed `firstExclude` and the helper function now throws if no keys are provided to exclude. I chose this as one should be using the C function `bson_copy` if this were the case.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -43,6 +43,13 @@ public class DocumentStorage {
 /// A struct representing the BSON document type.
 @dynamicMemberLookup
 public struct Document {
+    /// Error thrown when calls to bson_new return a null pointer.
+    internal static let InvalidBSONError = InternalError(message: "Invalid BSON data")
+
+    /// Error thrown when BSON buffer is too small.
+    internal static let BsonBufferTooSmallError =
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +211,36 @@ public class DocumentIterator: IteratorProtocol {
...
+                cStrings.forEach { free($0) }
+            }
+            // we must append a null pointer to the array of CVarArgs so that we know when CVaList terminates
+            let firstExclude = cStrings.removeFirst()
+            let cVarArgs = cStrings.map { $0 as CVarArg }
+            let nullPtr = unsafeBitCast(0, to: OpaquePointer.self)
+            withVaList(cVarArgs + [nullPtr]) {
+                bson_copy_to_excluding_noinit_va(srcBSON, dstBSON, firstExclude, $0)
+            }
+        } else {
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +211,36 @@ public class DocumentIterator: IteratorProtocol {
...
+                    throw InternalError(message: "Failed to copy strings")
+                }
+                return argv
+            }
+            defer {
+                cStrings.forEach { free($0) }
+            }
+            // we must append a null pointer to the array of CVarArgs so that we know when CVaList terminates
+            let firstExclude = cStrings.removeFirst()
+            let cVarArgs = cStrings.map { $0 as CVarArg }
```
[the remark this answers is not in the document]
JamesHeppenstall: You're right! I think I initially included the cast when debugging seg faults caused when the `CVaList` doesn't include a null pointer as the last element. I've removed it.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +164,34 @@ extension Document {
...
-                        try newSelf.setValue(for: pair.key, to: newValue)
+                        keysBefore.append(next.key)
+                        keysAfter.append(next.key)
                     } else {
-                        try newSelf.setValue(for: pair.key, to: pair.value)
+                        seen ? keysAfter.append(next.key) : keysBefore.append(next.key)
                     }
                 }
+
+                // To preserve order, we construct a Document excluding keys before the key and a Document excluding
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -127,23 +127,44 @@ public class DocumentIterator: IteratorProtocol {
...
+                break
+            }
+        }
+
+        while let next = iter.next() {
+            excludedKeys.append(next.key)
+        }
+
+        switch excludedKeys.count {
+        case 0:
```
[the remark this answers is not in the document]
JamesHeppenstall: I've included an additional test for indexing into documents since there didn't seem to be one already.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +164,34 @@ extension Document {
...
                 // otherwise, we just create a new document and replace this key
             } else {
-                // TODO: SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
-                var newSelf = Document()
+                guard let iter = DocumentIterator(forDocument: self) else {
+                    throw Document.BSONBufferTooSmallError
+                }
+
+                var keysBefore: [String] = []
+                var keysAfter: [String] = []
```
patrickfreed: based on these names, I would've guessed that `keysBefore` referred to the "keys that come before the set one", but based on their usage it seems to be the opposite. To reduce the confusion, could you just add comments to each of these indicating what they refer to?

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +212,35 @@ public class DocumentIterator: IteratorProtocol {
 #endif
     }
 
+    /// Internal helper function for copying elements from some source 'bson_t' to a destination 'bson_t' while
+    /// excluding a non-zero number of keys
+    internal static func copyElements(
```
patrickfreed: It might be a little more Swift-like to write this as an instance method on `Document` that takes in the source `Document`  as a parameter.
e.g.
```swift
let doc = Document()
doc.copyElements(from: otherDoc, excluding: ["key"])
```
To access the underlying pointers, you can use the free-floating [`withMutableBSONPointer`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/BSON/Document.swift#L512) function for mutable access or the [`_bson`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/BSON/Document.swift#L53) property on `Document` for read-only access. This makes me realize we should probably move both of these onto `Document` and use similar names for both of them.

In general, we try to limit working directly with libbson / libmongoc wherever possible and instead use the abstractions we've written around them. When we do need to drop into C, we try to do it within `withXPointer` blocks. An example of this is in [`ListDatabasesOperation`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/Operations/ListDatabasesOperation.swift#L59).

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +164,34 @@ extension Document {
...
+
+                // To preserve order, we construct a Document excluding keys after the key, set the new value for the
+                // key, and then append all keys after the key.
+                guard let bson = bson_new() else {
+                    throw Document.InvalidBSONError
+                }
+
+                try DocumentIterator.copyElements(from: self._bson, to: bson, excluding: keysAfter)
+                var newSelf = Document(stealing: bson)
+                try newSelf.setValue(for: key, to: newValue)
```
patrickfreed: we can pass in `checkForKey: false` here to skip the key checking branch, since we know we excluded it in the previous statement.

Sources/MongoSwift/BSON/DocumentIterator.swift
```diff
@@ -190,6 +212,35 @@ public class DocumentIterator: IteratorProtocol {
...
+        to dstBSON: OpaquePointer,
+        excluding keys: [String]
+    ) throws {
+        guard !keys.isEmpty else {
+            throw InternalError(message: "No keys to exclude, use 'bson_copy' instead")
+        }
+
+        // use strdup from C standard library to copy each input string
+        var cStrings: [UnsafeMutablePointer<CChar>] = try keys.compactMap {
+            guard let argv = strdup($0) else {
```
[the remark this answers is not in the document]
kmahar: fyi, we spent a while looking into this - [relevant blog post](https://oleb.net/blog/2016/10/swift-array-of-c-strings/)

see the end of the post for the solution that inspired this... that one has body taking in an array of optional pointers so it conveniently ignores the `strdup` failures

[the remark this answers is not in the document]
kmahar: see also: https://github.com/apple/swift/blob/ddcd8d324a309844b0e1c39bcd5f97ea957aca41/stdlib/private/SwiftPrivate/SwiftPrivate.swift#L60

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -157,17 +164,34 @@ extension Document {
 
                 // otherwise, we just create a new document and replace this key
             } else {
-                // TODO: SWIFT-224: use va_list variant of bson_copy_to_excluding to improve performance
-                var newSelf = Document()
+                guard let iter = DocumentIterator(forDocument: self) else {
```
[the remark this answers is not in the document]
JamesHeppenstall: I agree that having the document iterator throw would be more straightforward. My only issue with this is that all the current uses of `DocumentIterator` in `Document+Collection.swift` and `Document+Sequence.swift` would also need to be changed. Maybe we could file a new ticket to make these changes?

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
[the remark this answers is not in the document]
JamesHeppenstall: I've reverted `copyElements` back to the solution in the blog post with an additional step where I unwrap the optionals. It's not the most efficient but let me know what you think!

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -238,6 +255,34 @@ extension Document {
...
+                throw InternalError(message: "Failed to copy strings")
+            }
+            return argv
+        }
+
+        // we must append a null pointer to the array of C strings so that we know when CVaList terminates
+        let firstExclude = cStrings.removeFirst()
+        let nullPtr = unsafeBitCast(0, to: OpaquePointer.self)
+        withVaList(cStrings + [nullPtr]) {
+            bson_copy_to_excluding_noinit_va(self._bson, otherDoc._bson, firstExclude, $0)
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/DocumentIterator.swift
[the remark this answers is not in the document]
JamesHeppenstall: Thanks for the thorough overview about why calling C functions in Swift is not desirable. Using your suggestions and the attached links, I've modified `copyElements` to not call any C functions apart from `bson_copy_to_excluding_noinit_va`.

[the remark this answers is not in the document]
patrickfreed: Awesome, looks really good. 

Please note that what I said there is reflective of my opinions on this subject and is not necessarily representative of the Swift community as a whole. I was actually reading a thread on the Swift forums the other day where someone said they like to drop into C for a lot of the heavy lifting and only pop up to Swift at the API level. Similarly, in the world of Rust, there are a lot of discussions going on about how much `unsafe` code should be used and in what circumstances. Some people like to write large swaths of their programs in `unsafe` blocks and others disallow it completely in anything they write. There's a whole spectrum of opinions on this subject across languages, and I just so happen to fall on the safer side.

Also, I apologize for coming in here at the last minute and dropping all this feedback after the PR was really close to merging. I know that can be frustrating, but sometimes it happens during multi-stage reviews. I think that fighting scope creep by filing tickets (as you suggested in the other thread) is a good thing to do to prevent any last minute surprises and to keep things from dragging on.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -238,6 +255,35 @@ extension Document {
...
+    /// excluding a non-zero number of keys
+    internal func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
+        guard !keys.isEmpty else {
+            throw InternalError(message: "No keys to exclude, use 'bson_copy' instead")
+        }
+
+        let cStrings: [ContiguousArray<CChar>] = keys.map { $0.utf8CString }
+        var cPtrs: [UnsafePointer<CChar>] = []
+
+        for cString in cStrings {
```
patrickfreed: this can also be written as a map over `cStrings`

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -238,6 +255,35 @@ extension Document {
...
+            let bufferPtr: UnsafeBufferPointer<CChar> = cString.withUnsafeBufferPointer { $0 }
+            guard let cPtr = bufferPtr.baseAddress else {
```
patrickfreed: Just as a curiosity: does this compile when just passing in the buffer pointers (or even the `ContiguousArray`s themselves) to `withVaList`?

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: Done.

Sources/MongoSwift/BSON/Document.swift
JamesHeppenstall: No it won't because `withVaList` expects an array of [`CVarArg`](https://developer.apple.com/documentation/swift/cvararg)s and neither `ContiguousArray` nor `UnsafeBufferPointer` are conforming types.

PR #426, opened 2020-03-19
SWIFT-738 Make ReadPreference a struct

kmahar:
This change makes `ReadPreference` a struct which is backed by an instance of a private class wrapping the libmongoc type.


Sources/MongoSwift/ReadPreference.swift
```diff
@@ -109,61 +97,97 @@ public final class ReadPreference {
...
             }
-
-            let tags = try BSONEncoder().encode(Document(tagSets.map { .document($0) }))
-            mongoc_read_prefs_set_tags(self._readPreference, tags._bson)
+            let tags = Document(tagSets.map { .document($0) })
+            mongoc_read_prefs_set_tags(self.readPref, tags._bson)
         }
 
         if let maxStalenessSeconds = maxStalenessSeconds {
+            guard mode != .primary else {
```
kmahar: reading the docs I realized this is an invalid combo we had previously missed.

patrickfreed: One way to prevent this at the type level is to define `ReadPreference` as an enum where only the cases that can have thees fields have them defined. 
e.g.
```swift
enum ReadPreference {
    case primary,
    case secondary(maxStaleness: Int? = nil, tagSets: [Document]? = nil),
    case nearest(maxStaleness: Int? = nil, tagSets: [Document]? = nil),
    ...
}

let primary = ReadPreference.primary
let secondary = ReadPreference.secondary(maxStaleness: 123)
let compileError = ReadPrefernce.primary(maxStaleness: 123)
// etc
```
This has the unfortunate problem of not being able to have any private data, meaning the private class approach won't be able to work with it. I would say that from a user's perspective, this might be a mildly more convenient API. What do you think?

kmahar: hmmmm very interesting idea. I agree it does seem mildly better. should read concern also be an enum? all it is now is a struct wrapping an enum... 🤔 write concern could maybe be one too. enums really are addictive... I guess my main concern for all of these is, do we think we could end up needing to add more cases in the future? I don't know how much it would really matter if so because users probably aren't switching on these types themselves in the same way you would with e.g. errors.

patrickfreed: We have the `Mode` enum, so even if `ReadPreference` isn't an enum itself, we're still facing the "adding more cases" problem unless we add an `.other` case. It looks like going all the way back to 2.2 had [the same read preferences](https://docs.mongodb.com/v2.2/applications/replication/#read-preference), so we're probably good here.

I think `WriteConcern` is probably okay as a struct. Because each case of `w` is compatible with both journal and `wtimeoutMS`, we don't gain too much from making it an enum. The syntax would be a little less verbose, but it wouldn't really map to to the schema of a write concern:
```swift
let wc: WriteConcern = .majority(journal: true, wTimeoutMS: 100)
```
vs:
```
{ w: <value>, j: <boolean>, wtimeout: <number> }
```

`ReadConcern` does look like it could just be an enum. It already has the `.other` case, so its forward compatible. It looks like every version of MongoDB since read concern has been released has added a new read concern mode, so that case is definitely necessary.

patrickfreed: Something to note: the read concern spec is [relatively explicit](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#read-concern) about the layout of a read concern being a class with a single level field. [This section](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unknown-levels-and-additional-options-for-string-based-readconcerns) seems to indicate that new options (in addition to new levels) could potentially be added to read concerns in the future, meaning a straight enum might not be the best choice. There's also the benefit of consistency between read/write concerns if they're both enums.

kmahar: sorry for my delay on revisiting this. so I just spent a while speculatively making these changes to see how it would go. some things that came up:

1) so I defined the enum as follows:
```swift
public enum ReadPreference: Equatable {
    case primary
    case primaryPreferred(maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil)
    case secondary(maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil)
    case secondaryPreferred(maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil)
    case nearest(maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil)
}
```

usage comparison:
```swift
/// 1. creating a new primary read preference as a variable
let rp = ReadPreference(.primary) // old
let rp = ReadPreference.primary // new
let rp: ReadPreference = .primary // alternate new

/// 2. creating a new non primary read preference (with no additional opts) as a variable
let rp = ReadPreference(.secondary) // old
let rp = ReadPreference.secondary() // new, also unexpected to me that I needed the ()! 
let rp: ReadPreference = .secondary() // alt new

/// 3. creating a new non primary read preference (with additional opts) as a variable
let rp = try ReadPreference(.secondary, tagSets: [...]) // old
let rp = ReadPreference.secondary(tagSets: [...]) // new
let rp: ReadPreference = .secondary(tagSets: [...]) // alt new

/// 4. passing a *primary* read preference directly into an options initializer
let opts = RunCommandOptions(readPreference: ReadPreference(.primary)) // old
let opts = RunCommandOptions(readPreference: .primary) // new

/// 5. passing a non-primary read preference (no opts) directly into options initializer
let opts = RunCommandOptions(readPreference: ReadPreference(.secondary)) // old
let opts = RunCommandOptions(readPreference: .secondary()) // new

/// 6. passing a non-primary read preference (with additional opts) directly into options initializer
let opts = try RunCommandOptions(readPreference: ReadPreference(.secondary, tagSets: [...])) // old
let opts = RunCommandOptions(readPreference: .secondary(tagSets: [...])) // new
```

to me, the only place we really see much improvement here is in the case where you are directly passing a read preference into an options struct. this may be the most common usage, so maybe that should be prioritized, but I'm not positive. 

otherwise, I think the "enum with optional associated values" thing is a littler weird, because as I noted above it was unexpected to me that I had to add the parentheses in to make it compile. maybe people know that but I do wonder if it introduces a usability tradeoff.

2) although the enum solves the problem of "when you are allowed to use max staleness and tag sets", there isn't a great place to handle the error re: max staleness value being too small. we can't do it while initializing, so I guess we'd have to do it in the `withMongocReadPreference` helper, which would then have to throw unconditionally. that's not catastrophic, but does delay the error reporting until later on.


so overall I'm not totally sure where I stand here after actually trying this out. I think the enum is a bit more convenient in one situation syntactically but in others doesn't seem to help out much otherwise since there is still an error case we need to handle eventually.

cc @mbroadst as well since more API discussion 🙂

patrickfreed: I did a brief survey of other drivers to see how they handled this:
- PyMongo: separate class per read preference mode ([docs](https://api.mongodb.com/python/current/api/pymongo/read_preferences.html))
- Go: struct with mode / tag sets / maxStaleness, but it has various static methods for creating specific ones ([docs](https://godoc.org/go.mongodb.org/mongo-driver/mongo/readpref#ReadPref))
- Node: object with mode / tag sets / maxStaleness ([docs](https://mongodb.github.io/node-mongodb-native/api-generated/readpreference.html))
- C#: class with mode / tag sets / maxStaleness ([docs](https://mongodb.github.io/mongo-csharp-driver/2.7/apidocs/html/M_MongoDB_Driver_ReadPreference__ctor.htm))
- Java: abstract class with a bunch of static methods for each mode, no public constructor though ([docs](http://mongodb.github.io/mongo-java-driver/3.6/javadoc/com/mongodb/ReadPreference.html))

None of these languages have enums with associated values though, so I'm not sure how useful the survey results are to answering our question.

One interesting idea we could explore that go / java seem to do is to define a bunch of static methods on `ReadPreference` used to initialize  the various modes. This is similar to that error proposal you sent in slack a little while back. e.g.:
```swift
struct ReadPreference {
    public static func primary() -> ReadPreference { ... }
    public static func primaryPreferred(tagSets = nil, maxStaleness = nil) throws -> ReadPreference { ... }
    // etc...
}
```
This is essentially the same thing as an enum, but it allows us to do whatever validation we need to. I was originally leaning more towards enum because I didn't like the idea of throwing in the read preference initializer, but I've sort of gotten over that. "throwing" in Swift isn't like throwing an exception in C++ (where I think this idiom of not throwing in a constructor came from) or Java, it's just sugar around returning a `Result<T>`, so it's not like there's performance concerns. Also, we're already doing this in `WriteConcern`'s initializer and probably will continue to do so.

mbroadst: Does this work:
```
struct ReadPreference {
    let mode: Mode

    public static let primary    = ReadPreference(.primary)
    public static let secondary  = ReadPreference(.secondary)
    public static let nearest   = ReadPreference(.nearest)

    public enum Mode {
        case primary
        case secondary
        case nearest
    }
}
```

```
coll.insertOne({ a: 42 }, readPreference: .nearest);
```
?

kmahar: I like this idea, and yes @mbroadst we could have those in addition for the cases where you don't want to pass in any parameters. 

so if we define the type like
```swift
public struct ReadPreference {
    public enum Mode {
        case primary
        case secondary
        case nearest
    }

    public let mode: Mode
    public let maxStalenessSeconds: Int?
    public let tagSets: [Document]?

    public static let primary = ReadPreference(.primary)
    public static let secondary = ReadPreference(.secondary)
    public static let nearest = ReadPreference(.nearest)

    static func secondary(maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil) throws -> ReadPreference {
        return try ReadPreference(.secondary, maxStalenessSeconds: maxStalenessSeconds, tagSets: tagSets)
    }

    static func nearest(maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil) throws -> ReadPreference {
        return try ReadPreference(.nearest, maxStalenessSeconds: maxStalenessSeconds, tagSets: tagSets)
    }

    /// this init allows us to create the static properties without a `try` (we can't throw from that spot)
    internal init(_ mode: Mode) {
        self.mode = mode
        self.maxStalenessSeconds = nil
        self.tagSets = nil
    }
 
    /// all the static methods with parameters call through to here
    internal init(_ mode: Mode, maxStalenessSeconds: Int? = nil, tagSets: [Document]? = nil) throws {
        // do validation of max staleness etc here
        self.mode = mode
        self.maxStalenessSeconds = maxStalenessSeconds
        self.tagSets = tagSets
    }
}
```

then usage looks like:
```swift
let rp1 = ReadPreference.primary
let rp2 = ReadPreference.secondary
let rp3 = try ReadPreference.secondary(maxStalenessSeconds: 200)

let opts1 = RunCommandOptions(readPreference: .primary)
let opts2 = RunCommandOptions(readPreference: .secondary)
let opts3 = try RunCommandOptions(readPreference: .secondary(maxStalenessSeconds: 200))
```

this seems like a nice mix between the enum approach and the struct approach, I like it

kmahar: my only concern is if this API pattern with static methods and properties pretending to be enums is discoverable/intuitive to users since it's kind of tricky how it works, but I guess SwiftPM does use it 🙂

mbroadst: nice! Let's do it

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -134,36 +255,35 @@ public final class ReadPreference {
...
-}
 
-/// An extension of `ReadPreference` to make it `Equatable`.
-extension ReadPreference: Equatable {
-    public static func == (lhs: ReadPreference, rhs: ReadPreference) -> Bool {
-        lhs.mode == rhs.mode && lhs.tagSets == rhs.tagSets &&
-            lhs.maxStalenessSeconds == rhs.maxStalenessSeconds
+    fileprivate var maxStalenessSeconds: Int? {
+        let maxStaleness = mongoc_read_prefs_get_max_staleness_seconds(self.readPref)
+        return maxStaleness == MONGOC_NO_MAX_STALENESS ? nil : Int(exactly: maxStaleness)
```
kmahar, line 283: this could technically return nil if you were on a 32 bit machine and somehow got a hold of a libmongoc read preference with a max staleness value greater than Int32.max, but that seems pretty unlikely. but if you think it would be better to fatal error or something here let me know.

kmahar:
Ok, I've updated to use the approach we discussed above.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -50,81 +49,203 @@ public final class ReadPreference {
...
     }
 
-    /// The tags of this `ReadPreference`
-    public var tagSets: [Document] {
-        guard let bson = mongoc_read_prefs_get_tags(self._readPreference) else {
-            fatalError("Failed to retrieve read preference tags")
-        }
-        // we have to copy because libmongoc owns the pointer.
-        let wrapped = Document(copying: bson)
+    /// Optionally specified tag sets, indicating a member or members of a replica set to target.
```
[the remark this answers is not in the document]
kmahar: updated here and in the static method docstrings

PR #427, opened 2020-03-19
SWIFT-730 Make ChangeStream.kill result discardable

kmahar:
I got to the point of making the PR before I realized this, so I'm not sure we should actually merge this but I figured I'd open for discussion anyway...

while this can never fail for reasons related to the future itself having an issue, it _can_ still fail if, for example, you close the parent client before calling this method, or if you shut down the `EventLoopGroup` you provided to the client.

these are definitely weird edge cases, but, I'm not sure it's fair to say this *never* fails.

but I also don't remember if we already discussed this before and I just forgot to put the justification in the ticket.

what do you think?

### DIFF
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -241,6 +241,7 @@ public class ChangeStream<T: Codable>: CursorProtocol {
      * - Returns:
      *   An `EventLoopFuture` that evaluates when the change stream has completed closing. This future should not fail.
      */
+    @discardableResult
     public func kill() -> EventLoopFuture<Void> {
         return self.client.operationExecutor.execute {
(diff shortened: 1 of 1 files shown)


patrickfreed:
Hm yeah that's a good point. Also, in a pure Swift world, we'd probably send a killCursors command explicitly that we may want to return the result of. I think you're right and we probably shouldn't merge this. Going from non discardable to discardable isn't a breaking change, but if we ever wanted to come back from to non discardable, it would be. So it's probably just safer to leave these alone. 

Did we already have cursor's kill be discardable? We may want to revert that if we aren't going to merge this one.

kmahar:
> Did we already have cursor's kill be discardable? 

we didn't make it discardable so it should be fine as is. 

closing this PR and will close the ticket as won't fix.

PR #430, opened 2020-03-24
SWIFT-744 Update ClientSessionOptions to support transactions

JamesHeppenstall:


Sources/MongoSwift/Operations/StartSessionOperation.swift
```diff
@@ -23,9 +27,15 @@ private func withSessionOpts<T>(
     // swiftlint:disable:next force_unwrapping
     var opts = mongoc_session_opts_new()! // always returns a value
     defer { mongoc_session_opts_destroy(opts) }
+
     if let causalConsistency = options?.causalConsistency {
         mongoc_session_opts_set_causal_consistency(opts, causalConsistency)
     }
+
+    withMongocTransactionOpts(wrapping: options?.defaultTransactionOptions) {
```
patrickfreed: I shouldve mentioned this in the previous PR, but I think maybe `copying` would be a better label here, since we're creating a new `mongoc_transactions_opts_t`, not making a view on an existing one. I think `from` would also be fine. What do you guys think?

JamesHeppenstall: I think `copying` is a good label too. Will make the changes and then update the PR.

kmahar: SGTM

PR #431, opened 2020-03-25
SWIFT-745 Implement transaction operations

JamesHeppenstall:


Sources/MongoSwift/Operations/AbortTransactionOperation.swift
```diff
@@ -0,0 +1,21 @@
...
+
+/// An operation corresponding to aborting a transaction.
+internal struct AbortTransactionOperation: Operation {
+    internal func execute(using _: Connection, session: ClientSession?) throws {
+        guard let session = session else {
+            throw InternalError(message: "No session provided to AbortTransactionOperation")
+        }
+
+        // session either was not started or ended
+        guard case let .started(sessionPtr, _) = session.state else {
```
kmahar: since we handle starting the session ourselves in `OperationExecutor.execute` by calling `startIfNeeded` and chain the execution of this op onto that call, I would say it's an `InternalError` here if the session is not started. I think the same applies to the start and commit operations.

if the session has been ended, potentially that could be considered a logic error, as it would indicate the user tried call this after ending the session / didn't properly synchronize their calls to abort and then end the session. that said, maybe it's ok to just consider this a no-op... WDYT @patrickfreed ?

Sources/MongoSwift/MongoError.swift
```diff
@@ -261,6 +261,8 @@ private func parseMongocError(_ error: bson_error_t, reply: Document?) -> MongoE
         return LogicError(message: message)
     case (MONGOC_ERROR_PROTOCOL, MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION):
         return CompatibilityError(message: message)
+    case (MONGOC_ERROR_TRANSACTION, MONGOC_ERROR_TRANSACTION_INVALID_STATE):
```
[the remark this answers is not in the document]
JamesHeppenstall: The C driver [error reporting doc](http://mongoc.org/libmongoc/1.15.1/errors.html) states that this error occurs when "You attempted to start a transaction when one is already in progress, or commit or abort when there is no transaction". Essentially, an invalid transition in the [transaction state model](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#clientsession-changes)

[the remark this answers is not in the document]
kmahar: ahhh I forgot this is a switch on domain+code and this corresponds to a single error, I was thinking these were two separate error codes 👍

Sources/MongoSwift/Operations/AbortTransactionOperation.swift
JamesHeppenstall: Ok, I'll update this so that session not started throws an `InternalError` and session ended is a no-op for now.

patrickfreed: While it is harmless to call `abortTransaction` on an ended session, doing so is probably indicative of a bug in a user's program (e.g. they escaped a session from `withSession`). For that reason, I think we should be throwing a `LogicError`.

JamesHeppenstall: That makes sense @patrickfreed. I've updated `AbortTransactionOperation`.

patrickfreed: I think this also applies to the other operations as well. Also, instead of using `default` plus a comment, we can just explicitly match the ended case.

JamesHeppenstall: Ok, I've updated all the other operations as well.

kmahar: I agree it would be good to get rid of the `default`. if we ever add another state to `ClientSession`, it will prevent the compiler from telling us that we need to handle that case in these places.

Patrick's suggestion to explicitly match `ended` and return accomplishes that but I think stylistically I have a slight preference for not duplicating the matching on that case.

to condense the logic we could either nest the logic calling the mongoc method for a started session under the `.started` case in the `switch`, or, do something like this:
```swift
let sessionPtr: OpaquePointer
switch session.state {
case let .started(ptr, _):
    sessionPtr = ptr
case .notStarted:
    throw InternalError(message: "Session not started for AbortTransactionOperation")
case .ended:
    throw LogicError(message: "Tried to abort transaction on an ended session")
}

// use sessionPtr down here
```

JamesHeppenstall: Apologies for mixing up above - I've made the relevant changes now.

PR #432, opened 2020-03-25
SWIFT-746 Implement asynchronous transactions API

JamesHeppenstall:

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -227,11 +227,18 @@ public final class ClientSession {
      * Starts a multi-document transaction for all subsequent operations in this session. Any options provided in
      * `options` override the default transaction options for this session and any options inherited from
      * `MongoClient`. The transaction must be completed with `commitTransaction` or `abortTransaction`. An in-progress
-     * transaction is automatically aborted when `ClientSession.end()` is called. Throws an error if the session
-     * already has an in-progress transaction. Throws a `LogicError` if the session has ended.
+     * transaction is automatically aborted when `ClientSession.end()` is called.
      *
      * - Parameters:
      *   - options: The options to use when starting this transaction
+     *
+     * - Returns:
+     *    An `EventLoopFuture<Void>` that succeeds when `startTransaction` is successful.
+     *
+     *    If the future fails, the error is likely one of the following:
+     *    - `CommandError` if an error occurs that prevents the command from executing.
+     *    - `LogicError` if the session already has an in-progress transaction.
+     *    - `LogicError` if `startTransaction` is called on an ended session.
      */
     public func startTransaction(_ options: TransactionOptions?) -> EventLoopFuture<Void> {
         switch self.state {
@@ -244,8 +251,15 @@ public final class ClientSession {
     }
 
     /**
-     * Commits a multi-document transaction for this session. Throws an error if the session has no in-progress
-     * transaction, or if there is a server or network error. Throws a `LogicError` if the session has ended.
+     * Commits a multi-document transaction for this session. Server and network errors are not ignored.
+     *
+     * - Returns:
+     *    An `EventLoopFuture<Void>` that succeeds when `commitTransaction` is successful.
+     *
+     *    If the future fails, the error is likely one of the following:
+     *    - `CommandError` if an error occurs that prevents the command from executing.
+     *    - `LogicError` if the session has no in-progress transaction.
+     *    - `LogicError` if `commitTransaction` is called on an ended session.
      */
     public func commitTransaction() -> EventLoopFuture<Void> {
         switch self.state {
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -222,4 +222,52 @@ public final class ClientSession {
             }
         }
     }
+
+    /**
+     * Starts a multi-document transaction for all subsequent operations in this session. Any options provided in
```
kmahar: I think you've got all the important content in this docstring, just a wording/formatting nit: since the async methods don't actually throw, we word their docstrings a bit differently. I think we should match that here. for example, see [this docstring](https://github.com/mongodb/mongo-swift-driver/blob/ec2bba2fa2779360a427695425d7589f93e9f843/Sources/MongoSwift/MongoCollection.swift#L93) on `MongoCollection.drop`. key differences are the phrasing "if the future fails, the error is likely one of the following..." and the bulleted list of errors.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -222,4 +222,52 @@ public final class ClientSession {
...
+     * `options` override the default transaction options for this session and any options inherited from
```
kmahar: the mention of options precedence here made me wonder. what if I do the following:
- set no custom options on client
- for default txn options, set a custom read concern and write concern
- for options provided here, set a custom read concern, and do not specify a write concern in the options struct.

what is the write concern used for this txn? the custom one from the default options, or is it the default one since I didn't specify one in my options for this particular txn (yet did specify a read concern)?

I *think* consistency with our how we handle this inheritance in terms of DBs and collections inheriting from clients would mean that we should use the write concern from the default options: https://github.com/mongodb/mongo-swift-driver/blob/ec2bba2fa2779360a427695425d7589f93e9f843/Sources/MongoSwift/MongoClient.swift#L420

it may already work in that way but this comment made me realize we should probably verify that is the case.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -222,4 +222,52 @@ public final class ClientSession {
...
+        case .notStarted, .started:
+            let operation = StartTransactionOperation(options: options)
+            return self.client.operationExecutor.execute(operation, client: self.client, session: self)
+        case .ended:
+            return self.client.operationExecutor.makeFailedFuture(ClientSession.SessionInactiveError)
+        }
+    }
+
+    /**
+     * Commits a multi-document transaction for this session. Throws an error if the session has no in-progress
```
kmahar: similar comment applies here about formatting / error list

Sources/MongoSwift/ClientSession.swift
```diff
@@ -222,4 +222,52 @@ public final class ClientSession {
...
+        case .notStarted, .started:
+            let operation = CommitTransactionOperation()
+            return self.client.operationExecutor.execute(operation, client: self.client, session: self)
+        case .ended:
+            return self.client.operationExecutor.makeFailedFuture(ClientSession.SessionInactiveError)
+        }
+    }
+
+    /**
+     * Aborts a multi-document transaction for this session. Throws an error if the session has no in-progress
```
kmahar: similar comment applies here about formatting / error list

JamesHeppenstall: Done.

Sources/MongoSwift/ClientSession.swift
JamesHeppenstall: Done.

Sources/MongoSwift/ClientSession.swift
JamesHeppenstall: Done.

Sources/MongoSwift/ClientSession.swift
JamesHeppenstall: Yes, this is the case. The [transactions spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transactionoptions) explicitly state that "the writeConcern property is optional. The default value is NULL. If writeConcern is NULL the value will be inherited from this session’s defaultTransactionOptions. If defaultTransactionOptions itself is NULL or the writeConcern in defaultTransactionOptions is NULL, the writeConcern is inherited from the MongoClient associated with this session."

We set the default transactions options on the underlying `libmongoc` objects in `startTransaction`. I see that `MongoClient` does something similar. As such, `libmongoc` has all the necessary information to implement the logic described in the spec.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -222,4 +222,72 @@ public final class ClientSession {
...
+     * transaction is automatically aborted when `ClientSession.end()` is called.
+     *
+     * - Parameters:
+     *   - options: The options to use when starting this transaction
+     *
+     * - Returns:
+     *    An `EventLoopFuture<Void>` that succeeds when `startTransaction` is successful.
+     *
+     *    If the future fails, the error is likely one of the following:
+     *    - `CommandError` if an error occurs that prevents the command from executing.
```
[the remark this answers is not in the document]
JamesHeppenstall: I missed including the `LogicError` when there's already an in-progress transaction.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -222,4 +222,72 @@ public final class ClientSession {
             }
         }
     }
+
+    /**
```
patrickfreed: Each of the new methods in this field could benefit from a `SeeAlso` link to the official MongoDB documentation on transactions (https://docs.mongodb.com/manual/core/transactions). See [here](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/MongoCollection%2BChangeStreams.swift#L24) for an example of that.

JamesHeppenstall: Done.

PR #433, opened 2020-03-25
SWIFT-747 Implement synchronous transactions API

JamesHeppenstall:


Sources/MongoSwiftSync/ClientSession.swift
```diff
@@ -96,4 +96,45 @@ public final class ClientSession {
...
+     *
+     * - Parameters:
+     *   - options: The options to use when starting this transaction
+     *
+     * - Throws:
+     *   - `CommandError` if an error occurs that prevents the command from executing.
+     *   - `LogicError` if the session already has an in-progress transaction.
+     *   - `LogicError` if `startTransaction` is called on an ended session.
+     */
+    public func startTransaction(_ options: TransactionOptions?) throws {
```
kmahar: let's provide a default value of `nil` for the options here.

JamesHeppenstall: Ok, will do!

Sources/MongoSwiftSync/ClientSession.swift
```diff
@@ -96,4 +96,45 @@ public final class ClientSession {
...
+     *
+     * - Parameters:
+     *   - options: The options to use when starting this transaction
+     *
+     * - Throws:
+     *   - `CommandError` if an error occurs that prevents the command from executing.
+     *   - `LogicError` if the session already has an in-progress transaction.
+     *   - `LogicError` if `startTransaction` is called on an ended session.
+     */
+    public func startTransaction(_ options: TransactionOptions? = nil) throws {
```
patrickfreed: We typically don't omit the label for options parameters, so for consistency with elsewhere in the driver we probably shouldn't here either.

Sources/MongoSwiftSync/ClientSession.swift
```diff
@@ -96,4 +96,45 @@ public final class ClientSession {
     public func advanceOperationTime(to operationTime: Timestamp) {
         self.asyncSession.advanceOperationTime(to: operationTime)
     }
+
+    /**
```
patrickfreed: ditto comment in other PR about adding `SeeAlso` link to official docs.

JamesHeppenstall: Done.

Sources/MongoSwiftSync/ClientSession.swift
JamesHeppenstall: Done.

Sources/MongoSwiftSync/ClientSession.swift
patrickfreed: I think this change along with the underscore removal got reverted by mistake.

JamesHeppenstall: Thanks for catching that - will revert.

PR #434, opened 2020-03-25
SWIFT-750 Don't block in thread pool while waiting for a connection

kmahar:
Fixes the problem I found the other day where we can end up deadlocked if there are no connections free and all the threads in the thread pool are waiting for connections.

### DIFF
--- a/Sources/MongoSwift/Operations/Operation.swift
+++ b/Sources/MongoSwift/Operations/Operation.swift
@@ -11,6 +11,14 @@ internal protocol Operation {
     func execute(using connection: Connection, session: ClientSession?) throws -> OperationResult
 }
 
+/// Contains the result of attempting to execute an operation in the thread pool.
+private enum ExecuteResult<T> {
+    /// Indicates that the operation was successfully executed and returned the associated value.
+    case success(T)
+    /// Indicates that a connection was not available and the operation should be resubmitted to the pool.
+    case resubmit
+}
+
 /// Operation executor used by `MongoClient`s.
 internal class OperationExecutor {
     /// A group of event loops to use for running operations in the thread pool.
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -94,13 +132,6 @@ internal class OperationExecutor {
     }
 }
 
-/// Given a client and optionally a session associated which are to be associated with an operation, returns a
-/// connection for the operation to use. After the connection is no longer in use, it should be returned by
-/// passing it to `returnConnection` along with the same client and session that were passed into this method.
-internal func resolveConnection(client: MongoClient, session: ClientSession?) throws -> Connection {
```
kmahar: we only called this method in one place, and also its docstring was out of date. it seemed better to just move this logic into the place we called this, where it was already documented

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -33,4 +33,21 @@ final class MongoClientTests: MongoSwiftTestCase {
...
+        try self.withTestNamespace { _, _, coll in
+            let docs: [Document] = (1...10).map { ["x": .int32($0)] }
+            _ = try coll.insertMany(docs).wait()
+
+            let cursors = try (1...100).map { _ in try coll.find().wait() }
+
+            // queue up more operations
+            let waitingOperations = (1...MongoClient.defaultThreadPoolSize).map { _ in coll.countDocuments() }
+            // cursors can still make progress even though operations are waiting
+            _ = try cursors.map { try $0.toArray().wait() }
```
kmahar: I'm not sure how useful this test is because I don't think there's a way to guarantee that all of the `waitingOperations` have been submitted to the thread pool prior to all of these `toArray` calls getting submitted.

however, I can confirm that this test consistently deadlocks on master and never deadlocks on this branch (and that the same is true for the benchmark I originally discovered this with), so that seems like a decent indicator that the problem is fixed.

patrickfreed:
Looks like sourcery needs to be run, but other than that LGTM

PR #435, opened 2020-03-26
SWIFT-751 Add maxPoolSize option

kmahar:
@mbroadst looping you in since this is new public API.

I had mixed feelings on whether to put this option into a new options struct or just put them into `ClientOptions`, but I went with a new struct because:
- eventually we will have two more options for users to specify here, and this provides a nice way to group those together / separate them out like we did with TLS options
- CMAP seems to imply they belong in their own struct but maybe that's not required https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-options-1

let me know what you think.


Sources/MongoSwift/MongoClient.swift
```diff
@@ -4,6 +4,9 @@ import NIOConcurrencyHelpers
 
 /// Options to use when creating a `MongoClient`.
 public struct ClientOptions: CodingStrategyProvider, Decodable {
+    /// Specifies options this client should use when creating connection pools.
+    public var connectionPoolOptions: ConnectionPoolOptions?
```
mbroadst: I have a slight preference for keeping the `ClientOptions` a flat structure, with a rationale that our connection string doesn't have hierarchies. In node's distant history it supported nested options like this, and it became quite unwieldy and confusing to users and developers alike. I don't really have a sense if that's the case here, and I see there's already precedence for this with our TLS options, but I figured I'd bring it up anyway. 

Maybe this is where to start: do you anticipate that a user will create a struct and pass it in?
```
var cmapOptions = ConnectionPoolOptions(maxPoolSize: 10)
var clientOptions = ClientOptions(conectionPoolOptions: cmapOptions, readConcern: .secondary)
var client = MongoClient(clientOptions)
```
That feels a little more cumbersome to me, but maybe this is a moot point because we expect people to mostly use the connection string.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -162,6 +167,9 @@ public class MongoClient {
     /// Default size for a client's NIOThreadPool.
     public static let defaultThreadPoolSize = 5
 
+    /// Default maximum size for connection pools created by this client.
+    public static let defaultMaxConnectionPoolSize = 100
```
mbroadst: while we're here, is it useful for this to be a public value? Do we expect users to do something with it programmatically?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -221,7 +229,11 @@ public class MongoClient {
         initializeMongoc()
 
         let connString = try ConnectionString(connectionString, options: options)
-        self.connectionPool = try ConnectionPool(from: connString, options: options?.tlsOptions)
+        self.connectionPool = try ConnectionPool(
+            from: connString,
+            poolOptions: options?.connectionPoolOptions,
```
mbroadst: Related to above: if you move everything directly into `ClientOptions` you can more easily pass that single unit around, rather than needing to split up parts like this.

Sources/MongoSwift/MongoClient.swift
kmahar: matching the structure of the connection string makes sense to me. I don't really recall why we went with the separate struct approach for TLS. possibly a leakage of internals since it made it easier to pass just the TLS options to the pool rather than all of the options.

I'm fine with switching to a flat structure for everything, will make the update

Sources/MongoSwift/MongoClient.swift
kmahar: I think I copied this (and doing the same for the thread pool size) from `NonBlockingFileIO`: https://github.com/apple/swift-nio/blob/6cac95501b52884d6ac9bf63f9029d983b179512/Sources/NIO/NonBlockingFileIO.swift#L32
but thinking about it more in their case they require the user to manage the thread pool so it makes more sense to have it programmatically available. I don't really know what people would do with this in our situation though so I'm fine with removing both of the constants

kmahar:
I've condensed everything into a single client options struct. as part of that I renamed all of the TLS options to match what they are named in the URI options.

mbroadst:
@kmahar looks great! This does make me wonder if we should drag [SWIFT_641](https://jira.mongodb.org/browse/SWIFT-641) and [SWIFT-573](https://jira.mongodb.org/browse/SWIFT-573) into consideration for 1.0

kmahar:
I'm not opposed, it seems like both should be pretty straightforward.

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -45,6 +45,15 @@ internal class ConnectionPool {
             throw InternalError(message: "Could not configure error handling on client pool")
         }
 
+        if let maxPoolSize = options?.maxPoolSize {
+            guard maxPoolSize > 0 && maxPoolSize <= UInt32.max else {
+                throw InvalidArgumentError(
```
patrickfreed: We could just specify the option in `ClientOptions` as a `UInt32` to avoid this check.

Sources/MongoSwiftSync/Exports.swift
```diff
@@ -97,7 +97,6 @@
 @_exported import struct MongoSwift.ServerOpeningEvent
 @_exported import struct MongoSwift.ServerSelectionError
 @_exported import struct MongoSwift.Symbol
-@_exported import struct MongoSwift.TLSOptions
```
patrickfreed: for the sake of the changelog, maybe we should file a ticket that tracks this change and just close it when this gets merged in.

kmahar: good point. I opened [SWIFT-753](https://jira.mongodb.org/browse/SWIFT-753)

Sources/MongoSwift/ConnectionPool.swift
kmahar: I considered this. however, I think the community seems to have converged on still using `Int` in cases like this. Consider that, for example, `NIOThreadPool.init` and `MultiThreadedEventLoopGroup` initializers both take in `Int`s even though those values definitely need to be positive.
We do currently have the additional restriction of max pool size being capped at `UInt32.max`, but that's a constraint introduced by libmongoc. It doesn't seem like we would bother to cap the max allowed pool size at that value in a pure Swift implementation.

Also from the [Swift language guide](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html) (see UInt section):
> Use `UInt` only when you specifically need an unsigned integer type with the same size as the platform’s native word size. If this isn’t the case, `Int` is preferred, even when the values to be stored are known to be nonnegative. A consistent use of `Int` for integer values aids code interoperability, avoids the need to convert between different number types, and matches integer type inference, as described in Type Safety and Type Inference.

That said, if everyone is on board with adopting this guidance we probably need to do an audit of our API to correct existing violations.

patrickfreed: Oh cool, I didn't know about that section of the guide, good find. And yeah, we probably should adopt that guidance, since others likely are (e.g. NIO is).

PR #437, opened 2020-03-31
SWIFT-618, SWIFT-623: Drop Swift 5.0 support

kmahar:
This PR drops Swift 5.0 support and adds testing against Swift 5.2 on Evergreen.

Most of this diff was generated via `swiftlint autocorrect` and search and replace:
- I turned on swiftlint's implicit return rule everywhere so that it also applies to functions and getters now
- I did a search and replace for `type(of: self)` -> `Self`


Sources/MongoSwift/Operations/ListCollectionsOperation.swift
```diff
@@ -112,7 +112,7 @@ internal struct ListCollectionsOperation: Operation {
             opts["filter"] = .document(filterDoc)
 
             // If `listCollectionNames` is called with a non-name filter key, change server-side nameOnly flag to false.
-            if self.nameOnly && filterDoc.keys.contains { $0 != "name" } {
+            if self.nameOnly && filterDoc.keys.contains(where: { $0 != "name" }) {
```
kmahar: this changed due to an interesting new compiler warning:
```
warning: trailing closure in this context is confusable with the body of the statement; pass as a parenthesized argument to silence this warning
            if self.nameOnly && filterDoc.keys.contains { $0 != "name" } {
                                                       ~^
                                                       (where:          )
```

Sources/MongoSwift/BSON/CodableNumber.swift
```diff
@@ -140,11 +140,6 @@ extension UInt: CodableNumber {
 extension Double: CodableNumber {
     internal init?(from value: BSONValue) {
         switch value {
-        case let v as Int:
```
kmahar: the compiler now points out that a cast from `BSONValue` to `Int` always fails since `Int` no longer conforms to `BSONValue`.

kmahar:
evergreen patch: https://evergreen.mongodb.com/version/5e83a6aad6d80a23e2d3eaed

kmahar:
agreed on SWIFT-621, I'll schedule it

PR #438, opened 2020-04-01
SWIFT-434 Document how to use the driver with TLS/SSL

patrickfreed:
[SWIFT-434](https://jira.mongodb.org/browse/SWIFT-434)

As the ticket describes, this PR adds documentation for describing how to ensure clients are using the latest TLS protocols. As part of that,  figured it made sense to write the rest of the guide, since we didn't already have one. The guide was heavily influenced by [pymongo's guide](https://api.mongodb.com/python/current/examples/tls.html) on the same topic.

One API change that's important to note: I added a generic `tls` option to `ClientOptions`, since it seems reasonable to me that a user would want to enable TLS but not necessarily identify all the client's connecting to the database. This option is included in the URI options spec, and I think most drivers support it (pymongo does at least). cc @mbroadst

### DIFF
--- a/Guides/TLS.md
+++ b/Guides/TLS.md
@@ -32,7 +32,7 @@ To require that connections to MongoDB made by the driver use TLS/SSL, simply sp
 let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tls: true))
 ```
 
-Alternatively, `tls=true` can be specified in the [MongoDB Connection String]() passed to the initializer:
+Alternatively, `tls=true` can be specified in the [MongoDB Connection String](https://docs.mongodb.com/manual/reference/connection-string/) passed to the initializer:
 ```swift
 let client = try MongoClient("mongodb://example.com/?tls=true", using: elg)
 ```
@@ -47,7 +47,7 @@ A path to a file with either a single or bundle of certificate authorities to be
 let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tlsCAFile: URL(string: "/path/to/ca.pem")))
 ```
 
-Alternatively, the path can be specified via the `tlsCAFile` option in the [MongoDB Connection String]() passed to the client's initializer:
+Alternatively, the path can be specified via the `tlsCAFile` option in the [MongoDB Connection String](https://docs.mongodb.com/manual/reference/connection-string/) passed to the client's initializer:
 ```swift
 let client = try MongoClient("mongodb://example.com/?tlsCAFile=/path/to/ca.pem", using: elg)
 ```
@@ -67,7 +67,7 @@ let client = try MongoClient(
 )
 ```
 
-Alternatively, these options can be set via the `tlsCertificateKeyFile` and `tlsCertificateKeyFilePassword` options in the [MongoDB Connection String]() passed into the initializer:
+Alternatively, these options can be set via the `tlsCertificateKeyFile` and `tlsCertificateKeyFilePassword` options in the [MongoDB Connection String](https://docs.mongodb.com/manual/reference/connection-string/) passed into the initializer:
 ```swift
 let client = try MongoClient(
     "mongodb://example.com/?tlsCertificateKeyFile=/path/to/cert.pem&tlsCertificateKeyFilePassword=<password>"
--- a/Guides/TLS.md
+++ b/Guides/TLS.md
@@ -27,7 +27,7 @@ macOS 10.13 (High Sierra) and newer support TLS 1.1+.
 
 ## Basic Configuration
 
-To connect to MongoDB over TLS/SSL, simply specify `tls: true` in the `ClientOptions` passed to a `MongoClient`'s initializer:
+To require that connections to MongoDB made by the driver use TLS/SSL, simply specify `tls: true` in the `ClientOptions` passed to a `MongoClient`'s initializer:
 ```swift
 let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tls: true))
 ```
--- a/Guides/TLS.md
+++ b/Guides/TLS.md
@@ -49,7 +49,7 @@ let client = try MongoClient("mongodb://example.com", using: elg, options: Clien
 
 Alternatively, the path can be specified via the `tlsCAFile` option in the [MongoDB Connection String](https://docs.mongodb.com/manual/reference/connection-string/) passed to the client's initializer:
 ```swift
-let caFile = "/path/to/ca.pem".addingPercentEncoding(withAllowedCharacters: .urlFragmentAllowed)!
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -39,6 +39,11 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
      */
     public var threadPoolSize: Int?
 
+    /// Whether or not to require TLS for connections to the server
```
kmahar, line 42: nits: 1) maybe helpful to state default is to not require TLS? 2) end sentence with period

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
+# TLS/SSL and MongoSwift
+
+MongoSwift supports connecting to MongoDB over TLS/SSL. This guide covers the installation requirements and configuration options for enabling this functionality in MongoSwift. See the [server documentation](https://docs.mongodb.com/manual/tutorial/configure-ssl/) to configure MongoDB to use TLS/SSL.
```
kmahar, line 3: I think we have to be a little careful about using the name `MongoSwift` now that we have two modules and the library isn't named `MongoSwift`. It might lead you to think the sync API doesn't support TLS. I think it's probably ok to just say "the driver".

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+If the version number is less than 1.0.1, support for TLS 1.1 or newer is not available. Contact your operating system vendor for a solution, upgrade to a newer distribution, or manually upgrade your installation of OpenSSL.
+
+#### ...on macOS
+
+macOS 10.13 (High Sierra) and newer support TLS 1.1+.
+
+
+## Basic Configuration
+
+To connect to MongoDB over TLS/SSL, simply specify `tls: true` in the `ClientOptions` passed to a `MongoClient`'s initializer:
```
kmahar, line 30: maybe state explicitly that this will lead to *requiring* TLS for connections?

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+```swift
+let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tls: true))
+```
+
+Alternatively, `tls=true` can be specified in the [MongoDB Connection String]() passed to the initializer:
```
kmahar, line 35: missing URL

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+```swift
+let client = try MongoClient("mongodb://example.com/?tls=true", using: elg)
+```
+**Note:** Specifying any `tls` prefixed option in the connection string or `ClientOptions` will cause MongoSwift to attempt to connect using TLS.
+
+## Specifying a CA File
+
+MongoSwift can be configured to use a specific set of CA certificates. This is most often used with “self-signed” server certificates.
```
kmahar, line 43: weird quotation marks here (`“”` vs `""`)

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+## Specifying a CA File
+
+MongoSwift can be configured to use a specific set of CA certificates. This is most often used with “self-signed” server certificates. 
+
+A path to a file with either a single or bundle of certificate authorities to be considered trusted when making a TLS connection can be specified via the `tlsCAFile` option on `ClientOptions`:
+```swift
+let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tlsCAFile: URL(string: "/path/to/ca.pem")))
+```
+
+Alternatively, the path can be specified via the `tlsCAFile` option in the [MongoDB Connection String]() passed to the client's initializer:
```
kmahar, line 50: link missing again

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+If the private key is password protected, a password can be supplied via `tlsCertificateKeyFilePassword` on `ClientOptions`:
+```swift
+let client = try MongoClient(
+    "mongodb://example.com",
+    using: elg,
+    options: ClientOptions(tlsCertificateKeyFile: URL(string: "/path/to/cert.pem"), tlsCertificateKeyFilePassword: <password>)
+)
+```
+
+Alternatively, these options can be set via the `tlsCertificateKeyFile` and `tlsCertificateKeyFilePassword` options in the [MongoDB Connection String]() passed into the initializer:
```
kmahar, line 70: and link one more time :)

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+To connect to MongoDB over TLS/SSL, simply specify `tls: true` in the `ClientOptions` passed to a `MongoClient`'s initializer:
+```swift
+let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tls: true))
+```
+
+Alternatively, `tls=true` can be specified in the [MongoDB Connection String]() passed to the initializer:
+```swift
+let client = try MongoClient("mongodb://example.com/?tls=true", using: elg)
+```
+**Note:** Specifying any `tls` prefixed option in the connection string or `ClientOptions` will cause MongoSwift to attempt to connect using TLS.
```
kmahar, line 39: nit: `tls`-prefixed, also this could maybe be changed to use the "required" wording as well

Guides/TLS.md
```diff
@@ -0,0 +1,78 @@
...
+MongoSwift can be configured to use a specific set of CA certificates. This is most often used with “self-signed” server certificates. 
+
+A path to a file with either a single or bundle of certificate authorities to be considered trusted when making a TLS connection can be specified via the `tlsCAFile` option on `ClientOptions`:
+```swift
+let client = try MongoClient("mongodb://example.com", using: elg, options: ClientOptions(tlsCAFile: URL(string: "/path/to/ca.pem")))
+```
+
+Alternatively, the path can be specified via the `tlsCAFile` option in the [MongoDB Connection String]() passed to the client's initializer:
+```swift
+let client = try MongoClient("mongodb://example.com/?tlsCAFile=/path/to/ca.pem", using: elg)
```
mbroadst, line 52: did you try using this connection string? I would imagine you need to urlencode the path at very least

Guides/TLS.md
patrickfreed: fixed

Sources/MongoSwift/MongoClient.swift
patrickfreed: done and done

Guides/TLS.md
```diff
@@ -0,0 +1,81 @@
+# TLS/SSL and MongoSwift
```
kmahar, line 1: whoops I missed that `MongoSwift` was used in the title too :)

patrickfreed: updated to "Swift Driver TLS/SSL" guide. Also filed a ticket for updating the various other guides as well

PR #439, opened 2020-04-01
SWIFT-621 Provide default values for WriteModel options associated values

patrickfreed:
[SWIFT-621](https://jira.mongodb.org/browse/SWIFT-621)

I looked through the [rest of our enums](https://mongodb.github.io/mongo-swift-driver/MongoSwift/Enums.html) (ty jazzy) and didn't see anything else that needed this treatment besides `WriteModel`.


kmahar:
lgtm, though it just occurred to me we can probably remove `options: nil` from a lot of places in the bulk write tests now?

patrickfreed:
good point, done

PR #440, opened 2020-04-02
SWIFT-595: update libmongoc to 1.16.x

mbroadst:
This required slight modification to the sed scripts used during the vendoring process, mostly to remove prefixes for bson or mongoc headers.

Highlights:
- `libmongoc` uses the [PRId64](https://en.cppreference.com/w/cpp/types/integer) macro for printing `int64_t` types, and Swift's compiler [can't handle that](https://bugs.swift.org/browse/SR-6378). Instead I've opted to replace the references with an `%lld`, which will work but will also now warn because the variable isn't explicity cast to `long long`. Someone could fix this in the future by making the replacement a little smarter and casting.

- the build is currently broken, I imagine because the auth tests need to be updated for changes fixed in this version of `libmongoc`

- for now I've completely disabled CSFLE, we would need to bundle libmongocrypt as well which imo is more pain than it's worth at the moment.


kmahar:
Yes, the tests are failing now because of [SWIFT-636](https://jira.mongodb.org/browse/SWIFT-636). I believe you just need to delete a chunk of code I marked with a TODO SWIFT-636 in `Credential.swift` and it will fix the new failures.

kmahar:
> Someone could fix this in the future by making the replacement a little smarter and casting. 

should we open a ticket about this?

mbroadst:
Thanks for the tip on SWIFT-636 @kmahar :sparkles: Looks like it'll pass now, except on Travis because of an arcane openssl version over there... I propose we dump Travis now that EVG is set up for PR testing. 

Regarding the compiler warnings, I created [SWIFT-758](https://jira.mongodb.org/browse/SWIFT-758) to track the work. I think it's low priority.

mbroadst:
@patrickfreed I think you're wrong about that, the Travis ubuntu image is xenial (16.04) and our evergreen build is using bionic (18.04) - we're trying to use a feature that's too new for travis' version of ubuntu (also a version unlikely to be used for new swift on server apps IMO). Having said that, I think I found the bug and am trying it out now.

patrickfreed:
My previous comment was incorrect, the one travis is using is too old for us.
That said, it looks like we'll be able to support older versions of OpenSSL by doing something like:
```c
#if OPENSSL_VERSION_NUMBER < 0x10100000
# define MONGOC_HAVE_ASN1_STRING_GET0_DATA 0
#endif
```
in the generated headers. 

I think I'm also fine with us not supporting older OpenSSL as well, but If that's the case we should officially state that and I can add it to the TLS guide as a requirement.

PR #443, opened 2020-04-06
SWIFT-752 Implement transactions spec test runner

JamesHeppenstall:
I've implemented everything except the sharding-specific tests (mongos-pin-auto.json, mongos-recovery-token.json, and pin-mongos.json), which I will do in another PR. The retryable reads and transactions tests now run in a unified test runner.

### DIFF
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -258,7 +258,7 @@ extension Document {
 
     /// Helper function for copying elements from some source document to a destination document while
     /// excluding a non-zero number of keys
-    internal func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
+    public func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
         guard !keys.isEmpty else {
             throw InternalError(message: "No keys to exclude, use 'bson_copy' instead")
         }
--- a/Sources/MongoSwift/BSON/Document.swift
+++ b/Sources/MongoSwift/BSON/Document.swift
@@ -258,7 +258,7 @@ extension Document {
 
     /// Helper function for copying elements from some source document to a destination document while
     /// excluding a non-zero number of keys
-    public func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
+    internal func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
         guard !keys.isEmpty else {
             throw InternalError(message: "No keys to exclude, use 'bson_copy' instead")
         }
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -71,12 +71,12 @@ public final class ClientSession {
     internal var id: Document?
 
     /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
-    internal var serverId: Int? {
+    internal var serverId: UInt32? {
         switch self.state {
         case .notStarted, .ended:
             return nil
         case let .started(session, _):
-            return Int(mongoc_client_session_get_server_id(session))
+            return UInt32(mongoc_client_session_get_server_id(session))
         }
     }
 
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -66,7 +66,8 @@ public final class ClientSession {
     /// The client used to start this session.
     public let client: MongoClient
 
-    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
+    /// The session ID of this session. This is internal for now because we only have a value available after we've
+    /// started the libmongoc session.
     internal var id: Document?
 
     /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -67,10 +67,10 @@ public final class ClientSession {
     public let client: MongoClient
 
     /// The session ID of this session. We only have a value available after we've started the libmongoc session.
-    public var id: Document?
+    internal var id: Document?
 
     /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
-    public var serverId: Int? {
+    internal var serverId: Int? {
         switch self.state {
         case .notStarted, .ended:
             return nil
@@ -80,7 +80,7 @@ public final class ClientSession {
     }
 
     /// Enum tracking the state of the transaction associated with this session.
-    public enum TransactionState: String, Decodable {
+    internal enum TransactionState: String, Decodable {
         /// There is no transaction in progress.
         case none
         /// A transaction has been started, but no operation has been sent to the server.
@@ -126,7 +126,7 @@ public final class ClientSession {
     }
 
     /// The transaction state of this session.
-    public var transactionState: TransactionState? {
+    internal var transactionState: TransactionState? {
         switch self.state {
         case .notStarted, .ended:
             return nil
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -66,9 +66,74 @@ public final class ClientSession {
     /// The client used to start this session.
     public let client: MongoClient
 
-    /// The session ID of this session. This is internal for now because we only have a value available after we've
-    /// started the libmongoc session.
-    internal var id: Document?
+    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
+    public var id: Document?
+
+    /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
+    public var serverId: Int? {
+        switch self.state {
+        case .notStarted, .ended:
+            return nil
+        case let .started(session, _):
+            return Int(mongoc_client_session_get_server_id(session))
+        }
+    }
+
+    /// Enum tracking the state of the transaction associated with this session.
+    public enum TransactionState: String, Decodable {
+        /// There is no transaction in progress.
+        case none
+        /// A transaction has been started, but no operation has been sent to the server.
+        case starting
+        /// A transaction is in progress.
+        case inProgress
+        /// The transaction was committed.
+        case committed
+        /// The transaction was aborted.
+        case aborted
+
+        fileprivate var mongocTransactionState: mongoc_transaction_state_t {
+            switch self {
+            case .none:
+                return MONGOC_TRANSACTION_NONE
+            case .starting:
+                return MONGOC_TRANSACTION_STARTING
+            case .inProgress:
+                return MONGOC_TRANSACTION_IN_PROGRESS
+            case .committed:
+                return MONGOC_TRANSACTION_COMMITTED
+            case .aborted:
+                return MONGOC_TRANSACTION_ABORTED
+            }
+        }
+
+        fileprivate init(mongocTransactionState: mongoc_transaction_state_t) {
+            switch mongocTransactionState {
+            case MONGOC_TRANSACTION_NONE:
+                self = .none
+            case MONGOC_TRANSACTION_STARTING:
+                self = .starting
+            case MONGOC_TRANSACTION_IN_PROGRESS:
+                self = .inProgress
+            case MONGOC_TRANSACTION_COMMITTED:
+                self = .committed
+            case MONGOC_TRANSACTION_ABORTED:
+                self = .aborted
+            default:
+                fatalError("Unexpected transaction state: \(mongocTransactionState)")
+            }
+        }
+    }
+
+    /// The transaction state of this session.
+    public var transactionState: TransactionState? {
+        switch self.state {
+        case .notStarted, .ended:
+            return nil
+        case let .started(session, _):
+            return TransactionState(mongocTransactionState: mongoc_client_session_get_transaction_state(session))
+        }
+    }
 
     /// The most recent cluster time seen by this session. This value will be nil if either of the following are true:
     /// - No operations have been executed using this session and `advanceClusterTime` has not been called.
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -76,7 +76,7 @@ public final class ClientSession {
         case .notStarted, .ended:
             return nil
         case let .started(session, _):
-            return UInt32(mongoc_client_session_get_server_id(session))
+            return mongoc_client_session_get_server_id(session)
         }
     }
 
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -136,6 +136,14 @@ public final class ClientSession {
         }
     }
 
+    /// Indicates whether or not the session is in a transaction.
+    internal var inTransaction: Bool {
+        if let transactionState = self.transactionState {
+            return transactionState != .none
+        }
+        return false
+    }
+
     /// The most recent cluster time seen by this session. This value will be nil if either of the following are true:
     /// - No operations have been executed using this session and `advanceClusterTime` has not been called.
     /// - This session has been ended.
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -308,7 +308,7 @@ public final class ClientSession {
      * - SeeAlso:
      *   - https://docs.mongodb.com/manual/core/transactions/
      */
-    public func startTransaction(_ options: TransactionOptions?) -> EventLoopFuture<Void> {
+    public func startTransaction(options: TransactionOptions? = nil) -> EventLoopFuture<Void> {
         switch self.state {
         case .notStarted, .started:
             let operation = StartTransactionOperation(options: options)
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -3,16 +3,16 @@ import NIO
 import NIOConcurrencyHelpers
 
 /// Options to use when creating a `MongoClient`.
-public struct ClientOptions: CodingStrategyProvider, Decodable {
+public struct ClientOptions: CodingStrategyProvider {
     // swiftlint:disable redundant_optional_initialization
 
     /// Specifies the `DataCodingStrategy` to use for BSON encoding/decoding operations performed by this client and any
     /// databases or collections that derive from it.
-    public var dataCodingStrategy: DataCodingStrategy? = nil
+    public var dataCodingStrategy: DataCodingStrategy?
 
     /// Specifies the `DateCodingStrategy` to use for BSON encoding/decoding operations performed by this client and any
     /// databases or collections that derive from it.
-    public var dateCodingStrategy: DateCodingStrategy? = nil
+    public var dateCodingStrategy: DateCodingStrategy?
 
     /// The maximum number of connections that may be associated with a connection pool created by this client at a
     /// given time. This includes in-use and available connections. Defaults to 100.
@@ -22,7 +22,7 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     public var readConcern: ReadConcern?
 
     /// Specifies a ReadPreference to use for the client.
-    public var readPreference: ReadPreference? = nil
+    public var readPreference: ReadPreference?
 
     /// Determines whether the client should retry supported read operations (on by default).
     public var retryReads: Bool?
@@ -65,7 +65,7 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
 
     /// Specifies the `UUIDCodingStrategy` to use for BSON encoding/decoding operations performed by this client and any
     /// databases or collections that derive from it.
-    public var uuidCodingStrategy: UUIDCodingStrategy? = nil
+    public var uuidCodingStrategy: UUIDCodingStrategy?
 
     // swiftlint:enable redundant_optional_initialization
 
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -72,8 +72,19 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     /// Specifies a WriteConcern to use for the client.
     public var writeConcern: WriteConcern?
 
-    private enum CodingKeys: CodingKey {
-        case retryWrites, retryReads, readConcern, writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case retryWrites, retryReads, readConcern, writeConcern, w, readConcernLevel, mode = "readPreference"
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.readConcern = (try? container.decode(ReadConcern.self, forKey: .readConcern)) ??
+            (try? ReadConcern(container.decode(String.self, forKey: .readConcernLevel)))
+        self.readPreference = try? ReadPreference(container.decode(ReadPreference.Mode.self, forKey: .mode))
+        self.retryReads = try? container.decode(Bool.self, forKey: .retryReads)
+        self.retryWrites = try? container.decode(Bool.self, forKey: .retryWrites)
+        self.writeConcern = (try? container.decode(WriteConcern.self, forKey: .writeConcern)) ??
+            (try? WriteConcern(w: container.decode(WriteConcern.W.self, forKey: .w)))
     }
 
     /// Convenience initializer allowing any/all parameters to be omitted or optional.
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -72,19 +72,8 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     /// Specifies a WriteConcern to use for the client.
     public var writeConcern: WriteConcern?
 
-    private enum CodingKeys: String, CodingKey {
-        case retryWrites, retryReads, readConcern, writeConcern, w, readConcernLevel, mode = "readPreference"
-    }
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        self.readConcern = (try? container.decode(ReadConcern.self, forKey: .readConcern)) ??
-            (try? ReadConcern(container.decode(String.self, forKey: .readConcernLevel)))
-        self.readPreference = try? ReadPreference(container.decode(ReadPreference.Mode.self, forKey: .mode))
-        self.retryReads = try? container.decode(Bool.self, forKey: .retryReads)
-        self.retryWrites = try? container.decode(Bool.self, forKey: .retryWrites)
-        self.writeConcern = (try? container.decode(WriteConcern.self, forKey: .writeConcern)) ??
-            (try? WriteConcern(w: container.decode(WriteConcern.W.self, forKey: .w)))
+    private enum CodingKeys: CodingKey {
+        case retryWrites, retryReads, readConcern, writeConcern
     }
 
     /// Convenience initializer allowing any/all parameters to be omitted or optional.
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -4,8 +4,6 @@ import NIOConcurrencyHelpers
 
 /// Options to use when creating a `MongoClient`.
 public struct ClientOptions: CodingStrategyProvider {
-    // swiftlint:disable redundant_optional_initialization
-
     /// Specifies the `DataCodingStrategy` to use for BSON encoding/decoding operations performed by this client and any
     /// databases or collections that derive from it.
     public var dataCodingStrategy: DataCodingStrategy?
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -197,6 +197,14 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
         let opts = try encodeOptions(options: options, session: session)
         var insertedIds: [Int: BSON] = [:]
 
+        if let transactionState = session?.transactionState, transactionState != .none,
+            self.options?.writeConcern != nil {
+            throw InvalidArgumentError(
+                message: "Cannot specify a write concern on an individual helper in a " +
+                    "transaction. Instead specify it when starting the transaction."
+            )
+        }
+
         let (serverId, isAcknowledged): (UInt32, Bool) =
             try self.collection.withMongocCollection(from: connection) { collPtr in
                 guard let bulk = mongoc_collection_create_bulk_operation_with_opts(collPtr, opts?._bson) else {
@@ -216,14 +224,13 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
 
                 var writeConcernAcknowledged: Bool
                 if let transactionState = session?.transactionState, transactionState != .none {
-                    // Bulk write operations cannot have a write concern in a transaction. Default to
-                    // writeConcernAcknowledged = true. Since `libmongoc` returns a null write concern from bulk write
-                    // operations in a transaction, we cannot call `mongoc_bulk_operation_get_write_concern`.
-                    if self.options?.writeConcern != nil {
-                        throw InvalidArgumentError(
-                            message: "Bulk write operations cannot have a write concern in atransaction"
-                        )
-                    }
+                    // Bulk write operations in transactions must get their write concern from the session, not from
+                    // the `BulkWriteOptions` passed to the `bulkWrite` helper. `libmongoc` surfaces this
+                    // implementation detail by nulling out the write concern stored on the bulk write. To sidestep
+                    // this, we can only call `mongoc_bulk_operation_get_write_concern` out of a transaction.
+                    //
+                    // In a transaction, default to writeConcernAcknowledged = true. This is acceptable because
+                    // transactions do not support unacknowledged writes.
                     writeConcernAcknowledged = true
                 } else {
                     let writeConcern = WriteConcern(from: mongoc_bulk_operation_get_write_concern(bulk))
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -197,8 +197,7 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
         let opts = try encodeOptions(options: options, session: session)
         var insertedIds: [Int: BSON] = [:]
 
-        if let transactionState = session?.transactionState, transactionState != .none,
-            self.options?.writeConcern != nil {
+        if session?.inTransaction == true && self.options?.writeConcern != nil {
             throw InvalidArgumentError(
                 message: "Cannot specify a write concern on an individual helper in a " +
                     "transaction. Instead specify it when starting the transaction."
@@ -223,7 +222,7 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
                 }
 
                 var writeConcernAcknowledged: Bool
-                if let transactionState = session?.transactionState, transactionState != .none {
+                if session?.inTransaction == true {
                     // Bulk write operations in transactions must get their write concern from the session, not from
                     // the `BulkWriteOptions` passed to the `bulkWrite` helper. `libmongoc` surfaces this
                     // implementation detail by nulling out the write concern stored on the bulk write. To sidestep
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -218,6 +218,9 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
                 if let transactionState = session?.transactionState, transactionState != .none {
                     // Bulk write operations cannot have a write concern in a transaction. Default to
                     // writeConcernAcknowledged = true.
+                    if self.options?.writeConcern != nil {
+                        throw LogicError(message: "Bulk write operations cannot have a write concern in a transaction")
+                    }
                     writeConcernAcknowledged = true
                 } else {
                     let writeConcern = WriteConcern(from: mongoc_bulk_operation_get_write_concern(bulk))
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -214,8 +214,17 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
                     mongoc_bulk_operation_execute(bulk, replyPtr, &error)
                 }
 
-                let writeConcern = WriteConcern(from: mongoc_bulk_operation_get_write_concern(bulk))
-                return (serverId, writeConcern.isAcknowledged)
+                var writeConcernAcknowledged: Bool
+                if let transactionState = session?.transactionState, transactionState != .none {
+                    // Bulk write operations cannot have a write concern in a transaction. Default to
+                    // writeConcernAcknowledged = true.
+                    writeConcernAcknowledged = true
+                } else {
+                    let writeConcern = WriteConcern(from: mongoc_bulk_operation_get_write_concern(bulk))
+                    writeConcernAcknowledged = writeConcern.isAcknowledged
+                }
+
+                return (serverId, writeConcernAcknowledged)
             }
 
         let result = try BulkWriteResult(reply: reply, insertedIds: insertedIds)
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -217,9 +217,12 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
                 var writeConcernAcknowledged: Bool
                 if let transactionState = session?.transactionState, transactionState != .none {
                     // Bulk write operations cannot have a write concern in a transaction. Default to
-                    // writeConcernAcknowledged = true.
+                    // writeConcernAcknowledged = true. Since `libmongoc` returns a null write concern from bulk write
+                    // operations in a transaction, we cannot call `mongoc_bulk_operation_get_write_concern`.
                     if self.options?.writeConcern != nil {
-                        throw LogicError(message: "Bulk write operations cannot have a write concern in a transaction")
+                        throw InvalidArgumentError(
+                            message: "Bulk write operations cannot have a write concern in atransaction"
+                        )
                     }
                     writeConcernAcknowledged = true
                 } else {
--- a/Sources/MongoSwift/MongoError.swift
+++ b/Sources/MongoSwift/MongoError.swift
@@ -165,7 +165,7 @@ public struct WriteConcernFailure: Codable {
     public let code: ServerErrorCode
 
     /// A human-readable string identifying write concern error.
-    public let codeName: String?
+    public let codeName: String
 
     /// A document identifying the write concern setting related to the error.
     public let details: Document?
@@ -174,7 +174,7 @@ public struct WriteConcernFailure: Codable {
     public let message: String
 
     /// Labels that may describe the context in which this error was thrown.
-    public let errorLabels: [String]? = nil
+    public let errorLabels: [String]?
 
     private enum CodingKeys: String, CodingKey {
         case code
@@ -182,6 +182,27 @@ public struct WriteConcernFailure: Codable {
         case details = "errInfo"
         case message = "errmsg"
         case errorLabels
+    }
+
+    // TODO: can remove this once SERVER-36755 is resolved
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.code = try container.decode(ServerErrorCode.self, forKey: .code)
+        self.message = try container.decode(String.self, forKey: .message)
+        self.codeName = try container.decodeIfPresent(String.self, forKey: .codeName) ?? ""
+        self.details = try container.decodeIfPresent(Document.self, forKey: .details)
+        self.errorLabels = try container.decodeIfPresent([String].self, forKey: .errorLabels)
+    }
+
+    // TODO: can remove this once SERVER-36755 is resolved
+    internal init(
+        code: ServerErrorCode, codeName: String, details: Document?, message: String, errorLabels: [String]? = nil
+    ) {
+        self.code = code
+        self.codeName = codeName
+        self.message = message
+        self.details = details
+        self.errorLabels = errorLabels
     }
 }
 
--- a/Sources/MongoSwift/MongoError.swift
+++ b/Sources/MongoSwift/MongoError.swift
@@ -96,7 +96,7 @@ public struct InternalError: RuntimeError {
 /// An error thrown when encountering a connection or socket related error.
 /// May contain labels providing additional information on the nature of the error.
 public struct ConnectionError: RuntimeError, LabeledError {
-    internal let message: String
+    public let message: String
 
     public let errorLabels: [String]?
 
@@ -250,6 +250,18 @@ private func parseMongocError(_ error: bson_error_t, reply: Document?) -> MongoE
             codeName: codeName,
             message: message,
             errorLabels: errorLabels
+        )
+    case (MONGOC_ERROR_WRITE_CONCERN, _):
+        var writeConcernErrorLabels =
+            reply?["writeConcernError"]?.documentValue?["errorLabels"]?.arrayValue?.asArrayOf(String.self)
+        if let errorLabels = errorLabels {
+            writeConcernErrorLabels = Array(Set((writeConcernErrorLabels ?? []) + errorLabels))
+        }
+        return CommandError(
+            code: ServerErrorCode(code.rawValue),
+            codeName: codeName,
+            message: message,
+            errorLabels: writeConcernErrorLabels
         )
     case (MONGOC_ERROR_STREAM, _):
         return ConnectionError(message: message, errorLabels: errorLabels)
--- a/Sources/MongoSwift/MongoError.swift
+++ b/Sources/MongoSwift/MongoError.swift
@@ -165,7 +165,7 @@ public struct WriteConcernFailure: Codable {
     public let code: ServerErrorCode
 
     /// A human-readable string identifying write concern error.
-    public let codeName: String
+    public let codeName: String?
 
     /// A document identifying the write concern setting related to the error.
     public let details: Document?
@@ -173,11 +173,15 @@ public struct WriteConcernFailure: Codable {
     /// A description of the error.
     public let message: String
 
+    /// Labels that may describe the context in which this error was thrown.
+    public let errorLabels: [String]? = nil
+
     private enum CodingKeys: String, CodingKey {
         case code
         case codeName
         case details = "errInfo"
         case message = "errmsg"
+        case errorLabels
     }
 }
 
@@ -251,18 +255,6 @@ private func parseMongocError(_ error: bson_error_t, reply: Document?) -> MongoE
             message: message,
             errorLabels: errorLabels
         )
-    case (MONGOC_ERROR_WRITE_CONCERN, _):
-        var writeConcernErrorLabels =
-            reply?["writeConcernError"]?.documentValue?["errorLabels"]?.arrayValue?.asArrayOf(String.self)
-        if let errorLabels = errorLabels {
-            writeConcernErrorLabels = Array(Set((writeConcernErrorLabels ?? []) + errorLabels))
-        }
-        return CommandError(
-            code: ServerErrorCode(code.rawValue),
-            codeName: codeName,
-            message: message,
-            errorLabels: writeConcernErrorLabels
-        )
     case (MONGOC_ERROR_STREAM, _):
         return ConnectionError(message: message, errorLabels: errorLabels)
     case (MONGOC_ERROR_SERVER_SELECTION, MONGOC_ERROR_SERVER_SELECTION_FAILURE):
@@ -291,6 +283,7 @@ internal func extractMongoError(error bsonError: bson_error_t, reply: Document? 
     // if the reply is nil or writeErrors or writeConcernErrors aren't present, then this is likely a commandError.
     guard let serverReply: Document = reply,
         !(serverReply["writeErrors"]?.arrayValue ?? []).isEmpty ||
+        !(serverReply["writeConcernError"]?.documentValue?.keys ?? []).isEmpty ||
         !(serverReply["writeConcernErrors"]?.arrayValue ?? []).isEmpty else {
         return parseMongocError(bsonError, reply: reply)
     }
@@ -402,11 +395,14 @@ internal func extractBulkWriteError<T: Codable>(
 
 /// Extracts a `WriteConcernError` from a server reply.
 private func extractWriteConcernError(from reply: Document) throws -> WriteConcernFailure? {
-    guard let writeConcernErrors = reply["writeConcernErrors"]?.arrayValue?.compactMap({ $0.documentValue }),
-        !writeConcernErrors.isEmpty else {
+    if let writeConcernErrors = reply["writeConcernErrors"]?.arrayValue?.compactMap({ $0.documentValue }),
+        !writeConcernErrors.isEmpty {
+        return try BSONDecoder().decode(WriteConcernFailure.self, from: writeConcernErrors[0])
+    } else if let writeConcernError = reply["writeConcernError"]?.documentValue {
+        return try BSONDecoder().decode(WriteConcernFailure.self, from: writeConcernError)
+    } else {
         return nil
     }
-    return try BSONDecoder().decode(WriteConcernFailure.self, from: writeConcernErrors[0])
 }
 
 /// Internal function used by write methods performing single writes that are implemented via the bulk API. If the
--- a/Sources/MongoSwift/MongoSwiftVersion.swift
+++ b/Sources/MongoSwift/MongoSwiftVersion.swift
@@ -1,6 +1,4 @@
 // Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
 // DO NOT EDIT
 
-
-// swiftlint:disable:previous vertical_whitespace
 internal let MongoSwiftVersionString = "1.0.0-rc0"
--- a/Sources/MongoSwift/ReadPreference.swift
+++ b/Sources/MongoSwift/ReadPreference.swift
@@ -3,10 +3,10 @@ import CLibMongoC
 /// Represents a MongoDB read preference, indicating which member(s) of a replica set read operations should be
 /// directed to.
 /// - SeeAlso: https://docs.mongodb.com/manual/reference/read-preference/
-public struct ReadPreference {
+public struct ReadPreference: Decodable {
     /// An enumeration of possible read preference modes.
     /// - SeeAlso: https://docs.mongodb.com/manual/core/read-preference/#read-preference-modes
-    public enum Mode: String {
+    public enum Mode: String, Decodable {
         /// Default mode. All operations read from the current replica set primary.
         case primary
         /// In most situations, operations read from the primary but if it is unavailable, operations read from
@@ -91,6 +91,16 @@ public struct ReadPreference {
     /// A `ReadPreference` with mode `nearest`. With this mode, operations read from the member of the replica set with
     /// the least network latency, irrespective of the member’s type.
     public static let nearest = ReadPreference(.nearest)
+
+    private enum CodingKeys: String, CodingKey {
+        case mode
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let mode = try container.decode(Mode.self, forKey: .mode)
+        self.init(mode)
+    }
 
     /**
      * Initializes a new `ReadPreference` with the mode `primaryPreferred`. With this mode, in most situations
--- a/Sources/MongoSwift/ReadPreference.swift
+++ b/Sources/MongoSwift/ReadPreference.swift
@@ -3,10 +3,10 @@ import CLibMongoC
 /// Represents a MongoDB read preference, indicating which member(s) of a replica set read operations should be
 /// directed to.
 /// - SeeAlso: https://docs.mongodb.com/manual/reference/read-preference/
-public struct ReadPreference: Decodable {
+public struct ReadPreference {
     /// An enumeration of possible read preference modes.
     /// - SeeAlso: https://docs.mongodb.com/manual/core/read-preference/#read-preference-modes
-    public enum Mode: String, Decodable {
+    public enum Mode: String {
         /// Default mode. All operations read from the current replica set primary.
         case primary
         /// In most situations, operations read from the primary but if it is unavailable, operations read from
@@ -91,16 +91,6 @@ public struct ReadPreference: Decodable {
     /// A `ReadPreference` with mode `nearest`. With this mode, operations read from the member of the replica set with
     /// the least network latency, irrespective of the member’s type.
     public static let nearest = ReadPreference(.nearest)
-
-    private enum CodingKeys: String, CodingKey {
-        case mode
-    }
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        let mode = try container.decode(Mode.self, forKey: .mode)
-        self.init(mode)
-    }
 
     /**
      * Initializes a new `ReadPreference` with the mode `primaryPreferred`. With this mode, in most situations
--- a/Sources/MongoSwiftSync/ClientSession.swift
+++ b/Sources/MongoSwiftSync/ClientSession.swift
@@ -48,18 +48,6 @@ public final class ClientSession {
     /// The options used to start this session.
     public var options: ClientSessionOptions? { self.asyncSession.options }
 
-    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
-    public var id: Document? { self.asyncSession.id }
-
-    /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
-    public var serverId: Int? { self.asyncSession.serverId }
-
-    /// Enum tracking the state of the transaction associated with this session.
-    public typealias TransactionState = MongoSwift.ClientSession.TransactionState
-
-    /// The transaction state of this session.
-    public var transactionState: TransactionState? { self.asyncSession.transactionState }
-
     /// Initializes a new client session.
     internal init(client: MongoClient, options: ClientSessionOptions?) {
         self.client = client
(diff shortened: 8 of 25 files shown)


Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -57,19 +75,138 @@ struct TestOperationDescription: Decodable {
...
 
         do {
-            let result = try self.operation.execute(on: target, session: session)
+            let result = try self.operation.execute(on: target, sessionDict: sessionDict)
             expect(self.error ?? false)
                 .to(beFalse(), description: "expected to fail but succeeded with result \(String(describing: result))")
             if let expectedResult = self.result {
                 expect(result).to(equal(expectedResult))
             }
+        } catch let error as CommandError {
```
patrickfreed, line 99: In general, we need to derive the assertions we make from the properties of the expected error, not the properties of the actual error received. As an example, we should be asserting that "errorCodeName" matches if the spec test tells us to, not based on the type of error the operations actually resulted in.

To that end, we should first focus our efforts on determining what the spec test requires in the result. To do that, we need to introduce a new `TestOperationResult` case (e.g. `.error(ErrorDescription)`). The `.document` one we're currently using is intended for successful results that happen to return a document (e.g. a successful `findOneAndDelete`).

Once we have that, we can invert the conditions and do something like:
```swift
catch {
    expect(self.error ?? false).to(beTrue(), description: "expected no error, got \(error)")
    if case let .error(errorDescription) = self.result {
        errorDescription.validateMatches(error)
    }
}
```
Feel free to rename any method / struct /  variable I've mentioned so far. I just picked them for demonstration purposes.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -335,74 +525,101 @@ struct UpdateMany: TestOperation {
...
 struct InsertOne: TestOperation {
+    let session: String
     let document: Document
 
-    func execute(on target: TestOperationTarget, session _: ClientSession?) throws -> TestOperationResult? {
+    private enum CodingKeys: String, CodingKey { case session, document }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.session = (try? container.decode(String.self, forKey: .session)) ?? ""
```
patrickfreed, line 588: rather than nil coalescing, I think its more idiomatic to just store a `String?` on each of these operations. Also, instead of doing `try?`, we can use [`decodeIfPresent`](https://developer.apple.com/documentation/swift/keyeddecodingcontainer/2893445-decodeifpresent) to conditionally decode the session name.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +847,432 @@ struct RenameCollection: TestOperation {
...
+        }
+        let runCommandOptions = RunCommandOptions(readPreference: self.readPreference)
+        let result = try database.runCommand(
+            self.command,
+            options: runCommandOptions,
+            session: sessionDict[self.session]
+        )
+
+        var refinedResult = Document()
+        try result.copyElements(
```
patrickfreed, line 1117: instead of stripping out these extra fields by name, we can conform `TestOperationResult` to [`Matchable`](https://github.com/mongodb/mongo-swift-driver/blob/master/Tests/MongoSwiftSyncTests/SpecTestRunner/Match.swift), and update the assertion in the test runner to assert that the result matches the expected one instead of equaling it.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
+        }
+    }
+
+    internal func processCommandStartedEvent(
+        commandStartedEvent: CommandStartedEvent,
+        sessionIds: [String: Document]
+    )
+        throws -> TestCommandStartedEvent {
+        var event = TestCommandStartedEvent(from: commandStartedEvent)
+        try event.excludeKeys(keys: ["$db", "$clusterTime"])
```
patrickfreed, line 343: this exclusion shouldn't be necessary if we're checking these events using their [`Matchable`](https://github.com/mongodb/mongo-swift-driver/blob/master/Tests/MongoSwiftSyncTests/SpecTestRunner/Match.swift) conformance.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +847,432 @@ struct RenameCollection: TestOperation {
...
+    }
+}
+
+struct AbortTransaction: TestOperation {
+    func execute(on target: TestOperationTarget, sessionDict _: [String: ClientSession])
+        throws -> TestOperationResult? {
+        guard case let .session(session) = target else {
+            throw TestError(message: "session not provided to abortTransaction")
+        }
+        _ = try session.abortTransaction()
```
patrickfreed, line 1019: this doesnt return anything, so I don't think the `_ =` is needed. ditto for commit transaction.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +847,432 @@ struct RenameCollection: TestOperation {
...
+}
+
+struct AssertCollectionExists: TestOperation {
+    let database: String
+    let collection: String
+
+    func execute(on target: TestOperationTarget, sessionDict _: [String: ClientSession])
+        throws -> TestOperationResult? {
+        guard case let .testRunner(database) = target else {
+            throw TestError(message: "database not provided to assertCollectionExists")
```
patrickfreed, line 1135: rather than throwing errors here and in the other `.testRunner` operations, I think we should instead be doing actual assertions (i.e. `expect(...)`)

Tests/Specs/transactions/tests/abort.json
```diff
@@ -0,0 +1,621 @@
+{
```
patrickfreed, line 1: we generally commit the JSON straight to master ahead of time to reduce the diff in the PR with the actual test / code changes.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -258,7 +258,7 @@ extension Document {
 
     /// Helper function for copying elements from some source document to a destination document while
     /// excluding a non-zero number of keys
-    internal func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
+    public func copyElements(to otherDoc: inout Document, excluding keys: [String]) throws {
```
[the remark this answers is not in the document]
JamesHeppenstall: I didn't realize this was a thing. Thanks for the tip!

Sources/MongoSwift/ClientSession.swift
```diff
@@ -66,9 +66,74 @@ public final class ClientSession {
...
     public let client: MongoClient
 
-    /// The session ID of this session. This is internal for now because we only have a value available after we've
-    /// started the libmongoc session.
-    internal var id: Document?
+    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
+    public var id: Document?
+
+    /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
+    public var serverId: Int? {
```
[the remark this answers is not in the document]
JamesHeppenstall: Ok, I've made the changes and pulled out the references from `ClientSession` in `MongoSwiftSync`. They are now added to the extension of `MongoSwiftSync.ClientSession` in the test module.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -66,9 +66,74 @@ public final class ClientSession {
     /// The client used to start this session.
     public let client: MongoClient
 
-    /// The session ID of this session. This is internal for now because we only have a value available after we've
-    /// started the libmongoc session.
-    internal var id: Document?
+    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
+    public var id: Document?
```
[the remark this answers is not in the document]
JamesHeppenstall: Apologies for changing this. Reverted it back to its original form.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -214,8 +214,17 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
                     mongoc_bulk_operation_execute(bulk, replyPtr, &error)
                 }
 
-                let writeConcern = WriteConcern(from: mongoc_bulk_operation_get_write_concern(bulk))
-                return (serverId, writeConcern.isAcknowledged)
+                var writeConcernAcknowledged: Bool
```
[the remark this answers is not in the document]
JamesHeppenstall: As mentioned [here](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#users-cannot-pass-readconcern-or-writeconcern-to-operations-in-transactions), users cannot pass explicit read or write concerns to operations in transactions.

The C driver therefore nulls out the write concern of bulk write operations that are in a transaction. Calling `mongoc_bulk_operation_get_write_concern(bulk)`, however, asserts that the write concern associated with `bulk` is non-null, throwing the following error:
```
mongo-swift-driver/Sources/CLibMongoC/mongoc/mongoc-write-concern.c:132 mongoc_write_concern_journal_is_set(): precondition failed: write_concern'
```
To workaround this, we shouldn't call `mongoc_bulk_operation_get_write_concern(bulk)` for a bulk write operation in a transaction. 

Since we only need the write concern to check whether it is acknowledged or not, it is safe to default to `writeConcernAcknowledged=true` in a transaction because unacknowledged write concerns are [disallowed](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#writeconcern).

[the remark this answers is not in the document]
patrickfreed: I think you may have discovered a bug in libmongoc. I don't think `mongoc_bulk_operation_get_write_concern` is intended to ever return NULL. Instead I think it should return an empty write concern (e.g. `mongoc_write_concern_new()`).

So I think the comment should be updated to reflect what you said there, namely that libmongoc returns null write concerns from bulk operations during transactions, so we can only use the returned pointer in other cases. Another related thing, if we've determined the session is in a transaction, we need to throw an error if `self.options.writeConcern != nil` since libmongoc can't verify that for us.

Sources/MongoSwift/MongoError.swift
```diff
@@ -251,6 +251,18 @@ private func parseMongocError(_ error: bson_error_t, reply: Document?) -> MongoE
             message: message,
             errorLabels: errorLabels
         )
+    case (MONGOC_ERROR_WRITE_CONCERN, _):
+        var writeConcernErrorLabels =
+            reply?["writeConcernError"]?.documentValue?["errorLabels"]?.arrayValue?.asArrayOf(String.self)
+        if let errorLabels = errorLabels {
+            writeConcernErrorLabels = Array(Set((writeConcernErrorLabels ?? []) + errorLabels))
+        }
+        return CommandError(
```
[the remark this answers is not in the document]
JamesHeppenstall: The reason I introduced the new error here is because committing a transaction can result in a write concern error but no write error. Here's the error message thrown by the test **write concern error on commit** in **commit.json**:
```
{
    "writeConcernError" : {
        "code" : 100,
        "codeName" : "UnsatisfiableWriteConcern",
        "errmsg" : "Not enough data-bearing nodes",
        "errInfo" : {
            "writeConcern" : {
                "w" : 10,
                "wtimeout" : 0,
                "provenance" :
                "clientSupplied"
            }
        }
    },
    "ok" : 1.0,
    "$clusterTime" : {
        "clusterTime" : {
            "$timestamp" : {
                "t" : 1586298210,
                "i" : 17
            }
        },
        "signature" : {
            "hash" : {
                "$binary" : {
                    "base64": "AAAAAAAAAAAAAAAAAAAAAAAAAAA=",
                    "subType" : "00"
                }
            },
            "keyId" : 0
        }
    },
    "operationTime" : {
        "$timestamp" : {
            "t" : 1586298210,
            "i" : 17
        }
    }
}
```
In its current form, `extractMongoError` in `MongoError.swift` parses an error into a `WriteError` if and only if `serverReply` contains the fields `writeErrors` or `writeConcernErrors`. It doesn't look for `writeConcernError`. I've changed the logic to reflect the new field name.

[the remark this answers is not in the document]
JamesHeppenstall: This will cause the proper `WriteError` to be constructed.

Sources/MongoSwiftSync/ClientSession.swift
```diff
@@ -48,6 +48,18 @@ public final class ClientSession {
     /// The options used to start this session.
     public var options: ClientSessionOptions? { self.asyncSession.options }
 
+    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
+    public var id: Document? { self.asyncSession.id }
```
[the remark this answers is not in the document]
JamesHeppenstall: I removed these fields completely and added them to an extension of `MongoClientSync.ClientSession` in the test module.

Tests/MongoSwiftSyncTests/SpecTestRunner/Match.swift
```diff
@@ -69,8 +69,16 @@ extension Array: Matchable where Element: Matchable {
 extension Document: Matchable {
     internal func contentMatches(expected: Document) -> Bool {
         for (eK, eV) in expected {
-            guard let aV = self[eK], aV.matches(expected: eV) else {
-                return false
+            if eV != .null {
```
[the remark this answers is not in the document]
JamesHeppenstall: This is much better :)

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
+        if let outcome = self.outcome {
+            try self.checkOutcome(outcome: outcome, dbName: dbName, collName: collName!)
+        }
+    }
+
+    internal func processCommandStartedEvent(
+        commandStartedEvent: CommandStartedEvent,
+        sessionIds: [String: Document]
+    )
+        throws -> TestCommandStartedEvent {
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
         }
         defer { parent.disableActiveFailPoint() }
 
-        let client = try MongoClient.makeTestClient(options: clientOptions)
-        let monitor = client.addCommandMonitor()
-
-        let db = client.db(dbName)
-        var collection: MongoCollection<Document>?
+        // The spec tests refer to the sessions below. Proactively start them in case this spec test requires one.
+        let sessionsToStart = ["session0", "session1"]
```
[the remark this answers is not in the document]
JamesHeppenstall: Ok, I've made the requested modifications.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
 
-        if let collName = collName {
-            collection = db.collection(collName)
+        var sessionDict = [String: ClientSession]()
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/TransactionsTests.swift
```diff
@@ -0,0 +1,139 @@
...
+            guard skippedTestKeywords.allSatisfy({ !testFile.name.contains($0) }) else {
+                fileLevelLog("Skipping tests from file \(testFile.name)...")
+                continue
+            }
+            try testFile.runTests(parent: self)
+        }
+    }
+}
+
+extension DatabaseOptions: Decodable {
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/MongoError.swift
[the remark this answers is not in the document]
patrickfreed: Ah thats a really good catch. I think the `writeError_s_` comes from us funneling everything through the bulk API, and libmongoc appends an "s" to the end of the writeConcernError field because there could be many.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
+        sessionIds: [String: Document]
+    )
+        throws -> TestCommandStartedEvent {
+        var event = TestCommandStartedEvent(from: commandStartedEvent)
+        try event.excludeKeys(keys: ["$db", "$clusterTime"])
+
+        // If command started event has "lsid": Document(...), change the value to correpond to "session0",
+        // "session1", etc.
+        if let sessionDoc = event.command["lsid"]?.documentValue {
+            for session in sessionIds.keys {
```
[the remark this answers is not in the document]
JamesHeppenstall: You still need `sessionName` so looping over the key-value pairs is the way to go 👍

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
+            // Keep track of the session IDs assigned to each session.
+            // Deinitialize each session thereby implicitly ending them.
+            for session in sessionDict.keys {
+                sessionIds[session] = sessionDict[session]?.id
+                sessionDict[session] = nil
+            }
+        }
+
+        let events = try monitor.commandStartedEvents().map { commandStartedEvent in
+            try processCommandStartedEvent(commandStartedEvent: commandStartedEvent, sessionIds: sessionIds)
```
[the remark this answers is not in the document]
JamesHeppenstall: Made the changes and removed the `processCommandStartedEvent` and `excludeKeys` functions as they're no longer needed.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -259,35 +279,93 @@ extension SpecTest {
...
+        // If command started event has "lsid": Document(...), change the value to correpond to "session0",
+        // "session1", etc.
+        if let sessionDoc = event.command["lsid"]?.documentValue {
+            for session in sessionIds.keys {
+                if let sessionId = sessionIds[session], sessionId == sessionDoc {
+                    event.command["lsid"] = .string(session)
+                }
+            }
+        }
+        // If command is "findAndModify" and does not have key "new", add the default value "new": false.
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -4,7 +4,7 @@ import TestsCommon
 
 /// A enumeration of the different objects a `TestOperation` may be performed against.
 enum TestOperationObject: String, Decodable {
-    case client, database, collection, gridfsbucket
+    case client, database, collection, gridfsbucket, session0, session1, testRunner
```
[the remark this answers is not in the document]
JamesHeppenstall: I opted to use `RawRepresentable` for the enum. Let me know if this is an acceptable choice.

Tests/Specs/transactions/tests/abort.json
JamesHeppenstall: Ok, I'll go ahead and commit the JSON for the convenient api tests straight to master.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
JamesHeppenstall: Ok, I've updated the comment for now.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: I created a new `ErrorResult` struct and moved all the error checking code there. We now have a new `TestOperationResult` case `.error(ErrorResult)`.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Ok, I've added the expectations. Note that we still need the `TestError`s in the first guard statement since we expect the test runner to contain the `MongoDatabase`.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -72,8 +72,19 @@ public struct ClientOptions: CodingStrategyProvider, Decodable {
     /// Specifies a WriteConcern to use for the client.
     public var writeConcern: WriteConcern?
 
-    private enum CodingKeys: CodingKey {
-        case retryWrites, retryReads, readConcern, writeConcern
+    private enum CodingKeys: String, CodingKey {
+        case retryWrites, retryReads, readConcern, writeConcern, w, readConcernLevel, mode = "readPreference"
+    }
+
+    public init(from decoder: Decoder) throws {
```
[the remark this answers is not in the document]
JamesHeppenstall: This `Decodable` initializer was needed because `clientOptions` in the transactions spec tests have a weird format that can't be decoded with the default initializer. Here's an example from **transaction-options.json**:
```
{
      "description": "transaction options inherited from client",
      "clientOptions": {
        "w": 1,
        "readConcernLevel": "local"
      },
      "operations":
      ...
}
```
I agree that it's probably not good practice to add this decoding logic into the public API. I've therefore created a new struct called `TestClientOptions` in **SpecTest.swift** that decodes `clientOptions` in the spec tests properly and then has a function that converts it into a corresponding `ClientOptions` struct.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
JamesHeppenstall: I've also thrown a `LogicError` if `self.options.writeConcern != nil`. Let me know if you think it should be another error type.

JamesHeppenstall:
> fyi I don't think the latest changes have been pushed up yet

Just pushed them :)

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
```diff
@@ -120,3 +154,128 @@ extension DeleteResult: BulkWriteResultConvertible {
...
+
+    internal var errorContains: String?
+
+    internal var errorCodeName: String?
+
+    internal var errorLabelsContain: [String]?
+
+    internal var errorLabelsOmit: [String]?
+
+    public init(from doc: Document) {
```
patrickfreed, line 169: I think it would be a little more consistent if we still relied on `Decodable` to generate this struct rather than doing it ourselves. To avoid collisions with the `document` case, we can check that at least one of the keys is present via the `CodingKey`s, and throw an error if none are. We currently do that for [`BulkWriteResult`](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/MongoCollection%2BBulkWrite.swift#L299).

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
```diff
@@ -120,3 +154,128 @@ extension DeleteResult: BulkWriteResultConvertible {
...
+    }
+
+    internal static func == (lhs: ErrorResult, rhs: ErrorResult) -> Bool {
+        lhs.errorContains == rhs.errorContains &&
+            lhs.errorCodeName == rhs.errorCodeName &&
+            lhs.errorLabelsContain == rhs.errorLabelsContain &&
+            lhs.errorLabelsOmit == rhs.errorLabelsOmit
+    }
+
+    public func checkErrorResult(_ error: Error) throws {
```
patrickfreed, line 186: so we still have the problem where we're deriving our assertions from the type of the error the driver threw rather than the expectations from the test. So instead of casting the error first and checking the test requirements, we need to first check for the test requirements and cast as necessary to verify.

e.g.
```swift
if let messageSubstring = self.errorContains {
    expect(error.localizedDescription).to(contain(messageSubstring))
}

if let codeName = self.errorCodeName {
    if let commandError = error as? CommandError {
        expect(commandError.codeName).to(equal(codeName))
    } else if let ... = error as? ... { ... }
    else { XCTFail("\(error) doesn't contain a codeName") }
}

if let errorLabels = self.errorLabelsContain {
    guard let labeledError = error as? LabeledError else {
        XCTFail("\(error) doesn't contain errorLabels")
    }
    expect(labeledError.errorLabels).to(contain(errorLabels))
}

// etc
```
This is important because as it is now, a failing test could mistakeln pass depending on the error thrown. For example, if the test specifies `errorLabelsContain: ["some label"]` and the driver threw a `LogicError`, we would pass because `checkLogicError` doesn't verify errorLabels.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
```diff
@@ -71,12 +84,33 @@ enum TestOperationResult: Decodable, Equatable {
...
+        case let (.bulkWrite(bw), .bulkWrite(expectedBw)):
+            return bw.matches(expected: expectedBw)
+        case let (.int(int), .int(expectedInt)):
+            return int.matches(expected: expectedInt)
+        case let (.array(array), .array(expectedArray)):
+            return array.matches(expected: expectedArray)
+        case let (.document(doc), .document(expectedDoc)):
+            return doc.matches(expected: expectedDoc)
+        case let (.error(error), .error(expectedError)):
+            return error.matches(expected: expectedError)
```
patrickfreed, line 105: we actually don't want to use the `Matchable` conformance for handling errors, we want to use the other custom logic in `checkErrorResult`. So we should just return false in the `(.error, _)` case.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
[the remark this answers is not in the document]
patrickfreed: as part of this we can make the `command` field of `TestCommandStartedEvent` a `let` property again.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: bah, that's unfortunate. Could we not add an extension to `ClientOptions` into `CodableExtensions.swift` instead of introducing the new type?

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
kmahar: if this is a libmongoc bug we should open a CDRIVER ticket

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
JamesHeppenstall: Ahh apologies for the confusion, I've made the proper changes. One thing of note is that `localizedDescription`doesn't return the correct error message so I've had to resort to parsing the error message depending on the error type. Another thing of note is that the messages have case discrepancies so I've compared them in lowercase.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
JamesHeppenstall: I tried doing this but since `ClientOptions` already conforms to `Decodable`, you cannot provide another initializer of the form `init(from decoder: Decoder)`.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -66,10 +66,75 @@ public final class ClientSession {
     /// The client used to start this session.
     public let client: MongoClient
 
-    /// The session ID of this session. This is internal for now because we only have a value available after we've
-    /// started the libmongoc session.
+    /// The session ID of this session. We only have a value available after we've started the libmongoc session.
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -92,6 +92,16 @@ public struct ReadPreference {
     /// the least network latency, irrespective of the member’s type.
     public static let nearest = ReadPreference(.nearest)
 
+    private enum CodingKeys: String, CodingKey {
+        case mode
+    }
+
+    public init(from decoder: Decoder) throws {
```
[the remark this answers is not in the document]
JamesHeppenstall: Ok, I've removed the `Decodable` conformance from the public API and added it to `CodableExtensions.swift`.

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -23,13 +23,15 @@ struct ClientSessionOp {
 }
 
 extension MongoSwiftSync.ClientSession {
-    var active: Bool {
-        self.asyncSession.active
-    }
+    internal var active: Bool { self.asyncSession.active }
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: Oh I see. I think we can just remove the `Codable` conformance in `ClientOptions` declaration then--any "options" codable conformance exists right now only for testing purposes. We figured it saved us from enough boilerplate and that users probably wouldn't try to actually use the conformance, so it was worth it to bleed into the API. I think it must have been an earlier version of Swift or something that required us to conform in the driver rather than the tests, or maybe we just never tried lol.

Sources/MongoSwift/MongoError.swift
```diff
@@ -165,19 +165,23 @@ public struct WriteConcernFailure: Codable {
     public let code: ServerErrorCode
 
     /// A human-readable string identifying write concern error.
-    public let codeName: String
+    public let codeName: String?
```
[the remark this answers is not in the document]
patrickfreed: I think its fine to just leave as-is. Just to confirm though: one of the write concern errors returned by the server as part of these tests doesn't have a codeName, and this isn't for reading errors from a spec test json file, right?

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +775,405 @@ struct RenameCollection: TestOperation {
...
+}
+
+struct StartTransaction: TestOperation {
+    let options: TransactionOptions
+
+    private enum CodingKeys: CodingKey {
+        case options
+    }
+
+    init() {
```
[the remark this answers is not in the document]
patrickfreed: Oh, that's a bit annoying. Yeah I think this is a good approach then. My only comment would be to use `decodeIfPresent` rather than `try?` + `decode`

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
patrickfreed: Sounds good. I think we probably want our `localizedDescription` to contain a `message`, if there is one. Filed [SWIFT-769](https://jira.mongodb.org/browse/SWIFT-769) to fix that.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
```diff
@@ -120,3 +149,141 @@ extension DeleteResult: BulkWriteResultConvertible {
...
+                expect(labeledError.errorLabels).to(contain(label))
+            }
+        }
+
+        if let errorLabelsOmit = self.errorLabelsOmit {
+            guard let labeledError = error as? LabeledError else {
+                XCTFail("\(error) does not contain errorLabels")
+                return
+            }
+            if labeledError.errorLabels == nil {
```
patrickfreed, line 281: it would be a bit more idiomatic to write this as a `guard let`. I think this could alternatively be omitted as it is in the errorLabelsContain part.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
```diff
@@ -120,3 +149,141 @@ extension DeleteResult: BulkWriteResultConvertible {
...
+            }
+        }
+    }
+
+    // swiftlint:enable cyclomatic_complexity
+
+    internal func checkCodeName(_ error: Error) throws {
+        if let errorCodeName = self.errorCodeName {
+            if let commandError = error as? CommandError {
+                expect(commandError.codeName).to(satisfyAnyOf(equal(errorCodeName), equal("")))
```
patrickfreed, line 238: if a codeName was provided by the test file, I don't think we can accept the empty string as valid also.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -257,37 +328,70 @@ extension SpecTest {
...
 
         if let expectations = self.expectations {
             expect(events).to(match(expectations), description: self.description)
         }
+
+        try self.checkOutcome(dbName: dbName, collName: collName)
+    }
+
+    internal func checkOutcome(dbName: String, collName: String?) throws {
+        if let outcome = self.outcome {
```
patrickfreed, line 384: because this methods entire body is contained by this `if let`, we can just make it a `guard let` that returns on failure instead.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
patrickfreed: To clarify what I meant about updating the comment: bulk write operations in transactions can in fact use write concerns; it's just those write concerns must come from the session, not from the `BulkWriteOptions` passed to the `bulkWrite` helper. The problem we have here is that libmongoc surfaces this implementation detail by nulling out the write concern stored on the bulk write, even if a write concern will be used as part of the transaction. To sidestep this, we can only call `bulk_operation_get_write_concern` if we know we're not in a transaction.

With respect to the error message, this means we should be a bit more specific about what exactly the user did wrong. As it is now, it seems that they can't use a write concern at all, which isn't the case. It could say something like "Cannot specify a write concern on an individual helper in a transaction. Instead specify it when starting the transaction." or something along those lines.

Another thing I just realized is that we're throwing this argument error after we actually executed the bulk write. We should be doing this outside of the `withMongocCollection` block so that the write isn't attempted if we know we'll be throwing an error.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
JamesHeppenstall: I included this here because an error could conform to `LabeledError` but have `errorLabels=nil`, in which case the expectation fails.

JamesHeppenstall: I've changed it to a `guard let`.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
JamesHeppenstall: I spoke with @kmahar about this earlier and these empty strings are necessary until SERVER-36755 is resolved. I've added a TODO note in the `checkCodeName` function as a reminder.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
JamesHeppenstall: Done.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -70,6 +70,72 @@ public final class ClientSession {
     /// started the libmongoc session.
     internal var id: Document?
 
+    /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
+    internal var serverId: Int? {
+        switch self.state {
+        case .notStarted, .ended:
+            return nil
+        case let .started(session, _):
+            return Int(mongoc_client_session_get_server_id(session))
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -214,8 +214,23 @@ internal struct BulkWriteOperation<T: Codable>: Operation {
...
+                if let transactionState = session?.transactionState, transactionState != .none {
+                    // Bulk write operations cannot have a write concern in a transaction. Default to
+                    // writeConcernAcknowledged = true. Since `libmongoc` returns a null write concern from bulk write
+                    // operations in a transaction, we cannot call `mongoc_bulk_operation_get_write_concern`.
+                    if self.options?.writeConcern != nil {
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
```diff
@@ -0,0 +1,79 @@
+@testable import struct MongoSwift.ReadPreference
+import MongoSwiftSync
+
+extension DatabaseOptions: Decodable {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let readConcern = try? container.decode(ReadConcern.self, forKey: .readConcern)
```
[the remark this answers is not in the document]
JamesHeppenstall: Ok, I've updated all of these `try?` statements.

Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
```diff
@@ -0,0 +1,79 @@
...
+        )
+        self.init(causalConsistency: causalConsistency, defaultTransactionOptions: defaultTransactionOptions)
+    }
+
+    private enum CodingKeys: CodingKey {
+        case causalConsistency, defaultTransactionOptions
+    }
+}
+
+extension TransactionOptions: Decodable {
```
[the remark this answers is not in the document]
JamesHeppenstall: Yeah, the compiler throws the following error if you try to rely on generated codable conformance:
```
error: implementation of 'Decodable' cannot be automatically synthesized in an extension in a different file to the type
```

Sources/MongoSwift/MongoError.swift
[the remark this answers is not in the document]
patrickfreed: Hah, this is actually due to a libmongoc bug ([CDRIVER-3147](https://jira.mongodb.org/browse/CDRIVER-3147)) which I reported and subsequently forgot about. In the comments you can see we even said we'd run into this when implementing transactions, lol.

Anyways, since this is a libmongoc bug and therefore likely has a shorter path to being fixed, we should probably keep this codeName non-optional. We will have to add a custom decoder initializer like we do for `WriteFailure` [here](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/MongoError.swift#L146).

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
patrickfreed: One last thing: I think it would make the conditions slightly clearer if we added an `inTransaction` computed property (or something named like that) to `ClientSession` to replace `let transactionState = session?.transactionState, transactionState != .none`.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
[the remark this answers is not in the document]
patrickfreed: This still needs to be done

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
patrickfreed: Should we do the same thing for the errorLabelsContain portion then?

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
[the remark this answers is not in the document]
JamesHeppenstall: This is outdated. I force pushed after rebasing on master so maybe this is why you still see it.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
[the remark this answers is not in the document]
patrickfreed: I do see the other changes, but I still see `command` as a `let` property. I'm referring to this [line.](https://github.com/mongodb/mongo-swift-driver/blob/SWIFT-752/spec-test-runner/Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift#L10)

Sources/MongoSwift/MongoError.swift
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
JamesHeppenstall: I don't think so because we want the `expect` to fail in errorLabelsContain if `labeledError.errorLabels` is `nil`.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
patrickfreed: Oh right, of course. I got myself confused.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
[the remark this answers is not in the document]
patrickfreed: looks good!

I filed [CDRIVER-3607](https://jira.mongodb.org/browse/CDRIVER-3607) for the bug you discovered.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +769,339 @@ struct RenameCollection: TestOperation {
...
 
 struct ListDatabaseNames: TestOperation {
-    func execute(on target: TestOperationTarget, session: ClientSession?) throws -> TestOperationResult? {
+    func execute(on target: TestOperationTarget, sessions _: [String: ClientSession])
+        throws -> TestOperationResult? {
         guard case let .client(client) = target else {
             throw TestError(message: "client not provided to listDatabaseNames")
         }
-        return try .array(client.listDatabaseNames(session: session).map { .string($0) })
+        return try .array(client.listDatabaseNames(session: nil).map { .string($0) })
```
kmahar, line 795: nit throughout the file: session is an optional argument whose value defaults to `nil` so I think you can just replace all these `(session: nil)`s with `()`

that said, can one of you clarify why we were passing through a `session` argument previously here but don't need to anymore?

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +769,339 @@ struct RenameCollection: TestOperation {
...
-            _ = try client.watch(session: session)
+            _ = try client.watch(session: nil)
         case let .database(database):
-            _ = try database.watch(session: session)
+            _ = try database.watch(session: nil)
         case let .collection(collection):
-            _ = try collection.watch(session: session)
+            _ = try collection.watch(session: nil)
+        case .session, .testRunner:
+            break
```
kmahar, line 882: this should maybe be an error rather than a `break`?

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -599,122 +769,339 @@ struct RenameCollection: TestOperation {
...
+    init() {
+        self.options = TransactionOptions()
+    }
+
+    func execute(on target: TestOperationTarget, sessions _: [String: ClientSession])
+        throws -> TestOperationResult? {
+        guard case let .session(session) = target else {
+            throw TestError(message: "session not provided to startTransaction")
+        }
+        _ = try session.startTransaction(options: self.options)
```
kmahar, line 910: do we need the `_`? I thought this method returned nil already

Tests/MongoSwiftSyncTests/TransactionsTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    override func setUp() {
+        self.continueAfterFailure = false
+    }
+
+    func testTransactions() throws {
+        let skippedTestKeywords = [
+            "count", // skipped in RetryableReadsTests.swift
+            "mongos-pin-auto", // useMultipleMongoses, targetedFailPoint not implemented
```
kmahar, line 62: do we need a separate ticket tracking these?

Tests/MongoSwiftSyncTests/TransactionsTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        self.disableActiveFailPoint()
+    }
+
+    override func setUp() {
+        self.continueAfterFailure = false
+    }
+
+    func testTransactions() throws {
+        let skippedTestKeywords = [
+            "count", // skipped in RetryableReadsTests.swift
```
kmahar, line 61: so the reason we skip this in retryable reads is that the tests are written using the old count API, which was last minute deprecated right before the 4.0 release. we don't have the old API at all but some older drivers that haven't tagged a major version since then still do.

in this case, it looks like the txns count tests are checking that certain errors occur when you try to run `count` in a transactions, since that isn't supported by the server. since we don't provide an API for running the `count` command I think it's ok to skip this, but we could update the comment to be more descriptive.

we could consider converting it to `runCommand` and running count that way since a user could technically still do it, but that seems probably like overkill.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -71,12 +71,12 @@ public final class ClientSession {
...
 
     /// The server ID of the mongos this session is pinned to. A server ID of 0 indicates that the session is unpinned.
-    internal var serverId: Int? {
+    internal var serverId: UInt32? {
         switch self.state {
         case .notStarted, .ended:
             return nil
         case let .started(session, _):
-            return Int(mongoc_client_session_get_server_id(session))
+            return UInt32(mongoc_client_session_get_server_id(session))
```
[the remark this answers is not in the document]
JamesHeppenstall: Done.

Sources/MongoSwift/MongoError.swift
```diff
@@ -173,11 +173,36 @@ public struct WriteConcernFailure: Codable {
...
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.code = try container.decode(ServerErrorCode.self, forKey: .code)
+        self.message = try container.decode(String.self, forKey: .message)
+        self.codeName = try container.decodeIfPresent(String.self, forKey: .codeName) ?? ""
+        self.details = try container.decodeIfPresent(Document.self, forKey: .details)
+        self.errorLabels = try container.decodeIfPresent([String].self, forKey: .errorLabels)
+    }
+
+    // TODO: can remove this once SERVER-36755 is resolved
+    internal init(
```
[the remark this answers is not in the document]
JamesHeppenstall: Essentially, the `codeName` is not always set in `writeError`s and `writeConcernError`s so we need to manually satisfy codable conformance. This is until SERVER fixes issues in which `codeName` is missing.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperationResult.swift
```diff
@@ -120,3 +149,142 @@ extension DeleteResult: BulkWriteResultConvertible {
...
+            )
+        }
+
+        self.errorContains = try container.decodeIfPresent(String.self, forKey: .errorContains)
+        self.errorCodeName = try container.decodeIfPresent(String.self, forKey: .errorCodeName)
+        self.errorLabelsContain = try container.decodeIfPresent([String].self, forKey: .errorLabelsContain)
+        self.errorLabelsOmit = try container.decodeIfPresent([String].self, forKey: .errorLabelsOmit)
+    }
+
+    internal static func == (lhs: ErrorResult, rhs: ErrorResult) -> Bool {
```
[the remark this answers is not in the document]
JamesHeppenstall: You're right we don't need it - I've removed it.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -241,10 +295,27 @@ internal protocol SpecTest: Decodable {
...
+    /// Document describing the return value and/or expected state of the collection after the operation is executed.
+    var outcome: TestOutcome? { get }
+
+    /// Map of session names (e.g. "session0") to parameters to pass to `MongoClient.startSession()` when creating that
+    /// session.
+    var sessionOptions: [String: ClientSessionOptions]? { get }
+
+    /// Array of session names (e.g. "session0", "session1") that the test refers to. Each session is proactively
+    /// started in `run()`.
+    static var sessionNames: [String] { get }
```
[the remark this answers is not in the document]
JamesHeppenstall: There can only be sessions that exist here. The reason for the dictionary of session names (i.e. "session0") to `ClientSession`s is that the spec test assert that `CommandStartedEvent`s contain the field `"lsid": session0`. Since the `lsid` is generated automatically by the server and cannot be specified, we need to check that the generated `lsid` corresponds to that of "session0".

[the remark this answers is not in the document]
JamesHeppenstall: We've added this field so that you can specify only the session names that a spec test refers to. For instance, one test may refer to "session0" and "session1" whereas another might refer to "sessionA" and "sessionB". This gives us the flexibility to reflect any of these changes down the line.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -83,12 +148,18 @@ enum TestOperationTarget {
 
     /// Execute against the provided collection.
     case collection(MongoCollection<Document>)
+
+    /// Execute against the provided session.
+    case session(ClientSession)
+
+    /// Execute against the provided test runner.
```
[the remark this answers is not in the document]
JamesHeppenstall: That's a really good point about the db name. It actually isn't necessary at all so I've just removed it completely. The db name is directly passed to the test runner operations.

I've also updated the comment to reflect that the test runner refers to special test ops.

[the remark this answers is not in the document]
JamesHeppenstall: I also discovered a small error where `assertSessionPinned` and `assertSessionUnpinned` were inverted so have fixed that.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: You can't pass a session here because we are never provided a session name in this command. So it's impossible to know which session in the session dictionary to provide.

Also, none of the spec tests using this command refer to sessions so it's not even necessary to provide.

JamesHeppenstall: I've removed all references to `session: nil` in `TestOperation.swift`.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Good point, I've changed it to throw a `TestError`.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: You're right, I've removed it.

Tests/MongoSwiftSyncTests/TransactionsTests.swift
JamesHeppenstall: I think we need a separate ticket tracking the `useMultipleMongoses` part.

kmahar: ok let's create a ticket then - and I think I answered that in my comment below
> so the reason we skip this in retryable reads is that the tests are written using the old count API, which was last minute deprecated right before the 4.0 release. we don't have the old API at all but some older drivers that haven't tagged a major version since then still do.

JamesHeppenstall: Got it. I've actually implemented support for `useMultipleMongoses` and `targetedFailPoint`, which I've included in the most recent commit. Evergreen is still failing with some weird errors so I will comment out the tests for now. I'll create a ticket to remind us to finish this.

Tests/MongoSwiftSyncTests/TransactionsTests.swift
JamesHeppenstall: Ok, I've updated the comment to be more specific about why `count` is implemented.

Tests/MongoSwiftSyncTests/TransactionsTests.swift
JamesHeppenstall: Created [SWIFT-774](https://jira.mongodb.org/browse/SWIFT-774).

Tests/MongoSwiftSyncTests/TransactionsTests.swift
patrickfreed: So we actually do call `count` under the hood in `estimatedDocumentCount`. That being said, [`estimatedDocumentCount` is forbidden from accepting a session by the sessions spec](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#new-collection-methods-that-take-an-explicit-session) because `count` isn't supported in transactions, so it's a little puzzling why this tests exists in the first place.

So for that reason, I think we should continue to skip the test, but we should remove the session parameter from `estimatedDocumentCount`, and also perhaps add a line to its docstring mentioning that it cannot be used in a transaction.

kmahar: as for why the test exists - it just checks that you properly error when a user tries to call `count` in a transaction.

kmahar: ok I see your edit now jinx lol

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: So what I meant by this was to remove the `Decodable` conformance in [this line](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/MongoClient.swift#L6). Right now, that declaration is causing the compiler to "synthesize" a `Decodable` conformance in the driver code, which is why you're seeing the error complaining about multiple decodable conformances. Once that's removed, you can add it as an extension in the tests.

Sources/MongoSwift/MongoError.swift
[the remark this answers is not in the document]
patrickfreed: And because we add the initializer for manual `Codable` conformance, the compiler stops synthesizing the default initializer. For that reason I kind of wish there was an `@default` or something that explicitly created the initializer. It seems really easy to accidentally make a breaking change in this way.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -257,37 +328,78 @@ extension SpecTest {
...
+            try parent.activateFailPoint(failPoint)
+        }
+        defer { parent.disableActiveFailPoint() }
+
+        var sessions = [String: ClientSession]()
+        for session in Self.sessionNames {
+            sessions[session] = client.startSession(options: self.sessionOptions?[session])
         }
 
+        var sessionIds = [String: Document]()
```
patrickfreed, line 360: given the way this map is actually used, I think it should probably be flipped to a `[Document: String]`, since we're using the ids to look up session names.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -335,74 +456,102 @@ struct UpdateMany: TestOperation {
...
 }
 
 struct InsertOne: TestOperation {
+    let session: String?
     let document: Document
 
-    func execute(on target: TestOperationTarget, session _: ClientSession?) throws -> TestOperationResult? {
+    private enum CodingKeys: String, CodingKey { case session, document }
+
+    init(from decoder: Decoder) throws {
```
patrickfreed, line 518: I think this custom initializer isn't required. After looking over this file again, it seems its necessary for operations that decode top level arguments into an optoins struct (e.g. `DeleteOne` does this).

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
JamesHeppenstall: Ah I think I missed addressing this comment. I've removed `TestClientOptions` and simply written a decodable extension of `ClientOptions`.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -335,74 +456,102 @@ struct UpdateMany: TestOperation {
...
 
 struct InsertMany: TestOperation {
+    let session: String?
     let documents: [Document]
     let options: InsertManyOptions
 
-    func execute(on target: TestOperationTarget, session: ClientSession?) throws -> TestOperationResult? {
+    private enum CodingKeys: String, CodingKey { case session, documents }
+
+    init(from decoder: Decoder) throws {
```
patrickfreed, line 539: Despite having an options struct, this one doesn't actually need any special handling. If you look at the retryable-writes form of this operation, it has an `options` field:
```json
"operation": {
    "name": "insertMany",
    "arguments": {
        "documents": [],
        "options": {
            "ordered": true
        }
    }
}
```
In the transactions tests, it has a session field at the same level `options` would be:
```json
{
          "name": "insertMany",
          "object": "collection",
          "arguments": {
            "documents": [],
            "session": "session1"
          },
}
```
So they're compatible, and we can delete this initializer and just rely on the synthesized one.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
```diff
@@ -470,126 +619,150 @@ extension WriteModel: Decodable {
...
 
 struct BulkWrite: TestOperation {
+    let session: String?
     let requests: [WriteModel<Document>]
     let options: BulkWriteOptions
 
-    func execute(on target: TestOperationTarget, session: ClientSession?) throws -> TestOperationResult? {
+    private enum CodingKeys: CodingKey { case session, requests }
+
+    init(from decoder: Decoder) throws {
```
patrickfreed, line 628: my comment for `InsertMany` applies here as well

Tests/MongoSwiftSyncTests/TransactionsTests.swift
JamesHeppenstall: Ok, I've removed the session parameter from `estimatedDocumentCount` and added a line to the docstring mentioning that it can't be used in a transaction.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
JamesHeppenstall: Done. One thing to note is that I've had to nil-coalesce the `id` since it's optional in ClientSession. This default value should never occur, however, since we explicitly start the sessions above.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperation.swift
JamesHeppenstall: Done.

Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -44,7 +44,9 @@ final class SyncClientSessionTests: MongoSwiftTestCase {
         CollectionSessionOp(name: "aggregate") { _ = try $0.aggregate([], session: $1).next()?.get() },
         CollectionSessionOp(name: "distinct") { _ = try $0.distinct(fieldName: "x", session: $1) },
         CollectionSessionOp(name: "countDocuments") { _ = try $0.countDocuments(session: $1) },
-        CollectionSessionOp(name: "estimatedDocumentCount") { _ = try $0.estimatedDocumentCount(session: $1) }
+        CollectionSessionOp(name: "estimatedDocumentCount") { collection, _ in
```
kmahar, line 47: this should maybe be removed, but now I'm wondering how this test was passing to begin with if this op doesn't support sessions? cc @patrickfreed

kmahar: seems like a libmongoc bug that one is accepted. I tested this by passing in a session explicitly and using command monitoring to verify that session is used for the command
```
explicit session id: Optional({ "id" : { "$binary" : { "base64": "OWNGttajRqywX2fJDFdr1Q==", "subType" : "04" } } })

started(MongoSwift.CommandStartedEvent(command: { "count" : "MongoCollectionTests_testBlah", "$db" : "test", "lsid" : { "id" : { "$binary" : { "base64": "OWNGttajRqywX2fJDFdr1Q==", "subType" : "04" } } } }, databaseName: "test", commandName: "count", requestId: 10, operationId: 282475254, serverAddress: MongoSwift.Address(host: "127.0.0.1", port: 27017)))

succeeded(MongoSwift.CommandSucceededEvent(duration: 433, reply: { "n" : 2, "ok" : 1.0 }, commandName: "count", requestId: 10, operationId: 282475254, serverAddress: MongoSwift.Address(host: "127.0.0.1", port: 27017)))
```

kmahar: I opened [CDRIVER-3612](https://jira.mongodb.org/browse/CDRIVER-3612)

JamesHeppenstall: So it looks like `estimatedDocumentCount` is now failing some evergreen tests. Not too sure why yet but am looking into it.

JamesHeppenstall: I'm going to remove this test since it's failing after removing the session option.

kmahar: sounds good to me

Tests/MongoSwiftSyncTests/TransactionsTests.swift
```diff
@@ -0,0 +1,78 @@
...
+            "mongos-pin-auto", // TODO: see SWIFT-774
+            "mongos-recovery-token", // TODO: see SWIFT-774
+            "pin-mongos", // TODO: see SWIFT-774
+            "retryable-abort-errorLabels", // requires libmongoc v1.17 (see SWIFT-762)
+            "retryable-commit-errorLabels" // requires libmongoc v1.17 (see SWIFT-762)
+        ]
+
+        let tests = try retrieveSpecTestFiles(specName: "transactions", asType: TransactionsTestFile.self)
+        for (_, testFile) in tests {
+            guard skippedTestKeywords.allSatisfy({ !testFile.name.contains($0) }) else {
```
kmahar, line 71: it looks like we also have logic to skip tests by filename in `runTests`, do we need this here too?

Sources/MongoSwift/MongoSwiftVersion.swift
```diff
@@ -1,6 +1,4 @@
 // Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
 // DO NOT EDIT
 
-
```
kmahar, line 4: did swiftlint start complaining about this suddenly?

JamesHeppenstall: Yes, it was complaining about the vertical whitespace comments and saying they were superflous.

Tests/MongoSwiftSyncTests/TransactionsTests.swift
JamesHeppenstall: That's a good point. The same is done in the retryable reads tests but I'll remove them from both and just have `runTests` handle it.

JamesHeppenstall: Actually, `runTests` skipped based on words in the description while the tests themselves skip based on `name`. I think it's probably better to leave as is.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
patrickfreed: So it actually can occur because `startSession` doesn't actually start the session; that only happens upon operation execution (see the discussion in slack about this). So if the `validateExecution` loop didn't use a particular session, its id will be `nil`. Given that the only circumstance the id will be `nil` is if we don't need it anyways, I think we can just conditionally add the sessions with non-nil id's.

JamesHeppenstall: Ok will do.

PR #444, opened 2020-04-06
SWIFT-423 Provide transactions guide for docs

JamesHeppenstall:


kmahar:
my bad for not pointing this out before, but we have all of the official docs examples live in the project located at `Examples/DocsExamples`, so it would be great if we could have the code in there as well. that makes it easy for us to ensure our example code still builds in the future.

that said I think it's a good idea to have a guide in our docs on transaction use too so let's keep the markdown file, and we can expand it with convenient API examples later on.

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
...
+    try? elg.syncShutdownGracefully()
+}
+
+let client = try MongoClient(using: elg)
+let session = client.startSession()
+
+let txnOpts = TransactionOptions(
+    maxCommitTimeMS: 30,
+    readConcern: ReadConcern(.local),
+    readPreference: ReadPreference.primaryPreferred,
```
kmahar, line 61: nit: this can just be `.primaryPreferred`, the compiler will infer the `ReadPreference` part

same below

JamesHeppenstall: Done.

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
+# Swift Driver Transactions Guide
+
+`MongoSwift` 1.0.0 added support for [transactions](https://docs.mongodb.com/manual/core/transactions/), which allow applications to use multi-statement transactions that guarantee the atomicity of reads and writes to multiple documents (in a single or multiple collections). Applications can use transactions instead of implementing complex and error-prone ACID-compliant logic themselves, simplifying development and allowing developers to focus on the new features that really matter.
```
patrickfreed, line 3: >which allow applications to use multi-statement transactions

When defining "transactions", we probably shouldn't use the word "transactions" again. This could be rewritten to something like "which allow applications to execute multiple read and write operations atomically across multiple documents and/or collections."

>Applications can use transactions instead of implementing complex and error-prone ACID-compliant logic themselves, simplifying development and allowing developers to focus on the new features that really matter.

I think this is a little too marketing-speaky for our guides. I think it's also important to note, for users who are used to relational databases, that transactions are not always necessary in MongoDB. Maybe we could update it to something milder like: "Transactions reduce the need for complicated application logic when operating on several different documents simultaneously; however, because operations on single documents are always atomic, transactions are often not necessary.", or something like that.

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
...
+
+**Note**: Transactions only work with MongoDB replica sets (v4.0+) and sharded clusters (v4.2+).
+
+## Examples
+
+### Transaction that Atomically Moves a `Document` from One `MongoCollection` to Another
+```swift
+let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
```
patrickfreed, line 11: I think its probably not necessary to include the event loop group boilerplate in each one.

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
+# Swift Driver Transactions Guide
+
+`MongoSwift` 1.0.0 added support for [transactions](https://docs.mongodb.com/manual/core/transactions/), which allow applications to use multi-statement transactions that guarantee the atomicity of reads and writes to multiple documents (in a single or multiple collections). Applications can use transactions instead of implementing complex and error-prone ACID-compliant logic themselves, simplifying development and allowing developers to focus on the new features that really matter.
+
+**Note**: Transactions only work with MongoDB replica sets (v4.0+) and sharded clusters (v4.2+).
+
```
patrickfreed, line 6: I think this guide could use a little prose section here describing the general process of using transactions with the following key points:
- transactions are started on a session
- that session must be passed to each operation in the transaction
- the transaction must be committed
- the transaction can be aborted

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
...
+## Examples
+
+### Transaction that Atomically Moves a `Document` from One `MongoCollection` to Another
```
patrickfreed, line 9: each of these examples could use a little blurb describing what's going on in them. For this one, I think it might be useful to include an explanation of why a transaction is useful here, with perhaps a "bad" example of doing this non atomically with commentary pointing out the issues.

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
...
+    // do something
+}.flatMap { _ in
+    session.commitTransaction()
+}.whenFailure { error in
+    session.abortTransaction()
+    // handle error
+}
+```
+
+Note: Any transaction options provided directly to `startTransaction()` override the default transaction options for the session. More so, the default transaction options for the session override any options inherited from the client.
```
patrickfreed, line 108: i think it would be a little clearer to move this example to before the `Transaction with Custom Transaction Options` one, and to add a blurb to the top of the `Transaction with Custom Transaction Options` one with this note.

Guides/Transactions.md
```diff
@@ -0,0 +1,111 @@
...
+
```
patrickfreed, line 109: The [error handling examples from the spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#error-reporting-and-retrying-transactions) would be useful to have here. We can also link in the error handling guide's description of the various labels to this section once added.

Guides/Transactions.md
JamesHeppenstall: Done.

Guides/Transactions.md
JamesHeppenstall: Done.

Guides/Transactions.md
JamesHeppenstall: Done.

Guides/Transactions.md
JamesHeppenstall: Done.

Guides/Transactions.md
JamesHeppenstall: Done.

Guides/Transactions.md
JamesHeppenstall: I'll make a note to link the error handling examples when [SWIFT-347](https://jira.mongodb.org/browse/SWIFT-347) is pushed. I've currently got a PR open for these changes.

JamesHeppenstall:
Note that the tests to compile the examples will fail until SWIFT-752 is pushed to master. This is because you must currently specify the label `options` in `startTransaction()`.

Guides/Transactions.md
```diff
@@ -0,0 +1,87 @@
...
+Transactions in the driver must be started on a `ClientSession` using `startTransaction()`. The session must then be passed to each operation in the transaction. If the session is not passed to an operation, said operation will be executed outside the context of the transaction. Transactions must be committed or aborted using `commitTransaction()` or `abortTransaction()`, respectively. Ending a session *aborts* all in-progress transactions.
+
+**Note**: Transactions only work with MongoDB replica sets (v4.0+) and sharded clusters (v4.2+).
+
+## Examples
+
+### Transaction that Atomically Moves a `Document` from One `MongoCollection` to Another
+
+The transaction below atomically deletes the document `{ "hello": "world" }` from the collection `test.src` and inserts the document in the collection `test.dest`. This ensures that the document exists in either `test.src` or `test.dest`, but not both or neither. Exectuting the delete and insert non-atomically raises the following issues:
```
patrickfreed, line 13: typo: "Exectuting"

Guides/Transactions.md
```diff
@@ -0,0 +1,87 @@
...
+    destColl.insertOne(docToMove, session: session)
+}.flatMap { _ in
+    session.commitTransaction()
+}.whenFailure { error in
+    session.abortTransaction()
+    // handle error
+}
+```
+
+### Transaction with Default Transaction Options
```
patrickfreed, line 39: This could use a slight blurb as well. Something simple that just says the options specified here will be applied to any transaction started on the session.

Guides/Transactions.md
JamesHeppenstall: Ah my bad haha 😂

Guides/Transactions.md
JamesHeppenstall: Done.

Guides/Transactions.md
patrickfreed: Filed [SWIFT-776](https://jira.mongodb.org/browse/SWIFT-776) to cover this work.

PR #446, opened 2020-04-07
SWIFT-749 use ELG provider, update shutdown API

kmahar:
This PR updates our client initializer to accept an `EventLoopGroupProvider` rather than an `EventLoopGroup`.
As a result I needed to update our shutdown API since it can no longer return a future. 

I took a lot of inspiration writing shutdown code from:
- [NIOThreadPool](https://github.com/apple/swift-nio/blob/e876fb37410e0036b98b5361bb18e6854739572b/Sources/NIO/NIOThreadPool.swift#L233)
- [MultiThreadedEventLoopGroup](https://github.com/apple/swift-nio/blob/2ad77f9a9b7e4e61bd9a9639473f716e835091d9/Sources/NIO/EventLoop.swift#L834)
- [AsyncHTTPClient](https://github.com/swift-server/async-http-client/blob/master/Sources/AsyncHTTPClient/HTTPClient.swift)

Comments inline

### DIFF
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -1,4 +1,5 @@
 import CLibMongoC
+import Foundation
 import NIOConcurrencyHelpers
 
 /// A connection to the database.
@@ -22,6 +23,14 @@ internal class Connection {
     }
 }
 
+extension NSCondition {
+    fileprivate func withLock<T>(_ body: () throws -> T) rethrows -> T {
+        self.lock()
+        defer { self.unlock() }
+        return try body()
+    }
+}
+
 /// A pool of one or more connections.
 internal class ConnectionPool {
     /// Represents the state of a `ConnectionPool`.
@@ -40,7 +49,7 @@ internal class ConnectionPool {
     /// The number of connections currently checked out of the pool.
     private var connCount = 0
     /// Lock over `state` and `connCount`.
-    private let stateLock = Lock()
+    private let stateLock = NSCondition()
 
     /// Internal helper for testing purposes that returns whether the pool is in the `closing` state.
     internal var isClosing: Bool {
@@ -103,23 +112,18 @@ internal class ConnectionPool {
             case .closing, .closed:
                 throw Self.PoolClosedError
             }
-        }
 
-        // continually loop and wait to get all connections back before destroying the pool. release the lock on each
-        // iteration to allow other methods to acquire the lock.
-        var done = false
-        while !done {
-            try self.stateLock.withLock {
-                if self.connCount == 0 {
-                    switch self.state {
-                    case .open, .closed:
-                        throw InternalError(message: "ConnectionPool in unexpected state \(self.state) during close()")
-                    case let .closing(pool):
-                        mongoc_client_pool_destroy(pool)
-                        self.state = .closed
-                    }
-                    done = true
-                }
+            while self.connCount > 0 {
+                // wait for signal from checkIn().
+                self.stateLock.wait()
+            }
+
+            switch self.state {
+            case .open, .closed:
+                throw InternalError(message: "ConnectionPool in unexpected state \(self.state) during close()")
+            case let .closing(pool):
+                mongoc_client_pool_destroy(pool)
+                self.state = .closed
             }
         }
     }
@@ -140,7 +144,8 @@ internal class ConnectionPool {
     }
 
     /// Checks out a connection from the pool, or returns `nil` if none are currently available. Throws an error if the
-    /// pool is not open.
+    /// pool is not open. This method may block waiting on the state lock as well as libmongoc's locks and thus must be
+    // run within the thread pool.
     internal func tryCheckOut() throws -> Connection? {
         try self.stateLock.withLock {
             switch self.state {
@@ -157,13 +162,16 @@ internal class ConnectionPool {
     }
 
     /// Checks a connection into the pool. Accepts the connection if the pool is still open or in the process of
-    /// closing; throws an error if the pool has already finished closing.
+    /// closing; throws an error if the pool has already finished closing. This method may block waiting on the state
+    /// lock as well as libmongoc's locks, and thus must be run within the thread pool.
     fileprivate func checkIn(_ connection: Connection) throws {
         try self.stateLock.withLock {
             switch self.state {
             case let .open(pool), let .closing(pool):
                 mongoc_client_pool_push(pool, connection.clientHandle)
                 self.connCount -= 1
+                // signal to close() that we are updating the count.
+                self.stateLock.signal()
             case .closed:
                 throw Self.PoolClosedError
             }
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -125,7 +125,8 @@ internal class ConnectionPool {
     }
 
     /// Checks out a connection. This connection will return itself to the pool when its reference count drops to 0.
-    /// This method will block until a connection is available.
+    /// This method will block until a connection is available. Throws an error if the pool is in the process of
+    /// closing or has finished closing.
     internal func checkOut() throws -> Connection {
         try self.stateLock.withLock {
             switch self.state {
@@ -138,7 +139,8 @@ internal class ConnectionPool {
         }
     }
 
-    /// Checks out a connection from the pool, or returns `nil` if none are currently available.
+    /// Checks out a connection from the pool, or returns `nil` if none are currently available. Throws an error if the
+    /// pool is not open.
     internal func tryCheckOut() throws -> Connection? {
         try self.stateLock.withLock {
             switch self.state {
@@ -154,6 +156,8 @@ internal class ConnectionPool {
         }
     }
 
+    /// Checks  a connection into the pool. Accepts the connection if the pool is still open or in the process of
+    /// closing; throws an error if the pool has already finished closing.
     fileprivate func checkIn(_ connection: Connection) throws {
         try self.stateLock.withLock {
             switch self.state {
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -257,7 +257,9 @@ internal class ConnectionPool {
     }
 
     /// Sets the provided APM callbacks on this pool, using the provided client as the "context" value. **This method
-    /// may only be called before any connections are checked out of the pool.**
+    /// may only be called before any connections are checked out of the pool.** Ideally this code would just live in
+    /// `ConnectionPool.init`. However, the client we accept here has to be fully initialized before we can pass it
+    /// as the context. In order for it to be fully initialized its pool must exist already.
     internal func setAPMCallbacks(callbacks: OpaquePointer, client: MongoClient) {
         self.stateLock.withLock {
             switch self.state {
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -113,7 +113,7 @@ internal class ConnectionPool {
                 if self.connCount == 0 {
                     switch self.state {
                     case .open, .closed:
-                        throw InternalError(message: "ConnectionPool in unexpected state during close()")
+                        throw InternalError(message: "ConnectionPool in unexpected state \(self.state) during close()")
                     case let .closing(pool):
                         mongoc_client_pool_destroy(pool)
                         self.state = .closed
@@ -218,7 +218,7 @@ internal class ConnectionPool {
                 mongoc_client_pool_set_ssl_opts(pool, &opts)
             case .closing, .closed:
                 // if we get here, we must have called this method outside of `ConnectionPool.init`.
-                fatalError("ConnectionPool unexpectedly in .closing or .closed state")
+                fatalError("ConnectionPool in unexpected state \(self.state) while setting TLS options")
             }
         }
     }
@@ -266,7 +266,7 @@ internal class ConnectionPool {
             case .closing, .closed:
                 // this method is called via `initializeMonitoring()`, which is called from `MongoClient.init`.
                 // unless we have a bug it's impossible that the pool is already closed.
-                fatalError("ConnectionPool unexpectedly in .closed state")
+                fatalError("ConnectionPool in unexpected state \(self.state) while setting APM callbacks")
             }
         }
     }
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -156,7 +156,7 @@ internal class ConnectionPool {
         }
     }
 
-    /// Checks  a connection into the pool. Accepts the connection if the pool is still open or in the process of
+    /// Checks a connection into the pool. Accepts the connection if the pool is still open or in the process of
     /// closing; throws an error if the pool has already finished closing.
     fileprivate func checkIn(_ connection: Connection) throws {
         try self.stateLock.withLock {
@@ -216,14 +216,13 @@ internal class ConnectionPool {
             opts.allow_invalid_hostname = invalidHosts
         }
 
-        self.stateLock.withLock {
-            switch self.state {
-            case let .open(pool):
-                mongoc_client_pool_set_ssl_opts(pool, &opts)
-            case .closing, .closed:
-                // if we get here, we must have called this method outside of `ConnectionPool.init`.
-                fatalError("ConnectionPool in unexpected state \(self.state) while setting TLS options")
-            }
+        // lock isn't needed as this is called before pool is in use.
+        switch self.state {
+        case let .open(pool):
+            mongoc_client_pool_set_ssl_opts(pool, &opts)
+        case .closing, .closed:
+            // if we get here, we must have called this method outside of `ConnectionPool.init`.
+            fatalError("ConnectionPool in unexpected state \(self.state) while setting TLS options")
         }
     }
 
@@ -265,15 +264,14 @@ internal class ConnectionPool {
     /// `ConnectionPool.init`. However, the client we accept here has to be fully initialized before we can pass it
     /// as the context. In order for it to be fully initialized its pool must exist already.
     internal func setAPMCallbacks(callbacks: OpaquePointer, client: MongoClient) {
-        self.stateLock.withLock {
-            switch self.state {
-            case let .open(pool):
-                mongoc_client_pool_set_apm_callbacks(pool, callbacks, Unmanaged.passUnretained(client).toOpaque())
-            case .closing, .closed:
-                // this method is called via `initializeMonitoring()`, which is called from `MongoClient.init`.
-                // unless we have a bug it's impossible that the pool is already closed.
-                fatalError("ConnectionPool in unexpected state \(self.state) while setting APM callbacks")
-            }
+        // lock isn't needed as this is called before pool is in use.
+        switch self.state {
+        case let .open(pool):
+            mongoc_client_pool_set_apm_callbacks(pool, callbacks, Unmanaged.passUnretained(client).toOpaque())
+        case .closing, .closed:
+            // this method is called via `initializeMonitoring()`, which is called from `MongoClient.init`.
+            // unless we have a bug it's impossible that the pool is already closed.
+            fatalError("ConnectionPool in unexpected state \(self.state) while setting APM callbacks")
         }
     }
 }
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -1,4 +1,5 @@
 import CLibMongoC
+import NIOConcurrencyHelpers
 
 /// A connection to the database.
 internal class Connection {
@@ -13,11 +14,10 @@ internal class Connection {
     }
 
     deinit {
-        switch self.pool.state {
-        case let .open(pool):
-            mongoc_client_pool_push(pool, self.clientHandle)
-        case .closed:
-            assertionFailure("ConnectionPool was already closed")
+        do {
+            try self.pool.checkIn(self)
+        } catch {
+            assertionFailure("Failed to check connection back in: \(error)")
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -56,29 +73,47 @@ internal class OperationExecutor {
...
+            return self.makeFailedFuture(err)
         }
 
         // closure containing the work to run in the thread pool: obtain a connection and execute the operation.
         let doOperation = { () -> ExecuteResult<T.OperationResult> in
-            // it's possible that the client was closed in between submitting this task and it being executed, so we
-            // check again here.
-            guard !client.isClosed else {
-                throw MongoClient.ClosedClientError
+            let connResult: Result<Connection?, Error> = client.stateLock.withLock {
```
kmahar, line 89: this change basically makes it so we only give you a connection if the client is still open. we were attempting to do this before, but without using the lock there could be an interleaving where the client is closed in between you checking if its closed and actually trying to check out a connection, which could lead to an error when you attempt to check out a connection.

we could just let the connection pool report that error, but it seems kind of racy/weird since it could lead to calls you make after calling `MongoClient.shutdown` going through if the call to shut down the connection pool hasn't been executed yet.

related to this, you can end up with some bad issues if you still have cursors out in the wild when shutdown is called.... the connections' memory leaks because they aren't in the pool at the time `destroy` is called, and when they try to check themselves back in they fail to do so.
we might consider calling `mongoc_client_destroy` directly on the orphaned connection when those cases occur which is not documented as a valid solution but I think would clean them up, but I would also say this is somewhat user error and will only really happen when your application is terminating anyway, so 🤷‍♀

Sources/MongoSwift/MongoClient.swift
```diff
@@ -168,8 +170,23 @@ public class MongoClient {
...
+        /// The client is running and has not been shut down.
+        case upAndRunning
+        /// The client is in the process of shutting down.
+        case shuttingDown
+        /// The client has finished shutting down.
+        case shutDown
+    }
+
+    /// Indicates the state of the client.
+    internal private(set) var state: State
```
kmahar, line 184: this + the lock is to provide a little more fine grained and accurate info about what's going in the client at any given point and to let us prevent more operations from getting queued up once shutdown has been called.

kmahar:
also all this definitely necessitates updating guides/examples but I figured I'd see what you all have to say on the API before I do that

Sources/MongoSwift/MongoClient.swift
```diff
@@ -252,41 +274,109 @@ public class MongoClient {
...
+        if let eventLoop = MultiThreadedEventLoopGroup.currentEventLoop {
+            preconditionFailure("""
+            BUG DETECTED: syncShutdown() must not be called when on an EventLoop.
+            Calling syncShutdown() on any EventLoop can lead to deadlocks.
+            Current eventLoop: \(eventLoop)
+            """)
+        }
+        let errorStorageLock = Lock()
+        var errorStorage: Error?
+        let continuation = DispatchWorkItem {}
```
patrickfreed, line 365: This `DispatchWorkItem` seems like it could have been useful for the cursor stuff. The way it's being used here is very similar to a channel which I remember we wanted. Looking at the cursor code now, I can't really remember why exactly I thought a channel was necessary though.

Sources/MongoSwift/Operations/Operation.swift
patrickfreed: Hm yeah, that does seem like a bit of an issue. Maybe we should just document the possibility of a leak for now until we come up with something better. Given client's aren't thread safe, I can't really think of anything though. One way we can combat it is to check for the number of references to the `MongoClient`, and if it's greater than 1 just return an error. Not sure if that would be frustrating for users or not.

Sources/MongoSwiftSync/MongoClient.swift
```diff
@@ -43,7 +43,11 @@ public class MongoClient {
     public init(_ connectionString: String = "mongodb://localhost:27017", options: ClientOptions? = nil) throws {
         let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 5)
         do {
-            self.asyncClient = try MongoSwift.MongoClient(connectionString, using: eventLoopGroup, options: options)
+            self.asyncClient = try MongoSwift.MongoClient(
+                connectionString,
+                eventLoopGroupProvider: .shared(eventLoopGroup),
```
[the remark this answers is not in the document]
patrickfreed: I think it's probably more straightforward to use the `.createNew` behavior here and set the default number of threads to 5. I do have to mention that this opinion is a bit contentious though (we're discussing what to set this number to in Rust right now). Number of threads aside, another key difference is that in the `.shared` case, we have direct access to the async runtime in the sync driver if we ever need it. I'm not sure why we would, but it's something to consider I guess.

Sources/MongoSwift/Operations/Operation.swift
mbroadst: I don't consider that a leak. It's a positive aspect of our connection pool design that connections are allowed to live on after a pool is closed (in fact this is a feature depended on by the new "quiesce mode" in replication). The connections will be closed/destroyed when they are eventually returned to the pool, either because the operation completed or the cursor was explicitly closed. If a cursor is not exhausted or explicitly closed after a client/pool is closed then I would consider that user error

Sources/MongoSwiftSync/MongoClient.swift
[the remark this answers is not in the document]
mbroadst: I'm fine with this decision today, let's wait and see if anybody asks for more control

Sources/MongoSwift/Operations/Operation.swift
patrickfreed: I think in this case it actually is a leak because libmongoc considers destroying clients derived from a pool after the pool is closed to be UB. I may be confusing this with something else though.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -252,41 +274,109 @@ public class MongoClient {
...
+     *
+     * - Parameters:
+     *   - queue: a `DispatchQueue` on which to execute the provided callback, defaulting to the global queue.
+     *   - callback: A callback to handle any errors that occur in the process of shutdown.
+     *
+     * If the error provided to the callback is non-nil, it is likely one of the following:
+     * - `LogicError` if `syncShutdown` or `shutdown` has already been called on this client.
+     */
+    public func shutdown(queue: DispatchQueue = .global(), _ callback: @escaping (Error?) -> Void) {
+        let error: Error? = self.stateLock.withLock {
```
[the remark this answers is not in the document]
patrickfreed: I think a deadlock would be possible if the elg was single threaded. If two concurrent tasks happened to call `shutdown` simultaneously and the context switched while one had the lock, blocking here in the thread without the lock would deadlock. Also, even if deadlock doesn't occur, we'd be blocking an entire event loop group thread rather than relying on its cooperative scheduling functionality.

Sources/MongoSwift/Operations/Operation.swift
kmahar: as it stands, libmongoc's pool as no in-between state where it is closed but not destroyed. if you destroy a pool and there are still checked-out clients, they all become non-functional, because destroying the pool shuts down SDAM. 
consequently, our pool is the same way --`Connection`s have to be checked in before `ConnectionPool.close()` is called.

possibly, we could implement such behavior where connections can live on ourselves, adding some intermediate state where our pool is considered closed and won't let you check out new connections, but the underlying client pool has not yet been destroyed.

then the question becomes: when *do* we call `mongoc_client_pool_destroy`? that method is tricky and cannot simply be called in `deinit` because it needs to be run via the executor, as it does IO (sends `endSessions`.)
if we could come up with a way to have `ConnectionPool.close()` return a future that resolves once all the connections are returned to the pool and the pool has been destroyed, and then shut down the executor once that future has resolved I think it could work...

Sources/MongoSwift/MongoClient.swift
```diff
@@ -252,41 +274,109 @@ public class MongoClient {
...
         self.commandEventHandlers = []
+        self.state = .upAndRunning
+
         self.connectionPool.initializeMonitoring(client: self)
     }
 
     deinit {
-        assert(self.isClosed, "MongoClient was not closed before deinitialization")
+        assert(
+            self.state == .shutDown,
```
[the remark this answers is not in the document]
kmahar: since `deinit` is called when there are 0 references left to the client, I don't believe it's possible for there to be any races around access to this

Sources/MongoSwift/MongoClient.swift
```diff
@@ -168,8 +170,23 @@ public class MongoClient {
     /// Default maximum size for connection pools created by this client.
     internal static let defaultMaxConnectionPoolSize = 100
 
-    /// Indicates whether this client has been closed.
-    internal private(set) var isClosed = false
+    /// Indicates the state of this `MongoClient`.
+    internal enum State {
```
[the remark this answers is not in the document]
patrickfreed: I don't really have strong feelings on this one. I think you summed the situation up well, and its just to decide whether we want consistency across drivers or across ecosystem members. 

As a related topic: do we want to consider the naming of `MongoCursor`'s `isAlive`/`kill` names in these terms? From brief examination, it looks like most drivers use `isOpen`/`isClosed`/`close` instead (Java, Go, Ruby were the ones I looked at).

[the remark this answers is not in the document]
kmahar: for the sake of moving forward on this, it seems like @mbroadst feels the most strongly so I say we go back to open/closed naming for the client.

regarding cursors, I believe I copied the `alive` thing from pymongo`. they use `close` for the closing method.  mongoc uses `mongoc_cursor_more` and `mongoc_cursor_destroy`. 
node also uses `close`, and looks like they have an `end` event that gets emitted so I think that's how you can tell if there's potential for more data? possibly we can defer a decision on this to next week when I'd like to work out the potential API changes to line up with Rust (I opened [SWIFT-767](https://jira.mongodb.org/browse/SWIFT-767) about that)

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
kmahar: hmm... I'm not sure it's worth handling what happens when people call shutdown multiple times; we've explicitly stated you should call it once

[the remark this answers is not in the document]
patrickfreed: It's not restricted to just that case though. If any other thread holds the lock the same thing could occur. e.g. if an operation grabs it to check the state, `shutdown` is simultaneously called, and then the context switches to the `shutdown` thread. 

Regardless of all that though, I think the biggest problem is that blocking an event loop thread means it can't be used to run the rest of the things that are sharing the group.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
kmahar: went back to `close` over `shutdown` for the client and connection pool.

I did stick with `shutdown` for the executor because in that case it directly translates to calling 1 or possibly 2 `shutdown` methods on the thread pool and possibly ELG.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
kmahar: maybe I'm just confused about how event loops work, but how would the context switch in the middle of executing a single task?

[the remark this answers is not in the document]
patrickfreed: I think you're right, that doesn't seem to be possible. An event loop processes each task to completion before going on to the next one. I was thinking in an async/await mindset where the context switches on `await` (I think that's how this works), but that doesn't apply in Swift. The only scenario I can imagine that deadlocks is if we acquire the lock in one context and release it in the callback of another future, which we're not doing here.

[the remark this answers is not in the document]
mbroadst: I don't think we have an issue here. If two concurrent tasks called `shutdown` simultaneously, one would have to wait for the lock. The lock is incredibly short-lived, and its not calling to any code which might reenter this method (which would indeed be a deadlock). I propose we move on here.

Sources/MongoSwift/Operations/Operation.swift
mbroadst: Ah I see, I misunderstood libmongoc's pooling behavior. But still, this is the lock that I am most concerned about. You acquire the lock, and then do a _whole bunch_ of stuff, including potentially reentering this method (an operation can end up calling something that calls another operation, etc).

I think at very least what we want to do reduce the scope of the critical section as much as possible. It seems like you want to be able to "fail fast" if the client is closed, so you should be able to release the lock as soon as you check state, correct?

kmahar: all that happens in the body of `withLock` is:
```swift
guard client.state == .open else {
    return .failure(MongoClient.ClosedClientError)
}

do {
    // select a connection in following order of priority:
    // 1. connection specifically provided for use with this operation
    // 2. if a session was provided, use its underlying connection
    // 3. a new connection from the pool, if available
    return .success(
        try connection ??
        session?.getConnection(forUseWith: client) ??
        client.connectionPool.tryCheckOut()
    )
} catch {
    return .failure(error)
}
```
basically it's the state check, and then also we try to figure out a connection to use / if one is available. my reasoning for doing the connection part with the lock is without it, you could have the following interleaving:
1. we grab the lock here, see client is open, and then release it
2. user calls `MongoClient.close`, client state transitions to `.closing`
3. we let you check out a new connection here, as the operation to close the connection pool hasn't been executed yet, even though we've already marked the client as `.closing`

mbroadst: It's incredibly hard to see this with the large comment thread there, but I don't think that's right. From what I can [see](https://github.com/mongodb/mongo-swift-driver/blob/e85695953cb567ff4498954f7b0b3815a01a7266/Sources/MongoSwift/Operations/Operation.swift#L89-L107) it looks like the `withLock` statement contains the `do` statement. What you put in your comment above this one looks correct to me.

kmahar: you don't think your previous statement is right, or that what I said is right? lol 

the do/catch is just the connection logic

mbroadst: OK, I see now that the functional change here is that you are putting the lock around just the client state check and connection checkout. 

What's weird about this is that there are paths here where you don't actually try to check out a connection from the pool. How are you handling those cases? (connection is explicitly passed it, or bound to a session) The lock here has a smell.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: You're right that deadlocks are not really an issue here, but again my main concern is blocking an event loop that we're sharing with the rest of the user's application. The lock is not always short-lived--remember shutdown isn't the only thing acquiring this lock. Every operation acquires it when it checks out a connection, and if the pool is depleted, the lock may be held for a while while the pending operation waits for connection to become available. While `shutdown` blocks on this lock, nothing on that event loop can progress. SwiftNIO's docs repeatedly mentions the importance of not blocking in event loop threads for this reason ([\[1\]](https://github.com/apple/swift-nio#a-note-on-blocking), [\[2\]](https://apple.github.io/swift-nio/docs/current/NIO/Protocols/EventLoop.html)). It would be okay to block in here if we knew the call was never made from an event loop thread, but that's what `syncShutdown` is for. In fact, the main distinction between `syncShutdown` and `shutdown` is that one is blocking and the other is non-blocking. If they both block, then I don't really see the purpose of having two methods.

 I strongly think we should not move on here.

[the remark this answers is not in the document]
kmahar: we call `tryCheckout()` to get a new connection, which returns nil immediately if one isn't available, and then the lock is released, so blocking until a connection is available is not an issue.

Sources/MongoSwift/Operations/Operation.swift
kmahar: so I'm realizing we should probably just get rid of the explicit connection thing. it isn't used anywhere in the driver anymore.
we used to use it for `NextOperation` just to avoid checking out an extra connection that would go unused (we would pass in the cursor's underlying connection instead), but `NextOperation` still didn't even explicitly use the passed-in cursor connection as the `mongoc_cursor_t` keeps a reference to its source `mongoc_client_t`. we ended up getting rid of `NextOperation` altogether and now we just run that code via the `Operation.execute` block that takes in a closure and runs it via thread pool.
(this does make me realize that, since that `execute` method bypasses the check for the client being open, methods you call on a cursor created before closing the client are allowed to execute after you've called `MongoClient.close()`. so maybe we need to introduce a check on that path as well...)

going back to your original point, getting rid of the explicit connection case still leaves the session one. can you elaborate on what your concerns are there?
if you think it would be clearer we could move the call to `session.getConnection` outside of the lock code and only call `tryCheckout` in `withLock` if the session is nil.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: If you look at the [implementation of `mongoc_client_pool_try_pop`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-client-pool.c#L278), you can see that it still isn't immediate, since it blocks on the lock on the pool itself and also creates new clients if the pool is < maxPoolSize, which may block on the topology lock.

Sources/MongoSwift/Operations/Operation.swift
mbroadst: wrt explicit connections: SGTM.
wrt locking: I just want to make sure we aren't complicating our codebase to match the behavior of libmongoc, where we can avoid it. In a future where we don't use libmongoc, we would not want to lock here, and arguably the lock here is guiding behavior of the pool (as you note) so that is most likely better deferred to the pool. As it stands, it might not make a huge difference so I'm fine approving this as-is.

kmahar:
alright I've made significant changes based on our conversations, will make some comments inline

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -97,4 +97,56 @@ final class MongoClientTests: MongoSwiftTestCase {
             _ = try waitingOperations.map { try $0.wait() }
         }
     }
+
+    func testConnectionPoolClose() throws {
```
kmahar, line 101: I think this test demonstrates well the behavior I was talking about in triage today where cursors still work after calling `close()`.

the reason we get this behavior for cursors is that all of their API methods are currently run via the `OperationExecutor.execute` variant that just takes in a closure and nothing else. That method doesn't have access to the client and thus does not check if the client is open.

OTOH a "normal" API method / any that tries to get a new connection goes through the method that takes in a client too and checks that the client is still open.

`ClientSession.end` uses the no-validation method which is why that still works after closing. But other methods accepting sessions go through the validation route so they will error.

we might consider adding a way to optionally pass in a client to validate the state of to the non-validating method but maybe it doesn't matter that much.

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -67,38 +93,76 @@ internal class ConnectionPool {
...
+                self.connCount += 1
+                return Connection(clientHandle: mongoc_client_pool_pop(pool), pool: self)
+            case .closing, .closed:
+                throw Self.PoolClosedError
+            }
         }
     }
 
     /// Checks out a connection from the pool, or returns `nil` if none are currently available.
     internal func tryCheckOut() throws -> Connection? {
```
nbbeeken, line 142: nit: `try` implies a nil returning function but this also throws, might be worth adding that to the doc string. I know its in the declaration but I feel like its just slightly more future dev friendly.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -167,8 +169,10 @@ public class MongoClient {
     /// Default maximum size for connection pools created by this client.
     internal static let defaultMaxConnectionPoolSize = 100
 
-    /// Indicates whether this client has been closed.
-    internal private(set) var isClosed = false
+    /// Indicates whether this client has been closed. We don't need a lock because:
```
nbbeeken, line 172: Is this wording intentional for public facing docs? could be confusing, the driver doesn't need a lock or the user doesn't need a lock.

kmahar: this is a private variable, so it's just for our own reference. that said I've opened it to say "a lock around this variable is not needed" so there's less confusion around who "we" is

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -67,38 +93,80 @@ internal class ConnectionPool {
...
+                self.state = .closing(pool: pool)
+            case .closing, .closed:
+                throw Self.PoolClosedError
+            }
+        }
+
+        // continually loop and wait to get all connections back before destroying the pool. release the lock on each
+        // iteration to allow other methods to acquire the lock.
+        var done = false
+        while !done {
```
patrickfreed, line 111: Rather than using what is essentially a spin lock, it may be better to use something that can be signaled like a [condition variable](https://developer.apple.com/documentation/foundation/nscondition):
```swift
self.stateLock.lock()
while self.connCount > 0 {
    self.stateLock.wait() // waiting on a condition variable unlocks it
}
defer { self.stateLock.unlock() } // once the condition is met we hold the lock again
switch self.state {
    // do stuff
}
```
And in `checkIn:`
```swift
self.stateLock.lock()
defer { self.stateLock.unlock() }
...
self.connCount -= 1
self.stateLock.signal() // signal that the condition has changed, waking up any blocked `close`s
...
```
There's also [`NSConditionLock`](https://developer.apple.com/documentation/foundation/nsconditionlock) which is similar, but I don't think it would work for our case because we don't want a condition to be required for something to acquire the lock anywhere outside of `close`.

kmahar: maybe I'm misunderstanding your proposal, but I think we end up with a deadlock situation here:
```swift
self.stateLock.lock()
while self.connCount > 0 {
    self.stateLock.wait() // waiting on a condition variable unlocks it
}
```
we hold the lock and are waiting for `connCount` to drop to 0, but `close` needs to get the lock itself in order to update `connCount`.

kmahar: aha, I missed this from the docs:
> When a thread waits on a condition, the condition object unlocks its lock and blocks the thread. When the condition is signaled, the system wakes up the thread. The condition object then reacquires its lock before returning from the wait() or wait(until:) method. Thus, from the point of view of the thread, it is as if it always held the lock.

patrickfreed: `wait` unlocks the lock until it's woken up. The condition is then checked and if it doesn't pass, the `wait` is entered again and the lock is released again. If the condition passes, the lock will be held until the `defer` executes.

From the docs for `NSCondition`:
>When a thread waits on a condition, the condition object unlocks its lock and blocks the thread. When the condition is signaled, the system wakes up the thread. The condition object then reacquires its lock before returning from the wait() or wait(until:) method. Thus, from the point of view of the thread, it is as if it always held the lock.

With a normal lock, this is hard to implement without doing busy waiting like we're doing here or sleeping + periodic checks.

kmahar: jinx :)

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -67,38 +93,80 @@ internal class ConnectionPool {
...
+                return Connection(clientHandle: handle, pool: self)
+            case .closing, .closed:
+                throw Self.PoolClosedError
+            }
+        }
+    }
+
+    /// Checks  a connection into the pool. Accepts the connection if the pool is still open or in the process of
+    /// closing; throws an error if the pool has already finished closing.
+    fileprivate func checkIn(_ connection: Connection) throws {
```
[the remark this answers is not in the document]
patrickfreed: In the success case, there wouldn't really be any behavioral difference. The main benefit there would just be that the blocking code is done more explicitly, so there isn't a possibility we do it by accident. 

In the error case (i.e. when the user forgets to call `kill` or `end`), the `deinit` approach prevents a memory leak but may block the event loop, whereas if we only explicitly check in connections, the user would get a memory leak but not stall the event loop. It's not really a big deal either way though, since ideally users will be closing their resources.

Anyways, this is pretty minor so either way is fine. I figured it was worth consideration though.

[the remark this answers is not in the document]
kmahar: ah yeah. I don't really know which bad outcome is worse. I've opened [SWIFT-825](https://jira.mongodb.org/browse/SWIFT-825) so we can discuss in triage.

PR #447, opened 2020-04-10
SWIFT-574 Implement asynchronous convenient transactions API

JamesHeppenstall:
I've got a working implementation of the async convenient transactions API. I'm not quite sure what the best way to implement the synchronous API would be, given that you cannot just call `wait()`. I suppose that could be left to another PR.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -426,11 +426,23 @@ public final class ClientSession {
         options: TransactionOptions?,
         _ transactionBody: @escaping (ClientSession) throws -> EventLoopFuture<T>
     ) -> EventLoopFuture<T> {
-        self.startTransaction(options: options).flatMap { _ in
+        let maybeRetryOrFail = { (error: Error) -> EventLoopFuture<T> in
+            if let labeledError = error as? LabeledError, let errorLabels = labeledError.errorLabels,
+                retryTimeoutTime.timeIntervalSinceNow > 0 && errorLabels.contains("TransientTransactionError") {
+                return self.attemptTransaction(
+                    retryTimeoutTime: retryTimeoutTime,
+                    options: options,
+                    transactionBody
+                )
+            }
+            return self.client.operationExecutor.makeFailedFuture(error)
+        }
+
+        return self.startTransaction(options: options).flatMap { _ in
             do {
                 let transactionBodyFuture = try transactionBody(self)
                 return transactionBodyFuture.flatMap { value in
-                    if [.none, .committed, .aborted].contains(self.transactionState ?? .none) {
+                    if let state = self.transactionState, [.none, .committed, .aborted].contains(state) {
                         return self.client.operationExecutor.makeSucceededFuture(value)
                     }
                     return self.attemptTransactionCommit(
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -373,4 +373,146 @@ public final class ClientSession {
...
+                    if [.none, .committed, .aborted].contains(self.transactionState ?? .none) {
+                        return self.client.operationExecutor.makeSucceededFuture(value)
+                    }
+                    return self.attemptTransactionCommit(
+                        value: value,
+                        retryTimeoutTime: retryTimeoutTime,
+                        options: options,
+                        transactionBody
+                    )
+                }.flatMapError { error in
```
kmahar, line 442: I think the code in this block would be a little clearer and less repetitive if you encapsulate the call to `maybeRetryOrFail` in a closure. I did something similar a couple of times in `executeOperation` with  `doOperation` and `resubmitIfNeeded` to make constructing future chains with repetitive code a bit simpler.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -373,4 +373,146 @@ public final class ClientSession {
...
+                    if [.starting, .inProgress].contains(self.transactionState ?? .none) {
```
kmahar, line 443: If it compiles (I think it will....) I think this line would read a little clearer like
```swift
if let state = self.transactionState, [.starting, .inProgress].contains(state) {
```

Sources/MongoSwift/MongoError.swift
```diff
@@ -481,4 +481,28 @@ internal func wrongIterTypeError(_ iter: DocumentIterator, expected type: BSONVa
...
+    if let commandError = error as? CommandError, commandError.code == maxTimeMSExpiredErrorCode {
+        return true
+    } else if let writeError = error as? WriteError {
+        if writeError.writeFailure?.code == maxTimeMSExpiredErrorCode ||
+            writeError.writeConcernFailure?.code == maxTimeMSExpiredErrorCode {
+            return true
+        }
+    } else if let bulkWriteError = error as? BulkWriteError {
+        if let writeFailures = bulkWriteError.writeFailures {
+            for writeFailure in writeFailures where writeFailure.code == maxTimeMSExpiredErrorCode {
```
kmahar, line 496: this could be written a bit more succinctly like
```swift
return writeFailures.contains { $0.code == maxTimeMSExpiredErrorCode }
```

Sources/MongoSwift/ClientSession.swift
JamesHeppenstall: Done.

Sources/MongoSwift/MongoError.swift
JamesHeppenstall: Doesn't this mean that you would return `false` if no `writeFailures` match the code, skipping the check below for `writeConcernFailure`?

kmahar: oh whoops, good point :)

Sources/MongoSwift/ClientSession.swift
JamesHeppenstall: Ok, I've done what you suggested 👍

patrickfreed:
Both the synchronous and async convenient API will be implemented together as part of follow-on work.

PR #449, opened 2020-04-13
feat: Add allowDiskUse find option

nbbeeken:
[SWIFT-772](https://jira.mongodb.org/browse/SWIFT-772)
[SPEC](https://github.com/mongodb/specifications/commit/92255e0e57027646e8ef23399d3f33fdc114221f)

Added the option `allowDiskUse` and tests that mirror the CRUDv2 tests. I figured a new file was prudent given the any tests needed here can be dropped in the future when the CRUDv2 runner is implemented.

### DIFF
--- a/Sources/MongoSwift/Operations/AggregateOperation.swift
+++ b/Sources/MongoSwift/Operations/AggregateOperation.swift
@@ -3,7 +3,7 @@ import Foundation
 
 /// Options to use when executing an `aggregate` command on a `MongoCollection`.
 public struct AggregateOptions: Codable {
-    /// Enables the server to write to temporary files. When set to true, the find operation
+    /// Enables the server to write to temporary files. When set to true, the aggregate operation
     /// can write data to the _tmp subdirectory in the dbPath directory.
     public var allowDiskUse: Bool?
 
--- a/Sources/MongoSwift/Operations/AggregateOperation.swift
+++ b/Sources/MongoSwift/Operations/AggregateOperation.swift
@@ -3,7 +3,7 @@ import Foundation
 
 /// Options to use when executing an `aggregate` command on a `MongoCollection`.
 public struct AggregateOptions: Codable {
-    /// Enables writing to temporary files. When set to true, aggregation stages
+    /// Enables the server to write to temporary files. When set to true, the find operation
     /// can write data to the _tmp subdirectory in the dbPath directory.
     public var allowDiskUse: Bool?
 
--- a/Sources/MongoSwift/Operations/FindOperation.swift
+++ b/Sources/MongoSwift/Operations/FindOperation.swift
@@ -32,6 +32,9 @@ public enum CursorType {
 
 /// Options to use when executing a `find` command on a `MongoCollection`.
 public struct FindOptions: Codable {
+    /// Enables writing to temporary files on the server.
+    public var allowDiskUse: Bool?
+
     /// Get partial results from a mongos if some shards are down (instead of throwing an error).
     public var allowPartialResults: Bool?
 
@@ -130,6 +133,7 @@ public struct FindOptions: Codable {
 
     /// Convenience initializer allowing any/all parameters to be omitted or optional.
     public init(
+        allowDiskUse: Bool? = nil,
         allowPartialResults: Bool? = nil,
         batchSize: Int32? = nil,
         collation: Document? = nil,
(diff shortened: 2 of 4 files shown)


Tests/MongoSwiftSyncTests/CrudV2Tests.swift
```diff
@@ -0,0 +1,70 @@
...
+/// A place for CrudV2 Tests until the swift crud v2 runner is shipped
+final class MongoCrudV2Tests: MongoSwiftTestCase {
+    func testFindOptionsAllowDiskUseNotSpecified() throws {
+        let client = try MongoClient.makeTestClient()
+        let monitor = client.addCommandMonitor()
+
+        let db = client.db(Self.testDatabase)
+
+        let collection = db.collection("collection")
+        try collection.insertOne(["test": "blahblah"])
```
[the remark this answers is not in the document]
nbbeeken: I consolidated them but now command monitoring captures zero events, I pushed the changes, I'm investigating let me know if anything sticks out.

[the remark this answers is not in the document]
kmahar: So, `withTestNamespace` also sets up a new test client and db (those are currently being discarded by the `_, _`). 

It looks like you're enabling command monitoring on the client defined outside of that closure, and not the one that was used to create the collection you're doing `find`s on.

so you'll want to do something like
```swift
self.withTestNamespace { client, _, coll in 
    let monitor = client.addCommandMonitor()
   // ...
}
```

Tests/LinuxMain.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.17.0 — https://github.com/krzysztofzablocki/Sourcery
```
kmahar, line 1: so, this will fail CI right now since the version used on there is 0.16.1. for now let's just manually edit this string back to 0.1.6.1.
when we do [SWIFT-592](https://jira.mongodb.org/browse/SWIFT-592) and move all the tasks over to Evergreen we can bump the Sourcery version.

Sources/MongoSwift/Operations/FindOperation.swift
```diff
@@ -32,6 +32,9 @@ public enum CursorType {
 
 /// Options to use when executing a `find` command on a `MongoCollection`.
 public struct FindOptions: Codable {
+    /// Enables writing to temporary files on the server.
```
kmahar, line 35: I swear I posted this comment in my first review but I guess not.... I think the second sentence from the CRUD spec about this option would also be useful to tack on here as it clarifies the temp files are just for the duration of the find.
>  When set to true, the server can write temporary data to disk while executing the find operation.

Tests/MongoSwiftSyncTests/CrudV2Tests.swift
```diff
@@ -0,0 +1,47 @@
...
+
+                let optionAllowDiskUseTrue = FindOptions(allowDiskUse: true)
+                expect(try coll.find(["dog": "notCat"], options: optionAllowDiskUseTrue)).toNot(throwError())
+            }
+
+            let events = monitor.commandStartedEvents()
+            expect(events).to(haveCount(3))
+
+            let eventAllowDiskUseNil = events[0]
+            expect(eventAllowDiskUseNil).toNot(beNil())
```
kmahar, line 32: I think we can get rid of the the assertions that the events are non-nil. the return type of `commandStartedEvents()` is an array of non-optionals, so it's impossible that we ever end up with nil values for these variables.

kmahar:
lgtm again assuming evergreen is happy 🌲

Sources/MongoSwift/Operations/FindOperation.swift
```diff
@@ -32,6 +32,10 @@ public enum CursorType {
 
 /// Options to use when executing a `find` command on a `MongoCollection`.
 public struct FindOptions: Codable {
+    /// Enables writing to temporary files. When set to true, the find operation
```
patrickfreed, line 35: We may want to clarify that this option allows the _server_ to write to temporary files.

nbbeeken: I added the clarification to the aggregation option as well.
Also hopefully I finally got my commit hook right and that'll be the last lint fix commit I gotta make. 😅

Sources/MongoSwift/Operations/AggregateOperation.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 /// Options to use when executing an `aggregate` command on a `MongoCollection`.
 public struct AggregateOptions: Codable {
-    /// Enables writing to temporary files. When set to true, aggregation stages
+    /// Enables the server to write to temporary files. When set to true, the find operation
```
patrickfreed, line 6: find -> aggregate

PR #450, opened 2020-04-15
SWIFT-773: add authorizedDatabases optional option

nbbeeken:
Added the `authorizedDatabases` option to ListDatabasesOperation as an optional param. No changes needed to the helpers.
[Check out the flag docs here](https://docs.mongodb.com/manual/reference/command/listDatabases/#dbcmd.listDatabases)

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -358,10 +358,10 @@ public class MongoClient {
      */
     public func listDatabases(
         _ filter: Document? = nil,
-        _ options: ListDatabasesOptions = ListDatabasesOptions(),
+        options: ListDatabasesOptions? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<[DatabaseSpecification]> {
-        let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: false, options: options)
+        let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: nil, options: options)
         return self.operationExecutor.execute(operation, client: self, session: session).flatMapThrowing { result in
             guard case let .specs(dbs) = result else {
                 throw InternalError(message: "Invalid result")
@@ -388,10 +388,10 @@ public class MongoClient {
      */
     public func listMongoDatabases(
         _ filter: Document? = nil,
-        _ options: ListDatabasesOptions = ListDatabasesOptions(),
+        options: ListDatabasesOptions? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<[MongoDatabase]> {
-        self.listDatabaseNames(filter, options, session: session).map { $0.map { self.db($0) } }
+        self.listDatabaseNames(filter, options: options, session: session).map { $0.map { self.db($0) } }
     }
 
     /**
@@ -412,7 +412,7 @@ public class MongoClient {
      */
     public func listDatabaseNames(
         _ filter: Document? = nil,
-        _ options: ListDatabasesOptions = ListDatabasesOptions(),
+        options: ListDatabasesOptions? = nil,
         session: ClientSession? = nil
     ) -> EventLoopFuture<[String]> {
         let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: true, options: options)
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -343,6 +343,7 @@ public class MongoClient {
      *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
      *     on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
      *   - session: Optional `ClientSession` to use when executing this command.
+     *   - authorizedDatabases: Optional `Bool` disabling will have mongodb return an error if user in unauthorized to list databases.
      *
      * - Returns:
      *    An `EventLoopFuture<[DatabaseSpecification]>`. On success, the future contains an array of the specifications
@@ -357,9 +358,10 @@ public class MongoClient {
      */
     public func listDatabases(
         _ filter: Document? = nil,
-        session: ClientSession? = nil
+        session: ClientSession? = nil,
+        authorizedDatabases: Bool? = nil
     ) -> EventLoopFuture<[DatabaseSpecification]> {
-        let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: nil)
+        let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: nil, authorizedDatabases: authorizedDatabases)
         return self.operationExecutor.execute(operation, client: self, session: session).flatMapThrowing { result in
             guard case let .specs(dbs) = result else {
                 throw InternalError(message: "Invalid result")
@@ -374,6 +376,7 @@ public class MongoClient {
      * - Parameters:
      *   - filter: Optional `Document` specifying a filter on the names of the returned databases.
      *   - session: Optional `ClientSession` to use when executing this command
+     *   - authorizedDatabases: Optional `Bool` disabling will have mongodb return an error if user in unauthorized to list databases.
      *
      * - Returns:
      *    An `EventLoopFuture<[MongoDatabase]>`. On success, the future contains an array of `MongoDatabase`s that
@@ -385,9 +388,10 @@ public class MongoClient {
      */
     public func listMongoDatabases(
         _ filter: Document? = nil,
-        session: ClientSession? = nil
+        session: ClientSession? = nil,
+        authorizedDatabases: Bool? = nil
     ) -> EventLoopFuture<[MongoDatabase]> {
-        self.listDatabaseNames(filter, session: session).map { $0.map { self.db($0) } }
+        self.listDatabaseNames(filter, session: session, authorizedDatabases: authorizedDatabases).map { $0.map { self.db($0) } }
     }
 
     /**
@@ -396,6 +400,7 @@ public class MongoClient {
      * - Parameters:
      *   - filter: Optional `Document` specifying a filter on the names of the returned databases.
      *   - session: Optional `ClientSession` to use when executing this command
+     *   - authorizedDatabases: Optional `Bool` disabling will have mongodb return an error if user in unauthorized to list databases.
      *
      * - Returns:
      *    An `EventLoopFuture<[String]>`. On success, the future contains an array of names of databases that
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -343,6 +343,7 @@ public class MongoClient {
      *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
      *     on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
      *   - session: Optional `ClientSession` to use when executing this command.
+     *   - authorizedDatabases: Optional `Bool` disabling will have mongodb return an error if user in unauthorized to list databases.
```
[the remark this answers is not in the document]
patrickfreed: I think we should definitely put it in an options struct, both for forward compatibility and consistency with the rest of the driver. `filter` can stay as-is for the reasons you mention.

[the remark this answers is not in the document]
nbbeeken: Should `nameOnly` be in the options object too? and if so then I feel like we drop `listDatabaseNames()` since we'll be taking in an options object from the user:
```swift
/// Options for "listDatabases" operations
public struct ListDatabasesOperationOptions {
    /// Specifies whether to only return databases for which the user has privileges.
    public let authorizedDatabases: Bool?
    /// Specifies whether to only return the list of database names
    public let nameOnly: Bool?
}
```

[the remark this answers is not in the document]
kmahar: That is a good question, which we spent a while discussing back when I think @patrickfreed initially implemented the database enumeration spec.

I think we should continue to keep `nameOnly` hidden from the user and just expose it via the separate API methods.

IMO the biggest advantage of the current approach is that, since we use separate methods to accomplish nameOnly vs not-nameOnly, we can use different return types for them -- `listDatabaseNames` returns an `[String]` while `listDatabases` returns a `DatabaseSpecification`.

Were we to combine these into one method, we'd have to return the same type in both situations. This type would have to be more flexible to support both cases, and would cause us to lose specificity / information about what information should be present when. It would probably look like `DatabaseSpecification` but have all fields except `name` be optionals. (Alternatively we could make the `ListDatabasesResults` enum public.)
Consider a user who doesn't pass in `nameOnly` to the method. We can guarantee that we will have the `sizeOnDisk` field available in the object we return when they set that to true, yet it would somewhat confusingly to them be an optional.
Alternatively, if we returned a `ListDatabasesResults`, they would have to switch on the returned value and handle the case where `.names` is returned, even though we know that's actually impossible.
Splitting up the methods allows us to be more precise in both cases about the present information.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -357,9 +358,10 @@ public class MongoClient {
      */
     public func listDatabases(
         _ filter: Document? = nil,
+        _ options: ListDatabasesOptions = ListDatabasesOptions(),
```
kmahar, line 361: let's use a default value of `nil` here and below, for consistency with the rest of our API.

also, let's require the parameter label here and below. right now the underscore in `_ options` says users don't need to specify it, but in the rest of our API we do require the label.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -357,9 +358,10 @@ public class MongoClient {
...
         session: ClientSession? = nil
     ) -> EventLoopFuture<[DatabaseSpecification]> {
-        let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: nil)
+        let operation = ListDatabasesOperation(client: self, filter: filter, nameOnly: false, options: options)
```
kmahar, line 364: I think `nameOnly` can stay `nil` here. the idea behind that is we just don't send the flag at all when we don't want `nameOnly`.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -17,28 +17,43 @@ public struct DatabaseSpecification: Codable {
 }
 
 /// Internal intermediate result of a ListDatabases command.
-internal enum ListDatabasesResults {
+public enum ListDatabasesResults {
```
kmahar, line 20: I think this can stay `internal`?

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -17,28 +17,43 @@ public struct DatabaseSpecification: Codable {
...
 
     /// Only includes the names.
     case names([String])
 }
 
+/// Options for "listDatabases" operations
+public struct ListDatabasesOptions: Encodable {
+    /// Specifies whether to only return databases for which the user has privileges,
+    /// if false MongoDB will return an error for unauthorized users.
+    public let authorizedDatabases: Bool
```
kmahar, line 32: let's make this a `var`. that allows users to build options up programmatically like:
```swift
var opts = ListDatabasesOptions()
opts.authorizedDatabases = true
```

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -17,28 +17,43 @@ public struct DatabaseSpecification: Codable {
...
+
+    public init(_ authorizedDatabases: Bool = true) {
```
kmahar, line 34: I don't think we need to provide the default value of true here. generally, we try to avoid hard-coding any server default values into drivers. instead, when the user specifies no value, we send no value, and allow the server to handle what the default behavior should be.
this makes us more future-proofed against any change in the server default. also, server versions before this option was introduced might error if we send it when it isn't supported, so we should avoid doing so without the user explicitly choosing to send it.

instead, let's make the struct's `authorizedDatabases` property an optional, and this initializer can be written as
```swift
public init(authorizedDatabases: Bool? = nil) {
    self.authorizedDatabases = authorizedDatabases
}
```

similarly to my other comment, we want to require the parameter label here as well. if we add more options to this struct in the future the labels will make it clearer which parameter corresponds to which option.

Sources/MongoSwift/MongoClient.swift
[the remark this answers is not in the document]
patrickfreed: Another rationale in addition to all that is that `nameOnly: true` has performance implications server side due to having different locking requirements, so it makes sense to communicate that at the type level.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -17,28 +17,43 @@ public struct DatabaseSpecification: Codable {
...
+}
+
 /// An operation corresponding to a "listDatabases" command on a collection.
 internal struct ListDatabasesOperation: Operation {
     private let client: MongoClient
     private let filter: Document?
-    private let nameOnly: Bool?
+    private let options: ListDatabasesOptions?
+    /// Specifies whether to only return the list of database names
+    public let nameOnly: Bool?
```
kmahar, line 45: this can stay private

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -17,28 +17,43 @@ public struct DatabaseSpecification: Codable {
...
 
     internal init(
         client: MongoClient,
         filter: Document?,
-        nameOnly: Bool?
+        nameOnly: Bool?,
+        options: ListDatabasesOptions? = nil
```
kmahar, line 51: nit: this probably doesn't need a default value, since we already handle the default value of nil at the API level

Sources/MongoSwift/MongoClient.swift
kmahar: here's some relevant reading on argument labels: https://swift.org/documentation/api-design-guidelines/#argument-labels

we generally stick to this, with the addendum that if there is a first parameter where the meaning is already fairly clear from context (such as the filter here) we drop it.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
nbbeeken: I think I can lean on the default init so I don't need to make one anymore, or is it good practice to have one now incase its needed for different behavior in the future?

kmahar: unfortunately I think the default generated initializer has access level `internal` and we need to write one ourselves to make it public.
you can test this though by trying to create an instance of the struct from somewhere in the tests that does not `@testable` import. (we might want to add some tests for using this option anyway)

kmahar: this is kind of annoying to test behaviorally due to the nature of testing auth related stuff, but we could at least do some assertions in the style of those you wrote for `allowDiskUse` to see if we're correctly sending it in the commands to the server

nbbeeken: Yea I was just on the hunt for a nice `try self.withAuthTestNamespace` method 😅 maybe that's a good future helper!

kmahar:
just occurred to me, you'll need to make these same changes to support `ListDatabasesOptions` in the sync API as well! 
and you need to run `make exports` so that `ListDatabasesOptions` gets re-exported from the sync module

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
patrickfreed: Is the synthesized one `internal` even if all the fields are `public`? I was really hoping we'd be able to remove them all eventually; it would be disappointing if that never can happen. :cry:

nbbeeken: I just tested looks like that is the case, `internal` until declared otherwise.

Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -25,20 +25,35 @@ internal enum ListDatabasesResults {
     case names([String])
 }
 
+/// Options for "listDatabases" operations
+public struct ListDatabasesOptions {
+    /// Specifies whether to only return databases for which the user has privileges,
```
kmahar, line 30: I think my comment on this got lost before, but let's move the description of the potential error case out of here and into the individual method docstrings' error lists, something like:
"- An [XError] if the`authorizedDatabases` option was set to `false` but the user does not have the privilege to list all databases in the cluster."

(not sure what type of error that comes back as, maybe just a CommandError? worth checking)

Sources/MongoSwift/MongoClient.swift
```diff
@@ -352,14 +353,17 @@ public class MongoClient {
      *    - `LogicError` if the provided session is inactive.
      *    - `LogicError` if this client has already been closed.
      *    - `EncodingError` if an error is encountered while encoding the options to BSON.
+     *    - `CommandError` when options.authorizedDatabases is false:
```
kmahar, line 356: sorry to be pedantic...

1. can we phrase this more similarly to the previous bullet points? by that I mean the "`CommandError` if ...".  format. so something like "if options.authorizedDatabases is false and the user does not have listDatabases permissions."

2. listDatabase -> listDatabases

PR #451, opened 2020-04-15
SWIFT-784 Disable BSON_EXTRA_ALIGN

kmahar:
This was done experimentally while investigating the C interop issue in [SWIFT-779](https://jira.mongodb.org/browse/SWIFT-779). 
The required changes ended up being pretty minimal so we should just go ahead with them now.

### DIFF
--- a/Sources/CLibMongoC/include/CLibMongoC_bson-config.h
+++ b/Sources/CLibMongoC/include/CLibMongoC_bson-config.h
@@ -128,7 +128,7 @@
 /*
  * Define to 1 if you want extra aligned types in libbson
  */
-#define BSON_EXTRA_ALIGN 1
+#define BSON_EXTRA_ALIGN 0
 #if BSON_EXTRA_ALIGN != 1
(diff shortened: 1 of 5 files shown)


Sources/CLibMongoC/include/CLibMongoC_bson-config.h
```diff
@@ -128,7 +128,7 @@
 /*
  * Define to 1 if you want extra aligned types in libbson
  */
-#define BSON_EXTRA_ALIGN 1
+#define BSON_EXTRA_ALIGN 0
```
patrickfreed, line 131: do we need to define this here also, or is the one in `etc/generated_headers` sufficient?

mbroadst: The change in `etc/generated_headers` produces this output, so both are required

patrickfreed: oh, right :facepalm:

PR #453, opened 2020-04-17
SWIFT-779 Funnel all bson_t access through helper methods

kmahar:
Putting this up as a draft as we should probably talk through the approach more but I figured I'd let you all talk a look.
This seems like a mostly reasonable approach to me aside from the fact that we really end up in nested closure hell in places where a libmongoc method takes in 3+ arguments.

I think there is more work to do to ensure these problems don't plague any other types. for example I think we might be vulnerable to similar issues with `ReadPreference` as it follows the same "struct backed by a private class wrapping a C type" pattern.

To summarize main changes here:
- `BSONValue` protocol's `encode` method now takes in an `inout Document` rather than a `DocumentStorage`
- `DocumentStorage` is a `private` type, and its property `_bson` is now `fileprivate`. this means both are only accessible within `Document.swift`.
- the only way to access a `Document`'s underlying `bson_t` is via helpers, all of which use `withExtendedLifetime` to guarantee the `Document` stays valid for the duration of the provided closure. These are:
    - the existing `withMutableBSONPointer`
    - a new `withBSONPointer`, same as the previous but the pointer is immutable
    - a new `withOptionalBSONPointer`, which is to simplify usage in places where we have an optional document we may need to pass to a libmongoc API (basically this is just used for options documents)
- `DocumentIterator` previously stored a reference to the `DocumentStorage` it came from. however I don't think this is really different than it just storing a copy of the `Document` it's iterating over -- in both cases we should get CoW behavior if someone tries to modify a copy of that document which is backed by the same storage. changing this this enabled me to make the change in the previous bullet point.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONEncoder.swift
+++ b/Sources/MongoSwift/BSON/BSONEncoder.swift
@@ -754,8 +754,8 @@ private class MutableArray: BSONValue {
 
     /// methods required by the BSONValue protocol that we don't actually need/use. MutableArray
     /// is just a BSONValue to simplify usage alongside true BSONValues within the encoder.
-    fileprivate func encode(to _: DocumentStorage, forKey _: String) throws {
-        fatalError("`MutableArray` is not meant to be encoded to a `DocumentStorage`")
+    fileprivate func encode(to _: inout Document, forKey _: String) throws {
+        fatalError("`MutableArray` is not meant to be encoded to a `Document`")
     }
 
     internal static func from(iterator _: DocumentIterator) -> BSON {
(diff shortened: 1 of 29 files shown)


kmahar:
made a few changes in my latest commit:
- moved the helpers onto `Document` itself, excluding the convenience one, `withOptionalBSONPointer` since that needs to take in an optional. it is implemented by just delegating to the non optional helper. in addition to making the helper usages a bit more concise, this enabled me to make `_storage` private.
- nested the `DocumentStorage` type definition within `Document` and renamed it just `Storage`

PR #456, opened 2020-04-21
SWIFT-779 Safe access to ReadPreference pointer

kmahar:
@mbroadst I'm not sure if you want me to keep including you on these, LMK.

This PR switches `ReadPreference` over to the "create a temporary libmongoc equivalent, use it to execute a closure, and then clean it up" approach we follow for `ReadConcern`, `WriteConcern`, `MongoDatabase`, and `MongoCollection`. 

We had existing tests regarding `ReadConcern` and `WriteConcern` that verified these got set correctly on mongoc clients, dbs, and collections. I think we skipped these for `ReadPreference` at the time because it was just wrapping a `mongoc_read_prefs_t` so there was no need to test out the temporary creation of the libmongoc type.

I have now added similar tests for `ReadPreference`. As part of that I also moved libmongoc-dependent helper methods for getting the `ReadConcern`/`WriteConcern` on a temporary db/collection into the driver itself. This allowed me to remove the CLibMongoC dependency from `MongoSwiftTests` 🎉Now we just need to remove it from `BSONTests`. 

I also added some round trip testing between Swift and libmongoc types for all three of RC, WC, RP.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         .target(name: "AtlasConnectivity", dependencies: ["MongoSwiftSync"]),
         .target(name: "TestsCommon", dependencies: ["MongoSwift", "Nimble"]),
         .testTarget(name: "BSONTests", dependencies: ["MongoSwift", "TestsCommon", "Nimble", "CLibMongoC"]),
-        .testTarget(name: "MongoSwiftTests", dependencies: ["MongoSwift", "TestsCommon", "Nimble", "NIO", "CLibMongoC"]),
+        .testTarget(name: "MongoSwiftTests", dependencies: ["MongoSwift", "TestsCommon", "Nimble", "NIO"]),
         .testTarget(name: "MongoSwiftSyncTests", dependencies: ["MongoSwiftSync", "TestsCommon", "Nimble", "MongoSwift"]),
         .target(
             name: "CLibMongoC",
--- a/Sources/MongoSwift/ReadConcern.swift
+++ b/Sources/MongoSwift/ReadConcern.swift
@@ -62,7 +62,7 @@ public struct ReadConcern: Codable {
 
     // Initializes a new `ReadConcern` with the same level as the provided `mongoc_read_concern_t`.
     // The caller is responsible for freeing the original `mongoc_read_concern_t`.
-    internal init(from readConcern: OpaquePointer) {
+    internal init(copying readConcern: OpaquePointer) {
         if let level = mongoc_read_concern_get_level(readConcern) {
             self.level = Level(rawValue: String(cString: level))
         }
--- a/Sources/MongoSwift/ReadPreference.swift
+++ b/Sources/MongoSwift/ReadPreference.swift
@@ -232,7 +232,7 @@ public struct ReadPreference: Equatable {
         let rp = mongoc_read_prefs_new(self.mode.mongocMode)! // never returns nil
         defer { mongoc_read_prefs_destroy(rp) }
 
-        if let tagSets = self.tagSets, tagSets.isEmpty {
+        if let tagSets = self.tagSets, !tagSets.isEmpty {
             let tags = Document(tagSets.map { .document($0) })
             tags.withBSONPointer { tagsPtr in
                 mongoc_read_prefs_set_tags(rp, tagsPtr)
--- a/Sources/MongoSwift/ReadPreference.swift
+++ b/Sources/MongoSwift/ReadPreference.swift
@@ -3,7 +3,7 @@ import CLibMongoC
 /// Represents a MongoDB read preference, indicating which member(s) of a replica set read operations should be
 /// directed to.
 /// - SeeAlso: https://docs.mongodb.com/manual/reference/read-preference/
-public struct ReadPreference {
+public struct ReadPreference: Equatable {
     /// An enumeration of possible read preference modes.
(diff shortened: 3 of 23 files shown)


Sources/MongoSwift/ReadConcern.swift
```diff
@@ -62,7 +62,7 @@ public struct ReadConcern: Codable {
 
     // Initializes a new `ReadConcern` with the same level as the provided `mongoc_read_concern_t`.
     // The caller is responsible for freeing the original `mongoc_read_concern_t`.
-    internal init(from readConcern: OpaquePointer) {
+    internal init(copying readConcern: OpaquePointer) {
```
kmahar, line 65: `copying` seemed more descriptive and consistent with our other similar internal APIs

Tests/MongoSwiftTests/ReadConcernTests.swift
```diff
@@ -202,4 +177,23 @@ final class ReadConcernTests: MongoSwiftTestCase {
             try checkReadConcern(coll5, majority, "collection retrieved with majority RC from \(dbDesc)")
         }
     }
+
+    func testRoundTripThroughLibmongoc() throws {
+        let rcs: [ReadConcern] = [
+            ReadConcern(),
+            ReadConcern(.local),
```
kmahar, line 184: writing this out made me think we should switch to a similar API to the one we have now for `ReadPreference`. basically static properties corresponding to each `ReadConcern` mode, plus a static method for creating one with an unknown level for forward compatibility. thoughts?

Tests/MongoSwiftTests/ReadPreferenceTests.swift
```diff
@@ -3,6 +3,27 @@ import Nimble
 import TestsCommon
 import XCTest
 
+/// Indicates that a type has a read preference property, as well as a way to get a read preference from an instance
+/// of the corresponding mongoc type.
+private protocol ReadPreferenceable {
```
kmahar, line 8: somehow this sounds a lot worse than `ReadConcernable` but it's just for tests so I didn't spend too much time thinking about it lol

Tests/MongoSwiftTests/ReadConcernTests.swift
mbroadst: SGTM, you could probably also do it for `WriteConcern` for some expected common values (e.g. `majority`)

Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         .target(name: "AtlasConnectivity", dependencies: ["MongoSwiftSync"]),
         .target(name: "TestsCommon", dependencies: ["MongoSwift", "Nimble"]),
         .testTarget(name: "BSONTests", dependencies: ["MongoSwift", "TestsCommon", "Nimble", "CLibMongoC"]),
-        .testTarget(name: "MongoSwiftTests", dependencies: ["MongoSwift", "TestsCommon", "Nimble", "NIO", "CLibMongoC"]),
+        .testTarget(name: "MongoSwiftTests", dependencies: ["MongoSwift", "TestsCommon", "Nimble", "NIO"]),
```
nbbeeken, line 19: 🎉🎉🎉

kmahar:
whoops sorry i was trying to actually remove your name @mbroadst but i guess since you commented once already you are stuck on the list forever

Tests/MongoSwiftTests/ReadConcernTests.swift
kmahar: cool, I opened [SWIFT-815](https://jira.mongodb.org/browse/SWIFT-815) to track that work

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -193,104 +186,73 @@ public struct ReadPreference {
...
 
-        guard !wrapped.isEmpty else {
-            return nil
+        let maxStalenessValue = mongoc_read_prefs_get_max_staleness_seconds(pointer)
+        if maxStalenessValue != MONGOC_NO_MAX_STALENESS {
+            self.maxStalenessSeconds = Int(exactly: maxStalenessValue)
         }
+    }
 
+    internal func withMongocReadPreference<T>(body: (OpaquePointer) throws -> T) rethrows -> T {
```
patrickfreed, line 230: probably worth documenting that the pointer is only valid within the closure and will need its contents copied if preserved.

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -193,104 +186,73 @@ public struct ReadPreference {
...
+            mongoc_read_prefs_set_max_staleness_seconds(rp, Int64(maxStaleness))
+        }
+
+        return try body(rp)
     }
+}
 
-    deinit {
-        mongoc_read_prefs_destroy(readPref)
+internal func withOptionalMongocReadPreference<T>(
```
patrickfreed, line 250: For organizational purposes, would this be better as a static method on `ReadPreference`?

Sources/MongoSwift/ReadPreference.swift
kmahar: done

Sources/MongoSwift/ReadPreference.swift
kmahar: that's fine with me, done. we should probably make the same change for the optional document helper for consistency then. I'll do it in a separate PR after this is merged

PR #459, opened 2020-04-27
SWIFT-640: Add authentication options to ClientOptions

nbbeeken:
I made the Credential struct public so we can share the data structure with users specifying their authentication options. Since the URI in libmongoc is the source of truth about what settings the user wishes to have when connecting to mongo I set up a function to set all the authentication properties in the URI from the user specified Credential. Added tests to cover.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -114,7 +114,7 @@ internal class ConnectionString {
             return nil
         }
         let str = String(cString: mechanism)
-        return MongoCredential.Mechanism(name: str)
+        return MongoCredential.Mechanism(str)
     }
 
     /// Returns a document containing the auth mechanism properties if any were provided, otherwise nil.
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -202,18 +202,32 @@ internal class ConnectionString {
 
     /// Sets credential properties in the URI string
     internal func setCredential(from credential: Credential) throws {
-        guard mongoc_uri_set_username(self._uri, credential.username) else {
-            throw InvalidArgumentError(message: "Cannot set username to \(String(describing: credential.username))")
+        if let username = credential.username {
+            guard mongoc_uri_set_username(self._uri, username) else {
+                throw InvalidArgumentError(message: "Cannot set username to \(username).")
+            }
+        } else {
+            throw InvalidArgumentError(message: "Credential must specify username.")
         }
-        guard mongoc_uri_set_password(self._uri, credential.password) else {
-            throw InvalidArgumentError(message: "Cannot set password")
+
+        if let password = credential.password {
+            guard mongoc_uri_set_password(self._uri, password) else {
+                throw InvalidArgumentError(message: "Cannot set password")
+            }
         }
-        guard mongoc_uri_set_auth_source(self._uri, credential.source) else {
-            throw InvalidArgumentError(message: "Cannot set authSource to \(String(describing: credential.source))")
+
+        if let authSource = credential.source {
+            guard mongoc_uri_set_auth_source(self._uri, authSource) else {
+                throw InvalidArgumentError(message: "Cannot set authSource to \(authSource)")
+            }
         }
-        guard mongoc_uri_set_auth_mechanism(self._uri, credential.mechanism?.rawValue) else {
-            throw InvalidArgumentError(message: "Cannot set mechanism to \(String(describing: credential.mechanism))")
+
+        if let mechanism = credential.mechanism {
+            guard mongoc_uri_set_auth_mechanism(self._uri, mechanism.rawValue) else {
+                throw InvalidArgumentError(message: "Cannot set mechanism to \(mechanism))")
+            }
         }
+
         try credential.mechanismProperties?.withBSONPointer { mechanismPropertiesPtr in
             guard mongoc_uri_set_mechanism_properties(self._uri, mechanismPropertiesPtr) else {
                 throw InvalidArgumentError(
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -34,7 +34,7 @@ internal class ConnectionString {
         }
 
         if let credential = options?.credential {
-            try self.setCredential(credential)
+            try self.setMongoCredential(credential)
         }
     }
 
@@ -109,12 +109,12 @@ internal class ConnectionString {
     }
 
     /// Returns the auth mechanism if one was provided, otherwise nil.
-    internal var authMechanism: Credential.Mechanism? {
+    internal var authMechanism: MongoCredential.Mechanism? {
         guard let mechanism = mongoc_uri_get_auth_mechanism(self._uri) else {
             return nil
         }
         let str = String(cString: mechanism)
-        return Credential.Mechanism(rawValue: str)
+        return MongoCredential.Mechanism(name: str)
     }
 
     /// Returns a document containing the auth mechanism properties if any were provided, otherwise nil.
@@ -144,8 +144,8 @@ internal class ConnectionString {
     }
 
     /// Returns the credential configured on this URI. Will be empty if no options are set.
-    internal var credential: Credential {
-        Credential(
+    internal var credential: MongoCredential {
+        MongoCredential(
             username: self.username,
             password: self.password,
             source: self.authSource,
@@ -221,7 +221,7 @@ internal class ConnectionString {
         }
 
         if let mechanism = credential.mechanism {
-            guard mongoc_uri_set_auth_mechanism(self._uri, mechanism.rawValue) else {
+            guard mongoc_uri_set_auth_mechanism(self._uri, mechanism.name) else {
                 throw InvalidArgumentError(message: "Cannot set mechanism to \(mechanism)).")
             }
         }
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -32,6 +32,10 @@ internal class ConnectionString {
         if let rr = options?.retryReads {
             mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYREADS, rr)
         }
+
+        if let credential = options?.credential {
+            try self.setCredential(from: credential)
+        }
     }
 
     /// Initializes a new connection string that wraps a copy of the provided URI. Does not destroy the input URI.
@@ -194,5 +198,28 @@ internal class ConnectionString {
     /// Executes the provided closure using a pointer to the underlying `mongoc_uri_t`.
     internal func withMongocURI<T>(_ body: (OpaquePointer) throws -> T) rethrows -> T {
         try body(self._uri)
+    }
+
+    /// Sets credential properties in the URI string
+    internal func setCredential(from credential: Credential) throws {
+        guard mongoc_uri_set_username(self._uri, credential.username) else {
+            throw InvalidArgumentError(message: "Cannot set username to \(String(describing: credential.username))")
+        }
+        guard mongoc_uri_set_password(self._uri, credential.password) else {
+            throw InvalidArgumentError(message: "Cannot set password")
+        }
+        guard mongoc_uri_set_auth_source(self._uri, credential.source) else {
+            throw InvalidArgumentError(message: "Cannot set authSource to \(String(describing: credential.source))")
+        }
+        guard mongoc_uri_set_auth_mechanism(self._uri, credential.mechanism?.rawValue) else {
+            throw InvalidArgumentError(message: "Cannot set mechanism to \(String(describing: credential.mechanism))")
+        }
+        try credential.mechanismProperties?.withBSONPointer { mechanismPropertiesPtr in
+            guard mongoc_uri_set_mechanism_properties(self._uri, mechanismPropertiesPtr) else {
+                throw InvalidArgumentError(
+                    message: "Cannot set mechanismProperties to \(String(describing: credential.mechanismProperties))"
+                )
+            }
+        }
     }
 }
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -201,7 +201,7 @@ internal class ConnectionString {
     }
 
     /// Sets credential properties in the URI string
-    internal func setCredential(from credential: Credential) throws {
+    internal func setMongoCredential(_ credential: MongoCredential) throws {
         if let username = credential.username {
             guard mongoc_uri_set_username(self._uri, username) else {
                 throw InvalidArgumentError(message: "Cannot set username to \(username).")
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -34,7 +34,7 @@ internal class ConnectionString {
         }
 
         if let credential = options?.credential {
-            try self.setCredential(from: credential)
+            try self.setCredential(credential)
         }
     }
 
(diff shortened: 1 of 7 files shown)


kmahar:
Looping in @mbroadst since this is public API

Sources/MongoSwift/Credential.swift
```diff
@@ -1,5 +1,5 @@
 /// Represents an authentication credential.
-internal struct Credential: Decodable, Equatable {
+public struct Credential: Decodable, Equatable {
```
kmahar, line 2: the spec says we SHOULD name this `MongoCredential`: https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mongocredential

also, this type needs a public initializer now, and let's make all of its properties `public` and `var`s.

that said we did recently decide to move away from having a nested struct for TLS options and instead flattened it into the `ClientOptions` to match the connection string structure. So another option would be to nest this info directly in `ClientOptions` though that does go against the spec.

Opinions?

Sources/MongoSwift/Credential.swift
```diff
@@ -19,7 +19,7 @@ internal struct Credential: Decodable, Equatable {
 }
 
 /// Possible authentication mechanisms.
-internal enum AuthMechanism: String, Decodable {
+public enum AuthMechanism: String, Decodable {
```
kmahar, line 22: for forward compatibility we might want to add an `.other(String)` case. this would mean we don't have to make a breaking change if a new mechanism is added.
I don't have a good sense of how often we add auth mechanisms though - thoughts @patrickfreed @mbroadst ?

kmahar: maybe we should give this the full name now, i.e. `AuthenticationMechanism`

Sources/MongoSwift/MongoClient.swift
```diff
@@ -4,6 +4,9 @@ import NIOConcurrencyHelpers
 
 /// Options to use when creating a `MongoClient`.
 public struct ClientOptions: CodingStrategyProvider {
+    /// Authentication Credential
```
kmahar, line 7: I think this docstring could be more descriptive, something like "Specifies authentication options for use with the client.

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -188,4 +192,26 @@ internal class ConnectionString {
 
         return hosts
     }
+
+    internal func setCredential(from credential: Credential) throws {
+        guard mongoc_uri_set_username(self._uri, credential.username) else {
```
kmahar: we probably only want to be calling these libmongoc methods if the value we are passing in is non nil.

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -188,4 +192,26 @@ internal class ConnectionString {
 
         return hosts
     }
+
+    internal func setCredential(from credential: Credential) throws {
```
kmahar, line 196: nit: I'd probably just omit the argument label here since the one argument we pass in is the credential so I think the relationship is fairly clear at the call site. alternatively you could name the method something like `setAuthenticationOptions`

Sources/MongoSwift/ConnectionString.swift
kmahar: ah I see you pushed a fix for this right as I published my comments 🙂

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -194,18 +194,32 @@ internal class ConnectionString {
...
 
     internal func setCredential(from credential: Credential) throws {
-        guard mongoc_uri_set_username(self._uri, credential.username) else {
-            throw InvalidArgumentError(message: "Cannot set username to \(String(describing: credential.username))")
+        if let username = credential.username {
+            guard mongoc_uri_set_username(self._uri, username) else {
+                throw InvalidArgumentError(message: "Cannot set username to \(username).")
+            }
+        } else {
+            throw InvalidArgumentError(message: "Credential must specify username.")
```
kmahar, line 202: according to the spec, username is optional for MONGODB-X509 and MONGODB-AWS:
https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mongocredential

Sources/MongoSwift/Credential.swift
nbbeeken: Quick tour of the drivers shows how stretched should really is 😅 :
- [go](https://github.com/mongodb/mongo-go-driver/blob/master/x/mongo/driver/auth/cred.go#L10)
- [ruby](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/auth/user.rb#L23)
- [rust](https://github.com/mongodb/mongo-rust-driver/blob/master/src/client/auth/mod.rs#L181)
- [javascript](https://github.com/mongodb/node-mongodb-native/blob/master/lib/cmap/auth/mongo_credentials.js#L35)

I think that it does warrant its own sub-struct as it is a neat unit of information constrained to auth only. Additionally it does match other drivers in that way. But I defer to your call on that since tls opts were flattened

Sources/MongoSwift/Credential.swift
nbbeeken: Is it breaking to add a field to an enum? Would it make sense to leave out the flexibility of `.other(String)` to force an update for something as major as a new mechanism, basically making sure the driver supports it correctly?
Perhaps we could move the enum inside the Credential struct giving us  the `Credential.Mechanism` type?

kmahar: > Is it breaking to add a field to an enum?

adding a new case to an enum is technically breaking, yes, as any `switch` statements that were previously exhaustive are no longer exhaustive.
admittedly it's unlikely that someone would be switching on auth mechanism in their own code but it's possible.

> Would it make sense to leave out the flexibility of .other(String) to force an update for something as major as a new mechanism, basically making sure the driver supports it correctly?

we follow the `.other` pattern elsewhere, e.g. in `ReadConcern`: https://github.com/mongodb/mongo-swift-driver/blob/973acb25f748d7ca27d131c88b109193b839f803/Sources/MongoSwift/ReadConcern.swift#L20

If a new mechanism is added, we'll get new tests added to the auth spec and a ticket created for support + tests, similarly to what happens for driver options like those you've implemented recently. So I think that will satisfy your concern about making sure the driver supports it correctly.

> Perhaps we could move the enum inside the Credential struct giving us the Credential.Mechanism type?
I'm good with nesting the type if we stick with the separate struct approach 👍

patrickfreed: I wonder if it would be better to just make this an opaque struct. We seem to be adding new authentication mechanisms somewhat often, and the `.other` API isn't the greatest. Because the mechanisms themselves don't have any customization / associated values and don't need to be switched on, we can do this without any API cost. Also, unlike options that are sent to the server, the driver won't start supporting a new auth mechanism without an update of the driver, so the `.other` case only serves to preserve API compatibility but doesn't provide any forward behavioral compatibility.

e.g.:
```swift
public struct Mechanism {
    public static let gssApi = AuthMechanism(name: "GSSAPI")
    public static let scramSha1 = AuthMechanism(name: "SCRAM-SHA-1")

    private init(name: String) {
        self.name = name
    }
}
```
Then due to the nice syntactic stuff we've been using recently for other types, users can still specify the mechanism with the shorthand syntax if they want:
```swift
Credential(mechanism: .scramSha1)
Credential(mechanism: Credential.Mechanism.scramSha1)
```

Knowing that the syntax is nice even when using structs, I'm starting to get the feeling that if we do need an `.other` case, we _probably_ shouldn't be using an enum. While that strategy allow us to support new cases in the future without breaking API, it does so in an unfortunate way for users. I think that until non-exhaustive enums are available outside of the standard library, we should just stick to static structs for enum-like things that may need new cases in the future. As enums are now, one of their primary value-adds is their exhaustiveness, and if we define them for cases that cannot be exhaustively covered, then I don't think they add much. (I think I've reached my enum hangover...).

A survey of the driver for enums that have a `.other` case yields:
- [`CollectionType`](https://mongodb.github.io/mongo-swift-driver/MongoSwift/Enums/CollectionType.html): This is returned as part of the `CollectionSpecification` model in a `listCollections` response. Going back to 3.4, no value besides `collection` and `view` have been possible here.
    - My proposal: we simply drop the `.other` case due to low risk of new value being introduced. Maybe have to confirm with server team
- [`FullDocument`](https://mongodb.github.io/mongo-swift-driver/MongoSwift/Enums/FullDocument.html): The spec specifically requires that users be able to specify unknown values
    - My proposal: convert to struct with a static `updateLookup`, allow `FullDocument` to be initializable from `String` (and string literal)
- [`ReadConcern.Level`](https://mongodb.github.io/mongo-swift-driver/MongoSwift/Structs/ReadConcern/Level.html): New variants are added somewhat regularly (`snapshot` in 4.0 being the most recent)
   - My proposal: convert to struct with static values for the various levels, allow `Level` to be initializable from `String` (and string literal)

Sources/MongoSwift/Credential.swift
kmahar: ok, I'm cool with keeping a separate type then. it sounds like we've reached consensus on keeping the `Mongo` prefixes. so I'll resolve this thread.

Sources/MongoSwift/Credential.swift
patrickfreed: For my comments not related to this PR, I filed [SWIFT-827](https://jira.mongodb.org/browse/SWIFT-827)

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,67 @@
+/// Represents an authentication credential.
+public struct MongoCredential: Decodable, Equatable {
+    /// A string containing the username. For auth mechanisms that do not utilize a password, this may be the entire
```
patrickfreed, line 3: I personally prefer there were newlines between each of the properties, though I'm not sure if we're consistent about this in the codebase.

Sources/MongoSwiftSync/Exports.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.17.0 — https://github.com/krzysztofzablocki/Sourcery
```
patrickfreed, line 1: I think this needs to be 0.16.1 to not break travis. We should update this in the future to not include the sourcery version.

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Possible authentication mechanisms.
+    public struct Mechanism: Codable, Equatable {
+        /// See https://docs.mongodb.com/manual/core/kerberos/
+        public static let gssAPI = Mechanism(name: "GSSAPI")
+        /// Deprecated: see https://docs.mongodb.com/manual/release-notes/3.0-scram/
+        public static let mongodbCR = Mechanism(name: "MONGODB-CR")
+        /// See https://docs.mongodb.com/manual/core/security-x.509/#security-auth-x509
+        public static let mongodbX509 = Mechanism(name: "MONGODB-X509")
```
patrickfreed, line 41: should this be named `mongoDBX509`? (and same for mongoDBCR?) I honestly don't know.

nbbeeken: They mirror the names from the existing enum

https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/Credential.swift#L21-L29

Sources/MongoSwift/MongoCredential.swift
patrickfreed: ditto for `Mechanism` below.

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,71 @@
...
+        public static let plain = Mechanism(name: "PLAIN")
+        /// See https://docs.mongodb.com/manual/core/security-scram/#authentication-scram
+        public static let scramSHA1 = Mechanism(name: "SCRAM-SHA-1")
+        /// See https://docs.mongodb.com/manual/core/security-scram/#authentication-scram
+        public static let scramSHA256 = Mechanism(name: "SCRAM-SHA-256")
+
+        /// Name of the authentication mechanism.
+        public var name: String
+
+        public init(name: String) {
```
patrickfreed, line 56: we actually don't want this to be public, since a user will never be able to access a new mechanism without updating the driver like could be possible with parameters sent to the server (e.g. read concern).

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,71 @@
...
+        public static let mongodbX509 = Mechanism(name: "MONGODB-X509")
+        /// See https://docs.mongodb.com/manual/core/security-ldap/
+        public static let plain = Mechanism(name: "PLAIN")
+        /// See https://docs.mongodb.com/manual/core/security-scram/#authentication-scram
+        public static let scramSHA1 = Mechanism(name: "SCRAM-SHA-1")
+        /// See https://docs.mongodb.com/manual/core/security-scram/#authentication-scram
+        public static let scramSHA256 = Mechanism(name: "SCRAM-SHA-256")
+
+        /// Name of the authentication mechanism.
+        public var name: String
```
patrickfreed, line 54: I think this can probably be internal and we should just conform to `CustomStringConvertible`.

Sources/MongoSwiftSync/Exports.swift
kmahar: is it possible to make sourcery exclude that line?

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,71 @@
...
+            self.name = name
+        }
+
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.singleValueContainer()
+            let string = try container.decode(String.self)
+            self.name = string
+        }
+
+        public func encode(to encoder: Encoder) throws {
```
kmahar, line 66: do we ever actually need to encode this struct?

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,71 @@
...
+        self.source = source
+        self.mechanism = mechanism
+        self.mechanismProperties = mechanismProperties
+    }
+
+    /// Possible authentication mechanisms.
+    public struct Mechanism: Codable, Equatable {
+        /// See https://docs.mongodb.com/manual/core/kerberos/
+        public static let gssAPI = Mechanism(name: "GSSAPI")
+        /// Deprecated: see https://docs.mongodb.com/manual/release-notes/3.0-scram/
```
kmahar, line 42: should we remove MongoDB-CR support altogether? it was deprecated in 3.0 and removed from the server altogether in 4.0. cc @mbroadst

mbroadst: Yes, let's go ahead and remove it. We can add it back if someone asks for it, but I can't imagine who would

Sources/MongoSwiftSync/Exports.swift
patrickfreed: Appears not unfortunately: https://github.com/krzysztofzablocki/Sourcery/issues/399

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -0,0 +1,75 @@
...
+        self.source = source
+        self.mechanism = mechanism
+        self.mechanismProperties = mechanismProperties
+    }
+
+    /// Possible authentication mechanisms.
+    public struct Mechanism: Decodable, Equatable, CustomStringConvertible {
+        /// GSSAPI authentication.
+        /// - SeeAlso: https://docs.mongodb.com/manual/core/kerberos/
+        public static let gssAPI = Mechanism("GSSAPI")
```
mbroadst, line 42: nit: we've decided to use `ID` for `ObjectID` and all of these cases use acronyms or initialism, they should probably mostly be uppercased

patrickfreed: Given that these are properties, I think they still have to conform to the lower camel case convention for the first chunk and can go full capitalization for subsequent chunks. (reference: https://github.com/raywenderlich/swift-style-guide/issues/101#issuecomment-320200304)

mbroadst: feel free to ignore this comment, I don't have strong opinions and left an approval

PR #461, opened 2020-04-29
SWIFT-815: More convenient API for ReadConcern/WriteConcern creation

nbbeeken:
Added convenience properties to ReadConcern and WriteConcern.
Some notes:
- ReadConcern no longer has a public constructor, users can rely on the static other method to make unsupported ReadConcerns.
- WriteConcern now has a `fileprivate init` this was done because we needed a non-throwing constructor to add the `static let majority: WriteConcern` property.
- We now make use the convenience API internally so the changes are being tested.

### DIFF
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -19,7 +19,7 @@ private func causalConsistency() throws {
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
         readConcern: .majority,
-        writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
+        writeConcern: try WriteConcern.majority(wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
     let result1 = items.updateOne(
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -19,7 +19,7 @@ private func causalConsistency() throws {
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
         readConcern: .majority,
-        writeConcern: try WriteConcern.majority(wtimeoutMS: 1000)
+        writeConcern: try .majority(wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
     let result1 = items.updateOne(
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -18,7 +18,7 @@ private func causalConsistency() throws {
     let s1 = client1.startSession(options: ClientSessionOptions(causalConsistency: true))
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
-        readConcern: ReadConcern.majority,
+        readConcern: .majority,
         writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -18,7 +18,7 @@ private func causalConsistency() throws {
     let s1 = client1.startSession(options: ClientSessionOptions(causalConsistency: true))
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
-        readConcern: ReadConcern(.majority),
+        readConcern: ReadConcern.majority,
         writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
--- a/Examples/Docs/Sources/SyncExamples/main.swift
+++ b/Examples/Docs/Sources/SyncExamples/main.swift
@@ -12,7 +12,7 @@ private func causalConsistency() throws {
     let s1 = client1.startSession(options: ClientSessionOptions(causalConsistency: true))
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
-        readConcern: ReadConcern.majority,
+        readConcern: .majority,
         writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
--- a/Examples/Docs/Sources/SyncExamples/main.swift
+++ b/Examples/Docs/Sources/SyncExamples/main.swift
@@ -13,7 +13,7 @@ private func causalConsistency() throws {
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
         readConcern: .majority,
-        writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
+        writeConcern: try WriteConcern.majority(wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
     try items.updateOne(
--- a/Examples/Docs/Sources/SyncExamples/main.swift
+++ b/Examples/Docs/Sources/SyncExamples/main.swift
@@ -13,7 +13,7 @@ private func causalConsistency() throws {
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
         readConcern: .majority,
-        writeConcern: try WriteConcern.majority(wtimeoutMS: 1000)
+        writeConcern: try .majority(wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
     try items.updateOne(
--- a/Examples/Docs/Sources/SyncExamples/main.swift
+++ b/Examples/Docs/Sources/SyncExamples/main.swift
@@ -12,7 +12,7 @@ private func causalConsistency() throws {
     let s1 = client1.startSession(options: ClientSessionOptions(causalConsistency: true))
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
-        readConcern: ReadConcern(.majority),
+        readConcern: ReadConcern.majority,
         writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
     )
     let items = client1.db("test", options: dbOptions).collection("items")
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -449,7 +449,7 @@ public class MongoClient {
      * Gets a `MongoDatabase` instance for the given database name. If an option is not specified in the optional
      * `DatabaseOptions` param, the database will inherit the value from the parent client or the default if
      * the client’s option is not set. To override an option inherited from the client (e.g. a read concern) with the
-     * default value, it must be explicitly specified in the options param (e.g. ReadConcern(), not nil).
+     * default value, it must be explicitly specified in the options param (e.g. ReadConcern.empty, not nil).
      *
      * - Parameters:
      *   - name: the name of the database to retrieve
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -449,7 +449,7 @@ public class MongoClient {
      * Gets a `MongoDatabase` instance for the given database name. If an option is not specified in the optional
      * `DatabaseOptions` param, the database will inherit the value from the parent client or the default if
      * the client’s option is not set. To override an option inherited from the client (e.g. a read concern) with the
-     * default value, it must be explicitly specified in the options param (e.g. ReadConcern.empty, not nil).
+     * default value, it must be explicitly specified in the options param (e.g. ReadConcern.serverDefault, not nil).
      *
      * - Parameters:
      *   - name: the name of the database to retrieve
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -135,7 +135,7 @@ public struct MongoDatabase {
      * Access a collection within this database. If an option is not specified in the `CollectionOptions` param, the
      * collection will inherit the value from the parent database or the default if the db's option is not set.
      * To override an option inherited from the db (e.g. a read concern) with the default value, it must be explicitly
-     * specified in the options param (e.g. ReadConcern.empty, not nil).
+     * specified in the options param (e.g. ReadConcern.serverDefault, not nil).
      *
      * - Parameters:
      *   - name: the name of the collection to get
@@ -153,7 +153,7 @@ public struct MongoDatabase {
      * `MongoCollection` instance. If an option is not specified in the `CollectionOptions` param, the
      * collection will inherit the value from the parent database or the default if the db's option is not set.
      * To override an option inherited from the db (e.g. a read concern) with the default value, it must be explicitly
-     * specified in the options param (e.g. ReadConcern.empty, not nil).
+     * specified in the options param (e.g. ReadConcern.serverDefault, not nil).
      *
      * - Parameters:
      *   - name: the name of the collection to get
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -135,7 +135,7 @@ public struct MongoDatabase {
      * Access a collection within this database. If an option is not specified in the `CollectionOptions` param, the
      * collection will inherit the value from the parent database or the default if the db's option is not set.
      * To override an option inherited from the db (e.g. a read concern) with the default value, it must be explicitly
-     * specified in the options param (e.g. ReadConcern.serverDefault, not nil).
+     * specified in the options param (e.g. .serverDefault, not nil).
      *
      * - Parameters:
      *   - name: the name of the collection to get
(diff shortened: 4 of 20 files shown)


Examples/Docs/Sources/AsyncExamples/main.swift
```diff
@@ -18,7 +18,7 @@ private func causalConsistency() throws {
     let s1 = client1.startSession(options: ClientSessionOptions(causalConsistency: true))
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
-        readConcern: ReadConcern(.majority),
+        readConcern: ReadConcern.majority,
```
mbroadst, line 21: I thought the idea was that we could write like:
```
let dbOptions = DatabaseOptions(
    readConcern: .majority,
    ...
)
```

nbbeeken: Indeed that was the point! I gotta level up my swifiness 🏃 🏃 🏃

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -113,6 +116,13 @@ public struct WriteConcern: Codable {
         }
     }
 
+    /// Initializes a new `WriteConcern` without throwing any errors.
+    fileprivate init(journal: Bool?, w: W, wtimeoutMS: Int?) {
```
patrickfreed, line 120: rather than using this and perhaps silently making an invalid write concern, we can just re-use the other one with `try!`. Because it's a static variable, if this ever fails we'll know about it immediately.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,55 @@ import CLibMongoC
...
+    /// Available ReadConcern, see https://docs.mongodb.com/manual/reference/read-concern-available/.
+    public static let available = ReadConcern(.available)
+    /// Linearizable ReadConcern, see https://docs.mongodb.com/manual/reference/read-concern-majority/.
+    public static let linearizable = ReadConcern(.linearizable)
+    /// Majority ReadConcern, see https://docs.mongodb.com/manual/reference/read-concern-linearizable/.
+    public static let majority = ReadConcern(.majority)
+    /// Snapshot ReadConcern, see https://docs.mongodb.com/master/reference/read-concern-snapshot/.
+    public static let snapshot = ReadConcern(.snapshot)
+    /// Empty ReadConcern.
+    public static let empty = ReadConcern(nil)
```
patrickfreed, line 16: rather than calling this `empty`, I think `default` or `serverDefault` would be more accurate. Also, if we're going to do this with `ReadConcern`, we should do the same with `WriteConcern`.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,55 @@ import CLibMongoC
...
+        public static let linearizable = Level("linearizable")
         /// See https://docs.mongodb.com/master/reference/read-concern-snapshot/
-        case snapshot
-        /// Any other read concern level not covered by the other cases.
-        /// This case is present to provide forwards compatibility with any
-        /// future read concerns which may be added to new versions of MongoDB.
-        case other(level: String)
+        public static let snapshot = Level("snapshot")
+
+        public var name: String
```
patrickfreed, line 39: I don't think this needs to be public. If we want to expose this as a `String`, then we can conform the type to [`CustomStringConvertible`](https://developer.apple.com/documentation/swift/customstringconvertible).

Sources/MongoSwiftSync/ClientSession.swift
```diff
@@ -13,7 +13,7 @@ import MongoSwift
  *
  * e.g.
  *   ```
- *   let opts = CollectionOptions(readConcern: ReadConcern(.majority), writeConcern: try WriteConcern(w: .majority))
+ *   let opts = CollectionOptions(readConcern: .majority, writeConcern: WriteConcern.majority)
```
patrickfreed, line 16: the write concern bit can just be `.majority`

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -410,7 +410,7 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(try self.coll.countDocuments()).to(equal(1))
 
         // test using a write concern
-        let opts2 = FindOneAndDeleteOptions(writeConcern: try WriteConcern(w: .majority))
+        let opts2 = FindOneAndDeleteOptions(writeConcern: WriteConcern.majority)
```
patrickfreed, line 413: ditto these can be just `.majority`

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,55 @@ import CLibMongoC
...
+    /// Linearizable ReadConcern, see https://docs.mongodb.com/manual/reference/read-concern-majority/.
+    public static let linearizable = ReadConcern(.linearizable)
+    /// Majority ReadConcern, see https://docs.mongodb.com/manual/reference/read-concern-linearizable/.
+    public static let majority = ReadConcern(.majority)
+    /// Snapshot ReadConcern, see https://docs.mongodb.com/master/reference/read-concern-snapshot/.
+    public static let snapshot = ReadConcern(.snapshot)
+    /// Empty ReadConcern.
+    public static let empty = ReadConcern(nil)
+
+    public static func other(_ level: Level) -> ReadConcern {
```
patrickfreed, line 18: rather than having static `other` methods, I think using `init` like we have for `Level` serves us fine.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,55 @@ import CLibMongoC
...
-            case .majority:
-                return "majority"
-            case .linearizable:
-                return "linearizable"
-            case .snapshot:
-                return "snapshot"
-            case let .other(l):
-                return l
-            }
+        public init(_ name: String) {
```
patrickfreed, line 41: this could use a docstring explaining why it exists / why it's useful for forward compat. In general, we try to ensure that every public method / type / constant has a docstring.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -3,6 +3,9 @@ import Foundation
 
 /// A class to represent a MongoDB write concern.
 public struct WriteConcern: Codable {
+    /// Majority WriteConcern.
```
patrickfreed, line 6: This could be more specific, e.g. "A write concern with w: "majority" and the other fields unset."

Sources/MongoSwift/ReadConcern.swift
nbbeeken: I think the motivation for having an `other` method is so you can inline syntax like this:
`RunCommandOptions(readConcern: .other("blah"))`
instead of
`RunCommandOptions(readConcern: ReadConcern("blah"))`
Are we okay with the constructor/init syntax in this case?

Sources/MongoSwift/WriteConcern.swift
nbbeeken: I agree that would be cleaner, it does cause a linter error however (no `force_try`) I added a comment to avoid the error for now, do we wanna add skipping this rule globally?

Sources/MongoSwift/ReadConcern.swift
patrickfreed: I personally find the second one (using the initializer) more straightforward, though I'm curious to see what everyone thinks. 

cc @kmahar @mbroadst

Sources/MongoSwiftSync/MongoClient.swift
```diff
@@ -160,7 +160,7 @@ public class MongoClient {
      * Gets a `MongoDatabase` instance for the given database name. If an option is not specified in the optional
      * `DatabaseOptions` param, the database will inherit the value from the parent client or the default if
      * the client’s option is not set. To override an option inherited from the client (e.g. a read concern) with the
-     * default value, it must be explicitly specified in the options param (e.g. ReadConcern(), not nil).
+     * default value, it must be explicitly specified in the options param (e.g. ReadConcern.empty, not nil).
```
patrickfreed, line 163: this and similar comments in database will need to be updated to `.serverDefault` or whatever we settle on.

Sources/MongoSwift/WriteConcern.swift
patrickfreed: I like the link to the docs, but I think we still need to include information about the other fields. Calling it a "majority write concern" tells you about the `w` field, but it doesn't mention what the `j` or `wtimeout` fields are.

Also, you can replace "see" with `- SeeAlso:` to get nice formatting by our docs generator.

Sources/MongoSwift/WriteConcern.swift
patrickfreed: No, in most cases we never want to force try because it would crash a user application if it failed. This is okay here because it's in static space, so if our tests pass then we know it isn't failing.

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,69 @@ import CLibMongoC
...
-                return "snapshot"
-            case let .other(l):
-                return l
-            }
+        public static let snapshot = Level("snapshot")
+
+        private var name: String
+
+        /// Returns the string value of the Level for convenient conversion.
+        public var description: String { "\(self.name)" }
```
kmahar, line 51: can't we just return `name` here directly, why do we need string interpolation?

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,69 @@ import CLibMongoC
...
+        ReadConcern(level)
+    }
+
+    public static func other(_ level: String) -> ReadConcern {
+        ReadConcern(Level(level))
+    }
+
     /// An enumeration of possible ReadConcern levels.
-    public enum Level: RawRepresentable, Codable, Equatable {
+    public struct Level: Codable, Equatable, LosslessStringConvertible {
```
kmahar, line 32: do we need a public level type, or even a separate level type at that? it seems to me like we could just have `ReadConcern` store a string itself?

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -153,7 +153,7 @@ public struct MongoDatabase {
      * `MongoCollection` instance. If an option is not specified in the `CollectionOptions` param, the
      * collection will inherit the value from the parent database or the default if the db's option is not set.
      * To override an option inherited from the db (e.g. a read concern) with the default value, it must be explicitly
-     * specified in the options param (e.g. ReadConcern(), not nil).
+     * specified in the options param (e.g. ReadConcern.empty, not nil).
```
kmahar, line 156: also needs update to `serverDefault`

Sources/MongoSwift/MongoClient.swift
```diff
@@ -432,7 +432,7 @@ public class MongoClient {
      * Gets a `MongoDatabase` instance for the given database name. If an option is not specified in the optional
      * `DatabaseOptions` param, the database will inherit the value from the parent client or the default if
      * the client’s option is not set. To override an option inherited from the client (e.g. a read concern) with the
-     * default value, it must be explicitly specified in the options param (e.g. ReadConcern(), not nil).
+     * default value, it must be explicitly specified in the options param (e.g. ReadConcern.empty, not nil).
```
kmahar, line 435: should be updated to `serverDefault`, I think?

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -135,7 +135,7 @@ public struct MongoDatabase {
      * Access a collection within this database. If an option is not specified in the `CollectionOptions` param, the
      * collection will inherit the value from the parent database or the default if the db's option is not set.
      * To override an option inherited from the db (e.g. a read concern) with the default value, it must be explicitly
-     * specified in the options param (e.g. ReadConcern(), not nil).
+     * specified in the options param (e.g. ReadConcern.empty, not nil).
```
kmahar, line 138: also needs update to `serverDefault`

Sources/MongoSwift/ReadConcern.swift
```diff
@@ -2,53 +2,69 @@ import CLibMongoC
 
 /// A struct to represent a MongoDB read concern.
 public struct ReadConcern: Codable {
+    /// Local ReadConcern, see https://docs.mongodb.com/manual/reference/read-concern-local/.
```
kmahar, line 5: nit: can we make these seealso links, like:
```
/// Local read concern.
/// - SeeAlso: https://docs.mongodb.com/manual/reference/read-concern-local/

```

Sources/MongoSwift/ReadConcern.swift
kmahar: i think i prefer the `other` method to emphasize this is only for use in an edge case and to push people toward using the static properties whenever possible.

Sources/MongoSwift/WriteConcern.swift
patrickfreed: As another approach, we could take the path that we did with `ReadPreference` and make these static functions, optionally allowing the user to specify `wtimeout` and `j`.

Examples/Docs/Sources/SyncExamples/main.swift
```diff
@@ -12,8 +12,8 @@ private func causalConsistency() throws {
     let s1 = client1.startSession(options: ClientSessionOptions(causalConsistency: true))
     let currentDate = Date()
     var dbOptions = DatabaseOptions(
-        readConcern: ReadConcern(.majority),
-        writeConcern: try WriteConcern(w: .majority, wtimeoutMS: 1000)
+        readConcern: .majority,
+        writeConcern: try WriteConcern.majority(wtimeoutMS: 1000)
```
patrickfreed, line 16: I think this can still be `.majority(wtimeoutMS: 1000)`, though I could be wrong.

Sources/MongoSwift/ReadConcern.swift
patrickfreed: sgtm :+1:

Sources/MongoSwift/WriteConcern.swift
patrickfreed: >I like the link to the docs, but I think we still need to include information about the other fields. Calling it a "majority write concern" tells you about the w field, but it doesn't mention what the j or wtimeout fields are.

This portion of my comment still needs to be addressed. I also think the method could have some more info about how the user can optionally specify `j` or `wtimeout`, so as to differentiate it from the static property.

Examples/Docs/Sources/SyncExamples/main.swift
nbbeeken: Indeed it can! I wish the linter could pick up on this somehow, maybe someday

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -3,6 +3,28 @@ import Foundation
...
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/write-concern/#writeconcern._dq_majority_dq_
+    // swiftlint:disable force_try
+    public static let majority = try! WriteConcern(w: .majority)
+    // swiftlint:enable force_try
+
+    /**
+     * Returns a customized Majority WriteConcern.
+     *
+     * - Parameters:
+     *   - wtimeoutMS: This option specifies a time limit, in milliseconds, for the write concern.
```
patrickfreed, line 16: the official docs are a little lacking on this one, since they don't actually say what the time limit is for. We should expand on that to say "the maximum amount of time, in milliseconds, that the primary will wait for the write concern to be satisfied before returning a `WriteConcernError`."

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -3,6 +3,28 @@ import Foundation
...
+     *   - journal: requests acknowledgment that the mongod instances, have written to the on-disk journal.
```
patrickfreed, line 17: this comma is unnecessary

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -3,6 +3,28 @@ import Foundation
 
 /// A class to represent a MongoDB write concern.
 public struct WriteConcern: Codable {
+    /// Majority WriteConcern
```
patrickfreed, line 6: I don't mean to hammer the point, but I really think this docstring needs to be updated to mention that `wtimeout` and `j` are unset.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -3,6 +3,28 @@ import Foundation
...
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/write-concern/#writeconcern._dq_majority_dq_
+    // swiftlint:disable force_try
```
kmahar, line 8: I think the docstring won't get associated with the property if theres the comment in between. I think you can use // `swiftlint:disable:this` to put the disable on the same line as the definition, or `swiftlint:disable:previous` to put it on the line after.
also when we disable this rule we try to add a comment saying why it's ok.

Sources/MongoSwift/WriteConcern.swift
```diff
@@ -3,6 +3,28 @@ import Foundation
...
+     */
+    public static func majority(wtimeoutMS: Int? = nil, journal: Bool? = nil) throws -> WriteConcern {
+        try WriteConcern(journal: journal, w: .majority, wtimeoutMS: wtimeoutMS)
+    }
+
+    /// Server default WriteConcern.
+    public static let serverDefault = WriteConcern()
+
     /// An option to request acknowledgement that the write operation has propagated to specified mongod instances.
     public enum W: Codable, Equatable {
```
kmahar, line 29: i wonder if we should be worried about forward compatibility here 🤔 

whyyyy oh whyyyy must enums be frozen

patrickfreed: I think in this case we're actually okay, since it would technically be a breaking change on the server side to reserve a new write concern mode due to the custom mode accepting any string.

kmahar: oh ya, good point, I forgot tag is just an arbitrary string.

Sources/MongoSwift/WriteConcern.swift
nbbeeken: No worries about 🔨 I agree it can be clearer

kmahar:
looks like some merge conflicts were introduced by #468 but otherwise LGTM

Issue #448, opened 2020-04-12
`swift package update` fails with  `'mongoc' could not be found`


Rep2:
## Information

`swift package update` fails with `https://github.com/mongodb/swift-mongoc @ 2.0.0: error: manifest parse error: target 'mongoc' referenced in product 'mongoc' could not be found` 

Seems related to https://github.com/mongodb/swift-mongoc/issues/2

### Swift version

```
Apple Swift version 5.2 (swiftlang-1103.0.32.1 clang-1103.0.32.29)
Target: x86_64-apple-darwin19.3.0
```

Tested with multiple Swift versions 5.0 - 5.2.1.

### Operating system

macOS Catalina and Ubuntu 18.04

### Driver version

```
{
  "object": {
    "pins": [
      {
        "package": "Auth",
        "repositoryURL": "https://github.com/vapor/auth.git",
        "state": {
          "branch": null,
          "revision": "e6f61b403727ec124214beb3e57deff579f31d00",
          "version": "2.0.4"
        }
      },
      {
        "package": "Console",
        "repositoryURL": "https://github.com/vapor/console.git",
        "state": {
          "branch": null,
          "revision": "74cfbea629d4aac34a97cead2447a6870af1950b",
          "version": "3.1.1"
        }
      },
      {
        "package": "Core",
        "repositoryURL": "https://github.com/vapor/core.git",
        "state": {
          "branch": null,
          "revision": "1782c550512dc5a43d0bca405e28fc386d932bbf",
          "version": "3.10.0"
        }
      },
      {
        "package": "Crypto",
        "repositoryURL": "https://github.com/vapor/crypto.git",
        "state": {
          "branch": null,
          "revision": "df8eb7d8ae51787b3a0628aa3975e67666da936c",
          "version": "3.3.3"
        }
      },
      {
        "package": "DatabaseKit",
        "repositoryURL": "https://github.com/vapor/database-kit.git",
        "state": {
          "branch": null,
          "revision": "8f352c8e66dab301ab9bfef912a01ce1361ba1e4",
          "version": "1.3.3"
        }
      },
      {
        "package": "Fluent",
        "repositoryURL": "https://github.com/vapor/fluent.git",
        "state": {
          "branch": null,
          "revision": "783819d8838d15e1a05b459aa0fd1bde1e37ac26",
          "version": "3.2.1"
        }
      },
      {
        "package": "FluentSQLite",
[...]

PR #466, opened 2020-05-10
SWIFT-823 Use Swift 5.2-compatible way of including system libraries

kmahar:
Updates us to use the Swift 5.2-compatible way of linking to system libraries. Once we merge this in we can tag 0.3.0 as has been requested by users in #448.

Back when we tagged 0.2.0, we were not testing Swift 5.1 or 5.2 on Evergreen yet. In order to make a patch testing those I updated the versions in the config and then did this patch: 

https://evergreen.mongodb.com/version/5eb88a5f3066152a2dc1d163


kmahar:
alright that patch went horribly wrong because I didn't call `xcode-select` to switch to Xcode 11, and back then we still had that crazy problem with the sessions tests crashing the compiler now.

new patch including the necessary changes (just commented out the buggy sessions stuff, doesn't seem relevant to the changes we're making here): https://evergreen.mongodb.com/version/5eb88f350ae6062a81372c5d

patrickfreed:
It also looks like [this change](https://github.com/mongodb/mongo-swift-driver/commit/da86e1b45391c47d368b85cf2a7a3cb0216f5450#diff-06e3d33d9cd21e954fe14f92d1daaa6fR74) needs to be backported to fix the "latest" tests.

kmahar:
so I formatted/linted on the r0.3 branch but then bumping the Swift tools version here triggered a bunch of new linter violations.
new patch with those fixed + the test backport: https://evergreen.mongodb.com/version/5eb9e8b02fbabe5f91fa83e5

I'd like to tag 0.3.0 tomorrow, @mbroadst did you want to review this / will you have time to look at this by then?

PR #468, opened 2020-05-11
SWIFT-779 Safe access to connection pointer

kmahar:
Trying to finish up this ticket finally. This makes the `clientHandle` pointer underlying a `Connection` private, restricting access to a `withMongocConnection` method. Fortunately since we use `withMongocCollection` and `withMongocDatabase` in most places there wasn't a lot that needed to change here.

### DIFF
--- a/Sources/MongoSwift/Operations/ListDatabasesOperation.swift
+++ b/Sources/MongoSwift/Operations/ListDatabasesOperation.swift
@@ -73,19 +73,21 @@ internal struct ListDatabasesOperation: Operation {
         var reply = Document()
         var error = bson_error_t()
 
-        let success = cmd.withBSONPointer { cmdPtr in
-            readPref.withMongocReadPreference { rpPtr in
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/Operations/ListDatabasesOperation.swift
```diff
@@ -73,19 +73,21 @@ internal struct ListDatabasesOperation: Operation {
...
+                            mongoc_client_read_command_with_opts(
+                                connPtr,
+                                "admin",
+                                cmdPtr,
+                                rpPtr,
+                                optsPtr,
+                                replyPtr,
+                                &error
+                            )
+                        }
```
nbbeeken, line 90: \*_dreaming of pure swift_\* one day we won't have to nest this deep 😄

patrickfreed: ooh, I like how you can comment on ranges of lines now.

PR #463, opened 2020-05-06
SWIFT-774 Implement sharded transactions tests

patrickfreed:
[SWIFT-774](https://jira.mongodb.org/browse/SWIFT-774)

This PR adds support in the spec test runner for sharded transactions tests. It also refactors the runner to make skipping tests a little easier, repurposes the `FailPointConfigured` protocol to be more generally useful (it's now a requirement of the `SpecTest` protocol), and splits up `TestOperation.swift` into a subdirectory of different operations for easier code discovery.

The changes to `FailPointConfigured` will close out [SWIFT-755](https://jira.mongodb.org/browse/SWIFT-755) as part of this work.

PR #471, opened 2020-05-11
minor: refactor test operations

patrickfreed:
Opening a PR to get full ci run on subset of changes from #463


patrickfreed:
I ended up removing most of the `TestOperationResult` construction changes because they led to some wonky inference issues, so the changes here are restricted to the `TestOperation` protocol changes and changes resulting from that.

PR #472, opened 2020-05-13
SWIFT-779 Guard access to ClientSession internals

kmahar:
Final SWIFT-779 PR. This makes `ClientSession'`s state private and guards access to its underlying libmongoc session and connection through helpers.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -40,7 +40,7 @@ public final class ClientSession {
     )
 
     /// Enum for tracking the state of a session.
-    internal enum State {
+    private enum State {
         /// Indicates that this session has not been used yet and a corresponding `mongoc_client_session_t` has not
         /// yet been created. If the user sets operation time or cluster time prior to using the session, those values
         /// are stored here so they can be set upon starting the session.
@@ -53,7 +53,7 @@ public final class ClientSession {
     }
 
     /// Indicates the state of this session.
-    internal var state: State
+    private var state: State
 
     /// Returns whether this session is in the `started` state.
     internal var active: Bool {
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -197,9 +197,22 @@ public final class ClientSession {
     /// this session has already been ended.
     internal func startIfNeeded() -> EventLoopFuture<Void> {
         switch self.state {
-        case .notStarted:
+        case let .notStarted(opTime, clusterTime):
```
kmahar, line 200: as part of making the `State` private, I moved this logic directly into here rather than having it live in `StartSessionOperation`. That operation's work is now strictly limited to calling the corresponding libmongoc method (which blocks as it checks for sessions support, hence it must be an operation.)

Sources/MongoSwift/Operations/Operation.swift
```diff
@@ -86,13 +84,6 @@ internal class OperationExecutor {
         }
 
         if let session = session {
-            if case .ended = session.state {
```
kmahar, line 89: this logic is already handled in `session.getConnection`. I think it's overly defensive to check in both places. the only upside of checking here is that we find out a little sooner. but these are both edge cases/user error so I'm not sure it's worth optimizing for them.

Sources/MongoSwift/Operations/StartSessionOperation.swift
```diff
@@ -48,12 +48,10 @@ internal struct StartSessionOperation: Operation {
         self.session = session
     }
 
-    internal func execute(using connection: Connection, session _: ClientSession?) throws {
-        // session was already started
-        guard case let .notStarted(opTime, clusterTime) = self.session.state else {
```
kmahar, line 53: we already check the state before we fire off this operation. assuming users aren't concurrently using the session across threads (not allowed) there's no way for it to change between then and when execution happens here.

Issue #458, opened 2020-04-27
bson_malloc memory leak


valeriomazzeo:
It seems that some Documents are not released from within the `CreateIndexesOperation`.

![Screenshot 2020-04-27 at 15 17 57](https://user-images.githubusercontent.com/1882080/80382922-d3531500-889a-11ea-95d2-03229b7983a5.png)

### Swift version

What does this command give you?
```
$ swift --version
Apple Swift version 5.1.3 (swiftlang-1100.0.282.1 clang-1100.0.33.15)
Target: x86_64-apple-darwin19.4.0
```
### Operating system

What does this command give you?
```
$ uname -a
Darwin MBP 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
```
### Driver version

What does this command give you?
```
$ cat Package.resolved # Applies if you are using Swift package manager
{
        "package": "bson",
        "repositoryURL": "https://github.com/mongodb/swift-bson",
        "state": {
          "branch": null,
          "revision": "40f770ed265bb9958edbcd0c809a3c2afe75a3ba",
          "version": "2.1.0"
        }
},
{
        "package": "MongoSwift",
        "repositoryURL": "https://github.com/mongodb/mongo-swift-driver.git",
        "state": {
          "branch": null,
          "revision": "1b38ca8987d1938e0c9245f959fa2172b13cba60",
          "version": "0.2.0"
        }
}
```

### What is the version(s) of `mongod` that you are running with the driver?
Try running:
```
$ mongod --version
4.2.2
```

kmahar:
Hi @valeriomazzeo, thanks for the report!

Could you provide any more info on the code that triggered this, or a code sample? We have a number of index creation methods and it would be useful to know e.g. which is being used here, which options were specified for the command if any, etc.

Also, I don't develop in Xcode much so I'm not quite sure how to get to the view / tool you've shown in the screenshot, any more details on that would be very helpful.

Thanks!

valeriomazzeo:
The trigger from my driver should be [this one](https://github.com/asensei/vapor-fluent-mongo/blob/master/Sources/FluentMongo/MongoConnection.swift#L177), which leads to `MongoCollection+Indexes.swift` / `createIndexes` with model.

```swift
    @discardableResult
    public func createIndex(
        _ model: IndexModel,
        options: CreateIndexOptions? = nil,
[...]

PR #474, opened 2020-05-15
SWIFT-824 Fix leaks of reply bson_ts

kmahar:
The issue causing the leaks reported in #458 is that we were supposed to be passing in stack-allocated `bson_t`s anywhere libmongoc accepts a `reply` parameter. Instead, we were passing in heap-allocated `bson_t`s, whose original memory would be leaked.

I've solved this via a new helper method, `withStackAllocatedMutableBSONPointer`, which will create a stack-allocated `bson_t`, execute the provided closure, and then call `bson_destroy` to free it. In a number of places where we used to wrongly call `withMutableBSONPointer`, we now use this helper instead.

Most of the places I was fixing call through to libmongoc's `run_command` variants, and I noticed there was a lot of duplication between them so I introduced some helpers to clean those up as well. Likely we can clean up the operations + nested closure situation a lot more via introduction of further helpers such as these, but (relevant to our conversation in drivers meeting earlier, lol) I limited my refactoring to places I was already touching in this PR for now.

### DIFF
--- a/Sources/MongoSwift/Operations/MongocCommandHelpers.swift
+++ b/Sources/MongoSwift/Operations/MongocCommandHelpers.swift
@@ -1,0 +1,44 @@
+import CLibMongoC
+
+/// Signature for a Swift closure that wraps a mongoc run_command variant.
+internal typealias MongocCommandFunc =
+    (_ command: BSONPointer, _ opts: BSONPointer?, _ reply: MutableBSONPointer, _ error: inout bson_error_t) -> (Bool)
+
+/// Calls the provided mongoc command method using pointers to the specified command and options. Returns the resulting
+/// reply document from the server. If you do not need to use the reply document, `runMongocCommand` is preferable as
+/// it does not perform a copy of the reply.
+internal func runMongocCommandWithReply(
+    command: Document,
+    options: Document?,
+    body: MongocCommandFunc
+) throws -> Document {
+    try withStackAllocatedMutableBSONPointer { replyPtr in
+        try _runMongocCommand(command: command, options: options, replyPtr: replyPtr, body: body)
+        return Document(copying: replyPtr)
+    }
+}
+
+/// Calls the provided mongoc command method using pointers to the specified command and options.
+internal func runMongocCommand(command: Document, options: Document?, body: MongocCommandFunc) throws {
+    try withStackAllocatedMutableBSONPointer { replyPtr in
+        try _runMongocCommand(command: command, options: options, replyPtr: replyPtr, body: body)
+    }
+}
+
+/// Private helper to run the provided `MongocCommandFunc` using the specified location for a reply.
+private func _runMongocCommand(
+    command: Document,
+    options: Document?,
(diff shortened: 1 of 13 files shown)


Sources/MongoSwift/Operations/MongocCommandHelpers.swift
```diff
@@ -0,0 +1,58 @@
+import CLibMongoC
+
+/// Executes the provided closure using a stack-allocated, mutable bson_t. The bson_t is only valid for the body of the
+/// closure and must be copied if you wish to use it later on.
+internal func withStackAllocatedMutableBSONPointer<T>(body: (MutableBSONPointer) throws -> T) rethrows -> T {
+    var bson = bson_t()
+    defer {
+        withUnsafeMutablePointer(to: &bson) { ptr in
```
kmahar, line 8: as far as I can tell, there is no functional difference between using this `withUnsafeMutablePointer` helper and just using `&` directly when interoperating with C methods. however, I went with this approach as it seems more explicit.

patrickfreed:
One thought I just had--is there any way to incorporate the xcode memory leak analysis into CI? Or does it have to be done through the GUI? Also, does something like valgrind detect the same leaks?

kmahar:
That's a very good point. I can't find much info on this so I made a forums post: https://forums.swift.org/t/test-for-memory-leaks-in-ci/36526

Sources/MongoSwift/Operations/MongocCommandHelpers.swift
```diff
@@ -0,0 +1,58 @@
+import CLibMongoC
+
+/// Executes the provided closure using a stack-allocated, mutable bson_t. The bson_t is only valid for the body of the
+/// closure and must be copied if you wish to use it later on.
```
kevinAlbs, line 4: IIUC It may be worth noting that the closure must initialize the bson_t, otherwise the deferred `bson_destroy` would be accessing uninitialized memory (unless `bson_t()` also initializes).

kmahar: So according to [this](https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/using_imported_c_structs_and_unions_in_swift):
> If all imported members have default values, Swift also provides a default initializer that takes no arguments.

that's what we're using here. using the debugger it appears this just creates a totally zeroed-out `bson_t`. so yes, I believe we do have to initialize the `bson_t` in the closure.

kmahar:
Thanks for the info @kevinAlbs !

Apparently the clang/LLVM shipped with MacOS don't support ASAN leak detection. We can use ASAN via `--sanitize=address` but not that feature... Also apparently ASAN leak detection has a lot of [false positives with Swift](https://forums.swift.org/t/test-for-memory-leaks-in-ci/36526/9), sigh.

It doesn't seem like we have a quick and easy solution here so I opened [SWIFT-865](https://jira.mongodb.org/browse/SWIFT-865) about investigating this further when we have a chance. I'm going to go ahead and merge this.

PR #475, opened 2020-05-19
Avoid ConnectionPool leak in case of invalid options

kmahar:
I discovered yesterday that if you created a `MongoClient` with invalid `maxPoolSize` we would leak the `mongoc_client_pool_t`. This fixes that issue.

### DIFF
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -76,6 +76,15 @@ internal class ConnectionPool {
 
     /// Initializes the pool using the provided `ConnectionString` and options.
     internal init(from connString: ConnectionString, options: ClientOptions?) throws {
+        // validate option before we bother creating pool, so we don't have to destroy it on error.
+        if let maxPoolSize = options?.maxPoolSize {
+            guard maxPoolSize > 0 && maxPoolSize <= UInt32.max else {
+                throw InvalidArgumentError(
+                    message: "Invalid maxPoolSize \(maxPoolSize): must be between 1 and \(UInt32.max)"
+                )
+            }
+        }
+
         let pool: OpaquePointer = try connString.withMongocURI { uriPtr in
             guard let pool = mongoc_client_pool_new(uriPtr) else {
                 throw InternalError(message: "Failed to initialize libmongoc client pool")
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -76,27 +76,32 @@ internal class ConnectionPool {
 
     /// Initializes the pool using the provided `ConnectionString` and options.
     internal init(from connString: ConnectionString, options: ClientOptions?) throws {
+        // validate option before we bother creating pool, so we don't have to destroy it on error.
```
kmahar, line 79: I wanted to avoid calling `mongoc_client_pool_destroy` from the init. I *think* it would technically be ok because libmongoc wouldn't have started SDAM and wouldn't try to send `endSession` or anything but I don't know if we should rely on that. validating the option before pool creation seemed like the easiest way to avoid this.

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -76,27 +76,32 @@ internal class ConnectionPool {
...
-            guard maxPoolSize > 0 && maxPoolSize <= UInt32.max else {
-                throw InvalidArgumentError(
-                    message: "Invalid maxPoolSize \(maxPoolSize): must be between 1 and \(UInt32.max)"
-                )
-            }
             mongoc_client_pool_max_size(pool, UInt32(maxPoolSize))
         }
 
-        self.state = .open(pool: pool)
+        guard mongoc_client_pool_set_error_api(pool, MONGOC_ERROR_API_VERSION_2) else {
```
kmahar, line 101: I made this change to let us get away with not having to clean up the pool here due to the concerns noted in my previous comment...
this method only returns false if the error API is invalid. given that libmongoc will never break things, and we control the version now, I don't think it's possible that we would pass in an invalid error API version here. it was of more concern in the past because users controlled their libmongoc version and could plausibly try to use the driver with an old version without this API (though I'm not sure if the driver would have compiled with a version that old anyway since we use a lot of newer API methods.)

Sources/MongoSwift/ConnectionPool.swift
patrickfreed: I think a more future proof way of handling this might be to wrap the whole thing in a `do { ... } catch { ... }`, where we clean up the pool and then rethrow the error in the `catch`. This way we don't really have to worry about the ordering of things. What do you think?

kmahar: I think that still suffers from the issue of having to call potentially blocking code from this initializer to do cleanup, though?

patrickfreed: Oh, I completely misunderstood your comment. This makes sense now. Maybe we should add a comment saying that once the pool has been created, we should not throw to avoid leaking / blocking?

kmahar: I added a bit more context to the comment here about why calling the destroy method would be undesirable

PR #476, opened 2020-05-20
SWIFT-850 Replace all occurrences of Id with ID

kmahar:
This ticket was originally about `ObjectId` but `Id` actually comes up a lot in our public API so I replaced all occurrences.


patrickfreed:
Actually, it looks like the failures might be related. I think a result type stopped decoding properly because the result libmongoc gives us has `insertedId`in it.

kmahar:
fixed, I needed to add `CodingKeys` to `InsertOneResult` to fix the capitalization

PR #477, opened 2020-05-20
SWIFT-858: Duplicate pure Swift BSON breaking changes in driver

nbbeeken:
All changes listed in ticket:
https://jira.mongodb.org/browse/SWIFT-858

Additionally changed:
- `BSONType` is `UInt8` instead of UInt32

### DIFF
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -230,7 +230,7 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
         /// Initializes a `Subtype` with a custom value.
         /// Returns nil if rawValue within reserved range [0x07, 0x80).
         public init?(rawValue: UInt8) {
-            if rawValue > 0x06 && rawValue < 0x80 {
+            guard !(rawValue > 0x06 && rawValue < 0x80) else {
                 return nil
             }
             self.rawValue = rawValue
@@ -246,7 +246,7 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
                 throw InvalidArgumentError(message: "Cannot represent \(value) as UInt8")
             }
             guard byteValue >= 0x80 else {
-                throw InvalidArgumentError(message: "userDefined value must be greater than 0x80 got \(byteValue)")
+                throw InvalidArgumentError(message: "userDefined value must be greater than or equal to 0x80 got \(byteValue)")
             }
             guard let subtype = Subtype(rawValue: byteValue) else {
                 throw InvalidArgumentError(message: "Cannot represent \(byteValue) as Subtype")
--- a/Sources/MongoSwift/BSON/BSONValue.swift
+++ b/Sources/MongoSwift/BSON/BSONValue.swift
@@ -1,8 +1,12 @@
 import CLibMongoC
 import Foundation
+import NIO
+
+/// This shared allocator instance should be used for all underlying `ByteBuffer` creation.
+private let BSON_ALLOCATOR = ByteBufferAllocator()
 
 /// The possible types of BSON values and their corresponding integer values.
-public enum BSONType: UInt32 {
+public enum BSONType: UInt8 {
     /// An invalid type
     case invalid = 0x00
     /// 64-bit binary floating point
@@ -196,27 +200,36 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
     internal var bson: BSON { .binary(self) }
 
     /// The binary data.
-    public let data: Data
+    public let data: ByteBuffer
 
     /// The binary subtype for this data.
-    public let subtype: UInt8
+    public let subtype: Subtype
 
     /// Subtypes for BSON Binary values.
-    public enum Subtype: UInt8 {
+    public struct Subtype: Equatable, Codable, Hashable {
         /// Generic binary subtype
-        case generic,
-            /// A function
-            function,
-            /// Binary (old)
-            binaryDeprecated,
-            /// UUID (old)
-            uuidDeprecated,
-            /// UUID (RFC 4122)
-            uuid,
-            /// MD5
-            md5,
-            /// User defined
-            userDefined = 0x80
+        public static let generic = Subtype(0x00)
+        /// A function
+        public static let function = Subtype(0x01)
+        /// Binary (old)
+        public static let binaryDeprecated = Subtype(0x02)
+        /// UUID (old)
+        public static let uuidDeprecated = Subtype(0x03)
+        /// UUID (RFC 4122)
+        public static let uuid = Subtype(0x04)
+        /// MD5
+        public static let md5 = Subtype(0x05)
+        /// Encrypted BSON value
+        public static let encryptedValue = Subtype(0x06)
+
+        public let rawValue: UInt8
+
+        private init(_ value: UInt8) { self.rawValue = value }
+        internal init(_ value: bson_subtype_t) { self.rawValue = UInt8(value.rawValue) }
+
+        /// Create BSON Binary Subtype
+        /// Note: subtype 0x80-0xFF "User defined" subtype
+        public static func other(_ value: Int) -> Subtype { Subtype(UInt8(value)) }
     }
 
     /// Initializes a `BSONBinary` instance from a `UUID`.
@@ -238,8 +251,8 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
     /// Initializes a `BSONBinary` instance from a `Data` object and a `UInt8` subtype.
     /// - Throws:
     ///   - `InvalidArgumentError` if the provided data is incompatible with the specified subtype.
-    public init(data: Data, subtype: UInt8) throws {
-        if [Subtype.uuid.rawValue, Subtype.uuidDeprecated.rawValue].contains(subtype) && data.count != 16 {
+    public init(data: Data, subtype: Subtype) throws {
+        if [Subtype.uuid, Subtype.uuidDeprecated].contains(subtype) && data.count != 16 {
             throw InvalidArgumentError(
                 message:
                 "Binary data with UUID subtype must be 16 bytes, but data has \(data.count) bytes"
@@ -253,14 +266,16 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
     /// - Throws:
     ///   - `InvalidArgumentError` if the provided data is incompatible with the specified subtype.
     public init(data: Data, subtype: Subtype) throws {
-        try self.init(data: data, subtype: subtype.rawValue)
+        var buffer = BSON_ALLOCATOR.buffer(capacity: data.count)
+        buffer.writeBytes(data)
+        self.data = buffer
     }
 
-    /// Initializes a `BSONBinary` instance from a base64 `String` and a `UInt8` subtype.
+    /// Initializes a `Binary` instance from a base64 `String` and a `Subtype`.
     /// - Throws:
     ///   - `InvalidArgumentError` if the base64 `String` is invalid or if the provided data is
     ///     incompatible with the specified subtype.
-    public init(base64: String, subtype: UInt8) throws {
+    public init(base64: String, subtype: Subtype) throws {
         guard let dataObj = Data(base64Encoded: base64) else {
             throw InvalidArgumentError(
                 message:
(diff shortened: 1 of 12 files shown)


Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -812,9 +805,9 @@ public struct ObjectId: BSONValue, Equatable, CustomStringConvertible, Codable {
 
     /// Initializes an `ObjectId` from the provided hex `String`. Returns `nil` if the string is not a valid ObjectId.
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
-    public init?(_ hex: String) {
+    public init(_ hex: String) throws {
         guard bson_oid_is_valid(hex, hex.utf8.count) else {
-            return nil
+            throw InternalError(message: "Cannot create ObjectId from \(hex)")
```
kmahar, line 810: I think this should be an `InvalidArgumentError`, as it means the user gave us invalid input.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -196,27 +200,36 @@ public struct Binary: BSONValue, Equatable, Codable, Hashable {
...
+        public static let md5 = Subtype(0x05)
+        /// Encrypted BSON value
+        public static let encryptedValue = Subtype(0x06)
+
+        public let rawValue: UInt8
+
+        private init(_ value: UInt8) { self.rawValue = value }
+        internal init(_ value: bson_subtype_t) { self.rawValue = UInt8(value.rawValue) }
+
+        /// Create BSON Binary Subtype
```
kmahar, line 230: I think maybe `custom` or `userDefined` would be a better name than `other`. 

I also think we could make this throw if the subtype is not in 0x80-0xFF.

separately, let's make this docstring a bit more formal, e.g.
```
/// Initializes a `Subtype` with a custom value. This value must be in the range 0x80-0xFF.
/// Throws: ...
```

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -196,27 +200,36 @@ public struct Binary: BSONValue, Equatable, Codable, Hashable {
...
+        /// UUID (old)
+        public static let uuidDeprecated = Subtype(0x03)
+        /// UUID (RFC 4122)
+        public static let uuid = Subtype(0x04)
+        /// MD5
+        public static let md5 = Subtype(0x05)
+        /// Encrypted BSON value
+        public static let encryptedValue = Subtype(0x06)
+
+        public let rawValue: UInt8
```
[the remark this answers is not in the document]
nbbeeken: Yea I've been yelled at by the linter before for public undocumented things, idk why it missed this. I think you're right, I guess the other way they would have to inspect would be to wrap their type number in a subtype before testing `==` seems like rawValue is the nicer way out there.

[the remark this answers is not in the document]
kmahar: I wonder if we should just make `Subtype` conform to `RawRepresentable` ?

Tests/BSONTests/BSONValueTests.swift
```diff
@@ -209,20 +206,28 @@ final class BSONValueTests: MongoSwiftTestCase {
...
+            ),
+            BSONNumberTestCase(int: 0, double: 0.0, int32: Int32(0), int64: Int64(0), decimal: try Decimal128("0")),
+            BSONNumberTestCase(int: nil, double: 1.234, int32: nil, int64: nil, decimal: try Decimal128("1.234")),
+            BSONNumberTestCase(int: nil, double: -31.234, int32: nil, int64: nil, decimal: try Decimal128("-31.234"))
         ]
 
         cases.forEach { $0.run() }
     }
+
+    func testBSONDinarySubtype() {
```
kmahar, line 228: Dinary -> Binary? 🙂 

(also I think you need to run `make linuxmain`)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -812,9 +812,9 @@ public struct ObjectID: BSONValue, Equatable, CustomStringConvertible, Codable {
 
     /// Initializes an `ObjectID` from the provided hex `String`. Returns `nil` if the string is not a valid ObjectID.
```
kmahar, line 813: we need to update this docstring to indicate that we throw rather than returning nil.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -469,13 +478,9 @@ public struct Decimal128: BSONValue, Equatable, Codable, CustomStringConvertible
     /// Initializes a `Decimal128` value from the provided `String`. Returns `nil` if the input is not a valid
```
kmahar, line 478: we also need to update this docstring.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -196,27 +200,44 @@ public struct Binary: BSONValue, Equatable, Codable, Hashable {
...
+        /// MD5
+        public static let md5 = Subtype(0x05)
+        /// Encrypted BSON value
+        public static let encryptedValue = Subtype(0x06)
+
+        /// Subtype indicator value
+        public var rawValue: UInt8
+
+        public init(_ value: UInt8) { self.rawValue = value }
+        public init?(rawValue: UInt8) { self.rawValue = rawValue }
```
kmahar, line 229: sorry for the back and forth but I think maybe we should omit `RawRepresentable` conformance because the public initializers here allows users to bypass the validation of their custom subtype value. This is not great because a user could start using eg 0x07 as a custom subtype and then we could add a conflicting subtype in a future server version. 0x07-0x7F are reserved for future non-custom subtypes.

nbbeeken: Yea I think it would make more sense to not conform, I'll revert

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -196,27 +200,43 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
...
+        public static let uuidDeprecated = Subtype(0x03)
+        /// UUID (RFC 4122)
+        public static let uuid = Subtype(0x04)
+        /// MD5
+        public static let md5 = Subtype(0x05)
+        /// Encrypted BSON value
+        public static let encryptedValue = Subtype(0x06)
+
+        /// Subtype indicator value
+        public var value: UInt8
```
kmahar, line 226: I think this should be a `let` since we're restricting construction ofthis type to the constants and `userDefined` method

Tests/BSONTests/CodecTests.swift
```diff
@@ -297,21 +297,21 @@ final class CodecTests: MongoSwiftTestCase {
...
                 date: Date(timeIntervalSinceReferenceDate: 5000),
                 code: BSONCode(code: "hi"),
                 codeWithScope: BSONCodeWithScope(code: "hi", scope: ["x": .int64(1)]),
                 ts: BSONTimestamp(timestamp: 1, inc: 2),
                 int32: 5,
                 int64: 6,
-                dec: BSONDecimal128("1.2E+10")!,
-                minkey: BSONMinKey(),
+                dec: try BSONDecimal128("1.2E+10"),
+                minkey: MinKey(),
```
kmahar, line 309: should be `BSONMinKey`

Tests/BSONTests/CodecTests.swift
```diff
@@ -297,21 +297,21 @@ final class CodecTests: MongoSwiftTestCase {
...
                 maxkey: MaxKey(),
                 regex: BSONRegularExpression(pattern: "^abc", options: "imx"),
                 symbol: BSONSymbol("i am a symbol"),
                 undefined: BSONUndefined(),
-                dbpointer: BSONDBPointer(ref: "some.namespace", id: BSONObjectID("507f1f77bcf86cd799439011")!),
+                dbpointer: DBPointer(ref: "some.namespace", id: try BSONObjectID("507f1f77bcf86cd799439011")),
```
kmahar, line 314: should be `BSONDBPointer`

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -831,12 +842,13 @@ public struct BSONObjectID: BSONValue, Equatable, CustomStringConvertible, Codab
...
-    /// ObjectID.
+    /// Initializes an `BSONObjectID` from the provided hex `String`.
+    /// - Throws:
+    ///   - `InvalidArgumentError` if string passed is not a valid BSONObjectID
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
-    public init?(_ hex: String) {
+    public init(_ hex: String) throws {
         guard bson_oid_is_valid(hex, hex.utf8.count) else {
-            return nil
+            throw InvalidArgumentError(message: "Cannot create ObjectId from \(hex)")
```
patrickfreed, line 851: did we ever decide what we were going to do with BSON errors? As written here, we'd need to catch and rethrow any BSON errors as driver errors.

Sources/MongoSwift/BSON/BSONValue.swift
patrickfreed: Sorry to reignite this conversation, but could we not just return `nil` if a user specified a reserved but unused value? From the docs of `RawRepresentable`, this seems like the ideal case to conform to it (i.e. modeling a C style enum).

kmahar: that is a good point, didn't think of that lol. i'm fine with returning nil for reserved subtypes

Sources/MongoSwift/BSON/BSONValue.swift
kmahar: I thought we decided on catch and rethrow but I could be remembering wrong.

nbbeeken: I think catch and rethrow is what we gotta do to make swift-bson also be its own standalone library if that use case arises

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -196,27 +200,53 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
...
+
+        internal init(_ value: bson_subtype_t) { self.rawValue = UInt8(value.rawValue) }
+
+        private init(_ value: Int) { self.rawValue = UInt8(value) }
+
+        /// Initializes a `Subtype` with a custom value. This value must be in the range 0x80-0xFF.
+        /// - Throws:
+        ///   - `InvalidArgumentError` if value passed is outside of the range 0x80-0xFF
+        public static func userDefined(_ value: Int) throws -> Subtype {
+            guard value >= 0x80 && value <= 0xFF else {
```
patrickfreed, line 245: we can replace the upper bound check of this with a "cast" to `UInt8` via `UInt8.init(exactly:)`. This will also let us get rid of the private `Int` based initializer above.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -196,27 +200,53 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
...
+        /// Encrypted BSON value
+        public static let encryptedValue = Subtype(0x06)
+
+        /// Subtype indicator value
+        public let rawValue: UInt8
+
+        /// Initializes a `Subtype` with a custom value.
+        /// Returns nil if rawValue outside of the range 0x80-0xFF.
+        public init?(rawValue: UInt8) {
+            guard rawValue >= 0x80 && rawValue <= 0xFF else {
```
patrickfreed, line 231: This validation should only omit the values that are reserved but not allocated (e.g. `[0x07, 0x80)`)

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -207,45 +207,51 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
...
 
         /// Subtype indicator value
         public let rawValue: UInt8
 
         /// Initializes a `Subtype` with a custom value.
-        /// Returns nil if rawValue outside of the range 0x80-0xFF.
+        /// Returns nil if rawValue within reserved range [0x07, 0x80).
         public init?(rawValue: UInt8) {
-            guard rawValue >= 0x80 && rawValue <= 0xFF else {
+            if rawValue > 0x06 && rawValue < 0x80 {
```
patrickfreed, line 233: since this is returning early, I think its probably best written as a `guard`.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -207,45 +207,51 @@ public struct BSONBinary: BSONValue, Equatable, Codable, Hashable {
...
         /// - Throws:
         ///   - `InvalidArgumentError` if value passed is outside of the range 0x80-0xFF
         public static func userDefined(_ value: Int) throws -> Subtype {
-            guard value >= 0x80 && value <= 0xFF else {
-                throw InvalidArgumentError(message: "User defined Binary Subtypes must be between 0x80 and 0xFF")
+            guard let byteValue = UInt8(exactly: value) else {
+                throw InvalidArgumentError(message: "Cannot represent \(value) as UInt8")
+            }
+            guard byteValue >= 0x80 else {
+                throw InvalidArgumentError(message: "userDefined value must be greater than 0x80 got \(byteValue)")
```
patrickfreed, line 249: "greater than _or equal to_ 0x80"

PR #480, opened 2020-05-20
SWIFT-822 Prefix various type names to lessen likelihood of collisions

kmahar:
This PR makes the following changes:

#### Driver Types
- `Address` -> `ServerAddress`
- `Hint` -> `IndexHint`
- `ClientOptions` -> `MongoClientOptions`
- `DatabaseOptions` -> `MongoDatabaseOptions`
- `CollectionOptions` -> `MongoCollectionOptions`
- `CursorType` -> `MongoCursorType`

#### BSON Types
- `ObjectID` -> `BSONObjectID`
- `Timestamp` -> `BSONTimestamp`
- `Code` -> `BSONCode`
- `CodeWithScope` -> `BSONCodeWithScope`
- `Binary` -> `BSONBinary`
- `RegularExpression` -> `BSONRegularExpression`
- `Symbol` -> `BSONSymbol`

#### Types I did *not* prefix but we could prefix:
- `Document`
- `Decimal128`
- `DBPointer`

Arguably, we should just have all of our BSON types prefixed. But `Decimal128` and `DBPointer` don't seem very likely to clash to me, and `Document` would be a huge diff so I figured regardless I should leave that to a follow-up PR.

Thoughts?


Sources/MongoSwift/BSON/BSON.swift
```diff
@@ -307,20 +307,20 @@ extension BSON {
...
         MaxKey.self,
-        Symbol.self,
+        BSONSymbol.self,
         Double.self,
         String.self,
         Document.self,
-        Binary.self,
-        ObjectID.self,
+        BSONBinary.self,
+        BSONObjectID.self,
```
mbroadst, line 315: I feel like this is the most bikeshedy of bikeshedy territories, but a few things occur to me:
- `ObjectID` is unlikely to clash
- You've omitted `MinKey` and `MaxKey` which seem like they have equal "clash quotient"
- `Document` actually has quite a high "clash quotient" IMO
- I really dislike all the `BSON` prefixes on paper, but I think it may be argued similar to the JavaScript story that you're unlikely to really ever see these types in your daily programming. If we're prefixing some of them, should we just prefix _all_ of them? 

Does clashing even matter? Something like `Index` => `IndexHint`, or `Server` => `ServerAddress` are acceptable changes because they provide more contextual information while reducing the potential for collisions. Does the same hold true for `Timestamp`? Maybe.

Sources/MongoSwift/BSON/Document.swift
```diff
@@ -246,7 +246,7 @@ extension Document {
             return self
         }
 
-        var idDoc: Document = ["_id": .objectID(ObjectID())]
+        var idDoc: Document = ["_id": .objectID(BSONObjectID())]
```
mbroadst, line 249: I don't mean to re-open a can of worms here, but this looks cumbersome to me. Was it not possible to have something like:
```
var idDoc: Document = ["_id": .objectID(1) | .objectID("hexvalue") | .objectID()]
```
?

I understand changing the enum requires breaking changes, but the inputs for creating an `ObjectID` are unlikely to change, and you could still accept an `ObjectID` instance for those special cases (if they exist).

Sources/MongoSwift/BSON/BSON.swift
patrickfreed: >If we're prefixing some of them, should we just prefix all of them?

I also dislike the prefixing on paper, but I tend to agree with this. I think it could be surprising for users that some types aren't prefixed, even if those types are very unlikely to clash.

>Does clashing even matter?

Unfortunately the clashing matters a lot. If a user happens to add a new dependency that clashes, all their old code breaks and will need to be updated to be namespace prefixed. Worse yet, if a clashing type is adding to an _existing_ dependency in a minor version, it could break existing code.

e.g. The following would break if B introduces `Dog` in 1.1.
```swift
import A // 1.0
import B // 1.0

func main() {
    let x = Dog() // From A
}
```

kmahar: `MinKey` and `MaxKey` are `internal` types - they're just a place for the decoding/encoding logic for those to live, but users only interact with them via the `BSON` enum cases `.minKey` and `.maxKey`. So I didn't bother with prefixing those as the clashing issue doesn't apply, but we could prefix them for consistency internally anyway.

agree `Document` is one that is likely to clash - as I noted in the description I just left that off of this PR as it's a huge diff and I figured worth putting into a follow-up PR.

I'm fine with the all or nothing approach to prefixing on public types anyway.

yeah, the clashing issue is pretty frustrating, and makes semver feel a lot less useful. it's impossible to solve completely as someone *could* theoretically reuse any type name you come up with, no matter how unique, but I think we have to at least do our best to prevent issues with it.

Sources/MongoSwift/BSON/Document.swift
kmahar: to address each example you've given - 
* `"_id": .objectID(1)`: 1 isn't a valid ObjectID, so I'm not sure what you would expect the behavior to be there?
* `.objectID("hexvalue")`: we could actually support this by making `ObjectID` conform to `ExpressibleByStringLiteral`. but we lose the ability to validate your input string since we can't throw from that constructor. this is kind of problematic because in the case of constructing a document via dictionary literal I think we'd have to just fatal error if your input is invalid and can't be encoded to BSON.
* `.objectID()`: since Swift 5.1 now allows you to provide default values for enum associated values, we can do this easily by updating the case from `.objectID(ObjectID)` to `.objectID(ObjectID = ObjectID())`.

I don't love the `ExpressibleByStringLiteral` change for the reason I mentioned about validation, but the last one (adding default value) I am definitely in favor of.

mbroadst: cool, yes that default value approach is basically what I'm looking for!

Sources/MongoSwift/BSON/BSON.swift
mbroadst: OK, prefix all the things it is I guess :man_facepalming:

Sources/MongoSwift/BSON/Document.swift
kmahar: added. opened [SWIFT-870](https://jira.mongodb.org/browse/SWIFT-870) for changelog purposes

Sources/MongoSwift/BSON/BSON.swift
kmahar: kk, I'll do the remaining types in a second PR

PR #482, opened 2020-05-21
SWIFT-822 Prefix remaining BSON types

kmahar:
Leftovers from the last PR. Adds prefixes to `Document`, `DocumentIteraror`, `Decimal128`, and `DBPointer`.
Also prefixed `MinKey` and `MaxKey` for consistency, though they are `internal` types.


kmahar:
I'm making a pass through all of our guides right now and it's turning into a rather significant endeavor as a lot of stuff is outdated, so I'll handle that in a separate PR

kmahar:
update: put up #486 containing guides and examples changes.

PR #485, opened 2020-05-26
SWIFT-827: Rewrite enums with ".other" case as structs

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-827

`CollectionType` is out of scope (we discussed that since it's an administrative api we can leave it as is)
That only leaves `FullDocument` which is now a RawRepresentable struct with a public static other function

### DIFF
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -86,7 +86,7 @@ private func changeStreams() throws {
         let inventory = db.collection("inventory")
 
         // Option 1: use next() to iterate
-        let next = inventory.watch(options: ChangeStreamOptions(fullDocument: .other("blah")))
+        let next = inventory.watch(options: ChangeStreamOptions(fullDocument: .updateLookup))
             .flatMap { changeStream in
                 changeStream.next()
             }
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -86,7 +86,7 @@ private func changeStreams() throws {
         let inventory = db.collection("inventory")
 
         // Option 1: use next() to iterate
-        let next = inventory.watch(options: ChangeStreamOptions(fullDocument: .updateLookup))
+        let next = inventory.watch(options: ChangeStreamOptions(fullDocument: .other("blah")))
             .flatMap { changeStream in
                 changeStream.next()
             }
--- a/Sources/MongoSwift/ChangeStreamOptions.swift
+++ b/Sources/MongoSwift/ChangeStreamOptions.swift
@@ -11,6 +11,7 @@ public struct FullDocument: RawRepresentable, Codable {
 
     public var rawValue: String
 
+    /// Creates a `FullDocument`. Never returns nil.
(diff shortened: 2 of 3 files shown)


Examples/Docs/Sources/AsyncExamples/main.swift
```diff
@@ -86,7 +86,7 @@ private func changeStreams() throws {
         let inventory = db.collection("inventory")
 
         // Option 1: use next() to iterate
-        let next = inventory.watch(options: ChangeStreamOptions(fullDocument: .updateLookup))
+        let next = inventory.watch(options: ChangeStreamOptions(fullDocument: .other("blah")))
```
patrickfreed, line 89: these docs examples are meant to demonstrate basic behavior and will be included the [official docs page for change streams](https://docs.mongodb.com/manual/changeStreams/), so I think we should stick to `.updateLookup` rather than demonstrating the escape hatch.

kmahar, line 89: this is an example for the docs so I think we should leave it with the original type

Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -1,29 +1,18 @@
 /// Describes the modes for configuring the fullDocument field of a change stream document.
-public enum FullDocument: RawRepresentable, Codable {
+public struct FullDocument: RawRepresentable, Codable {
```
kmahar, line 2: this is another case where I'm questioning the value of `RawRepresentable`. I think we could do the same thing as subtype here and get rid of it, and just allow construction via the properties and the `other` method. WDYT @patrickfreed ?

Sources/MongoSwiftSync/Exports.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.18.0 — https://github.com/krzysztofzablocki/Sourcery
```
kmahar, line 1: usual comment here 😛

patrickfreed, line 1: unfortunately this change will probably break travis

kmahar:
lol double jinx 🙂

Sources/MongoSwift/ChangeStreamOptions.swift
patrickfreed: I guess the one benefit it gets us here is that the `Codable` conformance automatically operates just on the raw value. It also sort of signals that this struct really is just representing a string option.

On the flip side, the initializer being failable may indicate that we're doing some sort of validation, which we aren't.

I would say that I don't really have strong feelings either way and am happy with whatever you guys decide.

kmahar: yeah, the failable initializer is what is bothering me 🤔 I guess we could leave it and just document that it never returns nil.

Examples/Docs/Sources/AsyncExamples/main.swift
nbbeeken: Oops yea I just used this to make sure it would compile and forgot to revert

PR #487, opened 2020-05-27
SWIFT-855 Add minimum macOS version and document OS support policy

kmahar:
This adds a minimum macOS version to our package manifest, and clarifies supported / tested platforms in our README.
This also lets us remove the availability checks for 10.12+ we had to use for the date formatter.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,7 @@ import PackageDescription
 let package = Package(
     name: "mongo-swift-driver",
     platforms: [
-       .macOS(.v10_14)
+        .macOS(.v10_14)
     ],
     products: [
         .library(name: "MongoSwift", targets: ["MongoSwift"]),
--- a/Package.swift
+++ b/Package.swift
@@ -2,6 +2,9 @@
 import PackageDescription
 let package = Package(
     name: "mongo-swift-driver",
+    platforms: [
+       .macOS(.v10_14)
+    ],
     products: [
         .library(name: "MongoSwift", targets: ["MongoSwift"]),
         .library(name: "MongoSwiftSync", targets: ["MongoSwiftSync"])
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Bug reports in JIRA for all driver projects (i.e. NODE, PYTHON, CSHARP, JAVA) an
 Core Server (i.e. SERVER) project are **public**.
 
 ## Installation
-The driver supports use with Swift 5.1+ on MacOS and Linux.
+The driver supports use with Swift 5.1+. It is tested in continuous integration against macOS 10.14, Ubuntu 16.04, and Ubuntu 18.04.
 
 Installation is supported via [Swift Package Manager](https://swift.org/package-manager/).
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -34,7 +34,7 @@ Bug reports in JIRA for all driver projects (i.e. NODE, PYTHON, CSHARP, JAVA) an
 Core Server (i.e. SERVER) project are **public**.
 
 ## Installation
-The driver supports use with Swift 5.1+ on MacOS and Linux.
+The driver supports use with Swift 5.1+. The minimum macOS version required to build the driver is 10.14. The driver is tested in continuous integration against macOS 10.14, Ubuntu 16.04, and Ubuntu 18.04.
```
patrickfreed, line 37: So this mentions the minimum OS versions, but what is our OS support policy going forward? The last two macOS releases? I think for Linux we don't really need to do anything special since the runtime is bundled with the Swift distributions as we saw earlier.

kmahar: I don't know if we can really state a policy in that way since it depends on our major version release cadence as well as macOS release cadence. for now we support the last two versions... when 10.16 comes out we'll support that too and won't be able to drop 10.14 until 2.0. 
in practice I think this will likely translate to usually being ~ the last 2 releases, but not always. we could say at least the last 2, but imagine the following scenario: 10.16 is the latest release but came out several months months ago, and there is some feature/library we really want that is only available in 10.16+. we are about to tag 2.0 and probably won't tag 3.0 for a couple more years. do we drop 10.15? (I'd say yes)

patrickfreed: Ah okay, makes sense. And I know I've asked this before (maybe multiple times) so I apologize for forgetting, but why do we need/want to support 10.14 still?

kmahar: we can't actually test on 10.15 right now because evergreen has no hosts for it, so that eliminates the possibility of putting 10.15 as the minimum version in our `Package.swift`, as the driver would no longer compile on the 10.14 hosts.

that said we *could* have this at 10.14 but just say that we don't support 10.15, but I think that's confusing and users could easily miss it.
we don't really have a compelling reason to drop 10.14 right now anyway other than eventually needing `SwiftCrypto`, but my expectation is that we will be releasing any changes using that in 2.0 anyway.

Package.swift
```diff
@@ -2,6 +2,9 @@
 import PackageDescription
 let package = Package(
     name: "mongo-swift-driver",
+    platforms: [
+        .macOS(.v10_14)
```
patrickfreed, line 6: should this also list `v10_15`?

kmahar: confusingly, this list is meant to indicate minimum required platforms, not all supported platforms. having both of them gives an error about multiple declarations for the platform. see: https://github.com/apple/swift-package-manager/blob/master/Sources/PackageDescription/SupportedPlatforms.swift#L57

> /// Packages can choose to configure the minimum deployment
/// target version for a platform by using the APIs defined in this struct. The
/// Swift Package Manager emits appropriate errors when an invalid value is
/// provided for supported platforms, such as an empty array, multiple declarations
/// for the same platform, or an invalid version specification.

kmahar: the `SupportedPlatform` struct is also not to be confused with the `Platform` struct, which is used in other contexts in package manifests, e.g. how we use it to conditionally define `cSettings`: https://github.com/mongodb/mongo-swift-driver/blob/7e60c3c5a5fad0a6f1ca55170eed0909d2c7d062/Package.swift#L25

patrickfreed: Yeah, I think I had it confused with `Platform`.

PR #490, opened 2020-05-28
SWIFT-742 Implement unified replica set discovery behavior

kmahar:
This builds on my previous PR to add libmongoc support for `directConnection=true` to give us the desired discover behavior by default.

You can view a summary of the spec changes [here](https://github.com/mongodb/specifications/commit/e56f5eceed7729f8b9b43a4a1f76c7e5840db49f).

I was able to pull in the relevant initial DNS seedlist discovery tests as we implement that test runner, but the other tests use the SDAM and connection string test runners which we do not implement yet.

I did some manual local testing using SDAM monitoring and everything seems to work as expected in terms of initial topology assumptions as well as errors being reported in the right scenarios.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -35,7 +35,9 @@ internal class ConnectionString {
 
         // Per SDAM spec: If the ``directConnection`` option is not specified, newly developed drivers MUST behave as
         // if it was specified with the false value.
-        if !self.hasOption("directConnection") {
+        if let dc = options?.directConnection {
+            mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_DIRECTCONNECTION, dc)
+        } else if !self.hasOption("directConnection") {
             mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_DIRECTCONNECTION, false)
         }
 
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -33,6 +33,12 @@ internal class ConnectionString {
             mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYREADS, rr)
         }
 
+        // Per SDAM spec: If the ``directConnection`` option is not specified, newly developed drivers MUST behave as
+        // if it was specified with the false value.
+        if !self.hasOption("directConnection") {
+            mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_DIRECTCONNECTION, false)
+        }
+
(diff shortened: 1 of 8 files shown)


Tests/MongoSwiftSyncTests/SDAMMonitoringTests.swift
```diff
@@ -51,52 +47,48 @@ final class SDAMTests: MongoSwiftTestCase {
         }
         let hostAddress = try ServerAddress(host)
 
-        // check event count and that events are of the expected types
-        expect(receivedEvents.count).to(beGreaterThanOrEqualTo(5))
```
kmahar, line 55: in this test we previously saw:
- A topology opening event
- A topology changed event (change from unknown to single, no hosts in list)
- A server opening event
- A server changed event, from unknown to standalone
- A topology changed event, adding in the new server to the topology

We now see:
- A topology opening event
- A server opening event
- A server changed event, from unknown to standalone
- A topology changed event, from unknown to single

Previously in this case libmongoc by default would behave like the directConnection=true option was specified, which meant the topology would be assumed to be single right off the bat before discovering servers. Now that we specify directConnection=false that transition doesn't happen until the server has been discovered.

kmahar:
test added; was actually very easy thanks to the `getConnectionStringPerHost()` helper 🙂

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -33,6 +33,12 @@ internal class ConnectionString {
             mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYREADS, rr)
         }
 
+        // Per SDAM spec: If the ``directConnection`` option is not specified, newly developed drivers MUST behave as
+        // if it was specified with the false value.
+        if !self.hasOption("directConnection") {
```
patrickfreed, line 38: We should probably have an equivalent option in `MongoClientOptions` so that we can document its behavior

kmahar: done; also reworked the test to check all possible variations of unset / set via struct / set via URI

Sources/MongoSwift/MongoClient.swift
```diff
@@ -16,6 +16,12 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// databases or collections that derive from it.
     public var dateCodingStrategy: DateCodingStrategy?
 
+    /// When true, the client will connect directly to a single host. When false, the client will attempt to
+    /// automatically discover all replica set members if a replica set name is provided.
```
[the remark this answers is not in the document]
kmahar: whoops, I had that in here but it seems like I accidentally deleted it, good catch. added

PR #491, opened 2020-05-28
SWIFT-851 Namespace error types

patrickfreed:
[SWIFT-851](https://jira.mongodb.org/browse/SWIFT-851)

This PR moves all the error type definitoins under a new empty enum `MongoError`.
It also renames `MongoError` to `MongoErrorProtocol` and prefixes the other error protocols with `Mongo`.


Sources/MongoSwift/MongoError.swift
```diff
@@ -2,277 +2,287 @@ import CLibMongoC
 import Foundation
 
 /// An empty protocol for encapsulating all errors that this package can throw.
-public protocol MongoError: LocalizedError {}
-
-// TODO: update this link and the one below (SWIFT-319)
-/// A MongoDB server error code.
-/// - SeeAlso: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml
-public typealias ServerErrorCode = Int
+public protocol MongoErrorProtocol: LocalizedError {}
```
patrickfreed, line 5: to disambiguate from the namespace, I added the `Protocol` suffix that I've seen in a few places in the stdlib (e.g. [`LazySequenceProtocol`](https://developer.apple.com/documentation/swift/lazysequenceprotocol))

Sources/MongoSwift/MongoError.swift
```diff
@@ -2,277 +2,287 @@ import CLibMongoC
...
 
 /// Protocol conformed to by errors that may contain error labels.
-public protocol LabeledError: MongoError {
+public protocol MongoLabeledError: MongoErrorProtocol {
```
patrickfreed, line 8: There are limitations to this enum-as-namespace approach, namely protocols can't be defined that way. To avoid collisions, I added the `Mongo` prefix to all these protocols.

Sources/MongoSwift/MongoError.swift
```diff
@@ -2,277 +2,287 @@ import CLibMongoC
...
+public enum MongoError {
+    // TODO: update this link and the one below (SWIFT-319)
+    /// A MongoDB server error code.
+    /// - SeeAlso: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml
+    public typealias ServerErrorCode = Int
 
-    /// Labels that may describe the context in which this error was thrown.
-    public let errorLabels: [String]?
+    /// Thrown when commands experience errors on the server that prevent execution.
+    public struct CommandError: MongoServerError {
```
patrickfreed, line 30: We could even do sub-namespaces here for each of the error categories (e.g. `MongoError.ServerError.CommandError`), but that felt like overkill.

Sources/MongoSwift/MongoError.swift
```diff
@@ -2,277 +2,287 @@ import CLibMongoC
...
-/// - SeeAlso: https://docs.mongodb.com/manual/core/read-preference-mechanics/
-public struct ServerSelectionError: RuntimeError {
-    internal let message: String
+        public var errorDescription: String? { self.message }
+    }
 
-    public var errorDescription: String? { self.message }
-}
+    /// A struct to represent a single write error not resulting from an executed write operation.
+    public struct WriteFailure: Codable {
```
patrickfreed, line 155: these `-Failure` types aren't errors themselves, but they are contained within different error types. I wasn't sure where to put them so I just left them under `MongoError`.

PR #492, opened 2020-05-29
SWIFT-878: Define BSON Specific Error Types

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-878

Pretty much mirrors MongoError but with BSON. In order to nicely 🎣  catch and release errors I pulled out wrapper functions into a `BSONUtil.swift` file.


Sources/MongoSwift/BSON/BSONDocument.swift
```diff
@@ -145,10 +145,10 @@ extension BSONDocument {
      * presence first.
      *
      * - Throws:
-     *   - `MongoError.InternalError` if the new value is an `Int` and cannot be written to BSON.
-     *   - `MongoError.LogicError` if the new value is a `BSONDecimal128` or `BSONObjectID` and is improperly formatted.
-     *   - `MongoError.LogicError` if the new value is an `Array` and it contains a non-`BSONValue` element.
-     *   - `MongoError.InternalError` if the underlying `bson_t` would exceed the maximum size by encoding this
+     *   - `BSONError.InternalError` if the new value is an `Int` and cannot be written to BSON.
```
kmahar, line 148: this seems like an error we don't need to worry about anymore... I think it's probably a remnant of past API when `Int` was a `BSONValue`

Sources/MongoSwift/BSON/BSONDocument.swift
```diff
@@ -145,10 +145,10 @@ extension BSONDocument {
...
+     *   - `BSONError.LogicError` if the new value is a `BSONDecimal128` or `BSONObjectID` and is improperly formatted.
```
kmahar, line 149: I also don't think this can happen anymore, since we validate these when they are being initialized.

Sources/MongoSwift/BSON/BSONDocumentIterator.swift
```diff
@@ -177,8 +177,8 @@ public class BSONDocumentIterator: IteratorProtocol {
      * Overwrites the current value of this `BSONDocumentIterator` with the supplied value.
      *
      * - Throws:
-     *   - `MongoError.InternalError` if the new value is an `Int` and cannot be written to BSON.
-     *   - `MongoError.LogicError` if the new value is a `BSONDecimal128` or `BSONObjectID` and is improperly formatted.
+     *   - `BSONError.InternalError` if the new value is an `Int` and cannot be written to BSON.
```
kmahar, line 180: I think this and the next error listed here can also go away, not relevant for same reasons as above

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -70,17 +70,17 @@ internal protocol BSONValue: Codable {
      *   - key: A `String`, the key under which to store the value.
      *
      * - Throws:
-     *   - `MongoError.InternalError` if the `DocumentStorage` would exceed the maximum size by encoding this
+     *   - `BSONError.InternalError` if the `DocumentStorage` would exceed the maximum size by encoding this
```
kmahar, line 73: this line should probably just say `BSONDocument` rather than `DocumentStorage`

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -70,17 +70,17 @@ internal protocol BSONValue: Codable {
...
      *     key-value pair.
-     *   - `MongoError.LogicError` if the value is an `Array` and it contains a non-`BSONValue` element.
+     *   - `BSONError.LogicError` if the value is an `Array` and it contains a non-`BSONValue` element.
```
kmahar, line 75: this error case can go away too.

Sources/MongoSwift/BSON/Overwritable.swift
```diff
@@ -7,8 +7,8 @@ internal protocol Overwritable: BSONValue {
      * Overwrites the value at the current position of the iterator with self.
      *
      * - Throws:
-     *   - `MongoError.InternalError` if the `BSONValue` is an `Int` and cannot be written to BSON.
-     *   - `MongoError.LogicError` if the `BSONValue` is a `BSONDecimal128` or `BSONObjectID` and is improperly
+     *   - `BSONError.InternalError` if the `BSONValue` is an `Int` and cannot be written to BSON.
```
kmahar, line 10: these two can go away as well. in fact I think we could just make the protocol stop throwing

Sources/MongoSwift/BSON/BSONUtil.swift
```diff
@@ -0,0 +1,15 @@
+internal func getValue(from document: BSONDocument, for key: String) throws -> BSON? {
+    do {
+        return try document.getValue(for: key)
+    } catch let error as BSONError.InvalidArgumentError {
```
kmahar, line 4: what about e.g. internal errors?

nbbeeken: Do you mean emitted from here or in general? I don't think InternalError can be thrown from getValue or setValue
I tried sweeping the driver again it does look like there's one BSONInternalError that get's exposed when we handle boxing and unboxing binary types (specifically UUID) I'll push what I have but I'm unsure how we can 🎣  that

Sources/MongoSwift/BSON/BSONDocumentIterator.swift
```diff
@@ -173,19 +173,13 @@ public class BSONDocumentIterator: IteratorProtocol {
...
 
-    /**
-     * Overwrites the current value of this `BSONDocumentIterator` with the supplied value.
-     *
-     * - Throws:
-     *   - `BSONError.InternalError` if the new value is an `Int` and cannot be written to BSON.
-     *   - `BSONError.LogicError` if the new value is a `BSONDecimal128` or `BSONObjectID` and is improperly formatted.
-     */
+    /// Overwrites the current value of this `BSONDocumentIterator` with the supplied value.
     internal func overwriteCurrentValue(with newValue: Overwritable) throws {
```
kmahar, line 177: this can stop throwing now too i think?

Sources/MongoSwift/BSON/BSONUtil.swift
kmahar: it looks like `setValue` can throw an internal error if we exceed the maximum allowed BSON size.
also,`withID()` can also throw if `merge()` fails, for the same reason. we call `withID()` from `BulkWriteOperation` so I think we need to handle that too.

nbbeeken: Indeed, I missed it, I added the case, but now I realize all the encode methods for each BSONValue can throw BSONError.InternalError. What's the best way to wrap those calls? Is there something I can do with an extension?

kmahar: do you mean the `BSONValue.encode` method implementations? I believe those are only ever called via `setValue` which you are handling, so we are probably ok?

nbbeeken: Isn't it called when we encode options and such for operations? For example `BulkWriteOperation` has a `BSONEncoder`, won't that call `BSONValue` encode methods?

kmahar: ok so...

there are two `encode` methods on each `BSONValue` type. one of them comes from the `Encodable` protocol and the other from `BSONValue`.

The `Encodable` variants that we implement all just throw an `EncodingError` because right now we don't support using those types with non-BSON encoders. and the `BSONEncoder` itself special-cases all of those types so that the `Encodable.encode` method is never actually called on them.

As for the `BSONValue.encode` implementations...

at a high level what the encoder does is construct an ordered map of `[String: BSONValue]`. once that has been fully filled out, the final step is that we call this method `toDocument()` that (recursively if needed) converts that map to an equivalent `BSONDocument`: https://github.com/mongodb/mongo-swift-driver/blob/ca167f2474524ebbbac0aadb8bfe145e1b0889c8/Sources/MongoSwift/BSON/BSONEncoder.swift#L806

notice that method builds up the document via the subscript setter (which basically calls `getValue` but `fatalError`s on failure). I think this is actually not great, because we'll crash instead of erroring when e.g. the document we create gets too large, so it would probably be wise of us to update that to use `setValue` instead.

in either case, we never directly call the `BSONValue.encode` methods from the encoder. those methods are basically just implementation details of `BSONDocument` and are not at all related to `Codable`. in retrospect I think we should have given this a different name, that's why I went with `write(to buffer: ByteBuffer)` in the basically equivalent method in the new library. 

assuming we make the switch to `setValue` from within the encoder, which means those errors could actually bubble up when encoding a struct now, I think there are two different ways we could handle the driver error conversion: 

1) I'd argue `BSONEncoder` should maybe actually only ever throw `EncodingError`s, and therefore we should wrap the `setValue` call in `toDocument()` in a try/catch that converts any error it sees into an `EncodingError`.
2) alternatively if we do want to propagate BSON errors from the encoder you could extend `BSONEncoder` to add some internal method which does the try/catch, and we use that anywhere in the driver code that we do encoding.

personally I like option 1, what do you think? cc @patrickfreed 

we also have a symmetric decoding problem. the decoder uses `getValue` for retrieving values from `BSONDocument`s. depending on what we decide re: errors, we can either try/catch within the decoder itself and convert those to `DecodingError`s, or add an extension method that does the conversion.

patrickfreed: Definitely agree on option 1. When I did the error work way back when as part of my rotation, I think I tried to ensure that the encoder/decoder only threw `EncodingError` / `DecodingError`, though I see I may have missed a few cases.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -774,8 +784,7 @@ private class MutableArray: BSONValue {
 private class MutableDictionary: BSONValue {
     fileprivate static var bsonType: BSONType { .document }
 
-    // This is unused
-    fileprivate var bson: BSON { .document(self.toDocument()) }
+    fileprivate var bson: BSON { .document((try? self.toDocument()) ?? [:]) }
```
nbbeeken, line 787: This is used in testing, not sure what to do here since prop getters can't throw, do y'all think this is fine? maybe I can add a comment?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -149,7 +149,7 @@ public class BSONEncoder {
 
         let encoder = _BSONEncoder(options: self.options)
 
-        guard let boxedValue = try encoder.box_(value) else {
+        guard let boxedValue = try? encoder.box_(value) else {
```
kmahar, line 152: I don't think we should swallow the actual thrown error here, we lose information about what went wrong. I think we should try/catch and if we catch something you can put the resulting error info into the `debugDescription`.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -169,7 +169,17 @@ public class BSONEncoder {
             )
         }
 
-        return dict.toDocument()
+        guard let doc = try? dict.toDocument() else {
```
kmahar, line 172: same point as above. one error you might get here is that the resulting document would be too large to be valid BSON. we should use try/catch and propagate that error message to the user.

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -149,27 +149,38 @@ public class BSONEncoder {
...
 
-        guard let boxedValue = try encoder.box_(value) else {
-            throw EncodingError.invalidValue(
-                value,
-                EncodingError.Context(
-                    codingPath: [],
-                    debugDescription: "Top-level \(T.self) did not encode any values."
+        do {
+            let optionalBoxedValue = try encoder.box_(value)
+            guard let boxedValue = optionalBoxedValue else {
```
kmahar, line 154: I think you can just one line this, `guard let boxedValue = try encoder.box_(value) else { ...`?

Sources/MongoSwift/BSON/BSONUtil.swift
```diff
@@ -0,0 +1,17 @@
+/*
+ * BSONUtil contains helpers to wrap the underlying BSON library to assist in providing a consistent API
+ */
+
+/// We don't want driver users to handle any BSONErrors
+/// this will convert BSONError.* thrown from `fn` to MongoError.* and rethrow
+internal func convertingBSONErrors<T>(_ fn: () throws -> T) rethrows -> T {
```
kmahar, line 7: tiny nit, we basically always use `body` as the parameter name in these cases

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -803,10 +828,16 @@ private class MutableDictionary: BSONValue {
     }
 
     /// Converts self to a `BSONDocument` with equivalent key-value pairs.
-    fileprivate func toDocument() -> BSONDocument {
+    fileprivate func toDocument() throws -> BSONDocument {
         var doc = BSONDocument()
         for i in 0..<self.keys.count {
-            doc[self.keys[i]] = self.values[i].bson
+            try convertingBSONErrors {
```
kmahar, line 834: do we need this `convertingBSONErrors`, given that at the top level we are converting everything to an `EncodingError` anyway?

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -733,7 +744,16 @@ extension _BSONEncoder: SingleValueEncodingContainer {
 private class MutableArray: BSONValue {
     fileprivate static var bsonType: BSONType { .array }
 
-    fileprivate var bson: BSON { .array(self.array.map { $0.bson }) }
+    fileprivate var bson: BSON {
```
kmahar, line 747: so I actually think we should just get rid of these `.bson` properties' implementations altogether and `fatalError` when you try to use them. and we can just special case both of these and have the logic in `toDocument`. maybe you could just move this into a like `toBSONArray()` throwing method on this type? then your `toDocument()` logic becomes something like

```swift
for i in 0..<self.keys.count {
    let value = self.values[i]
    switch value {
    case let val as MutableDictionary:
        try doc.setValue(for: self.keys[i], to: val.toDocument().bson)
    case let val as MutableArray:
        let array = try val.toBSONArray()
        try doc.setValue(for: self.key[i], to: array.bson)
    default:
        try doc.setValue(for: self.keys[i], to: value.bson)
    }
}
```

Sources/MongoSwift/BSON/BSONEncoder.swift
```diff
@@ -149,27 +149,45 @@ public class BSONEncoder {
...
+                    EncodingError.Context(
+                        codingPath: [],
+                        debugDescription: "Top-level \(T.self) was not encoded as a complete document."
+                    )
+                )
+            }
 
-        guard let dict = boxedValue as? MutableDictionary else {
+            return try dict.toDocument()
+        } catch let error as BSONErrorProtocol {
```
kmahar, line 174: `BSONErrorProtocol` inehrits from `localizedError` so I'd think you'd just be able to directly use the `.errorDescription` property on the caught error here?

Sources/MongoSwift/BSON/BSONDecoder.swift
```diff
@@ -532,7 +532,7 @@ private struct _BSONKeyedDecodingContainer<K: CodingKey>: KeyedDecodingContainer
     /// Private helper function to check for a value in self.container. Returns the value stored
     /// under `key`, or throws an error if the value is not found.
     private func getValue(forKey key: Key) throws -> BSON {
-        guard let entry = try self.container.getValue(for: key.stringValue) else {
+        guard let entry = (try convertingBSONErrors { try self.container.getValue(for: key.stringValue) }) else {
             throw DecodingError.keyNotFound(
```
nbbeeken, line 536: This seems to be the only place where we'd have an issue, but I'm suspicious 🤔  sanity check?

Guides/Error-Handling.md
```diff
@@ -15,8 +15,7 @@
...
-
+The driver uses errors to communicate that an operation failed, an assumption wasn't met, or that the user did something incorrectly. Applications that use the driver can in turn catch these errors and respond appropriately without crashing or resulting in an otherwise inconsistent state. To correctly model the different sources of errors, the driver defines three separate categories of errors (`MongoServerError`, `MongoUserError`, `MongoRuntimeError`), each of which are protocols that inherit from the `MongoErrorProtocol` protocol. These protocols are defined in `MongoError.swift`, and the structs that conform to them are outlined here. The documentation for every public function that throws lists some of the errors that could possibly be thrown and the reasons they might be. The errors listed there are not comprehensive but will generally cover the most common cases.
```
nbbeeken, line 18: hard to pick out but this was just: `caregories` -> `categories`

kmahar:
whoops I don't know why that comment posted like that, but 🤷‍♀️

Sources/MongoSwift/BSON/BSONError.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    /// An error thrown when the BSON library is incorrectly used.
+    public struct LogicError: BSONUserError {
+        internal let message: String
+
+        public var errorDescription: String? { self.message }
+    }
+}
+
+internal func bsonTooLargeError(value: BSONValue, forKey: String) -> BSONErrorProtocol {
```
patrickfreed, line 37: Since this error could reasonably be caused by the user, maybe it makes sense to define a specific error for it (e.g. `BSONDocumentTooLarge` or `BSONDocumentExceededSizeLimit` or something. `InternalError` should generally be used for system issues (e.g. out of memory causes allocation to fail) or invariant failures (i.e. bugs in the BSON library).

Sources/MongoSwift/BSON/BSONError.swift
```diff
@@ -0,0 +1,114 @@
+import Foundation
+
+/// An empty protocol for encapsulating all errors that BSON package can throw.
+public protocol BSONErrorProtocol: LocalizedError {}
+
+/// A protocol describing errors caused by improper usage of the BSON library by the user.
+public protocol BSONUserError: BSONErrorProtocol {}
```
patrickfreed, line 7: Given that we only have 3 total error types, it might be better to just leave the "sub" protocols out for now. As we add more errors we can introduce a hierarchy to organize them.

Guides/Error-Handling.md
```diff
@@ -77,6 +76,14 @@ As part of the driver, `BSONEncoder` and `BSONDecoder` are implemented according
 
 See the official documentation for both [`EncodingErrors`](https://developer.apple.com/documentation/swift/encodingerror) and [`DecodingErrors`](https://developer.apple.com/documentation/swift/decodingerror) for more information.
 
+### BSON Errors
+
+The BSON library has its own subset of errors that communicate issues when constructing or using BSON.
+BSON Errors can be found in [Sources/MongoSwift/BSON/BSONError.swift](Sources/MongoSwift/BSON/BSONError.swift) and are as follows:
```
patrickfreed, line 82: Oh I didn't know you could link like this on GitHub, very cool.

Sources/MongoSwift/BSON/BSONError.swift
nbbeeken: Sure I agree, I think I'll still wanna add a catch and translate for this new error what do you think is the best MongoError to translate to?

patrickfreed: I think `MongoError.InternalError` still works fine probably, since if we're creating the document in the driver and make it too big, it's probably our fault.

nbbeeken: Indeed, I also added a generic BSONErrorProtocol catcher for future potential bson errors:

https://github.com/mongodb/mongo-swift-driver/blob/30fe38686000a37f7c18eb13dd1b711162564578/Sources/MongoSwift/BSON/BSONUtil.swift#L16-L18

Sources/MongoSwift/BSON/BSONError.swift
kmahar: fine with me

Sources/MongoSwift/BSON/BSONError.swift
kmahar: also fine with me

Sources/MongoSwift/BSON/BSONError.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /// An error thrown when the BSON library is incorrectly used.
+    public struct LogicError: BSONErrorProtocol {
+        internal let message: String
+
+        public var errorDescription: String? { self.message }
+    }
+
+    public struct TooLargeError: BSONErrorProtocol {
```
patrickfreed, line 30: This type needs a docstring. Also, its name could be specified a little bit, since its specifically referring to a document being too big (e.g. `DocumentTooLargeError`)

Guides/Error-Handling.md
```diff
@@ -77,6 +76,14 @@ As part of the driver, `BSONEncoder` and `BSONDecoder` are implemented according
...
+
+- `BSONError.InvalidArgumentError` - This error is thrown when a BSON type is being incorrectly constructed.
```
patrickfreed, line 84: the new error case should be documented here.

Sources/MongoSwift/BSON/BSONDocument.swift
```diff
@@ -145,10 +145,7 @@ extension BSONDocument {
      * presence first.
      *
      * - Throws:
-     *   - `MongoError.InternalError` if the new value is an `Int` and cannot be written to BSON.
-     *   - `MongoError.LogicError` if the new value is a `BSONDecimal128` or `BSONObjectID` and is improperly formatted.
-     *   - `MongoError.LogicError` if the new value is an `Array` and it contains a non-`BSONValue` element.
-     *   - `MongoError.InternalError` if the underlying `bson_t` would exceed the maximum size by encoding this
+     *   - `BSONError.InternalError` if the underlying `bson_t` would exceed the maximum size by encoding this
```
patrickfreed, line 148: this docstring should use the new toolarge error case.

Sources/MongoSwift/BSON/BSONValue.swift
```diff
@@ -70,17 +70,16 @@ internal protocol BSONValue: Codable {
      *   - key: A `String`, the key under which to store the value.
      *
      * - Throws:
-     *   - `MongoError.InternalError` if the `DocumentStorage` would exceed the maximum size by encoding this
+     *   - `BSONError.InternalError` if the `BSONDocument` would exceed the maximum size by encoding this
```
patrickfreed, line 73: ditto updating this comment. I'd just do a check and make sure all appropriate docstrings are updated to reflect the new error type.

Sources/MongoSwift/BSON/BSONError.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// An error thrown when the BSON library is incorrectly used.
+    public struct LogicError: BSONErrorProtocol {
+        internal let message: String
+
+        public var errorDescription: String? { self.message }
+    }
+
+    /// An error thrown when a document exceeds the maximum BSON encoding size
```
patrickfreed, line 30: needs period at end of sentence

Sources/MongoSwift/BSON/BSONValue.swift
nbbeeken: I went around updating so that's all good but then I found:
https://github.com/mongodb/mongo-swift-driver/blob/cd45de9df57a3a71ec5ffe1da39d3355b1c56abe/Sources/MongoSwift/BSON/BSONDocument.swift#L209-L214
You think it makes sense to also make a TooSmall version too? or maybe go generic DocumentSizeError?

patrickfreed: I think this is fine as an internal error, since there really should be no way for a user to create a document that is smaller than 5 bytes.

Guides/Error-Handling.md
```diff
@@ -84,6 +84,7 @@ BSON Errors can be found in [Sources/MongoSwift/BSON/BSONError.swift](Sources/Mo
 - `BSONError.InvalidArgumentError` - This error is thrown when a BSON type is being incorrectly constructed.
 - `BSONError.InternalError` - This error is thrown when there is an issue that is a result of system failure (e.g, allocation issue).
 - `BSONError.LogicError` - This error is thrown when there is an unexpected usage of the the API.
+- `BSONError.DocumentTooLargeError` - This error is thrown when there is an unexpected usage of the the API.
```
patrickfreed, line 87: the description is still copy/pasted from logic error, it needs its own

PR #496, opened 2020-06-03
SWIFT-882 Convert extJSON properties to methods

patrickfreed:
[SWIFT-882](https://jira.mongodb.org/browse/SWIFT-882)

This PR converts the extJSON computed properties to methods, since they may be computationally expensive. @mbroadst noted that they would be superseded by `ExtendedJSONEncoder`, and that we would probably deprecate them once that was introduced. Since these produce `String`s, they could perhaps be useful as conveniences over invoking the encoder, so I also added "String" as a suffix to their method names.

These changes are not release blockers, but if we do want to make them we need to do so today, so I just put up the PR now to get the discussion rolling.

### DIFF
--- a/Sources/MongoSwift/BSON/BSONDocument.swift
+++ b/Sources/MongoSwift/BSON/BSONDocument.swift
@@ -300,7 +300,7 @@ extension BSONDocument {
 
     /// Returns the relaxed extended JSON representation of this `BSONDocument`.
     /// On error, an empty string will be returned.
-    public var extendedJSON: String {
+    public func toExtendedJSON() -> String {
         let result = self.withBSONPointer { ptr in
             bson_as_relaxed_extended_json(ptr, nil)
         }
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/BSON/BSONDocument.swift
```diff
@@ -300,7 +300,7 @@ extension BSONDocument {
 
     /// Returns the relaxed extended JSON representation of this `BSONDocument`.
     /// On error, an empty string will be returned.
-    public var extendedJSON: String {
+    public func toExtendedJSONString() -> String {
```
kmahar, line 303: I was going to say it would be better to throw than to return an empty string now that these are methods, but OTOH i think in the new library we don't expect JSON serialization to fail unless it's explicitly unvalidated BSON. in which case i think fatal error / empty string is acceptable

PR #498, opened 2020-06-22
SWIFT-573 Implement URI options spec test runner

kmahar:
Implements the URI options spec tests: https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/README.rst

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -174,7 +174,28 @@ internal class ConnectionString {
         guard let optsDoc = mongoc_uri_get_options(self._uri) else {
             return nil
         }
-        return BSONDocument(copying: optsDoc)
+        var copy = BSONDocument(copying: optsDoc)
+
+        if let authSource = self.authSource {
+            copy.authsource = .string(authSource)
+        }
+        if let authMechanism = self.authMechanism {
+            copy.authmechanism = .string(authMechanism.name)
+        }
+        if let authMechanismProperties = self.authMechanismProperties {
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -174,7 +174,28 @@ internal class ConnectionString {
         guard let optsDoc = mongoc_uri_get_options(self._uri) else {
             return nil
         }
-        return BSONDocument(copying: optsDoc)
+        var copy = BSONDocument(copying: optsDoc)
+
+        if let authSource = self.authSource {
```
kmahar, line 179: libmongoc doesn't put a lot of these into the returned document so we have to add them manually to get the full list

patrickfreed: Is this a bug in libmongoc then?

kmahar: they document the method as only returning the options provided after the ? of a URI. 

all of the auth info in a URI comes before that. 

however, that still leaves a few cases, all of which they appear to be aware of as they work around them in their own version of these tests:
- tag sets and compressors: https://github.com/mongodb/mongo-c-driver/blob/c64955f98d7ddb6ab25e14501ecd7d175acc4ab1/src/libmongoc/tests/test-mongoc-connection-uri.c#L127
- read concern level: https://github.com/mongodb/mongo-c-driver/blob/c64955f98d7ddb6ab25e14501ecd7d175acc4ab1/src/libmongoc/tests/test-mongoc-connection-uri.c#L227

PR #501, opened 2020-06-23
minor: ignore MongoSwiftVersion.swift in swiftlint

nspektor:
`swiftlint autocorrect` now ignores `MongoSwiftVersion.swift`


kmahar:
What was the issue here? We already have a `// swiftlint:disable:previous vertical_whitespace` line in `MongoSwiftVersion.swift` so I would think that would take care of our linting issues?

nspektor:
The reason this came up was that running `swiftlint autocorrect` edited the `MongoSwiftVersion.swift` file

kmahar:
oh interesting, that seems to be a new change as of 0.39.2. LGTM!

PR #502, opened 2020-06-25
SWIFT-898 Add replicaSet option, and tests for replicaSet + appName options

kmahar:
This adds support for the replica set option and tests for it. Writing the tests made me realize I should probably at least add a simple test about options precedence for appName so I went ahead and added that too.


patrickfreed:
Oh, looks like sourcery needs to be run but otherwise lgtm still

PR #503, opened 2020-06-30
Swift 538/infer topology in tests

nspektor:


Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -254,15 +253,22 @@ final class ChangeStreamSpecTests: MongoSwiftTestCase {
...
+                )
 
-                guard version >= test.minServerVersion else {
-                    print("Skipping test case \"\(test.description)\": minimum required server " +
-                        "version \(test.minServerVersion) not met.")
+                let unmetRequirements = try globalClient.meetsRequirements(testRequirements)
+                guard unmetRequirements == nil else {
+                    switch unmetRequirements {
+                    case .minServerVersion:
+                        print("Skipping test case \"\(test.description)\": minimum required server " +
```
patrickfreed, line 265: It would be fine just to replace this message with the one from `printRequirementNotMetMessage`.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -298,8 +304,12 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
     /// Prose test 1 of change stream spec.
     /// "ChangeStream must continuously track the last seen resumeToken"
     func testChangeStreamTracksResumeToken() throws {
-        guard MongoSwiftTestCase.topologyType != .single else {
-            print(unsupportedTopologyMessage(testName: self.name))
+        let testRequirements = TestRequirement(
+            acceptableTopologies: ["replicaSet", "sharded"].map { TestTopologyConfiguration(from: $0) })
```
patrickfreed, line 308: here is where the linter is complaining I think. If you move the last `)` on line 308 to its own line, I think it should be good to go.

patrickfreed, line 308: Rather than mapping from strings, this can just be `[.replicaSet, .sharded]`, using the enum cases directly.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -332,13 +342,21 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
...
-        guard MongoSwiftTestCase.topologyType != .single else {
-            print(unsupportedTopologyMessage(testName: self.name))
-            return
-        }
-
-        guard try MongoClient.makeTestClient().serverVersion() < ServerVersion(major: 4, minor: 3, patch: 3) else {
-            print("Skipping test; see SWIFT-722")
+        let testRequirements = TestRequirement(
+            maxServerVersion: ServerVersion(major: 4, minor: 3, patch: 3),
+            acceptableTopologies: ["replicaSet", "sharded"].map { TestTopologyConfiguration(from: $0) }
```
patrickfreed, line 347: ditto using enum cases directly.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -369,8 +387,12 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
      * pipeline and options, except for the addition/update of a resumeToken.
      */
     func testChangeStreamAutomaticResume() throws {
-        guard MongoSwiftTestCase.topologyType != .single else {
-            print(unsupportedTopologyMessage(testName: self.name))
+        let testRequirements = TestRequirement(
+            acceptableTopologies: ["replicaSet", "sharded"].map { TestTopologyConfiguration(from: $0) })
```
patrickfreed, line 391: ditto enum cases directly and `)` formatting

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -430,8 +452,12 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
      * ChangeStream will not attempt to resume on any error encountered while executing an aggregate command.
      */
     func testChangeStreamFailedAggregate() throws {
-        guard MongoSwiftTestCase.topologyType != .single else {
-            print(unsupportedTopologyMessage(testName: self.name))
+        let testRequirements = TestRequirement(
+            acceptableTopologies: ["replicaSet", "sharded"].map { TestTopologyConfiguration(from: $0) })
```
patrickfreed, line 456: ditto

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -480,8 +506,12 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
      * 136 (CappedPositionLost), or 237 (CursorKilled) while executing a getMore command.
      */
     func testChangeStreamDoesntResume() throws {
-        guard MongoSwiftTestCase.topologyType != .single else {
-            print(unsupportedTopologyMessage(testName: self.name))
+        let testRequirements = TestRequirement(
+            acceptableTopologies: ["replicaSet", "sharded"].map { TestTopologyConfiguration(from: $0) })
```
patrickfreed, line 510: ditto
This applies to a few other places in this file, so I'll stop commenting on every one.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -747,14 +801,22 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
...
+        let testRequirements = TestRequirement(
+            minServerVersion: ServerVersion(major: 4, minor: 0),
+            acceptableTopologies: ["replicaSet", "sharded"].map { TestTopologyConfiguration(from: $0) }
+        )
+
+        let unmetRequirements = try client.meetsRequirements(testRequirements)
+        guard unmetRequirements == nil else {
+            switch unmetRequirements {
+            case .minServerVersion:
+                print("Skipping test case for server version \(try client.serverVersion())")
```
patrickfreed, line 814: I think the message from `printRequirementNotMetMessage` can be used instead of this

Tests/MongoSwiftTests/AsyncTestUtils.swift
```diff
@@ -38,6 +38,16 @@ extension MongoClient {
             return try ServerVersion(versionString)
         }
     }
+
+    /// Determine whether server version and topology requirements for a certain test are met
+    internal func meetsRequirements(_ testRequirement: TestRequirement) throws -> UnmetRequirements? {
```
patrickfreed, line 43: since this is no longer returning a boolean value, I think refactoring it to something like `checkRequirements` may make more sense.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -214,6 +303,25 @@ public func sortedEqual(_ expectedValue: BSONDocument?) -> Predicate<BSONDocumen
     }
 }
 
+/// Prints a message if a server version or topology requirement is not met and a test is skipped
+public func printRequirementNotMetMessage(
```
patrickfreed, line 307: I think `printSkipMessage` would be a bit briefer and more accurate name for this.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,95 @@ open class MongoSwiftTestCase: XCTestCase {
...
+
+/// Enumerates different possible unmet requirements that can be returned by meetsRequirements
+public enum UnmetRequirements {
+    case minServerVersion(actual: ServerVersion, required: ServerVersion)
+    case maxServerVersion(actual: ServerVersion, required: ServerVersion)
+    case topology(actual: TestTopologyConfiguration, required: [TestTopologyConfiguration])
+}
+
+/// Determines the topologyType of a client based on the reply returned by running an isMaster command
+public func topologyType(_ isMasterReply: BSONDocument) throws -> TestTopologyConfiguration {
```
patrickfreed, line 154: rather than a free floating function, this would be a little more discoverable as an initializer on `TestTopologyConfiguration`.

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
```diff
@@ -63,6 +63,16 @@ extension MongoClient {
         return try ServerVersion(versionString)
     }
 
+    /// Determine whether server version and topology requirements for a certain test are met
+    internal func meetsRequirements(_ testRequirement: TestRequirement) throws -> UnmetRequirements? {
+        let reply = try self.db("admin").runCommand(["isMaster": 1])
+        return try commonMeetsRequirements(
+            testRequirement: testRequirement,
+            serverVersion: serverVersion(),
```
patrickfreed, line 71: As part of the driver's style, we require explicit `self` to signal where method calls are going to. For here that means doing `self.serverVersion()`.

Sources/TestsCommon/CommonTestUtils.swift
nspektor: done

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,95 @@ open class MongoSwiftTestCase: XCTestCase {
...
+    topologyType: TestTopologyConfiguration
+) throws -> UnmetRequirements? {
+    testRequirement.isMet(by: serverVersion, topologyType)
+}
+
+/// Struct representing conditions that a deployment must meet in order for a test file to be run.
+public struct TestRequirement: Decodable {
+    private let minServerVersion: ServerVersion?
+    private let maxServerVersion: ServerVersion?
+    private let topology: [String]?
```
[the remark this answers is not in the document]
nspektor: nice! this is much cleaner.

Sources/TestsCommon/CommonTestUtils.swift
nspektor: ok done

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
nspektor: done

Tests/MongoSwiftTests/AsyncTestUtils.swift
nspektor: done

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: done

nspektor: done

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: did this for all similar places also

nspektor: nice :)

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: mhm

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: done

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: done

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: done

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: done

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,94 @@ open class MongoSwiftTestCase: XCTestCase {
     }
 }
 
+/// Enumerates the different topology configurations that are used throughout the tests
+public enum TestTopologyConfiguration: Decodable {
+    public init(from decoder: Decoder) throws {
```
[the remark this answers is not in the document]
nspektor: ok

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,94 @@ open class MongoSwiftTestCase: XCTestCase {
...
+        let str = try decoder.singleValueContainer().decode(String.self)
+        self.init(from: str)
+    }
+
+    case sharded
+    case replicaSet
+    case single
+    case unknown
+
+    public init(from str: String) {
```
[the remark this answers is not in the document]
patrickfreed: The only usage of it is in `testChangeStreamSpec`, and now that `TestTopologyConfiguration` is `Decodable`, we can update the `topology` field of `ChangeStreamTest` to be `[TestTopologyConfiguration]` instead of `[String]` as you suggest, and it should just work. Then this initializer will no longer be required.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -692,14 +737,22 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
...
+        let testRequirements = TestRequirement(
+            minServerVersion: ServerVersion(major: 4, minor: 0),
+            acceptableTopologies: [.replicaSet, .sharded]
+        )
+
+        let unmetRequirements = try client.checkRequirements(testRequirements)
+        guard unmetRequirements == nil else {
+            switch unmetRequirements {
+            case .minServerVersion:
+                print("Skipping test case for server version \(try client.serverVersion())")
```
patrickfreed, line 750: this one looks like it can just use the regular print method too.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,86 @@ open class MongoSwiftTestCase: XCTestCase {
     }
 }
 
+/// Enumerates the different topology configurations that are used throughout the tests
+public enum TestTopologyConfiguration: String, Decodable {
+    case sharded
+    case replicaSet
```
patrickfreed, line 128: From looking at the [evergreen failures](https://evergreen.mongodb.com/task_log_raw/mongo_swift_driver_tests_all__os_fully_featured~ubuntu_18.04_swift_version~5.1_ssl_auth~nossl_noauth_test_latest_replica_set_patch_2ce49dbfa192c24a9c02842ea0d8d7f50538a903_5efe2b4a3e8e861a667e11f1_20_07_02_18_45_31/0?type=T#L5611), it looks like the JSON tests actually specify this string as "replicaset" (no camelCase). To alleviate this, you can just specify the string value for this case (e.g. `case replicaSet = "replicaset"` ).

Sources/TestsCommon/CommonTestUtils.swift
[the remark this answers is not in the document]
nspektor: done

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
nspektor: ok

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,73 @@ open class MongoSwiftTestCase: XCTestCase {
...
+        if isMasterReply["msg"] != "isdbgrid" &&
+            isMasterReply["setName"] == nil &&
+            isMasterReply["isreplicaset"] != true {
+            self = .single
+        } else if isMasterReply["msg"] == "isdbgrid" {
+            self = .sharded
+        } else if isMasterReply["ismaster"] == true && isMasterReply["setName"] != nil {
+            self = .replicaSet
+        } else {
+            fatalError("Invalid test topology configuration")
```
kmahar, line 143: it might be helpful to include `isMasterReply` in the error string here, so we can use it to diagnose what went wrong if we ever hit this case.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,73 @@ open class MongoSwiftTestCase: XCTestCase {
...
+        }
+    }
+}
+
+/// Enumerates different possible unmet requirements that can be returned by meetsRequirements
+public enum UnmetRequirements {
```
kmahar, line 149: naming nit: can we drop the s, so it's just `UnmetRequirement`? an instance of this type just conveys a single unmet requirement.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -122,6 +122,73 @@ open class MongoSwiftTestCase: XCTestCase {
...
+        maxServerVersion: ServerVersion? = nil,
+        acceptableTopologies: [TestTopologyConfiguration]? = nil
+    ) {
+        self.minServerVersion = minServerVersion
+        self.maxServerVersion = maxServerVersion
+        self.topology = acceptableTopologies
+    }
+
+    /// Determines if the given deployment meets this requirement.
+    public func isMet(by version: ServerVersion, _ topology: TestTopologyConfiguration) -> UnmetRequirements? {
```
kmahar, line 172: something that is really emphasized in Swift API design is clarity at the point of use of a method. this method was originally named `isMet(by:)` which I think made sense given that it returned a boolean. however, I wonder if we can rename it now to make it clearer that what we are returning is an unmet requirement, if one exists?

the same applies to the specific `meetsRequirements` helpers we have for the sync and async modules; when I see e.g. 
```swift
 let unmetRequirements = try client.checkRequirements(testRequirements)
            guard unmetRequirements == nil else {
```
in a test it's not immediately apparent that a nil return value from `checkRequirements` corresponds to "all requirements are met"; you just have to know that, or go look at the method to see.

a first stab at a new name would be `getUnmetRequirement` or something along those lines - I think in that case it's more clear that we either return an unmet requirement, or that we return nil there is no unmet requirement.

relevant reading if you haven't seen this yet: https://swift.org/documentation/api-design-guidelines/

nspektor: oh okay yeah this makes a lot of sense. Thanks for linking them the api design guidelines. I renamed both `checkRequirements` and `isMet` to `getUnmetRequirement`, is it confusing/bad that they have the same name?

kmahar: great, thanks! I think the overlapping names are fine.

PR #505, opened 2020-07-02
SWIFT-928 Consolidate setting of client options

kmahar:
This consolidates all of our options settings for clients into a single place, making it easier to reason about and laying the groundwork for doing our own validation of options in a future PR (see [SWIFT-929](https://jira.mongodb.org/browse/SWIFT-929)).

Previously, we set TLS options and the pool size option directly on the pool rather than on the `mongoc_uri_t`.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -6,7 +6,6 @@ internal class ConnectionString {
     private let _uri: OpaquePointer
 
     /// Initializes a new `ConnectionString` with the provided options.
-    // swiftlint:disable:next cyclomatic_complexity
     internal init(_ connectionString: String, options: MongoClientOptions? = nil) throws {
         var error = bson_error_t()
         guard let uri = mongoc_uri_new_with_error(connectionString, &error) else {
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -6,52 +6,124 @@ internal class ConnectionString {
...
-            mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYWRITES, rw)
+        if let replicaSet = options?.replicaSet {
+            guard mongoc_uri_set_option_as_utf8(self._uri, MONGOC_URI_REPLICASET, replicaSet) else {
+                throw MongoError.InvalidArgumentError(message: "Failed to set replicaSet to \(replicaSet)")
+            }
         }
 
         if let rr = options?.retryReads {
-            mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYREADS, rr)
+            guard mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_RETRYREADS, rr) else {
```
kmahar, line 70: turns out these setter methods return bools indicating success, which we have been ignoring 😱

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -65,17 +65,29 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
 
             // TODO: DRIVERS-796 or DRIVERS-990: unskip this test
             guard fileName != "txt-record-with-overridden-uri-option.json" else {
-                return
+                continue
```
kmahar, line 68: this was a bug too 😬 fortunately we pass all of the ones that came after with no issues.

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -65,17 +65,29 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
...
             // parsing the URI, and other times they are not until we try to select a server.
             func doTest() throws -> ConnectionString {
-                let opts = MongoClientOptions(
+                var opts = MongoClientOptions(
                     tlsAllowInvalidCertificates: true,
                     tlsCAFile: URL(string: MongoSwiftTestCase.sslCAFilePath ?? ""),
                     tlsCertificateKeyFile: URL(string: MongoSwiftTestCase.sslPEMKeyFilePath ?? "")
                 )
+
+                // NOTE: since we have to connect to the replica set and attempt server selection to get the SRV
```
kmahar, line 80: here's why we didn't see this before:

Before, the URI had ssl=false, but the client pool had the TLS opts set above (CA file etc.). The client pool having those options set on it let to libmongoc inferring that to mean SSL=true for the pool, which overrode the URI's ssl=false.

Now that we set everything on the URI though, the URI has explicit SSL=false plus the options above. In that case it seems the ssl=false wins out, and no override occurs.

patrickfreed: This seems like we may be sidestepping the purpose of the test though. Aren't we supposed to be verifying that srv resolution works even if the user explicitly disables ssl?

kmahar: so I copied this from what libmongoc does in their equivalent tests:
https://github.com/mongodb/mongo-c-driver/blob/49d5590b0164e629c6adc1ed1b88e6852f62bba9/src/libmongoc/tests/test-mongoc-dns.c#L210

in other drivers, it looks like they are able to implement this test by not actually sending commands and just parsing the connection string and resolving the SRV and TXT records. 

for example, Node implements these tests via a `parseConnectionString` method: https://github.com/mongodb/node-mongodb-native/blob/5808e12f13e39bfddb26cc15fe9b4dd751e1f363/test/unit/core/mongodb_srv.test.js#L32

and it looks like yall do the same in Rust?
https://github.com/mongodb/mongo-rust-driver/blob/a497db1150082fba85880d6b5a652ce8a78932c4/src/test/spec/initial_dns_seedlist_discovery.rs#L40

so I think what we are doing here with checking `tempConnStr.options?["tls"]` is basically equivalent to the Node and Rust tests, the only difference is that we do it prior to SRV/TXT record resolution. note that the SSL/TLS option can not be set via TXT record per the spec though, so it's not possible that this option would change from doing resolution.

taking the route here of overriding the ssl option allows us to make the host list assertions for this test successfully, as doing that requires us to successfully start SDAM and inspect the resulting topology events. an alternative route would be to not override the option, but skip the host list assertion for this case. 

cc @kevinAlbs in case he has any insight

kmahar: I think the point of the test is to verify that users can override the default SSL=true behavior that they get when using a mongodb+srv URI.
we could alternatively maybe only run this particular test case when SSL is turned off for the test replica set? I think that would allow us to confirm that the ssl=false option is successfully respected.

patrickfreed: I think the drivers that just test the parsing are likely not run the tests the way they're supposed to be run. For instance, it looks like the Rust runner is just ignoring the `hosts` field of the spec test file completely.

>we could alternatively maybe only run this particular test case when SSL is turned off for the test replica set? I think that would allow us to confirm that the ssl=false option is successfully respected.

This seems like the right path.  As the spec is written right now, it seems to have conflicting requirements because it says both that the replica set MUST be started with ssl but also that drivers should be able to do an SDAM pass with the given options. Something like a `runOn` with `ssl: true` would be useful to include in the spec test format for stuff like this.

kevinAlbs: I don't have much additional insight, but I agree it seems like the spec has conflicting requirements. Perhaps this warrants a spec change. Since I don't think all drivers can set SSL options after client construction from a URI, maybe the test should omit the list of hosts?

IMO following suit with libmongoc's tests seem reasonable to me, but so does Patrick's suggestion. Either way you'll need a special case for this test.

kmahar: I've updated to run this test without SSL now. I also opened [DRIVERS-1324](https://jira.mongodb.org/browse/DRIVERS-1324) regarding the issues with the spec tests.

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -61,21 +57,35 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
...
-                return
+                continue
+            }
+
+            // this particular test case requires SSL is disabled. see DRIVERS-1324.
+            let requiresTLS = fileName != "txt-record-with-overridden-ssl-option.json"
+
+            // TLS requirement for this test case is not met.
+            guard (requiresTLS && MongoSwiftTestCase.ssl) || (!requiresTLS && !MongoSwiftTestCase.ssl) else {
+                continue
```
patrickfreed, line 70: for this skip and above, I think we should add a print statement mentioning that its being skipped to ensure the cases are being skipped under the appropriate conditions and that we have a constant reminder that we should unskip these at some point.

kmahar: done

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -61,21 +57,37 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
...
+            print("Running test file \(fileName)...")
+
+            let topologyWatcher = TopologyDescriptionWatcher()
+
             // Enclose all of the potentially throwing code in `doTest`. Sometimes the expected errors come when
             // parsing the URI, and other times they are not until we try to select a server.
             func doTest() throws -> ConnectionString {
-                let opts = MongoClientOptions(
-                    tlsAllowInvalidCertificates: true,
-                    tlsCAFile: URL(string: MongoSwiftTestCase.sslCAFilePath ?? ""),
```
kmahar, line 76: I realized that setting the CA file and key file was already handled by `makeTestClient` so we can remove these.

PR #507, opened 2020-07-10
SWIFT-906 Add localThresholdMS option

kmahar:
Adds support for specifying localThresholdMS via client options.

We can't really test this now beyond confirming its set correctly on the URI/overridden correctly. Once we eventually implement server selection tests, it will be tested more thoroughly there.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -54,6 +54,21 @@ internal class ConnectionString {
             }
         }
 
+        let invalidThresholdMsg = "Invalid localThresholdMS %d: must be between 0 and \(Int32.max)"
+        if let localThresholdMS = options?.localThresholdMS {
+            guard let value = Int32(exactly: localThresholdMS), value >= 0 else {
+                throw MongoError.InvalidArgumentError(message: String(format: invalidThresholdMsg, localThresholdMS))
+            }
+
+            guard mongoc_uri_set_option_as_int32(self._uri, MONGOC_URI_LOCALTHRESHOLDMS, value) else {
+                throw MongoError.InvalidArgumentError(message: "Failed to set localThresholdMS to \(value)")
+            }
+            // libmongoc does not validate an invalid value for localThresholdMS set via URI. if it was set that way and
+            // not overridden via options struct, validate it ourselves here.
+        } else if let uriValue = self.options?[MONGOC_URI_LOCALTHRESHOLDMS]?.int32Value, uriValue < 0 {
+            throw MongoError.InvalidArgumentError(message: String(format: invalidThresholdMsg, uriValue))
+        }
+
         if let maxPoolSize = options?.maxPoolSize {
             guard let value = Int32(exactly: maxPoolSize), value > 0 else {
                 throw MongoError.InvalidArgumentError(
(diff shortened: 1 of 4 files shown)


Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,39 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+        // options struct overrides string
+        let connStr3 = try ConnectionString("mongodb://localhost:27017/?localThresholdMS=50", options: opts)
+        expect(connStr3.options?["localthresholdms"]?.int32Value).to(equal(100))
+
+        let tooSmall = -10
+        expect(try ConnectionString(
+            "mongodb://localhost:27017",
+            options: MongoClientOptions(localThresholdMS: tooSmall)
+        )).to(throwError(errorType: MongoError.InvalidArgumentError.self))
+        // unfortunately libmongoc does not validate a negative value provided via URI.
```
kmahar, line 285: (we can check for this and handle it ourselves in the validation logic though)

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,39 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+        // option is parsed correctly from string
+        let connStr2 = try ConnectionString("mongodb://localhost:27017/?localThresholdMS=100")
+        expect(connStr2.options?["localthresholdms"]?.int32Value).to(equal(100))
+
+        // options struct overrides string
+        let connStr3 = try ConnectionString("mongodb://localhost:27017/?localThresholdMS=50", options: opts)
+        expect(connStr3.options?["localthresholdms"]?.int32Value).to(equal(100))
+
+        let tooSmall = -10
+        expect(try ConnectionString(
```
kmahar, line 281: I'm actually not sure if we want to error here, the URI options spec says we should log a warning instead. though we don't have a way to do that now.
libmongoc also just seems to ignore but not log a warning in the case of a negative value here.

what do you think?

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -40,6 +40,16 @@ internal class ConnectionString {
             }
         }
 
+        if let localThresholdMS = options?.localThresholdMS {
+            guard localThresholdMS >= 0 && localThresholdMS <= Int32.max else {
```
patrickfreed, line 44: ditto cast here

Tests/MongoSwiftTests/ConnectionStringTests.swift
patrickfreed: Given our conversation, I think we decided on erroring, right?

Sources/MongoSwift/ConnectionString.swift

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: yup. since we also learned libmongoc does nothing on an invalid value of this option I've updated `ConnectionString` to inspect the value if its set via connection string and assert its non-negative.

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -40,6 +40,24 @@ internal class ConnectionString {
             }
         }
 
+        let invalidLocalThresholdError = MongoError.InvalidArgumentError(
+            message: "localThresholdMS must be between 0 and \(Int32.max)"
```
patrickfreed, line 44: can this include the value the user set?

PR #508, opened 2020-07-10
SWIFT-905 Add heartbeatFrequencyMS option

kmahar:
Adds support for specifying the heartbeatFrequencyMS option via options struct.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -41,30 +41,27 @@ internal class ConnectionString {
         }
 
         if let heartbeatFrequencyMS = options?.heartbeatFrequencyMS {
-            guard heartbeatFrequencyMS >= 500 && heartbeatFrequencyMS <= Int32.max else {
-                throw MongoError.InvalidArgumentError(message: "localThresholdMS must be between 500 and \(Int32.max)")
+            guard let value = Int32(exactly: heartbeatFrequencyMS), value >= 500 else {
+                throw MongoError.InvalidArgumentError(
+                    message: "heartbeatFrequencyMS must be between 500 and \(Int32.max)"
+                )
             }
 
-            guard mongoc_uri_set_option_as_int32(
-                self._uri,
-                MONGOC_URI_HEARTBEATFREQUENCYMS,
-                Int32(heartbeatFrequencyMS)
-            ) else {
+            guard mongoc_uri_set_option_as_int32(self._uri, MONGOC_URI_HEARTBEATFREQUENCYMS, value) else {
                 throw MongoError.InvalidArgumentError(
-                    message: "Failed to set heartbeatFrequencyMS to \(heartbeatFrequencyMS)"
+                    message: "Failed to set heartbeatFrequencyMS to \(value)"
                 )
             }
         }
 
         if let maxPoolSize = options?.maxPoolSize {
-            // libmongoc casts the value to a uint32_t
-            guard maxPoolSize > 0 && maxPoolSize <= UInt32.max else {
+            guard let value = Int32(exactly: maxPoolSize), value > 0 else {
                 throw MongoError.InvalidArgumentError(
-                    message: "Invalid maxPoolSize \(maxPoolSize): must be between 1 and \(UInt32.max)"
+                    message: "Invalid maxPoolSize \(maxPoolSize): must be between 1 and \(Int32.max)"
                 )
             }
-            guard mongoc_uri_set_option_as_int32(self._uri, MONGOC_URI_MAXPOOLSIZE, Int32(maxPoolSize)) else {
-                throw MongoError.InvalidArgumentError(message: "Failed to set maxPoolSize to \(maxPoolSize)")
+            guard mongoc_uri_set_option_as_int32(self._uri, MONGOC_URI_MAXPOOLSIZE, value) else {
+                throw MongoError.InvalidArgumentError(message: "Failed to set maxPoolSize to \(value)")
             }
         }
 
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -40,6 +40,22 @@ internal class ConnectionString {
             }
         }
 
+        if let heartbeatFrequencyMS = options?.heartbeatFrequencyMS {
+            guard heartbeatFrequencyMS >= 500 && heartbeatFrequencyMS <= Int32.max else {
```
kmahar, line 44: 500 is minHeartbeatFrequencyMS as defined in [SDAM](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#minheartbeatfrequencyms).

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,89 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+            default:
+                return
+            }
+        }
+
+        fileprivate init() {}
+    }
+
+    func testHeartbeatFrequencyMSWithMonitoring() throws {
+        guard MongoSwiftTestCase.topologyType == .single else {
```
kmahar, line 324: rather than worrying about heartbeats to multiple servers it was simpler to just run this against standalones.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,89 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+        // started events at the end with no corresponding succeeded event
+        let succeeded = watcher.succeeded
+        let started = watcher.started[0..<succeeded.endIndex]
+
+        // the last started time should be 2s after the second-to-last succeeded time.
+        let lastStart = started.last!
+        let secondToLastSuccess = succeeded[succeeded.count - 2]
+
+        let difference = lastStart.timeIntervalSince1970 - secondToLastSuccess.timeIntervalSince1970
+        expect(difference.rounded()).to(equal(2.0))
```
kmahar, line 348: there's a small amount of variance here like on the order of 1/1000th of a second

Sources/MongoSwift/ConnectionString.swift
patrickfreed, line 44: ditto comment in #509 about doing the cast in the `guard`

Tests/MongoSwiftTests/ConnectionStringTests.swift
patrickfreed: ditto using `beCloseTo` for this

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,89 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+    func testHeartbeatFrequencyMSWithMonitoring() throws {
+        guard MongoSwiftTestCase.topologyType == .single else {
+            print(unsupportedTopologyMessage(testName: self.name))
+            return
+        }
+
+        let watcher = HeartbeatWatcher()
+
+        // verify that we can speed up the heartbeat frequency
+        try self.withTestClient(options: MongoClientOptions(heartbeatFrequencyMS: 2000)) { client in
```
patrickfreed, line 332: also very nice test :+1:

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: done

Sources/MongoSwift/ConnectionString.swift
kmahar: aha, I was thinking I couldn't do it because I need to compare against 500 too, but I forgot I can one-line it with
`guard let value = Int32(exactly: heartbeatFrequencyMS), value >= 500 else { ...`

done

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -40,15 +40,28 @@ internal class ConnectionString {
...
+                throw MongoError.InvalidArgumentError(
+                    message: "Failed to set heartbeatFrequencyMS to \(value)"
+                )
+            }
+        }
+
         if let maxPoolSize = options?.maxPoolSize {
-            // libmongoc casts the value to a uint32_t
-            guard maxPoolSize > 0 && maxPoolSize <= UInt32.max else {
+            guard let value = Int32(exactly: maxPoolSize), value > 0 else {
```
kmahar, line 58: your comment made me realize I could apply the same logic here to do the cast/error check at once. that made me realize we had a bug here (introduced by me): `UInt32.max` is larger than `Int32.max`. if you provided a value that was larger that could not fit into an `Int32` but could fit into a `UInt32`, we wouldn't error here, but we would then crash doing the cast below.
so while libmongoc stores the value as a `uInt32_t`, the API restricts the max value to `Int32.max`.

patrickfreed: nice!

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -40,15 +40,28 @@ internal class ConnectionString {
             }
         }
 
+        if let heartbeatFrequencyMS = options?.heartbeatFrequencyMS {
+            guard let value = Int32(exactly: heartbeatFrequencyMS), value >= 500 else {
+                throw MongoError.InvalidArgumentError(
+                    message: "heartbeatFrequencyMS must be between 500 and \(Int32.max)"
```
patrickfreed, line 46: can this message include the value they set similar to the maxPoolSize error message below?

PR #509, opened 2020-07-13
SWIFT-907 Add serverSelectionTimeoutMS option

kmahar:
Adds support for specifying serverSelectionTimeoutMS via options struct.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -91,6 +91,21 @@ internal class ConnectionString {
             }
         }
 
+        let invalidSSTimeoutMsg = "Invalid serverSelectionTimeoutMS %d: must be between 1 and \(Int32.max)"
+        if let ssTimeout = options?.serverSelectionTimeoutMS {
+            guard let value = Int32(exactly: ssTimeout), value > 0 else {
+                throw MongoError.InvalidArgumentError(message: String(format: invalidSSTimeoutMsg, ssTimeout))
+            }
+
+            guard mongoc_uri_set_option_as_int32(self._uri, MONGOC_URI_SERVERSELECTIONTIMEOUTMS, value) else {
+                throw MongoError.InvalidArgumentError(
+                    message: "Failed to set serverSelectionTimeoutMS to \(value)"
+                )
+            }
+        } else if let uriValue = self.options?[MONGOC_URI_SERVERSELECTIONTIMEOUTMS]?.int32Value, uriValue <= 0 {
+            throw MongoError.InvalidArgumentError(message: String(format: invalidSSTimeoutMsg, uriValue))
+        }
+
         if let tls = options?.tls {
             guard mongoc_uri_set_option_as_bool(self._uri, MONGOC_URI_TLS, tls) else {
                 throw MongoError.InvalidArgumentError(message: "Failed to set tls to \(tls)")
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -78,6 +78,24 @@ internal class ConnectionString {
             }
         }
 
+        if let ssTimeout = options?.serverSelectionTimeoutMS {
+            guard ssTimeout > 0 && ssTimeout <= Int32.max else {
```
patrickfreed, line 82: maybe instead of comparing against the max and doing another cast below, we can do it in one step and just do `let timeout = Int32(exactly: ssTimeout)`

Sources/MongoSwift/MongoClient.swift
```diff
@@ -45,6 +45,10 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// Determines whether the client should retry supported write operations (on by default).
     public var retryWrites: Bool?
 
+    /// Specifies how long the driver should attempt to select a server for before throwing an error. Defaults to 30
+    /// seconds (30000 ms).
+    public var serverSelectionTimeoutMS: Int?
```
patrickfreed, line 50: is this not one of the timeouts we were going to defer for a while? Or is bernie still thinking about it? I can't remember where we left of with that.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,54 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+
+    func testServerSelectionTimeoutMSWithCommand() throws {
+        let opts = MongoClientOptions(serverSelectionTimeoutMS: 1000)
+        try self.withTestClient("mongodb://localhost:27099", options: opts) { client in
+            let start = Date()
+            expect(try client.listDatabases().wait()).to(throwError(errorType: MongoError.ServerSelectionError.self))
+            let end = Date()
+
+            let difference = end.timeIntervalSince1970 - start.timeIntervalSince1970
+            expect(difference.rounded()).to(equal(1.0))
```
patrickfreed, line 312: could use `beCloseTo` here.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -262,4 +262,54 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+        expect(try ConnectionString("mongodb://localhost:27017/?serverSelectionTimeoutMS=\(tooLarge)"))
+            .to(throwError(errorType: MongoError.InvalidArgumentError.self))
+
+        expect(try ConnectionString(
+            "mongodb://localhost:27017",
+            options: MongoClientOptions(serverSelectionTimeoutMS: tooLarge)
+        )).to(throwError(errorType: MongoError.InvalidArgumentError.self))
+    }
+
+    func testServerSelectionTimeoutMSWithCommand() throws {
```
patrickfreed, line 304: this is a nice test :+1:

Sources/MongoSwift/MongoClient.swift
kmahar: it seems like we are safe to have. the likely outcome of the new timeout project is that this option will stick around but have a slightly different (and backwards compatible) new meaning

Sources/MongoSwift/ConnectionString.swift

PR #510, opened 2020-07-14
SWIFT-912 Error if minPoolSize is specified

kmahar:
Throw an error if users provide this option, due to problematic libmongoc implementation. It feels a little weird to throw on this option but not any other unsupported options, but I don't really know what else we can do.


kmahar:
good catch, fixed 👍  will wait to see it actually run on evg before I merge

PR #511, opened 2020-07-15
SWIFT-903, SWIFT-909: Add compression options to MongoClientOptions

kmahar:
This ended up having some API-related concerns so adding @mbroadst as well.

This adds support for specifying compressors (currently only zlib is supported), as well as an optional level to use for zlib compression, via `MongoClientOptions`.

### DIFF
--- a/Sources/MongoSwift/Compressor.swift
+++ b/Sources/MongoSwift/Compressor.swift
@@ -1,32 +1,38 @@
 /// Specifies a library to use for network compression.
-public struct Compressor {
+public struct Compressor: CustomStringConvertible {
+    internal enum _Compressor {
+        case zlib(level: Int32? = nil)
+    }
+
+    /// The compressor to use.
+    internal let _compressor: _Compressor
+
+    private init(_ compressor: _Compressor) {
+        self._compressor = compressor
+    }
+
     /// Use zlib for data compression.
     /// - SeeAlso: https://docs.mongodb.com/manual/reference/glossary/#term-zlib
-    public static let zlib = Compressor("zlib")
+    public static let zlib = Compressor(.zlib())
 
     /// Use zlib for data compression, with the specified compression level.
     /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.zlibCompressionLevel
     public static func zlib(level: Int) throws -> Compressor {
-        try Compressor("zlib", level: level)
-    }
-
-    /// Compressor name.
-    internal let name: String
-
-    internal let zLibLevel: Int32?
-
-    private init(_ compressor: String) {
-        self.name = compressor
-        self.zLibLevel = nil
-    }
-
-    private init(_ compressor: String, level: Int) throws {
         guard (-1...9).contains(level) else {
             throw MongoError.InvalidArgumentError(
                 message: "Invalid zlib compression level \(level): must be between -1 and 9"
             )
         }
-        self.name = compressor
-        self.zLibLevel = Int32(level)
+        return Compressor(.zlib(level: Int32(level)))
+    }
+
+    public var description: String {
+        switch self._compressor {
+        case let .zlib(level):
+            guard let level = level else {
+                return "zlib"
+            }
+            return "zlib(level:\(level))"
+        }
     }
 }
--- a/Sources/MongoSwift/Compressor.swift
+++ b/Sources/MongoSwift/Compressor.swift
@@ -1,0 +1,32 @@
+/// Specifies a library to use for network compression.
+public struct Compressor {
+    /// Use zlib for data compression.
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/glossary/#term-zlib
+    public static let zlib = Compressor("zlib")
+
+    /// Use zlib for data compression, with the specified compression level.
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.zlibCompressionLevel
+    public static func zlib(level: Int) throws -> Compressor {
+        try Compressor("zlib", level: level)
+    }
+
+    /// Compressor name.
+    internal let name: String
+
+    internal let zLibLevel: Int32?
+
+    private init(_ compressor: String) {
+        self.name = compressor
+        self.zLibLevel = nil
+    }
+
+    private init(_ compressor: String, level: Int) throws {
+        guard (-1...9).contains(level) else {
+            throw MongoError.InvalidArgumentError(
+                message: "Invalid zlib compression level \(level): must be between -1 and 9"
+            )
+        }
+        self.name = compressor
+        self.zLibLevel = Int32(level)
+    }
+}
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -33,25 +33,24 @@ internal class ConnectionString {
                 return
             }
 
-            // otherwise, the only valid inputs are a length 1 array containing either zlib or zlib with a level.
-            let nonZlib = compressors.filter { $0.name != "zlib" }
-            guard nonZlib.isEmpty else {
-                throw MongoError.InvalidArgumentError(message: "Unsupported compressor \(nonZlib[0].name)")
-            }
-
+            // otherwise, the only valid inputs is a length 1 array containing either zlib or zlib with a level.
             guard compressors.count == 1 else {
                 throw MongoError.InvalidArgumentError(message: "zlib compressor provided multiple times")
             }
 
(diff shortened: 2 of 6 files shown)


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -24,6 +24,38 @@ internal class ConnectionString {
             }
         }
 
+        if let compressors = options?.compressors {
+            // user specified an empty array, so we should nil out any compressors set via connection string.
+            guard !compressors.isEmpty else {
+                guard mongoc_uri_set_compressors(self._uri, nil) else {
```
kmahar, line 30: I'm not sure if it's actually OK/intended behavior that passing in nil here works. I did test passing in an empty string instead but that generates a libmongoc warning. I asked in #mongoc so we'll see what they say.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -441,4 +441,49 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+        )
+        expect(connStr.compressors).to(equal(["zlib"]))
+        expect(connStr.options?["zlibcompressionlevel"]?.int32Value).to(equal(6))
+
+        // duplicate compressors should error
+        opts.compressors = [.zlib, try .zlib(level: 1)]
+        expect(try ConnectionString("mongodb://localhost:27017", options: opts))
+            .to(throwError(errorType: MongoError.InvalidArgumentError.self))
+
+        // unfortunately, we can't error on an unsupported compressor provided via URI. libmongoc will generate a
```
kmahar, line 486: this is unfortunate. I suppose we could actually parse the URI ourselves and extract the `compressors=x` part if we wanted to in order to check this though. let me know what you think

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -441,4 +441,49 @@ final class ConnectionStringTests: MongoSwiftTestCase {
             options: MongoClientOptions(localThresholdMS: tooLarge)
         )).to(throwError(errorType: MongoError.InvalidArgumentError.self))
     }
+
+    func testCompressionOptions() throws {
+        // zlib level validation
+        expect(try Compressor.zlib(level: -2)).to(throwError(errorType: MongoError.InvalidArgumentError.self))
```
kmahar, line 447: debated getting rid of the level label so you'd just do `.zlib(6)` etc. but I wasn't sure if the meaning of the value was obvious enough without it. thoughts?

patrickfreed: One way to alleviate this would be to define an enum with cases for `.disabled`, `.defaultLevel`, or `.level(Int)`. Then it'd look like `.zlib(.level(6))`.

Sources/MongoSwift/Compressor.swift
```diff
@@ -0,0 +1,32 @@
...
+    /// Use zlib for data compression, with the specified compression level.
+    /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.zlibCompressionLevel
+    public static func zlib(level: Int) throws -> Compressor {
+        try Compressor("zlib", level: level)
+    }
+
+    /// Compressor name.
+    internal let name: String
+
+    internal let zLibLevel: Int32?
```
patrickfreed, line 16: this `Compressor` struct could be used for more than just zlib, so we probably don't want a zlib specific value here. I think a protocol or something that encapsulates the behavior for attaching the compressor to the URI could be stored here instead.

Tests/MongoSwiftTests/ConnectionStringTests.swift
patrickfreed: `mongoc_uri_get_compressors` returns just the valid ones then, right?

Sources/MongoSwift/Compressor.swift
kmahar: what I originally had here was an internal enum like
```swift
enum _Compressor {
    case zlib(level: Int? = nil)
    case zstd
}
```
I got rid of it because there was only one case, but I could bring it back now for the sake of making adding more compressors in the future easier.

Sources/MongoSwift/ConnectionString.swift
kmahar: update: Kevin said it's safe to rely on this, and he's opened [CDRIVER-3751](https://jira.mongodb.org/browse/CDRIVER-3751) about mentioning it in the documentation and adding a test to ensure it continues working.

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: correct

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: would this give the user the ability to do anything they can't already now?
given the override behavior, it seems to me that passing in both `compressors: []` and `compressors: [.zlib(.disabled))]` would have the same effect.

`.default` also seems somewhat contradictory to our usual practice of unspecified=default

Sources/MongoSwift/Compressor.swift
patrickfreed: Yeah I think bringing this back would be nice.

Tests/MongoSwiftTests/ConnectionStringTests.swift
patrickfreed: So default in this case really means "enabled with the default level", so maybe we make `level` argument an `Int? = nil` one. As far as `.disabled` goes, I was just translating the option 0 = disabled, but I agree that having it included doesn't really add much so we can probably just go back to `Int`.

Tests/MongoSwiftTests/ConnectionStringTests.swift
patrickfreed: Are there many other options like this where we can't verify? If not then I'm personally okay with just not verifying this one. If there are, then maybe some manual parsing of the connection string would be valuable.

Sources/MongoSwift/Compressor.swift
mbroadst: If we end up adding more properties to `zlib` we would just use a different case, e.g. `zlibV2(level: Int, comment: String)` ?

Tests/MongoSwiftTests/ConnectionStringTests.swift
mbroadst: I think the label looks nice, just my two cents

Tests/MongoSwiftTests/ConnectionStringTests.swift
mbroadst: Let's defer to a later project where you develop a connection string parser in pure swift. I think the changes will be additive (including potentially throwing an error) right? So we can take our time to think it through more holistically there.

Sources/MongoSwift/Compressor.swift
kmahar: switched to backing this struct with an internal enum.

since the enum is internal and we control all usages, we could just add a new property to the original zlib case. the main benefit is just that we get to be a little more strict about ensuring we don't have a level set for some future compressor that doesn't allow one.

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: ya I'm good with keeping the label.

I used the static property `zlib` plus the throwing static method `zlib(level:)` so that users can just do `.zlib` and not `try .zlib()` when they aren't providing a level

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: I think under the philosophy that these should have been validated all along but aren't currently, you could consider the addition of such an error/validation to be a bug fix. 
but it is a bit breaking in the sense that if you were using an connection string containing an unsupported compressor before and relying on the driver's behavior to do that, it would no longer be the case.

I'm fine with considering the current inability to validate a bug and just opening a ticket to remember that we don't validate in this case right now. while completing the [validation ticket](https://jira.mongodb.org/browse/SWIFT-929) (last one left in the epic after this) it's possible I might encounter more cases that would require connection string parsing to properly validate, but I can't think of any off the top of my head that I've seen yet.

kmahar: FYI I opened [SWIFT-951](https://jira.mongodb.org/browse/SWIFT-951) to track that

PR #512, opened 2020-07-16
SWIFT-902 Add Atlas connection examples

patrickfreed:
[SWIFT-902](https://jira.mongodb.org/browse/SWIFT-902)

This PR adds examples to be used in the Atlas connection UI for both SCRAM and X509 authentication mechanisms.

### DIFF
--- a/Examples/Atlas/Sources/AtlasExamples/main.swift
+++ b/Examples/Atlas/Sources/AtlasExamples/main.swift
@@ -1,0 +1,49 @@
+import Foundation
+import MongoSwift
+import NIO
+
+/// Atlas connection string examples
+
+func scramExample() throws {
+    // START SCRAM EXAMPLE HERE
+    let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
+    let client = try MongoClient(
+        "mongodb+srv://<user>:<password>@<host>/<dbname>?retryWrites=true&w=majority",
+        using: elg
+    )
+    defer {
+        try? client.syncClose()
+        cleanupMongoSwift()
+        try? elg.syncShutdownGracefully()
+    }
+
+    let db = client.db("library")
+    let collection = db.collection("books")
+    // do something with collection
+    // END SCRAM EXAMPLE HERE
+}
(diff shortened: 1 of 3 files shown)


Examples/Atlas/Sources/AtlasExamples/main.swift
```diff
@@ -0,0 +1,49 @@
...
+import MongoSwift
+import NIO
+
+/// Atlas connection string examples
+
+func scramExample() throws {
+    // START SCRAM EXAMPLE HERE
+    let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
+    let client = try MongoClient(
+        "mongodb+srv://<user>:<password>@<host>/<dbname>?retryWrites=true&w=majority",
```
patrickfreed, line 11: The atlas UI currently wants drivers to use the default db here, but it's really more of a shell-ism since it's basically just a less clear `authSource` for drivers. I left it in for consistency with other drivers' examples, but I think this should probably be changed.

kmahar: hmmm, yeah weird. I'm not sure what the correct way to bring that up would be / who to ask about it - @mbroadst do you know?
anyway I agree we should go with consistency with other drivers for now.

mbroadst: A good start might be to @mention Pia in the ticket

patrickfreed: I left a comment on the SWIFT ticket explaining the situation.

patrickfreed:
1. I asked Pia and she said just to do async examples, since they haven't been doing multiple examples for other drivers.

2. Done

Examples/Atlas/Sources/AtlasExamples/main.swift
patrickfreed: Pia responded saying to keep the dbname in there for consistency with other driver examples, but that they'll look into updating it for all examples.

PR #515, opened 2020-07-23
SWIFT-901, SWIFT-929 Add tlsInsecure option, validate options

kmahar:
This PR does the following: 

1. Added `tlsInsecure` option to client options.

2. Adds validation for various combinations of options / usages of options that are invalid, as specified in the URI options spec (see [this section](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#conflicting-tls-options) and the two following it.) When these situations are present in the initial string, libmongoc handles validating them for us, but when we override any of the involved options after the `mongoc_uri_t` is created we do not necessarily get validation. See [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723) for details.

3. I also realized that the "lack of libmongoc errors on invalid timeouts in the input string" issue (mentioned in [CDRIVER-3167](https://jira.mongodb.org/browse/CDRIVER-3167)) encountered when adding `serverSelectionTimeoutMS` and `localThresholdMS` was relevant for the options we've decided to only support via connection string. It seemed like we should at least be consistent and error on invalid values for these, too. It also applies to `maxStalenessSeconds` when specified via connection string.

4. I noticed that libmongoc does not support any of maxIdleTimeMS, waitQueueMultiple, and waitQueueTimeoutMS. Since these do show up in the connection string docs and other drivers support them, it seemed like it might make sense to error when they are provided. Though on the other hand, maybe we should treat them like any other unrecognized option and just do nothing. (libmongoc does at least log on other unrecognized options but does not for these. I am trying to open a CDRIVER ticket about that right now but Jira seems to be on the verge of a crash so I guess I'll wait until later.)

### DIFF
--- a/Sources/MongoSwift/ReadPreference.swift
+++ b/Sources/MongoSwift/ReadPreference.swift
@@ -197,8 +197,8 @@ public struct ReadPreference: Equatable {
         if let maxStaleness = maxStalenessSeconds {
             guard maxStaleness >= MONGOC_SMALLEST_MAX_STALENESS_SECONDS else {
                 throw MongoError.InvalidArgumentError(
-                    message: "Expected maxStalenessSeconds to be >= " +
-                        " \(MONGOC_SMALLEST_MAX_STALENESS_SECONDS), \(maxStaleness) given"
+                    message: "Invalid \(MONGOC_URI_MAXSTALENESSSECONDS) \(maxStaleness): " +
+                        "must be at least \(MONGOC_SMALLEST_MAX_STALENESS_SECONDS)"
                 )
             }
         }
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -96,21 +96,10 @@ internal class ConnectionString {
 
     /// Sets and validates TLS-related on the underlying `mongoc_uri_t`.
     private func applyAndValidateTLSOptions(_ options: MongoClientOptions?) throws {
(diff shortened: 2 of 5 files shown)


Sources/MongoSwift/ReadPreference.swift
```diff
@@ -197,8 +197,8 @@ public struct ReadPreference: Equatable {
         if let maxStaleness = maxStalenessSeconds {
             guard maxStaleness >= MONGOC_SMALLEST_MAX_STALENESS_SECONDS else {
                 throw MongoError.InvalidArgumentError(
-                    message: "Expected maxStalenessSeconds to be >= " +
-                        " \(MONGOC_SMALLEST_MAX_STALENESS_SECONDS), \(maxStaleness) given"
+                    message: "Invalid \(MONGOC_URI_MAXSTALENESSSECONDS) \(maxStaleness): " +
```
kmahar, line 200: this was just an update to make it match the same error thrown when this happens in the connection string

kmahar:
manual evg patch since PR integration is down: https://evergreen.mongodb.com/version/5f1a155c7742ae5c064c61f5

kmahar:
I ended up just introducing a helper to generate messages about options being out of range, which removed the need to use String formatting stuff. to use that on Linux you apparently need to use `withCString` to pass in the string.

kmahar:
yeah, I agree about erroring on unknown options. we just don't have the ability to do that now unfortunately since libmongoc only exposes the known options via its API after parsing.

PR #519, opened 2020-08-19
SWIFT-957 Fill in namespace on invalidate events

kmahar:
This PR works around the fact that invalid events do not have associated namespaces by accepting a `MongoNamespace` in `ChangeStream.init`, and then storing that namespace under `ns` in the `ChangeStream'`s `BSONDecoder'`s `userInfo`.

`ChangeStreamEvent` then looks for the matching key when decoding, and falls back to it if there is no key present in the original document.

Per conversation in driver-devs it appears that as of now anyway it's impossible to encounter an invalidate event on a client.

### DIFF
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -50,6 +50,13 @@ public struct ResumeToken: Codable, Equatable {
     }
 }
 
+// TODO: SWIFT-981: Remove this.
+/// The key we use for storing a change stream's namespace in it's `userInfo`. This allows types
+/// using the decoder e.g. `ChangeStreamEvent` to access the namespace even if it is not present in the raw
+/// document the server returns. Ok to force unwrap as initialization never fails.
+// swiftlint:disable:next force_unwrapping
+internal let changeStreamNamespaceKey = CodingUserInfoKey(rawValue: "namespace")!
+
 // sourcery: skipSyncExport
 /// A MongoDB change stream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
@@ -87,6 +94,7 @@ public class ChangeStream<T: Codable>: CursorProtocol {
         stealing changeStreamPtr: OpaquePointer,
         connection: Connection,
         client: MongoClient,
+        namespace: MongoNamespace,
         session: ClientSession?,
         decoder: BSONDecoder,
         options: ChangeStreamOptions?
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/ChangeStream.swift
```diff
@@ -50,6 +50,13 @@ public struct ResumeToken: Codable, Equatable {
     }
 }
 
+// TODO: SWIFT-981: Remove this.
+/// The key we use for storing a change stream's namespace in it's `userInfo`. This allows types
+/// using the decoder e.g. `ChangeStreamEvent` to access the namespace even if it is not present in the raw
+/// document the server returns. Ok to force unwrap as initialization never fails.
+// swiftlint:disable:next force_unwrapping
+internal let changeStreamNamespaceKey = CodingUserInfoKey(rawValue: "namespace")!
```
kmahar, line 58: ideally this would have lived on `ChangeStream`, but you can't have stored static properties on generic types 🤔

Sources/MongoSwift/ChangeStream.swift
```diff
@@ -99,7 +107,8 @@ public class ChangeStream<T: Codable>: CursorProtocol {
             type: .tailableAwait
         )
         self.client = client
-        self.decoder = decoder
+        self.decoder = BSONDecoder(copies: decoder, options: nil)
```
kmahar, line 110: since we are updating the userInfo I figured it was a bit safer to make a copy.

this did make me think about the fact that right now users can't give us userInfo that we should use when decoding their custom types, even though they can configure all of the other encoder/decoder options via the strategies. I don't think that's too urgent though, we can wait until if/when someone asks for this capability.

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -73,4 +74,29 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.operationType = try container.decode(OperationType.self, forKey: .operationType)
+        self._id = try container.decode(ResumeToken.self, forKey: ._id)
+
+        do {
+            self.ns = try container.decode(MongoNamespace.self, forKey: .ns)
+        } catch {
+            guard let ns = decoder.userInfo[changeStreamNamespaceKey] as? MongoNamespace else {
+                throw error
```
kmahar, line 93: I considered just filling in with an empty namespace or something here, but that seemed like it would hide errors from the server in cases where a namespace is unexpectedly missing.

kmahar:
someday I'll actually remember to sync it 🤦‍♀️ I should probably set up a githook

kmahar:
@patrickfreed it turns out that on 3.6 the behavior is slightly different when a collection is dropped; there is no `drop` event, only an `invalidate` event. I updated the test accordingly.
when testing locally on 3.6 I also noticed that the skip message printed in the second part of the test was misleading and made it look like the whole test was skipped rather than just the second half, so I ended up splitting this up into two tests.

PR #520, opened 2020-08-21
SWIFT-969 Manually clean up mongoc_uri_t when options validation fails

kmahar:


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -4,6 +4,8 @@ import CLibMongoC
 internal class ConnectionString {
     /// Pointer to the underlying `mongoc_uri_t`.
     private let _uri: OpaquePointer
+    /// Tracks whether we have already destroyed the above pointer.
+    private var destroyedPtr = false
```
kmahar, line 8: I considered going with a `State` enum approach. however, this would have required wide-sweeping changes throughout the type since we use `self._uri` _a lot_, including in various computed properties where we can't throw. 

Given that the only time we manually destroy the pointer is in `init` (and it's only in a case where the init throws, and therefore `self` doesn't exist anymore), it's not possible for us to accidentally use a destroyed `self._uri` anywhere else in the type, so it seems fairly low risk to go with this approach.

open to any other opinions or ideas you have though, I figured I would just start out with the simplest solution.

Sources/MongoSwift/Operations/CountDocumentsOperation.swift
```diff
@@ -64,21 +64,28 @@ internal struct CountDocumentsOperation<T: Codable>: Operation {
...
         var error = bson_error_t()
-        var reply = BSONDocument()
-        let count = self.collection.withMongocCollection(from: connection) { collPtr in
-            self.filter.withBSONPointer { filterPtr in
-                withOptionalBSONPointer(to: opts) { optsPtr in
-                    ReadPreference.withOptionalMongocReadPreference(from: self.options?.readPreference) { rpPtr in
-                        reply.withMutableBSONPointer { replyPtr in
-                            mongoc_collection_count_documents(collPtr, filterPtr, optsPtr, rpPtr, replyPtr, &error)
+
+        return try self.collection.withMongocCollection(from: connection) { collPtr in
```
kmahar, line 68: this is actually totally unrelated, but I encountered it when running the manual leak checking via the atExit handler hack described here. the problem is the leak of the reply `bson_t`.
we actually fixed all other occurrences of this problem back in https://github.com/mongodb/mongo-swift-driver/commit/b7c310a1a32c26053b1c3cc4a3d3c025896d30d5 / SWIFT-824; I'm not sure how I missed this one then. 🤔

PR #521, opened 2020-08-22
SWIFT-958 Ensure nil is returned from next() before ClosedCursorError

patrickfreed:
SWIFT-958

This fixes a bug where cursors would iterate forever after encountering a non-`DecodingError` by ensuring `next` returns `nil` once after the error. This does not apply to `tryNext` or `toArray`, since those are `throws`. Also, in the case of `tryNext`, `nil` doesn't necessarily imply that the cursor is dead.

### DIFF
--- a/Tests/MongoSwiftSyncTests/MongoCursorTests.swift
+++ b/Tests/MongoSwiftSyncTests/MongoCursorTests.swift
@@ -241,7 +241,12 @@ final class MongoCursorTests: MongoSwiftTestCase {
     }
 
     func testCursorTerminatesOnError() throws {
-        try self.withTestNamespace { _, _, coll in
+        try self.withTestNamespace { client, _, coll in
+            guard try client.supportsFailCommand() else {
+                printSkipMessage(testName: self.name, reason: "failCommand not supported")
+                return
+            }
+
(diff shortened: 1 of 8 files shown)


Tests/MongoSwiftSyncTests/MongoCursorTests.swift
```diff
@@ -238,4 +238,40 @@ final class MongoCursorTests: MongoSwiftTestCase {
...
+            let cursor = try coll.find([:], options: FindOptions(batchSize: 1))
+
+            let fp = FailPoint.failCommand(failCommands: ["getMore"], mode: .times(1), errorCode: 10)
+            try fp.enable()
+            defer { fp.disable() }
+
+            var count = 0
+            for result in cursor {
+                expect(count).to(beLessThan(2))
+                if count >= 2 {
```
patrickfreed, line 256: I could not get these tests to halt after failing an `expect`, so I manually added breaks. I tried setting `self.continueAfterFailure = false`, but it didn't seem to work.

kmahar: huh, weird 🤔

PR #522, opened 2020-08-27
SWIFT-974 Fix leak in Collection.findOne when encountering DecodingError

patrickfreed:
SWIFT-974

This PR fixes a bug in `Collection.findOne` where the cursor would get leaked if a `DecodingError` occurred while decoding the result.

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Read.swift
+++ b/Sources/MongoSwift/MongoCollection+Read.swift
@@ -58,7 +58,7 @@ extension MongoCollection {
     ) -> EventLoopFuture<T?> {
         let options = options.map { FindOptions(from: $0) }
         return self.find(filter, options: options, session: session).flatMap { cursor in
-            cursor.next().afterSuccess { _ in cursor.kill() }
+            cursor.next().always { _ in _ = cursor.kill() }
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoCollection+Read.swift
```diff
@@ -58,7 +58,7 @@ extension MongoCollection {
     ) -> EventLoopFuture<T?> {
         let options = options.map { FindOptions(from: $0) }
         return self.find(filter, options: options, session: session).flatMap { cursor in
-            cursor.next().afterSuccess { _ in cursor.kill() }
+            cursor.next().always { _ in _ = cursor.kill() }
```
patrickfreed, line 61: I opted to just kill the cursor in the background, since I don't think we actually care if it succeeds or not, and we don't need to wait until it finishes to return the result.

kmahar: I think that's fine. there is the possible edge case that it doesn't complete before your EventLoopGroup shuts down, like if you immediately called client.close() after this method, but in that case you are probably shutting down your application anyway.
can we add a test for findOne + a decoding error though?

PR #523, opened 2020-08-28
SWIFT-961 Add withBSONPointer to new BSONDocument

patrickfreed:
SWIFT-961

This PR adds [`swift-bson`](https://github.com/mongodb/swift-bson) as a dependency to the driver and adds some extensions to allow interop with libbson's `bson_t`.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -342,7 +342,8 @@ public final class ClientSession {
         }
 
         let sessionDoc = BSONDocument(copying: bson)
-        doc["lsid"] = sessionDoc["lsid"]
+        // key that libmongoc uses to store the client session id in options documents
+        doc["sessionId"] = sessionDoc["sessionId"]
     }
 
     /**
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -331,12 +331,18 @@ public final class ClientSession {
             throw ClientSession.SessionInactiveError
         }
 
-        var error = bson_error_t()
-        try doc.withMutableBSONPointer { docPtr in
-            guard mongoc_client_session_append(session, docPtr, &error) else {
-                throw extractMongoError(error: error)
-            }
+        guard let bson = bson_new() else {
+            fatalError("failed to allocate bson_t")
         }
+        defer { bson_destroy(bson) }
+
+        var error = bson_error_t()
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -331,12 +331,18 @@ public final class ClientSession {
             throw ClientSession.SessionInactiveError
         }
 
+        guard let bson = bson_new() else {
+            fatalError("failed to allocate bson_t")
+        }
+        defer { bson_destroy(bson) }
+
         var error = bson_error_t()
-        try doc.withMutableBSONPointer { docPtr in
```
patrickfreed, line 335: I opted to only add the immutable version, since I don't think there would be any way for us to track the writer index properly if we exposed a mutable pointer. This is the only place that is using `withMutableBSONPointer` that would need to continue to do so once we swapped out libraries, so I just added a manual implementation that drops into libbson.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -331,12 +331,18 @@ public final class ClientSession {
...
-            guard mongoc_client_session_append(session, docPtr, &error) else {
-                throw extractMongoError(error: error)
-            }
+        guard mongoc_client_session_append(session, bson, &error) else {
+            throw extractMongoError(error: error)
         }
+
+        let sessionDoc = BSONDocument(copying: bson)
+        doc["lsid"] = sessionDoc["lsid"]
```
patrickfreed, line 345: this is a little hack that I know will work just because I went and looked at the implementation of `mongoc_client_session_append`.

Tests/BSONTests/BSONPointerUtilsTests.swift
```diff
@@ -0,0 +1,40 @@
+import CLibMongoC
+import Foundation
+@testable import MongoSwift
+import Nimble
+@testable import SwiftBSON
+import TestsCommon
+import XCTest
+
+final class BSONPointerUtilsTests: MongoSwiftTestCase {
```
patrickfreed, line 9: these tests won't really be necessary once we merge the library in and start using it, but they were helpful as sanity checks to make sure the functions I implemented actually worked.

Sources/MongoSwift/ClientSession.swift
patrickfreed: When I left this comment, I was actually using the wrong key ("lsid"), it should be "sessionId".

PR #526, opened 2020-09-17
SWIFT-766 Upgrade to libmongoc 1.17.0

patrickfreed:
SWIFT-766

This PR updates the vendored version of the wrapped C driver to 1.17.0. This brings with it a number of under-the-hood improvements to topology monitoring and connection establishment, with the most notable changes being:
- Support for streamable server monitoring to reduce recovery time when the server topology changes
- Support for Online Certificate Status Protocol (OCSP).
- Reduced number of round trips for SCRAM and x509 authentication


.evergreen/config.yml
```diff
@@ -11,7 +11,7 @@ command_type: system
 # Protect ourself against rogue test case, or curl gone wild, that runs forever
 # Good rule of thumb: the averageish length a task takes, times 5
 # That roughly accounts for variable system performance for various buildvariants
-exec_timeout_secs: 1800 # 30 minutes is the longest we'll ever run
+exec_timeout_secs: 3600 # 60 minutes is the longest we'll ever run
```
patrickfreed, line 14: Now that libmongoc is using the streaming protocol in 4.4+, the failpoint based spec tests take a lot longer (see SERVER-45686 and SERVER-50500). Unfortunately, the 4.4+ auth sharded tests now run over the 30 minute mark and fail due to timing out even with sub-500ms minHeartbeatFrequency, necessitating this change.

From looking at the evergreen runs of the C driver, I couldn't find a 4.4+ sharded auth+ssl variant to compare against. I also checked the Go driver's evergreen runs, and their 4.4+ sharded auth+ssl tests only take about 6m, which seems impossible (?). I will do some more digging tomorrow to see if our tests are slower than the norm.

Sources/MongoSwift/ClientSession.swift
```diff
@@ -103,7 +103,7 @@ public final class ClientSession {
         /// A transaction has been started, but no operation has been sent to the server.
         case starting
         /// A transaction is in progress.
-        case inProgress
+        case inProgress = "in_progress"
```
patrickfreed, line 106: The spec tests were (annoyingly) updated to use snake_case here instead of camelCase, requiring this change.

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -7,6 +7,10 @@ internal class ConnectionString {
     /// Tracks whether we have already destroyed the above pointer.
     private var destroyedPtr = false
 
+    /// Minimum possible value for a heartbeatFrequencyMS specified via the URI or options.
+    /// This may be overriden in the tests via an internal option on `MongoClientOptions`.
+    private var minHeartbeatFrequencyMS: Int32 = 500
```
patrickfreed, line 12: I added a patch to our vendored libmongoc to allow setting this number really low for the tests, so I also had to move up the validation of this option to the Swift layer.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -111,10 +114,6 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// Specifies a WriteConcern to use for the client.
     public var writeConcern: WriteConcern?
 
-    private enum CodingKeys: CodingKey {
```
patrickfreed, line 114: this was duplicated in the tests

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -75,12 +75,17 @@ open class MongoSwiftTestCase: XCTestCase {
 
     /// Generates a unique collection name of the format "<Test Suite>_<Test Name>_<suffix>". If no suffix is provided,
     /// the last underscore is omitted.
+    ///
+    /// For compatibility with older servers, this name will be truncated so the entire namespace fits in 120
```
patrickfreed, line 79: some of the newer spec tests resulted in names that were too long, so I added some truncation logic.

Tests/MongoSwiftSyncTests/RetryableWritesTests.swift
```diff
@@ -108,14 +112,19 @@ final class RetryableWritesTests: MongoSwiftTestCase {
                 }
 
                 if test.outcome.error ?? false {
-                    expect(seenError).toNot(beNil(), description: test.description)
```
patrickfreed, line 111: retryable writes tests now have error labels in them and assertions based on their presence, so this runner was updated to do that.

Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
```diff
@@ -64,33 +64,54 @@ extension TransactionOptions: Decodable {
     }
 }
 
-extension ReadPreference.Mode: Decodable {}
+extension ReadPreference.Mode: Decodable {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        var string = try container.decode(String.self)
+        // spec tests capitalize first letter of mode, so need to account for that.
```
patrickfreed, line 71: the spec tests changed how they specify the mode too, so a custom decodable initializer was also necessary here.

Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
```diff
@@ -64,33 +64,54 @@ extension TransactionOptions: Decodable {
...
+            let mode = try container.decode(ReadPreference.Mode.self)
+            self.init(mode)
+        }
     }
 }
 
 extension MongoClientOptions: Decodable {
     private enum CodingKeys: String, CodingKey {
-        case retryReads, retryWrites, w, readConcernLevel, mode = "readPreference"
+        case retryReads, retryWrites, w, readConcernLevel, readPreference, heartbeatFrequencyMS
```
patrickfreed, line 102: heartbeatFrequencyMS is specified in some of the transactions tests and we weren't picking it up. Once I started lowering it all the way to 50, some tests started to fail. Now it only lowers to 50 if the spec tests don't specify a heartbeatFrequencyMS.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/TestOperationResult.swift
```diff
@@ -184,66 +184,71 @@ struct ErrorResult: Equatable, Decodable {
         self.errorLabelsOmit = try container.decodeIfPresent([String].self, forKey: .errorLabelsOmit)
     }
 
-    public func checkErrorResult(_ error: Error) throws {
```
patrickfreed, line 187: To help debugging, I added a description to all of the assertions in here. I also removed the `throws` from everything because nothing in here actually throws an error.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -407,7 +407,12 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
                 }
 
                 // notMaster error
-                let failPoint = FailPoint.failCommand(failCommands: ["getMore"], mode: .times(1), errorCode: 10107)
```
patrickfreed, line 410: 4.4+ only checks for error labels, so I updated these to include them.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -104,10 +104,6 @@ let shouldWarnButLibmongocErrors: [String: [String]] = [
         "Non-numeric wTimeoutMS causes a warning",
         "Too low wTimeoutMS causes a warning",
         "Invalid journal causes a warning"
-    ],
-    "valid-warnings.json": [
-        "Empty integer option values are ignored",
-        "Empty boolean option value are ignored"
```
patrickfreed, line 110: libmongoc stopped erroring in these cases. I would prefer if we continued to error too, but I'm not sure if it's possible short of us writing our own URI parser.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -346,16 +345,15 @@ final class ConnectionStringTests: MongoSwiftTestCase {
             sleep(5) // sleep to allow heartbeats to occur
         }
 
-        // in case there's an uneven number of events (i.e. client was closed mid-heartbeat) drop any
```
patrickfreed, line 349: in the streaming protocol, heartbeat started events are fired as soon as the previous heartbeat succeeds, so this needed to be updated to check the time between successes instead of between the last successes instead of between the last success and the previous heartbeat's start.

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -76,70 +76,66 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
...
+                    guard testCase.error != true else {
+                        XCTFail("Expected error for test case \(testCase.comment ?? ""), got none")
+                        return
+                    }
+
+                    // "You MUST verify that the set of ServerDescriptions in the client's TopologyDescription
+                    // eventually matches the list of hosts."
+                    // This needs to be done before the client leaves scope to ensure the SDAM machinery
+                    // keeps running.
+                    expect(topologyWatcher.lastTopologyDescription?.servers.map { $0.address })
```
patrickfreed, line 104: now that the SDAM machinery is started in the background and is more async, a race was happening between this scope exiting and the full topology being discovered. To ensure the topology is fully discovered every time, I moved all the assertions into the same scope as the client.

etc/inttypes-non-modular-header-workaround.diff
```diff
@@ -1,239 +0,0 @@
-diff --git a/Sources/CLibMongoC/bson/bson-iso8601.c b/Sources/CLibMongoC/bson/bson-iso8601.c
```
patrickfreed, line 1: @mbroadst This patch wasn't applying cleanly, so I tried just removing it and the driver seems to build and run fine on all our evergreen platforms. Do you think we're okay just leaving it off?

etc/vendor-libmongoc.sh
```diff
@@ -103,11 +103,13 @@ echo "RENAMING header files"
                   \
```
patrickfreed, line 103: a few more imports need their format updated

kmahar (changes requested):
looks great, just some small things and questions. thanks for doing this work - I didn't anticipate it was going to require so many test changes!

regarding OCSP, I admittedly know very little, but I know you've spent some time thinking about it for Rust. is it problematic that we're bringing in support and might not necessarily be able to support it to the same degree later? I'm not seeing anything regarding OCSP/any way to configure it in [swift-nio-ssl](https://github.com/apple/swift-nio-ssl)'s API, though given that it wraps BoringSSL, it seems like it could potentially be exposed in the future.

.evergreen/config.yml
patrickfreed: After doing some digging, I found that the deinitiliziar of a sync client takes about 0.5s to complete every time, and in the spec test runner we were creating a lot of short-lived clients for things like fail points or verifying the test outcomes. We were also running `distinct` on every shard for every test, even when it wasn't totally necessary. I made some changes to stop the extra `distinct` calls and also to try to reuse the clients as much as possible and was able to get significant performance improvements.

On a 4.4 sharded cluster, I was able to actually make some decent improvements even on the test times we're seeing on master, particularly for retryable reads:

|                  | master | 1.17 first pass | 1.17 after optimization |   |
|------------------|--------|-----------------|-------------------------|---|
| Transactions     | 233s   | timeout         | 212s                    |   |
| Retryable Writes | 16s    | 88s             | 19s                     |   |
| Retryable Reads  | 150s   | 558s            | 42s                     |   |

The optimized column are the tests running on my machine, so they'll probably be a bit slower on evergreen.

There's still a lot to be improved on, so I filed SWIFT-1008 to track work for further improving the test suite.

patrickfreed (commented):
BoringSSL itself seems to support OCSP (https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Certificate-verification), so it shouldn't be that hard to expose it in NIO's wrapper I think. In the meantime, we could try to disable it in libmongoc by manually setting `tlsDisableCertificateRevocationCheck` to true, though this appears to only work on OpenSSL but not SecureChannel (macOS) or LibreSSL (http://mongoc.org/libmongoc/current/configuring_tls.html#server-certificate-verification). We could also patch our vendored libmongoc to manually disable it completely, though that could be a lot of work to effectively remove a feature. I wonder if, by the time we plan to release the pure Swift version of the driver, OCSP will be required anyways, so we may as well just leave it on until then.

Sources/MongoSwift/Operations/RunCommandOperation.swift
```diff
@@ -35,15 +40,53 @@ internal struct RunCommandOperation: Operation {
...
+        options: RunCommandOptions?,
+        serverAddress: ServerAddress? = nil
+    ) {
         self.database = database
         self.command = command
         self.options = options
+        self.serverAddress = serverAddress
     }
 
     internal func execute(using connection: Connection, session: ClientSession?) throws -> BSONDocument {
```
patrickfreed, line 57: I added a `runCommand` overload that allows commands be sent to specific servers based on address to be used from the tests. This allows us to reuse existing clients that are connected to muliple mongoses to route failpoints to individual servers, instead of creating clients that are only connected to individual servers themselves.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -30,33 +39,28 @@ open class MongoSwiftTestCase: XCTestCase {
     /// Gets the connection string to use from the environment variable, $MONGODB_URI. If the variable does not exist,
     /// will return a default of "mongodb://127.0.0.1/". If singleMongos is true and this is a sharded topology, will
     /// edit $MONGODB_URI as needed so that it only contains a single host.
-    public static func getConnectionString(singleMongos: Bool = true) -> String {
-        let uri = Self.uri
-
+    public static func getConnectionString(singleMongos: Bool = true) -> ConnectionString {
```
patrickfreed, line 42: I updated these helpers to operate over our `ConnectionString` type rather than `String` since it has a lot of useful utilities.

etc/lower-minheartbeatfrequencyms.diff
```diff
@@ -0,0 +1,13 @@
+diff --git a/Sources/CLibMongoC/mongoc/mongoc-topology-private.h b/Sources/CLibMongoC/mongoc/mongoc-topology-private.h
+index 350b3b5..0e8ea28 100644
+--- a/Sources/CLibMongoC/mongoc/mongoc-topology-private.h
++++ b/Sources/CLibMongoC/mongoc/mongoc-topology-private.h
+@@ -28,7 +28,7 @@
+ #include "mongoc-client-session-private.h"
+ #include "mongoc-crypt-private.h"
+ 
+-#define MONGOC_TOPOLOGY_MIN_HEARTBEAT_FREQUENCY_MS 500
++#define MONGOC_TOPOLOGY_MIN_HEARTBEAT_FREQUENCY_MS 50
```
mbroadst, line 10: This seems dangerous, are you enforcing this minimum elsewhere? It may speed up test runs, but in actual deployments this could result in very tight loops in e.g. failover scenarios.

etc/vendor-libmongoc.sh
```diff
@@ -103,11 +103,13 @@ echo "RENAMING header files"
...
                   -e '/private/! s+include "mongoc.h"+include "CLibMongoC_mongoc.h"+' \
                   -e '/private\|defs/! s+include "mongoc-+include "CLibMongoC_mongoc-+' \
+                  -e '/private/! s+include "mongoc/+include "CLibMongoC_+' \
                   -e '/private/! s+include <mongoc-+include <CLibMongoC_mongoc-+' \
                   -e '/private/ s+include "mongoc/+include "+' \
                   \
                   -e 's+CLibMongoC_utlist.h+utlist.h+' \
-                  -e 's+PRId64+\"lld\"+'
+                  -e 's+PRId64+\"lld\"+' \
+                  -e 's+include <common-thread-private.h>+include "CLibMongoC_common-thread-private.h"+'
```
mbroadst, line 112: nit: I tried to group the sed rules by folder, `common` goes before `bson` and `mongoc`

etc/vendor-libmongoc.sh
```diff
@@ -130,16 +132,9 @@ echo "RENAMING header files"
...
-# how clang is building the module
-echo "PATCHING libmongoc"
-git apply "${ETC_DIR}/inttypes-non-modular-header-workaround.diff"
-
-# These patches are temporary workarounds to give us early access to the directConnection URI option and a bug fix.
-# This should be removed from the script when we update our vendored libmongoc to 1.17 via SWIFT-766.
-git apply "${ETC_DIR}/CDRIVER-3532-directConnection.diff"
-git apply "${ETC_DIR}/CDRIVER-3623-fix-setting-apm-in-pooled-client.diff"
+# Here we would apply any number of larger patches that don't fit into a single sed line.
+# echo "PATCHING libmongoc"
```
mbroadst, line 136: looks like the `echo` was left commented out

etc/inttypes-non-modular-header-workaround.diff
mbroadst: hmm, I guess so! I can't remember what motivated the change originally, maybe it was something on Travis

Sources/MongoSwift/Operations/RunCommandOperation.swift
mbroadst: That seems like something you'd want to decide in a different piece of work, it's not really "upgrading libmongoc" right?

.evergreen/config.yml
mbroadst: Oh here's where I can comment :smile: In order to make those tests run faster you need to reduce your `heartbeatFrequencyMS`. I can't find any reference to this in the specs, and it seems like a major oversight if we forgot to add it, but that value is what's used for the `maxAwaitTimeMS` of the call to stream "hellos" so you need to set it a relatively small values in tests where you expect failure in order to speed up cancellation of those streams.

kmahar: I think it's hard to see this because the PR is quite large and Patrick already did some iterating on the initial work, but he already adjusted `heartbeatFrequencyMS` down significantly for tests at the start, and we were still experiencing the timeouts, what he describes here is follow-on work.

regarding the clients, that seems weird. this program (using the async API directly, but doing exactly what the sync client would be doing in its shutdown) outputs on average ~0.002s for me locally:
```swift
let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
let client = try MongoClient(using: elg)
let _ = try client.db("test").runCommand(["isMaster": 1]).wait()

let start = Date()
try client.syncClose()
try elg.syncShutdownGracefully()
print("time: \(Date().timeIntervalSince(start))")
```

Tests/MongoSwiftTests/ConnectionStringTests.swift
kmahar: ah, so I meant that I think we should:
1) accept the libmongoc changes to comply with the spec as-is rather than trying to work around them, but also
2) plan that when we do write our own URI parser, we start erroring here again (I agree with you that I'd prefer to error)

I think this is consistent with how we handled https://jira.mongodb.org/browse/SWIFT-951; in general (in Swift) we have opted to error on invalid values of known options

.evergreen/config.yml
patrickfreed: Hm, I ran your example and got the similar ~0.5s time. Almost all the time seems to be spent in the call to `mongoc_client_pool_destroy`. Are you using the 1.17.0 branch?

Tests/MongoSwiftTests/ConnectionStringTests.swift
patrickfreed: Af okay, makes sense. Filed SWIFT-1009.

Sources/MongoSwift/Operations/RunCommandOperation.swift
patrickfreed: >That seems like something you'd want to decide in a different piece of work, it's not really "upgrading libmongoc" right? I'm not sure you would come up with this solution in isolation, maybe you would want to use a custom server selector for instance.

This overload is testing-only and not exposed to the public API. I agree that a comprehensive solution is preferable and that what is currently implemented would not be it, but this is just a temporary measure to make these tests to run faster in the meantime, since we couldn't merge the libmongoc upgrade otherwise.

>I'm also not sure how to comment on your other comment above (re: closing sync clients), but it's likely that the massive gains you saw in your testing could also relate to reducing the hardcoded minHeartbeatFrequencyMS to 50ms. 

Lowering to 50ms certainly did help, but the tests were still timing out. Reducing client re-usage got them to run without timing out.

>How were you measuring the deinit of the client? 0.5s seems like a very long time, and probably indicative of a bug, rather than necessitating sharing clients across tests.

I just created `Date`s before and after running the `deinit`. From looking further at individual statements in the deinit, it seems the culprit is `mongoc_client_pool_destroy`, and it taking that long could certainly be a bug. I need to look into it some more.

Also note, we're not sharing clients across tests, we're now just reusing clients within individual tests.

etc/lower-minheartbeatfrequencyms.diff
patrickfreed: We are enforcing the minimum in Swift now, see [here](https://github.com/mongodb/mongo-swift-driver/pull/526#discussion_r489829243).

etc/vendor-libmongoc.sh
patrickfreed: fixed

etc/vendor-libmongoc.sh
patrickfreed: fixed

etc/lower-minheartbeatfrequencyms.diff
patrickfreed: The previous validation wasn't sufficient for limiting immediate SDAM checks; updated to allow dynamic setting of the minHeartbeatFrequencyMS via another patch in mongoc. Now we manually set it for every client with a default of 500 which can be configured as in previous patches of this PR.

Sources/MongoSwift/Operations/RunCommandOperation.swift
```diff
@@ -10,6 +10,10 @@ public struct RunCommandOptions: Encodable {
     /// command that reads.
     public var readPreference: ReadPreference?
 
+    /// Opaque value representing a server to run the command on.
+    /// For usage in tests only.
+    internal var serverId: Int?
```
kmahar, line 15: it seems a little confusing to me that this exists on the struct but then gets filled in via the runCommand operation. I could see someone trying to provide an id directly via the options, like
```swift
var opts = RunCommandOptions()
opts.serverId = 123
```

only to have it overwritten when the command is actually being executed. it might be a little clearer to just add the server id directly to the document returned from `encodeOptions`

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -215,14 +223,16 @@ extension SpecTestFile {
 
             // Due to strange behavior in mongos, a "distinct" command needs to be run against each mongos
             // before the tests run to prevent certain errors from ocurring. (SERVER-39704)
-            if MongoSwiftTestCase.topologyType == .sharded, let collName = self.collectionName {
-                for connStr in MongoSwiftTestCase.getConnectionStringPerHost() {
-                    let client = try MongoClient.makeTestClient(connStr)
-                    _ = try client.db(self.databaseName).collection(collName).distinct(fieldName: "_id")
+            if MongoSwiftTestCase.topologyType == .sharded,
+                let collName = self.collectionName,
+                test.description.contains("distinct") {
```
kmahar, line 228: this is new. why does this only apply to test cases where the description contains "distinct"? is that just the only command that suffers from the linked SERVER issue?

kmahar (commented):
looks good! just some small questions / nits

Sources/MongoSwift/SDAM.swift
```diff
@@ -146,6 +149,7 @@ public struct ServerDescription {
     /// mongoc_server_description_t.
     internal init(_ description: OpaquePointer) {
         self.address = ServerAddress(mongoc_server_description_host(description))
+        self.serverId = Int(mongoc_server_description_id(description))
```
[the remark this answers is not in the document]
patrickfreed: fixed

Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
```diff
@@ -63,22 +72,39 @@ internal struct FailPoint: Decodable {
...
+    internal func enable() throws {
+        let client = try MongoClient.makeTestClient()
+        try self.enable(using: client)
+    }
+
+    internal func disable(using client: MongoClient? = nil, on address: ServerAddress? = nil) {
         do {
-            let client = try MongoClient.makeTestClient()
-            try client.db("admin").runCommand(["configureFailPoint": .string(self.name), "mode": "off"])
+            let clientToUse: MongoClient
```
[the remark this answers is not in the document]
patrickfreed: Nope, `try` statements "cannot appear to the right of a non-assignment operator".

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
patrickfreed: Yep, see ["Why do tests that run distinct sometimes fail with StaleDbVersion"](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#why-do-tests-that-run-distinct-sometimes-fail-with-staledbversion). Originally, we just ran it before every single test run for simplicity, but it turns out this loop was taking a long time because of all the client creation and destruction. Now that we can do `runCommand` per server and don't need to construct new clients each time, it's less important that we only do this for a subset of tests, but it's an easy enough optimization that we might as well do it.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/TestOperation.swift
```diff
@@ -117,8 +125,10 @@ struct TestOperationDescription: Decodable {
     ///
     /// The operation may be mutated over the course of this execution (e.g. by enabling a failpoint), so it
     /// must be passed as an `inout` parameter.
+    // swiftlint:disable:next function_parameter_count
```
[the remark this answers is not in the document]
patrickfreed: I actually think it's a decent one to have, but for the tests I don't really mind being more lax on code style.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -288,7 +287,7 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         let connStr3 = try ConnectionString("mongodb://localhost:27017/?heartbeatFrequencyMS=20000", options: opts)
         expect(connStr3.options?["heartbeatfrequencyms"]?.int32Value).to(equal(50000))
 
-        let tooSmall = 10
+        let tooSmall = 60
```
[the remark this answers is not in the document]
patrickfreed: Yep, that's exactly it.

Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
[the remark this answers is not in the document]
kmahar: huh oh yeah that error sounds familiar...

kmahar (approved):
great job, LGTM!

PR #527, opened 2020-09-17
SWIFT-854 Reorganize documentation sections

kmahar:
Similar PR to the BSON repo, with the addition of changes such that the doc generation script commits docs to the `gh-pages` branch instead.


etc/filter_sourcekitten_output.py
```diff
@@ -0,0 +1,43 @@
+import json
+
+# use Exports.swift as the canonical list of re-exported symbols we need to document.
```
kmahar, line 3: the alternative would be to maintain a list of the symbols we mark :nodoc: in multiple places. this approach seemed less likely to lead to issues in the future

.jazzy.yml
```diff
@@ -0,0 +1,174 @@
...
+  - name: Core Types
+    children:
+      - MongoClient
+      - MongoClientOptions
+      - MongoDatabase
+      - MongoDatabaseOptions
+      - MongoCollection
+      - MongoCollectionOptions
+
+  - name: BSON Types
```
kmahar, line 25: I went ahead and included the BSON symbols here since it seems like it will be possible for us to keep including them in the driver docs even when they're coming from the separate repo.

I figured I would wait to deal with updating the script to try to do that cross-repo combination until we've deleted the BSON code in the driver and switched to depending on the new library so I opened [SWIFT-1007](https://jira.mongodb.org/browse/SWIFT-1007).

etc/filter_sourcekitten_output.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    filtered_symbols = []
+    # key.substructure is an array with one element per symbol in the file.
+    for symbol in swift_file[file_name]['key.substructure']:
+        symbol_name = symbol['key.name']
+        if symbol_name in reexported_symbols:
+            filtered_symbols.append(symbol)
+
+    # make a copy of the file and update it with our filtered data.
+    filtered_file = swift_file
```
patrickfreed, line 35: This doesn't actually make a copy; it just makes a new reference to `swift_file`. I don't think we need to use `original_data` after this loop though, so it's okay if we mutate it's contents.

etc/generate-docs.sh
```diff
@@ -9,23 +9,53 @@ if ! command -v jazzy > /dev/null; then
   gem install jazzy || { echo "ERROR: Failed to locate or install jazzy; please install yourself with 'gem install jazzy' (you may need to use sudo)"; exit 1; }
 fi
 
+if ! command -v sourcekitten > /dev/null; then
+  gem install jazzy || { echo "ERROR: Failed to locate SourceKitten; please install yourself"; exit 1; }
```
patrickfreed, line 13: is `gem install jazzy` supposed to be something SourceKitten related here?

etc/generate-docs.sh
```diff
@@ -9,23 +9,53 @@ if ! command -v jazzy > /dev/null; then
...
+jazzy "${args[@]}"
+
+rm mongoswift-docs.json
+rm mongoswift-filtered.json
+rm mongoswiftsync-docs.json
+
+git checkout gh-pages
+
+rm -rf docs/*
+cp -r docs-temp/* docs/
```
patrickfreed, line 50: we could make this a `mv` instead of a `cp` + `rm -rf`, which lowers the stress of running this script by a little bit :smile:

etc/generate-docs.sh
```diff
@@ -9,23 +9,53 @@ if ! command -v jazzy > /dev/null; then
...
+rm -rf docs-temp
+
+echo '<html><head><meta http-equiv="refresh" content="0; url=MongoSwift/index.html" /></head></html>' > docs/index.html
+
+git add docs/
 
-modules=( MongoSwift MongoSwiftSync )
+git commit -m "${version} docs"
+git push
```
patrickfreed, line 58: what about leaving the push out of this script so we can test the docs out and see what they look like first? I think the release script might be more suited to actually pushing the docs.

etc/generate-docs.sh
kmahar: lol true. I've replaced this and the next line with
```
mv docs-temp/* docs/
rm -d docs-temp
```

(note that per comment below I moved this logic into the release script)

etc/generate-docs.sh
kmahar: that's fair. I've refactored the two scripts such that all this one docs is generate the docs and put them into `docs-temp`. the release script now handles switching to the `gh-pages` branch, moving them into `docs`, committing and pushing them.

patrickfreed (approved):
LGTM! Great job getting the different symbols together with SourceKitten, that's pretty cool functionality.

PR #528, opened 2020-09-24
SWIFT-1010 Test 5.3 on Ubuntu variants

kmahar:
While it seems like macOS testing may be blocked for a while, we can go ahead and add 5.3 testing on Linux now.

Evergreen patch in progress (added some 5.1 and 5.2 tests since I bumped the patch versions on those):
https://spruce.mongodb.com/version/5f6d04e9d1fe0779e9e01612/tasks

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -408,7 +408,7 @@ axes:
         display_name: "Ubuntu 16.04"
         run_on: ubuntu1604-test
 
-      - id: macos-1014
+      - id: macos-10.14
         display_name: "macOS 10.14"
         run_on: macos-1014
 
@@ -433,11 +433,15 @@ axes:
       - id: "5.1"
         display_name: "Swift 5.1"
         variables:
-           SWIFT_VERSION: "5.1.4"
+           SWIFT_VERSION: "5.1.5"
       - id: "5.2"
         display_name: "Swift 5.2"
         variables:
-           SWIFT_VERSION: "5.2"
+           SWIFT_VERSION: "5.2.5"
+      - id: "5.3"
+        display_name: "Swift 5.3"
+        variables:
+          SWIFT_VERSION: "5.3"
 
   - id: ssl-auth
     display_name: SSL and Auth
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/CursorCommon.swift
```diff
@@ -148,7 +148,7 @@ internal class Cursor<CursorKind: MongocCursorWrapper> {
             return nil
         }
 
-        var replyPtr = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
+        let replyPtr = UnsafeMutablePointer<BSONPointer?>.allocate(capacity: 1)
```
kmahar, line 151: with each release the compiler seems to get slightly better at finding these 🙂

.evergreen/config.yml
```diff
@@ -482,7 +488,49 @@ buildvariants:
...
+  display_name: "Atlas Connectivity ${swift-version} ${os-fully-featured}"
+  tasks:
+    - ".atlas-connect"
+
+
+# define separate matrices for swift 5.3 that excludes macOS 10.14. unfortunately
+# as of now we cannot remove entire buildvariants via rule from the matrices above
+# so we need to split this up due to inability to test 5.3 on macOS 10.14.
+# see EVG-13092
+- matrix_name: "tests-all"
```
patrickfreed, line 503: does this matrix having the same name as the other cause any issues? Or is it okay to share names?

patrickfreed (approved):
patch LGTM, one question about the matrix name

.evergreen/config.yml
kmahar: as far as I can tell it's fine - it doesn't seem like matrix name actually gets used for anything besides being prepended at the start of each buildvariant name, e.g. `tests-all__os-fully-featured~macos-1014_swift-version~5.2_ssl-auth~nossl-noauth`.
I think it would only be an issue if duplicate variants were generated by the two matrices since then their names would be identical.

PR #529, opened 2020-09-29
SWIFT-991 Logging specification POC

kmahar:
_Note: I've created a separate branch including the vendored libmongoc code so that my own changes atop it can be more easily reviewed._

This is very much in flux still and is therefore a draft PR - I've left a lot of inline comments with todos, extra commentary for the purpose of reviewers, etc. But I wanted to get it up sooner rather than later to get any feedback you may have on the general approach I've taken.

This is a POC of the general logging spec. In parallel to writing those requirements I've also been working on defining what command logging should look like so we have messages to test with.

What we add support for here is essentially just what is required by the specification: a way for users to enable/disable/configure logging on a per-component basis without having to recompile their application. 
(Note there is some conversation in flux about how we handle certain things like possible truncation/omission of large commands/replies in messages so some modifications to the env variable situation may be necessary.)

The spec also requires support for logging to stdout, stderr, as well as output files; however, it permits drivers to consider those supported if there are logging backends available in the ecosystem which do those for you, so I do not believe we need to do anything there. So long as a user has a backend bootstrapped in their application and runs their application with applicable environment variables set, logs will get sent to the backend.

That said... **this is not at all close to what I think the extent of our logging support should look like in 2.0**. To shed some light on what I know about where usage of `swift-log` is going:

`Logger`s are very lightweight objects (just structs storing mutable per-logger metadata). The general idea is to use a separate logger for each request in an application. Vapor does this, AsyncHTTPClient supports this, etc. As a logger gets passed around by libraries, each of them stores metadata on it, which means that e.g. a log message we produce deep down inside our code base (e.g. when checking out a connection) can be correlated with a Vapor request, SwiftNIO, etc... 

Of course, there are sometimes things that we will log that don't correspond to a particular request and are more client-level messages; for example, those relating to SDAM. In those cases, we might require users to provide us logger(s) when creating a `MongoClient` that we should use for such messages.

How does this get passed around? As I mentioned a while back, there has been ongoing work on coming up with a standardized way of context passing. These context objects would store loggers, along with tracing-related info.

Here's a [rather lengthy forums thread](https://forums.swift.org/t/the-context-passing-problem/39162) discussing how this should be handled API-wise by libraries. I have only read about half of it so far, but the consensus is definitely converging on explicit context passing to each API method, the same way that Go does it (though as the last parameter, not the first). 

In 2.0 I think we'll likely want to add context parameters, assuming the context-passing API has stabilized.

The stopgap solution some libraries are using (e.g. `AsyncHTTPClient`) is to accept a `Logger` directly for API methods. However, given the design of libmongoc's logging (single global callback), per-request logging would simply be impossible for us: by the time we receive a log message via our single global callback, we have no idea at all what request it came from or what logger to emit it to.

We _could_ think about at least adding a way for a user to register a custom global logger/loggers, however, I don't think that is an API we would want to persist into 2.0, and the spec does not require any particular programmatic API , so I lean toward not doing so.

I do not believe that what I propose adding here is incompatible with the future API though: even if we could support that ideal API now, we would still need to support this env var configuration in order to meet the "turn on logging for the driver with no code changes" requirement.

Anyway, sorry for the essay.... You may find it helpful to look over the logging spec draft (I won't post the link here since it's internal still, but you can find the link on DRIVERS-1204) as well as swift-log's [README](https://github.com/apple/swift-log) and [documentation](https://apple.github.io/swift-log/docs/current/Logging/index.html).

I also found this [doc](https://github.com/swift-server/async-http-client/blob/main/docs/logging-design.md) on the implementation of logging in AsyncHTTPClient to be useful.

### DIFF
--- a/Sources/MongoSwift/Logging.swift
+++ b/Sources/MongoSwift/Logging.swift
@@ -54,7 +54,7 @@ private func makeLogger(envVar: String, label: String) -> Logger? {
 private let COMMAND_LOGGER = makeLogger(envVar: "MONGODB_LOGGING_COMMAND", label: "MongoSwift.command")
 private let SDAM_LOGGER = makeLogger(envVar: "MONGODB_LOGGING_SDAM", label: "MongoSwift.sdam")
 private let SERVER_SELECTION_LOGGER = makeLogger(
-    envVar: "MONGODB_LOGGING_SERVER_SELECTION", 
+    envVar: "MONGODB_LOGGING_SERVER_SELECTION",
     label: "MongoSwift.serverSelection"
 )
 private let CONNECTION_LOGGER = makeLogger(envVar: "MONGODB_LOGGING_CONNECTION", label: "MongoSwift.connection")
@@ -62,10 +62,10 @@ private let CONNECTION_LOGGER = makeLogger(envVar: "MONGODB_LOGGING_CONNECTION",
 /// For testing purposes, we define fallback loggers which can be mutated as needed to temporarily turn logging on/off
 /// during tests. Arguably you could just handle this by making the original variables mutable, but it seemed a bit
 /// safer to keep those locked down with their initial values.
-internal var TEST_COMMAND_LOGGER: Logger? = nil
-internal var TEST_SDAM_LOGGER: Logger? = nil
-internal var TEST_SERVER_SELECTION_LOGGER: Logger? = nil
-internal var TEST_CONNECTION_LOGGER: Logger? = nil
+internal var TEST_COMMAND_LOGGER: Logger?
+internal var TEST_SDAM_LOGGER: Logger?
+internal var TEST_SERVER_SELECTION_LOGGER: Logger?
+internal var TEST_CONNECTION_LOGGER: Logger?
 
 extension mongoc_structured_log_component_t {
     /// Returns the Swift logger corresponding to the mongoc log component. This value is nil if the user did not
@@ -102,7 +102,8 @@ internal func handleMongocStructuredLogMessage(entry: OpaquePointer!, _: UnsafeM
     let msg = BSONDocument(copying: mongoc_structured_log_entry_get_message(entry))
     logger.log(
         level: level,
-        "\(msg["message"]!.stringValue!)", // assumes message is always present and a string. TODO: perhaps codify in spec?
+        // swiftlint:disable:next force_unwrapping
+        "\(msg["message"]!.stringValue!)", // assumes message is always present and a string. TODO: codify in spec?
         metadata: msg.toLoggerMetadata()
     )
 }
--- a/Sources/MongoSwift/Logging.swift
+++ b/Sources/MongoSwift/Logging.swift
@@ -1,0 +1,139 @@
+import CLibMongoC
+import Foundation
+import Logging
+
+extension mongoc_structured_log_level_t {
+    fileprivate var swiftLogLevel: Logger.Level {
+        switch self {
+        // .critical is the most severe level available in Swift so we must
+        // map the more severe levels emergency and alert down to it.
+        case MONGOC_STRUCTURED_LOG_LEVEL_EMERGENCY,
+             MONGOC_STRUCTURED_LOG_LEVEL_ALERT,
+             MONGOC_STRUCTURED_LOG_LEVEL_CRITICAL:
+            return .critical
+        case MONGOC_STRUCTURED_LOG_LEVEL_ERROR:
+            return .error
+        case MONGOC_STRUCTURED_LOG_LEVEL_WARNING:
+            return .warning
+        case MONGOC_STRUCTURED_LOG_LEVEL_NOTICE:
+            return .notice
+        case MONGOC_STRUCTURED_LOG_LEVEL_INFO:
+            return .info
+        case MONGOC_STRUCTURED_LOG_LEVEL_DEBUG:
+            return .debug
+        case MONGOC_STRUCTURED_LOG_LEVEL_TRACE:
+            return .trace
+        default:
+            fatalError("Unrecognized libmongoc log level \(self)")
+        }
+    }
+}
+
+private enum LogComponent {
+    case command, sdam, serverSelection, connection
+}
+
+extension mongoc_structured_log_component_t {
+    fileprivate var swiftLogComponent: LogComponent {
+        switch self {
+        case MONGOC_STRUCTURED_LOG_COMPONENT_COMMAND:
+            return .command
+        case MONGOC_STRUCTURED_LOG_COMPONENT_SDAM:
+            return .sdam
+        case MONGOC_STRUCTURED_LOG_COMPONENT_SERVER_SELECTION:
+            return .serverSelection
+        case MONGOC_STRUCTURED_LOG_COMPONENT_CONNECTION:
+            return .connection
+        default:
+            // maybe should just do nothing instead, since this presumably
+            // means its a component we don't support yet?
+            fatalError("Unrecognized libmongoc log component \(self)")
+        }
+    }
+}
+
+internal class CommandLogger {
+    internal static let global = CommandLogger()
+
+    internal let logger: Logger?
+
+    private init() {
+        // how to handle invalid value here? currently just ignore
+        // todo: handle if user specifies alert or emergency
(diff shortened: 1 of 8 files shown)


Tests/MongoSwiftTests/LoggingTests.swift
```diff
@@ -0,0 +1,158 @@
+import Foundation
```
kmahar, line 1: There are eventually going to be spec tests for this, and our testing needs and approach might change quite a bit. 

This was just enough basic code to consume log messages in the tests and confirm things are working as I would expect.

Sources/MongoSwift/Logging.swift
```diff
@@ -0,0 +1,132 @@
...
+            return CONNECTION_LOGGER ?? TEST_CONNECTION_LOGGER
+        default:
+            // ignore an unrecognized component. (TODO: codify in spec what should happen here?)
+            return nil
+        }
+    }
+}
+
+/// Callback for handling a mongoc structured log message. This is registered one time via `initializeMongoc()`.
+internal func handleMongocStructuredLogMessage(entry: OpaquePointer!, _: UnsafeMutableRawPointer?) {
```
patrickfreed, line 91: what is the second argument that we ignore here? could we name it something to describe it (still with underscore prefix though)?

Sources/MongoSwift/MongoSwift.swift
```diff
@@ -8,6 +8,8 @@ private final class MongocInitializer {
 
     private init() {
         mongoc_init()
+        mongoc_log_set_handler(handleMongocLogMessage, nil)
```
patrickfreed, line 11: is the unstructured logging the legacy / existing mongoc log behavior? If so, I imagine we'll never expect to make use of this logging and only implement the structured logs, right?

patrickfreed (commented):
Overall, it looks like a good approach. I just have a few clarifying questions.

>So long as a user has a backend bootstrapped in their application and runs their application with applicable environment variables set, logs will get sent to the backend.

And this is just achieved via the global `LoggingSystem.boostrap` invocation, right? Is it possible for users to direct all the MongoDB logs to a particular backend and their logs from other portions of their application to a different one? Or is there just one global backend?

>I do not believe that what I propose adding here is not incompatible with the future API though: even if we could support that ideal API now, we would still need to support this env var configuration in order to meet the "turn on logging for the driver with no code changes" requirement.

I'm having a bit of a hard time parsing this sentence. Is it saying that in the context-based API, we will still have to support the environment variables, and that may cause issues for us (i.e. which logger to use for a particular API invocation)?

kmahar:
> And this is just achieved via the global `LoggingSystem.boostrap` invocation, right? 

Yes, that's correct. If no handler is bootstrapped, there is a default `StreamLogHandler` that is used, which just logs to stdout.

To provide a little more context on how it all fits together, since I found it hard to figure some of this out without actually reading through the logger source code and I think it's all finally come together for me today and I want to write it down for myself anyway:

1. The application calls `LoggingSystem.bootstrap` and passes in a `factory`, which is a closure that returns an instance of a type conforming to the `LogHandler` protocol. A `LogHandler` type should be a struct.
2. Any time a logger is initialized (either directly in the application or in one of its dependencies), a new `Logger` (also a struct) is created, storing a reference to an instance of whatever `LogHandler` type was registered via boostrap. Both of these types being structs means there is a 1:1 logger:handler relationship, and updating a logger/handler does not affect any other loggers/handlers.
3. The `Logger.Level` for a logger is just the log level for its corresponding `LogHandler`. The getter/setter for `Logger.logLevel` just delegate to those on the logger's handler.
4. When `logger.log(level, ...)` is called, the `Logger` compares the provided `level` to the level for itself/its handler. If the message is less severe than the logger's level, this is a no-op. Otherwise, `LogHandler.log` is called with the message.
5. `LogHandler` types are permitted to have global overrides for log levels. For example, its `getter` for log level could check an environment variable or a configuration file to look up a level (In fact, Vapor's default logging backend supports this via environment variable) and the setter could just do nothing. This means that in practice (due to #3) it's possible to control the log level behavior of all `Logger`s via their handlers, since they delegate to their handlers for level.

> Is it possible for users to direct all the MongoDB logs to a particular backend and their logs from other portions of their application to a different one? Or is there just one global backend?

You only bootstrap once. That said, there is a `MultiplexLogHandler` built into swift-log which is just a wrapper for one or more log handlers that all get all the messages pushed to them. 
They also just recently released support in swift-log for a log message to include a `source` which is automatically populated with the module name that a log message came from.
The combination of these supports what you're asking about: you can have a multiplex handler and then use separate handlers that filter messages based on the source, so that (for example) your driver logs go to one file and your Vapor ones go to another.

> I'm having a bit of a hard time parsing this sentence.

Whoops! I added an extra "not" in there! It should read

> I do not believe that what I propose adding here is ~not~ incompatible with the future API though: even if we could support that ideal API now, we would still need to support this env var configuration in order to meet the "turn on logging for the driver with no code changes" requirement.

(Updated the original to avoid confusing future readers...) 

> Is it saying that in the context-based API, we will still have to support the environment variables, and that may cause issues for us (i.e. which logger to use for a particular API invocation)?

I do think you raise a good point here about how we might handle such a situation. 

I think generally,  the "turn on/configure logging with no code changes requirement" is really meant to serve people who have not set up logging programmatically in the driver already. In libmongoc the environment variables are completely ignored if a custom handler is registered, and I believe the Go driver is taking a similar approach.

To that end, I would say that if a user has provided a logger via API that is appropriate to use for a message, that one should always win out.

We will have a slightly more complicated situation due to Swift's many-logger vs. single logger situation, but I think the same principle applies. When emitting an SDAM message e.g., we would first check if the user gave us a logger for the client; if not, check if they set the environment variable / the global SDAM logger exists. If neither exist we do nothing (in line with the spec behavior that the default is to do nothing). 

One thing to consider is whether we would allow the environment variables to influence the log levels on the user provided loggers. For example, say a user sets `MONGODB_COMMAND_LOGGING=debug` and also passes us a logger to `insertOne`, however that logger has `info` level set on it (and thus ignores any messages at debug level). We could up the log level for the logger to `debug` before emitting the message.

That's also something that can be handled by logging backends though - as I noted in #5 above, a backend can globally override log levels, so that might be better taken care of at the handler level.

That's probably something we would want to ask to the swift-log maintainers about, but I think either direction (either completing ignoring the variables and assuming the log level will be dynamically configured via backend, or still using them to dynamically determine per-component levels) seems reasonable. Regardless users have to opt-in to passing in loggers via the API so I think it would be ok if the behavior of the environment variables is different when they decide to do that.

kmahar:
lol did not intend to link to PRs #3 and #5 above, throwback to the really old days of sketching out the initial DB and collection APIs!

Sources/MongoSwift/MongoSwift.swift
kmahar: yes, this is the unstructured/legacy logging, and I don't think we will ever use it.

I believe Andreas is still thinking about how the old and new APIs will interact with each other. 

It's possible that if structured logging is turned on the legacy logging would be automatically turned off, in which case we could get rid of this line.

Sources/MongoSwift/Logging.swift
kmahar: So the second parameter to `mongoc_structured_log_set_handler` is an `UnsafeMutableRawPointer`, which is basically a pointer to anything. the idea is that you could use it to register some contextual object, which is then passed into the callback here whenever it's executed.

This is similar to the parameter accepted by `mongoc_client_pool_set_apm_callbacks` [here](https://github.com/mongodb/mongo-swift-driver/blob/c3125ffee1919fb758cd7c8d3f04f77cb804eeda/Sources/MongoSwift/ConnectionPool.swift#L237), where we pass in the `MongoClient` so that when executing the callback we know which client it was associated with.

In this case since the callback is global rather than per-client, all context can just be tracked globally in Swift and I didn't find a need to use it.

swiftformat seems to want to use `_` for the parameter name since it's unused, so I've added `context` as a label and kept `_` as the parameter name.

kmahar:
Pushed some minor changes:
* support use of a new MONGODB_LOGGING_ALL variable that applies to all components
* test out truncation of commands/replies (this may end up being handled by libmongoc but doing it manually for now)
* update tests based on changes to fields included in command log messages

kmahar:
going to close for now given deferral of logging project but will keep the branch around.

PR #530, opened 2020-09-30
Fix inttype non-modular header issue when building from an .xcodeproj

kmahar:
The old patch just seemed to be too out of date to apply cleanly to about half of the files; I think a lot of them had added a number of include statements at the top so the line numbers were off.

I just manually updated the necessary files here and then used that to generate a new patch.


patrickfreed (approved):
diff lgtm, I assume you tried this out and it works with xcode now?

kmahar:
Yes, it builds successfully. I did run into a new issue which I think may either be an Xcode 12 or Swift 5.3 bug, where we seem to only pick up on about half of the spec test files. For example Xcode says the BSON corpus test ones are missing, but it finds the command monitoring ones and runs those tests just fine.... I can't figure out what is different about the cases where it works and where it doesn't. Unfortunately the transactions tests are among those it is not finding, which is what led me down the road of trying to build in Xcode in the first place. 🤦‍♀️ 

Anyway, that seemed like a separate issue so I have opened [SWIFT-1011](https://jira.mongodb.org/browse/SWIFT-1011) to track fixing that.

PR #531, opened 2020-10-07
SWIFT-1015 Only create monitoring events if they are being consumed

kmahar:
This is a longstanding inefficiency in the driver, but only came up as a problem when using the new BSON library.

Even if the user has no handlers registered, we create a corresponding Swift event, which in the case of e.g. a command started event for a large command is quite expensive (especially now that - at least for the moment - by default we validate incoming documents.)

This PR amends that by introducing a new concept of a monitoring "component" for each event type. Depending on the component of the event to be published, we inspect the client to see if any corresponding handlers exist.

I am feeling like this file is getting a little wild, using a lot of different types and protocols, but at the same time I couldn't figure out a clean way to accomplish this without the addition of the new type. I technically could have checked if the type conformed to `CommandEventProtocol` or not and acted accordingly but that felt kind of hacky. If you have any ideas let me know.

Note that as part of this I also resolved SWIFT-524 (see inline comment about test changes).

This problem might have somewhat "gone away" if we stopped validating the command documents (which we should probably also do), but this was a very quick win that I think is worth implementing regardless.

Comparison of runtimes, in seconds:
| Benchmark                               | Old   | New Unoptimized | With this fix |
|-----------------------------------------|-------|-----------------|---------------|
| Small doc bulk insert 10k copies        | .096  | 4258.673        | 0.424         |
| Large doc bulk insert 10 copies         | 0.332 | 59.637          | 2.79          |
| small doc insertOne 10k times           | 3.258 | 8.276           | 4.481         |
| findOne by _id 10k times                | 4.07  | 229.698         | 14.597        |
| find() and empty cursor, 10k small docs | 0.468 | 10436.246       | 1.025         |
| find() and empty cursor, 10 large docs  | 0.032 | 105.365         | 3.568         |

Note that the first and second to last ones literally got 10,000 times faster...!! My spawn host was working hard last night 🙂 

I need to do some further investigation into what the remaining bottlenecks for these benchmarks are, which should be much easier now that their runtimes are significantly lower. I suspect validation continues to be a main culprit, which we still encounter via e.g. parsing reply documents and reading documents from cursors. I will probably end up writing a couple of new small benchmarks that specifically test BSON validation in order to speed up iterating on that code.

### DIFF
--- a/Sources/MongoSwift/APM.swift
+++ b/Sources/MongoSwift/APM.swift
@@ -24,9 +24,15 @@ private protocol MongoSwiftEvent {
     associatedtype MongocEventType: MongocEvent
     associatedtype PublishableEventType: Publishable
 
+    static var monitoringComponent: MonitoringComponent { get }
+
     init(mongocEvent: MongocEventType)
 
     func toPublishable() -> PublishableEventType
+}
+
+private enum MonitoringComponent {
+    case command, sdam
 }
 
 /// A protocol for libmongoc event wrappers to implement.
@@ -117,6 +123,8 @@ public struct CommandStartedEvent: MongoSwiftEvent, CommandEventProtocol {
         }
     }
 
+    fileprivate static var monitoringComponent: MonitoringComponent { .command }
+
     /// The command.
     public let command: BSONDocument
 
@@ -165,6 +173,8 @@ public struct CommandSucceededEvent: MongoSwiftEvent, CommandEventProtocol {
             mongoc_apm_command_succeeded_get_context(self.ptr)
         }
     }
+
+    fileprivate static var monitoringComponent: MonitoringComponent { .command }
 
     /// The execution time of the event, in microseconds.
     public let duration: Int
@@ -215,6 +225,8 @@ public struct CommandFailedEvent: MongoSwiftEvent, CommandEventProtocol {
             mongoc_apm_command_failed_get_context(self.ptr)
         }
     }
+
+    fileprivate static var monitoringComponent: MonitoringComponent { .command }
 
     /// The execution time of the event, in microseconds.
     public let duration: Int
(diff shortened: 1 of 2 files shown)


Sources/MongoSwift/APM.swift
```diff
@@ -683,19 +718,19 @@ private func publishEvent<T: MongoSwiftEvent>(type: T.Type, eventPtr: OpaquePoin
     }
     let client = Unmanaged<MongoClient>.fromOpaque(context).takeUnretainedValue()
 
-    let event = type.init(mongocEvent: mongocEvent)
-
-    // TODO: SWIFT-524: remove workaround for CDRIVER-3256
-    if let tdChanged = event as? TopologyDescriptionChangedEvent,
```
kmahar, line 689: this CDRIVER ticket was resolved as part of 1.17.

Tests/MongoSwiftSyncTests/SDAMTests.swift
```diff
@@ -46,23 +46,27 @@ final class SDAMTests: MongoSwiftTestCase {
             return
         }
 
-        expect(receivedEvents.count).to(equal(4))
+        expect(receivedEvents.count).to(equal(5))
```
kmahar, line 49: so, the CDRIVER ticket mentioned above only actually resolved the duplicate server changed event issue. The topology changed one still existed. however, according to [DRIVERS-944](https://jira.mongodb.org/browse/DRIVERS-944), publishing the event is consistent with the SDAM monitoring tests.

I'm kind of 50/50 on whether to make the change, because as Kevin points out in the comment, there is no visible change to the user from the topology description. But to not publish it would be inconsistent with the spec and what other drivers seem to do if they pass the test.

patrickfreed: Yeah I kind of agree that we probably shouldn't emit events in this case, but also think for simplicity and conformity we should just do what the tests require. If that ticket ever gets resolved in the other direction we could just stop emitting the event--I don't think it would constitute a breaking behavioral change.

patrickfreed (approved):

PR #533, opened 2020-10-28
SWIFT-400 Improve Evergreen test result output

kmahar:
Putting this up as I'm calling it finished, but no pressure to review it during Skunkworks.

This adds a script written in Swift which reads input (which should be `swift test` output) from stdin, parses it into data structures, converts those structures to XML, and then writes the resulting XML to stdout.

This script is now used when running our tests on Evergreen to upload the results. 

With a small modification (updating list of suite names to skip) this should be reusable for the BSON repo.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -122,6 +122,7 @@ functions:
         script: |
           ${PREPARE_SHELL}
           sh ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh
+
   "upload-mo-artifacts":
     - command: shell.exec
       params:
@@ -152,7 +153,12 @@ functions:
           SSL=${SSL} \
           AUTH=${AUTH} \
           SWIFT_VERSION=${SWIFT_VERSION} \
-            sh ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
+            ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
+    
+  "upload test results":
+    - command: attach.xunit_results
+      params:
+        file: ./src/testResults.xml
 
   "run atlas tests":
     - command: shell.exec
@@ -199,6 +205,7 @@ pre:
 post:
   - func: "stop mongo-orchestration"
   - func: "upload-mo-artifacts"
+  - func: "upload test results"
   - func: "cleanup"
 
 tasks:
--- a/etc/convert-test-results.swift
+++ b/etc/convert-test-results.swift
@@ -1,5 +1,7 @@
 import Foundation
 
+// swiftlint:disable explicit_acl
+
 /// Represents a test suite.
 struct TestSuite {
     /// Name of the test suite.
@@ -20,9 +22,10 @@ struct TestSuite {
     /// Converts this test suite to XML.
     func toXML() -> String {
         var output =
-        """
-        <testsuite tests="\(self.count)" failures="\(self.failureCount)" errors="0" time="\(self.time)" name="\(self.name)">\n
-        """
+            """
+            <testsuite tests="\(self.count)" failures="\(self.failureCount)" \
+            errors="0" time="\(self.time)" name="\(self.name)">\n
+            """
 
         for test in self.tests {
             output += test.toXML()
@@ -64,7 +67,7 @@ struct TestCase {
                 .replacingOccurrences(of: ">", with: "﹥")
                 .replacingOccurrences(of: "&", with: "﹠")
 
-            output += 
+            output +=
                 """
                 <failure message="\(escapedFailure)"></failure>\n
                 """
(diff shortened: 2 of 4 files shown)


.evergreen/config.yml
```diff
@@ -152,7 +153,12 @@ functions:
           SSL=${SSL} \
           AUTH=${AUTH} \
           SWIFT_VERSION=${SWIFT_VERSION} \
-            sh ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
+            ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
```
kmahar, line 156: `pipefail` isn't supported by dash, which was being used instead of bash here on Ubuntu

etc/convert-test-results.swift
```diff
@@ -0,0 +1,332 @@
...
+    let tests: [TestCase]
+
+    /// Count of tests in the suite.
+    let count: Int
+
+    /// Count of failed tests in the suite.
+    let failureCount: Int
+
+    /// Converts this test suite to XML.
+    func toXML() -> String {
```
kmahar, line 21: I considered actually using an XML encoder as there are a few Codable-style ones floating around, but that would require making this a library rather than a standalone script. since the XML is so simple it didn't seem worth it.

etc/convert-test-results.swift
```diff
@@ -0,0 +1,332 @@
...
+#endif
+    static let testSuiteStartedRegex = try! NSRegularExpression(pattern:#"Test Suite '(.+)' started"#)
+    static let testSuiteStatusRegex = try! NSRegularExpression(pattern:#"Test Suite '(.+)' (passed|failed)"#)
+    static let testSuiteDetailsRegex = try! NSRegularExpression(pattern:#"Executed (\d+) tests?, with (\d+) failures? \((\d+) unexpected\) in (.+) \("#)
+
+    /// Processes a new line of test output and updates self accordingly.
+    mutating func processLine(_ line: String) throws {
+        let fullRange = NSRange(line.startIndex..<line.endIndex, in: line)
+
+        if let match = Self.testCaseStartedRegex.firstMatch(in: line, range: fullRange) {
```
kmahar, line 159: I tried to store a map of regex: handler function here so this could be done with a for loop rather than separate statements, but the compiler did not like my attempting to store mutating instance methods in a dictionary. I was running into the error described [here](https://stackoverflow.com/questions/37488316/partial-application-of-mutating-method-is-not-allowed), which I had not seen before but is a rather interesting case.

etc/convert-test-results.swift
```diff
@@ -0,0 +1,332 @@
...
+            """
+
+        if let failure = self.failure {
+            // hack to replace disallowed XML characters with very similar unicode ones.
+            // evergreen doesn't render the escaped XML characters correctly so this preserves
+            // readability while keeping the XML valid.
+            let escapedFailure = failure
+                .replacingOccurrences(of: "\"", with: "＂")
+                .replacingOccurrences(of: "'", with: "＇")
+                .replacingOccurrences(of: "<", with: "﹤")
```
kmahar, line 63: TIL there are "small" versions of many unicode symbols, lol. the "right" way to do this in XML is to use e.g. `&quot;` instead of `"` but that doesn't end up being very readable in the HTML logs.

etc/convert-test-results.swift
```diff
@@ -0,0 +1,332 @@
...
+    /// containing test pass/fail counts for the suite.
+    case awaitingSuiteDetails(name: String, completeTests: [TestCase], completeSuites: [TestSuite])
+    /// We are in the middle of a test with the given name in the given suite. `output` contains any output the test
+    /// case has produced.
+    case inTest(suite: String, name: String, output: [String], completeTests: [TestCase], completeSuites: [TestSuite])
+    /// None of the above. This means we are between test suites.
+    case none(completeSuites: [TestSuite])
+
+    // account for variation in formatting of test output on platforms. this assumes you are running this script on the
+    // same platform where you ran the tests.
```
patrickfreed, line 143: this is super frustrating and I think one of those warts exposed by the ObjC implementation on macOS. I remember looking into this a while back for some reason.

etc/convert-test-results.swift
patrickfreed, line 159: the spacing of this `if / else` chain is pretty spread out--was that done on purpose?

patrickfreed (approved):
Looks great! I have one question on some formatting but otherwise LGTM (don't need to review again). I really like the state-machine approach to parsing the tests.

etc/convert-test-results.swift
kmahar: nah, I think I originally had some comments or something that I deleted, removed. this actually made me realize I forgot to lint the file so I did that and made a few small fixes

PR #534, opened 2020-11-04
SWIFT-913 Make required changes to TestRequirement type for unified runner

kmahar:
For readability purposes, trying to break this PR up into separate chunks when I can.

This PR introduces changes to our `TestRequirement` and `TestTopologyConfiguration` types to support usage with the new unified test format.

### DIFF
--- a/Sources/TestsCommon/CommonTestUtils.swift
+++ b/Sources/TestsCommon/CommonTestUtils.swift
@@ -136,47 +136,50 @@ open class MongoSwiftTestCase: XCTestCase {
 }
 
 /// Enumerates the different topology configurations that are used throughout the tests
-public enum TestTopologyConfiguration: String, Decodable {
+public enum TestTopologyConfiguration: RawRepresentable, Decodable {
     case sharded
-    case replicaSet = "replicaset"
-    case shardedReplicaSet = "sharded-replicaset"
+    case replicaSet
     case single
 
-    /// Determines the topologyType of a client based on the reply returned by running an isMaster command and the
-    /// contents of the config.shards collection.
-    public init(isMasterReply: BSONDocument, shards: [BSONDocument]) throws {
+    public init?(rawValue: String) {
+        switch rawValue {
+        // as of MongoDB 3.6 (the lowest version this driver supports), shards must be replica sets, so for our
+        // testing purposes there is no distinction between these topology types.
+        case "sharded", "sharded-replicaset":
+            self = .sharded
+        case "replicaset":
+            self = .replicaSet
+        case "single":
+            self = .single
+        default:
+            return nil
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -136,12 +136,38 @@ open class MongoSwiftTestCase: XCTestCase {
...
     case sharded
-    case replicaSet = "replicaset"
+    case replicaSet
     case single
 
-    /// Determines the topologyType of a client based on the reply returned by running an isMaster command
+    public init?(rawValue: String) {
+        switch rawValue {
+        // as of MongoDB 3.6 (the lowest version this driver supports), shards must be replica sets, so for our
+        // testing purposes there is no distinction between these topology types.
```
patrickfreed, line 147: I thought this too when I was reviewing the Rust runner, but it seems like the docs might not be accurate here (https://github.com/mongodb/mongo-rust-driver/pull/249#discussion_r491147383).

kmahar: ah interesting. I originally changed the code to actually do a check by inspecting the contents of `config.shards`, but I deleted the code since it seemed superfluous. I guess I'll just revert my last commit then 🙂

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -139,21 +139,45 @@ open class MongoSwiftTestCase: XCTestCase {
...
+                )
+            }
+            for shard in shards {
+                guard let host = shard["host"]?.stringValue else {
+                    throw TestError(message: "config.shards document \(shard) unexpectedly missing host string")
+                }
+                // If the shard is backed by a single server, this field will contain a single host (e.g. localhost:27017).
+                // If the shard is backed by a replica set, this field will contain the name of the replica followed by a
+                // forward slash and a comma-delimited list of hosts.
+                let replSetHostRegex = try NSRegularExpression(pattern: #"^.*\/.*:\d+$"#)
```
kmahar, line 166: inspecting documents in config.shards is the way [the spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#determining-if-a-sharded-cluster-uses-replica-sets) instructs you to check for this.

borrowed this regex approach from [PHP](https://github.com/mongodb/mongo-php-library/blob/6c66a1a291f4309e910e1c29998550aa971de220/tests/FunctionalTestCase.php#L333).

I'm not sure if it's possible to have a sharded cluster that is backed by a mix of standalones and replica sets, but for testing purposes it seems to me that we shouldn't count it as being fully backed by replica sets, so we'll mark it `.sharded` here if that is the case.

patrickfreed (approved):

PR #536, opened 2020-11-10
SWIFT-913 Add EntityDescription and ExpectedEventsForClient types for unified test runner

kmahar:
Adding some more data structures and support for (at least some types) for verifying only expected options are present. I'll add some commentary inline.

### DIFF
--- a/Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
+++ b/Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
@@ -1,8 +1,36 @@
 @testable import struct MongoSwift.ReadPreference
 import MongoSwiftSync
+import TestsCommon
 
-extension MongoDatabaseOptions: Decodable {
+/// Allows a type to specify a set of known keys and check whether any unknown top-level keys are found in a decoder.
+internal protocol StrictDecodable: Decodable {
+    associatedtype CodingKeysType: CodingKey, CaseIterable, RawRepresentable where CodingKeysType.RawValue == String
+
+    /// Checks whether the top-level container in the decoder contains any keys that do not match up with cases in the
+    /// associated CodingKeysType.
+    static func checkKeys(using decoder: Decoder) throws
+}
+
+extension StrictDecodable {
+    /// Default implementation of checkKeys.
+    static func checkKeys(using decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let rawKeys = try container.decode(BSONDocument.self).keys
+        let supportedKeys = Set(Self.CodingKeysType.allCases.map { $0.rawValue })
+        for key in rawKeys {
+            guard supportedKeys.contains(key) else {
+                throw TestError(message: "Unsupported key \(key) found while decoding instance of \(Self.self)")
+            }
+        }
+    }
+}
+
+extension MongoDatabaseOptions: StrictDecodable {
+    internal typealias CodingKeysType = CodingKeys
+
     public init(from decoder: Decoder) throws {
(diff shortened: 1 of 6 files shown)


Tests/MongoSwiftSyncTests/SpecTestRunner/CodableExtensions.swift
```diff
@@ -1,35 +1,72 @@
 @testable import struct MongoSwift.ReadPreference
 import MongoSwiftSync
+import TestsCommon
+
+/// Allows a type to specify a set of known keys and check whether any unknown top-level keys are found in a decoder.
+internal protocol StrictDecodable: Decodable {
```
kmahar, line 6: Couple notes:

1:
The spec doesn't actually strictly require that we check that we support the options for all of these types. The requirements are only that we error on unsupported operations and arguments. However, it seemed wise to me to make such checks for these options anyway as a safeguard. Since their `Decodable` implementations are handwritten and as they stand do not actually try decoding every single possible property, only the ones we have seen in tests before, it seems easy for them to get out of sync. Also, the unified test runner spec states that the parameter names in the URI options spec will be used for everything in `MongoClientOptions`. Given that we don't support every single URI option as a `MongoClientOption` period this seemed like an easy place for us to miss a parameter, so for that type in particular these checks are valuable.

2:
This approach works decently well for types where we define a custom `Decodable` init already in a test target (basically everything in this file). I think the only downside is that you could forget to add a call to `checkKeys` in the `decode` method.

That said, I don't think this exact approach will make as much sense for `Operation` types. 

For one thing, a decent number of them don't currently need a manual `decode` method at all, but using this approach would require we write all of those out.

For another, those that _do_ need manual decode methods need one because they have associated options, which (unfortunately for us) are flattened into the same struct as the required parameters. This means that we can't rely on a nested approach to checking all the arguments are supported, like I do below for `ClientSessionOptions` and its `defaultTransactionOptions`: the same document will contain keys for both structs, and therefore we need to compare against a merged list of the operation parameters and the option struct's properties, and that won't correspond exactly to the CodingKeys for either of the structs.

I think we can use a similar approach though anyway, where each `Operation` defines a property containing supported arguments (both parameters and stuff that goes into options) and then in the place where we actually decode a concrete operation type (a giant switch statement over operation names) we can use similar logic to that below in `checkKeys` to compare the operation type's supported keys against the found ones.

Depending on what we end up doing there I may revise the logic here to match, but this seemed to be the cleanest way to accomplish strictly decoding for these particular types.

patrickfreed: I think the approach included here and the proposed one for operations both sound good to me :+1:

Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -0,0 +1,117 @@
...
+        let id: String
+
+        /// Database entity from which this collection will be created.
+        let database: String
+
+        /// Collection name.
+        let collectionName: String
+
+        /// Options to use for this collection.
+        let options: MongoCollectionOptions?
```
patrickfreed, line 68: according to the [schema](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/schema-1.0.json#L105), these options types are prefixed with the name of the entity theyre configuring. e.g. this one would be "collectionOptions". Ditto for session, database, bucket, etc.

kmahar: 🤦‍♀️  great catch, fixed.

patrickfreed (approved):
lgtm!

PR #537, opened 2020-11-11
SWIFT-913 Introduce a UnifiedOperation type

kmahar:
Adds another data structure, some comments inline

### DIFF
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
@@ -39,7 +39,7 @@ struct UnifiedOperation: Decodable {
     let result: UnifiedOperationResult?
 
     private enum CodingKeys: String, CodingKey {
-        case name, object, arguments, expectError, expectResult, saveResultAsEntity
+        case name, object, arguments
     }
 
     public init(from decoder: Decoder) throws {
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
@@ -1,0 +1,129 @@
+import Foundation
+import MongoSwiftSync
+
+struct UnifiedOperation: Decodable {
+    /// Represents an object on which to perform an operation.
+    enum Object: RawRepresentable, Decodable {
+        /// Used for special test operations.
+        case testRunner
+        /// An entity name e.g. "client0".
+        case entity(String)
+
+        public var rawValue: String {
+            switch self {
+            case .testRunner:
+                return "testRunner"
+            case let .entity(s):
+                return s
+            }
+        }
+
+        public init(rawValue: String) {
+            switch rawValue {
+            case "testRunner":
+                self = .testRunner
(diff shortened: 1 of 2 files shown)


Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+import MongoSwiftSync
+
+struct UnifiedOperation: Decodable {
```
kmahar, line 4: I looked into trying to reuse the old type but there are some notable differences, e.g. here "object" can be _any_ string since it can an arbitrary entity name, there's now the option to save the result as an entity, etc.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/// Represents the expected result of an operation.
+enum UnifiedOperationResult: Decodable {
+    /// One or more assertions for an error expected to be raised by the operation.
+    case error(ExpectedError)
+    /// - result: A value corresponding to the expected result of the operation.
+    /// - saveAsEntity: If specified, the actual result returned by the operation (if any) will be saved with this
+    ///       name in the Entity Map.
+    // TODO: SWIFT-913: consider using custom type to represent results
```
kmahar, line 71: I need to look into a bit more, but we might want do something similar to `TestOperationResult` from the old runner here, though it would need to be a bit more complex because there can now be the various custom operators e.g. $$unsetOrMatches within the result. if I understand correctly though it looks like the Rust runner just uses its BSON enum here though, so maybe that won't be needed.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+import MongoSwiftSync
+
+struct UnifiedOperation: Decodable {
+    /// Represents an object on which to perform an operation.
+    enum Object: RawRepresentable, Decodable {
```
kmahar, line 6: I made some somewhat arbitrary feeling choices about which types to nest here or not, e.g. `Object` is nested but `UnifiedOperationResult` is not, `ExpectedError` could arguably be nested in `UnifiedOperationResult` but isn't, etc.
Ultimately it came down to how complex a type "felt" to me; `Object` is quite simple but the other two are less so. 

I don't know if we've ever really standardized on when to do this in test code though, let me know if you have opinions.

patrickfreed: I've just been doing it by feel also. Or if a type is only used in the context of another type.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
patrickfreed: I think it'll depend how we want to implement matching. I'm okay with just leaving this TODO here and determine the structure here as needed in later PRs.

patrickfreed (approved):

PR #538, opened 2020-11-12
SWIFT-913 Introduce first batch of operations and strategy for verifying arguments

kmahar:
[review #537 first]

This is a first batch of support for parsing out the operations.

In a couple of cases (`InsertOne`, `Find`) I have opted to reuse operation structs from the existing test runner, in others a new struct was needed because there is no struct for the existing operation. I believe there are also some I haven't added here yet where the key names don't match up (e.g. a property named `databaseName` rather than `database`) so those might need a new unified version.

I'm not sure if this is actually going to save much time, because they will need different execute methods, and down the road we're going to have to go through and delete/consolidate code. So we could also opt to just duplicate everything... I don't really have strong feelings either way, let me know what you think.

More comments inline


Sources/MongoSwift/ChangeStreamOptions.swift
```diff
@@ -56,7 +56,7 @@ public struct ChangeStreamOptions: Codable {
      * - SeeAlso: https://docs.mongodb.com/master/changeStreams/#change-stream-start-after
      */
     // TODO: SWIFT-519 - Make this public when support is added for 4.2 change stream features.
-    internal var startAfter: ResumeToken?
+    // internal var startAfter: ResumeToken?
```
kmahar, line 59: I commented this out because I use `Mirror` elsewhere to determine the valid arguments for `createChangeStream`, and having this property here could inadvertently make it seem like we support `startAfter`.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
```diff
@@ -234,10 +249,34 @@ struct DeleteOne: TestOperation {
...
+    let options: InsertOneOptions
+
+    private enum CodingKeys: String, CodingKey, CaseIterable {
+        case document, session
+    }
+
+    static var knownArguments: Set<String> {
+        Set(
+            CodingKeys.allCases.map { $0.rawValue } +
+                Mirror(reflecting: InsertOneOptions()).children.map { $0.label! }
```
kmahar, line 269: in practice I haven't found any cases where the label is nil, so I'm not sure why this is optional 🤔

Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
```diff
@@ -0,0 +1,40 @@
...
+
+struct UnifiedFailPoint: UnifiedOperationProtocol {
+    /// The configureFailpoint command to be executed.
+    let failPoint: BSONDocument
+
+    /// The client entity to use for setting the failpoint.
+    let client: String
+
+    static var knownArguments: Set<String> {
+        ["failPoint", "client"]
```
kmahar, line 11: I just listed out the values here because, in order to use `Mirror`, you actually need an instance of the type. I guess I could have created an instance with dummy values for both of the properties, but given there are a small number it seemed OK to just write them out.

same applies for rest of this file

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedChangeStreamOperations.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import MongoSwiftSync
+
+struct CreateChangeStream: UnifiedOperationProtocol {
```
kmahar, line 4: this is basically just a `watch` operation, but the unified runner is fairly adamant about using this name instead because some drivers don't actually send the aggregate command when you call `watch` (we do).

so we could have reused the old `watch`, however as written it has no properties stored at all, so I figured I'd just go ahead and add the better-named version now.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
```diff
@@ -80,18 +81,32 @@ struct Distinct: TestOperation {
     }
 }
 
-struct Find: TestOperation {
-    let session: String?
+struct Find: TestOperation, UnifiedOperationProtocol {
```
patrickfreed, line 84: I personally prefer to duplicate all the operations (even if it involves just copy/pasting) since the two runners are completely separate otherwise, and we'll eventually have to remove the old one anyways as you point out.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
```diff
@@ -0,0 +1,40 @@
...
+        ["session"]
+    }
+}
+
+struct AssertDifferentLsidOnLastTwoCommands: UnifiedOperationProtocol {
+    /// Identifier for the client to perform the assertion on.
+    let client: String
+
+    static var knownArguments: Set<String> {
+        Set(["client"])
```
patrickfreed, line 29: is there any reason these ones use the `Set` initializer and the other ones use the literal?

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
kmahar: fine with me, I've reverted the changes to these and introduced duplicates prefixed with `Unified` for now.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
kmahar: good catch, no, I think I just realized partway through that I could use a literal and forgot to fix all of them

patrickfreed (approved):

PR #539, opened 2020-11-12
Update Sourcery, SwiftLint, SwiftFormat versions

kmahar:
Updates our various dependencies and makes some minor changes to appease formatter/linter.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -36,3 +36,9 @@
 
 # the ordering doesn't match linter's
 --disable specifiers
+
+# conflicts with our Swiftlint explicit ACL rule
+--disable extensionAccessControl
+
+# conflicts with Swiftlint opening_brace rule
+--disable wrapMultilineStatementBraces
--- a/.swiftformat
+++ b/.swiftformat
@@ -40,5 +40,3 @@
 # conflicts with our Swiftlint explicit ACL rule
 --disable extensionAccessControl
 
-# conflicts with Swiftlint opening_brace rule
---disable wrapMultilineStatementBraces
--- a/Sources/MongoSwift/BSON/BSONDocument.swift
+++ b/Sources/MongoSwift/BSON/BSONDocument.swift
@@ -255,7 +255,7 @@ extension BSONDocument {
             throw BSONError.InternalError(message: "No keys to exclude, use 'bson_copy' instead")
         }
 
-        let cStrings: [ContiguousArray<CChar>] = keys.map { $0.utf8CString }
+        let cStrings: [ContiguousArray<CChar>] = keys.map(\.utf8CString)
 
         var cPtrs: [UnsafePointer<CChar>] = try cStrings.map { cString in
             let bufferPtr: UnsafeBufferPointer<CChar> = cString.withUnsafeBufferPointer { $0 }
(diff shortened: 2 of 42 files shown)


.swiftformat
```diff
@@ -36,3 +36,9 @@
 
 # the ordering doesn't match linter's
 --disable specifiers
+
+# conflicts with our Swiftlint explicit ACL rule
+--disable extensionAccessControl
+
+# conflicts with Swiftlint opening_brace rule
+--disable wrapMultilineStatementBraces
```
kmahar, line 44: SwiftFormat wanted us to start using a newline for opening braces following multiline conditionals, e.g.
```swift
if let a = x,
    let b = y {
    // ...
}
```

would become 
```swift
if let a = x,
    let b = y 
{
    ...
}
```

SwiftLint did not like this because it wants the opening brace to be on the same line as the end of the statement. I have like a 3/10 preference for the way we've always done it so stuck with that, but if you feel strongly the other way happy to switch

Sources/MongoSwift/BSON/BSONDocument.swift
```diff
@@ -255,7 +255,7 @@ extension BSONDocument {
             throw BSONError.InternalError(message: "No keys to exclude, use 'bson_copy' instead")
         }
 
-        let cStrings: [ContiguousArray<CChar>] = keys.map { $0.utf8CString }
+        let cStrings: [ContiguousArray<CChar>] = keys.map(\.utf8CString)
```
kmahar, line 258: new SwiftFormat rule that prefers using keypaths where possible -- I like it so I left it on for now, LMK what you think

Tests/MongoSwiftSyncTests/CommandMonitoringTests.swift
```diff
@@ -102,12 +102,12 @@ private struct CMTest: Decodable {
     // `context` dictionary so we can access it in future events for the same test
     var context = [String: Any]()
 
-    var expectations: [ExpectationType] { try! expectationDocs.map { try makeExpectation($0) } }
+    var expectations: [ExpectationType] { try! self.expectationDocs.map { try makeExpectation($0) } }
```
kmahar, line 105: explicit self checking seems to have improved

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -1,3 +1,4 @@
+// swiftlint:disable duplicate_imports
```
kmahar, line 1: I think this is because it doesn't distinguish between these two lines, even though they have different meanings:
```
import MongoSwiftSync
@testable import struct MongoSwiftSync.MongoClientOptions
```

kmahar:
Evg failures should be fixed by #540

.swiftformat
patrickfreed: I prefer the latter (more than 3/10 :smile: ) because otherwise it's kinda hard to parse the difference between the `if` portion of the statement and the actual body (e.g. here https://github.com/mongodb/mongo-swift-driver/blob/636cd541225754a8956e42419461a433bd23b113/Sources/MongoSwift/MongoError.swift#L354).

Ideally the various `let` keywords would be aligned too, but I'm not sure if that's an option.

kmahar: that's a good point. let's turn it on then 🙂

patrickfreed (approved):

PR #541, opened 2020-11-13
SWIFT-913 Support for remaining test operations

kmahar:
This PR is a bit larger than the past ones, but is for the most part very repetitive so hopefully is not too difficult to review. 

I'm sure we are missing some operations right now, and/or optional arguments for operations, that might be added at some point as I did not exhaustively comb through every spec, but the design of the runner here where we error on both unsupported operation names and unsupported argument names should make it very easy to identify those cases in the future.

Comments inline

### DIFF
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/HasListableProperties.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/HasListableProperties.swift
@@ -1,0 +1,22 @@
+import MongoSwiftSync
+
+protocol HasListableProperties {
+    var propertyNames: [String] { get }
+}
+
+extension HasListableProperties {
+    var propertyNames: [String] {
+        Mirror(reflecting: self).children.map { $0.label! }
+    }
+}
+
+extension BulkWriteOptions: HasListableProperties {}
+extension FindOneAndReplaceOptions: HasListableProperties {}
+extension FindOneAndUpdateOptions: HasListableProperties {}
+extension DeleteOptions: HasListableProperties {}
+extension ReplaceOptions: HasListableProperties {}
+extension InsertOneOptions: HasListableProperties {}
+extension DeleteModelOptions: HasListableProperties {}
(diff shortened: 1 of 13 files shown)


Tests/MongoSwiftSyncTests/UnifiedTestRunner/HasListableProperties.swift
```diff
@@ -0,0 +1,21 @@
+import MongoSwiftSync
+
+protocol HasListableProperties {
```
kmahar, line 3: there doesn't seem to be a way to extend an arbitrary object. In retrospect I guess I could have added a top level function which took in `Any` but this feels a little more idiomatic and readable

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCollectionOperations.swift
```diff
@@ -1,7 +1,81 @@
...
+    let session: String?
+
+    static var knownArguments: Set<String> {
+        ["name", "keys", "session"]
+    }
+}
+
+struct UnifiedBulkWrite: UnifiedOperationProtocol {
+    /// Writes to perform.
+    let requests: [WriteModel<BSONDocument>]
```
patrickfreed, line 52: Instead of relying on `WriteModel`'s `Decodable` implementation for both the old runner and the unified one, what do you think about defining a new intermediate type that has a a unified-runner-specific `Decodable` implementation and using that here instead? Then we don't need to mix implementations.

patrickfreed (commented):
overall looks good, just have one suggestion re: the combined `Decodable` implementation for `WriteModel`

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCollectionOperations.swift
kmahar: it does seem a bit cleaner. done

patrickfreed (approved):
lgtm!

PR #542, opened 2020-11-14
SWIFT-592 Move format/lint/sourcery check to Evergreen

kmahar:
My branch name was a bit premature as code coverage is still being generated through Travis, but this moves everything else over and stops running tests on Travis altogether. 

It would be good to move code coverage too, though I think that will be a bit more complicated so I've opened another ticket to handle that separately at a later time: SWIFT-1019. The coverage isn't very accurate at this point because it only tests against a standalone on 3.6. We should be generating a number of reports across different Evergreen runs and merging them together. I don't think we really rely much on code coverage feedback in PRs anyway so it seems less essential to get timely feedback on.

Sample successful patch: https://spruce.mongodb.com/version/5faf1bc09ccd4e21e67cd68d/tasks

Sample patch where I intentionally broke formatting/linting/sourcery: https://spruce.mongodb.com/version/5faf1e443e8e8655c850e2ea/tasks

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -188,19 +188,56 @@ functions:
             perl -p -i -e "s|ABSOLUTE_PATH_REPLACEMENT_TOKEN|${DRIVERS_TOOLS}|g" $filename
           done
 
-  "install dependencies":
+  "install swift":
     - command: shell.exec
       params:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-
           SWIFT_VERSION=${SWIFT_VERSION} \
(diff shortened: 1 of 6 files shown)


patrickfreed (approved):

patrickfreed:
Oh I just noticed that it doesn't seem like the lint gets run as part of the PR--can we do that here or would we configure that after this gets merged?

kmahar:
I have to update the Evergreen project to run certain tasks based on a regex which I will do after merging. Thanks for the reminder, probably would have forgotten!

PR #543, opened 2020-11-19
SWIFT-913 Add support for test runner and test case setup

kmahar:
This adds support for all of the setup code that is necessary to run before executing operations, e.g. populating entity map, running necessary commands to ensure the topology is in a good state, etc.

### DIFF
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
@@ -137,7 +137,7 @@ struct UnifiedTestClient {
             opts.minHeartbeatFrequencyMS = 50
             opts.heartbeatFrequencyMS = 50
         }
-        self.client = try MongoClient.makeTestClient(connStr, options: clientDescription.uriOptions)
+        self.client = try MongoClient.makeTestClient(connStr, options: opts)
         self.commandMonitor = UnifiedTestCommandMonitor(
             observeEvents: clientDescription.observeEvents,
             ignoreEvents: clientDescription.ignoreCommandMonitoringEvents
@@ -247,8 +247,7 @@ extension Array where Element == EntityDescription {
                 map[dbDesc.id] = .database(clientEntity.client.db(
                     dbDesc.databaseName,
                     options: dbDesc.databaseOptions
-                )
-                )
+                ))
             case let .collection(collDesc):
                 guard let db = try map[collDesc.database]?.asDatabase() else {
                     throw TestError(message: "No database with id \(collDesc.database) found in entity map")
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
@@ -1,5 +1,7 @@
 import Foundation
+@testable import struct MongoSwift.MongoClientOptions
 import MongoSwiftSync
+import TestsCommon
 
 /// Represents a description of an entity (client, db, etc.).
 enum EntityDescription: Decodable {
@@ -30,7 +32,7 @@ enum EntityDescription: Decodable {
 
         /// Types of events that can be observed for this client. Unspecified event types MUST be ignored by this
         /// client's event listeners.
-        let observeEvents: [String]?
+        let observeEvents: [CommandEvent.EventType]?
 
         /// Command names for which the test runner MUST ignore any observed command monitoring events. The command(s)
         /// will be ignored in addition to configureFailPoint and any commands containing sensitive information (per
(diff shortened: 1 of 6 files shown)


Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -1,5 +1,7 @@
 import Foundation
+@testable import struct MongoSwift.MongoClientOptions
```
kmahar, line 2: I was going to rename this file to just Entity.swift since it contains various Entity-related code now but GitHub considered it too big a diff and was treating it as a new file. so I will do that in a future PR

Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -115,3 +117,155 @@ enum EntityDescription: Decodable {
...
+        var opts = clientDescription.uriOptions ?? MongoClientOptions()
+        // If the test might execute a configureFailPoint command, for each target client the test runner MAY
+        // specify a reduced value for heartbeatFrequencyMS (and minHeartbeatFrequencyMS if possible) to speed
+        // up SDAM recovery time and server selection after a failure; however, test runners MUST NOT do so for
+        // any client that specifies heartbeatFrequencyMS in its uriOptions.
+        if opts.heartbeatFrequencyMS == nil {
+            opts.minHeartbeatFrequencyMS = 50
+            opts.heartbeatFrequencyMS = 50
+        }
+        self.client = try MongoClient.makeTestClient(connStr, options: clientDescription.uriOptions)
```
patrickfreed, line 140: we should use `opts` here instead of going back to the uriOptions one in case we set heartbeat frequency stuff.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -115,3 +117,155 @@ enum EntityDescription: Decodable {
...
+                map[clientDesc.id] = try .client(UnifiedTestClient(clientDesc))
+            case let .database(dbDesc):
+                guard let clientEntity = try map[dbDesc.client]?.asTestClient() else {
+                    throw TestError(message: "No client with id \(dbDesc.client) found in entity map")
+                }
+                map[dbDesc.id] = .database(clientEntity.client.db(
+                    dbDesc.databaseName,
+                    options: dbDesc.databaseOptions
+                )
+                )
```
patrickfreed, line 251: this double paren is a little weird, I guess SwiftFormat doesn't mind it though. Could we move them to one line like the similar case below?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -0,0 +1,140 @@
...
+        // tests. Using the internal MongoClient, execute the killAllSessions command on either the primary or, if
+        // connected to a sharded cluster, all mongos servers.
+        switch self.topologyType {
+        case .single:
+            return
+        case .replicaSet:
+            let admin = self.internalClient.db("admin")
+            for address in MongoSwiftTestCase.getHosts() {
+                let isMaster = try admin.runCommand(["isMaster": 1], on: address)["ismaster"]!.boolValue!
+                if isMaster {
```
patrickfreed, line 28: instead of iterating over the hosts like this, could we just use a primary read preference?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -0,0 +1,140 @@
...
+                fileLevelLog("Running test \"\(test.description)\" from file \"\(file.description)\"")
+
+                // If initialData is specified, for each collectionData therein the test runner MUST drop the
+                // collection and insert the specified documents (if any) using a "majority" write concern. If no
+                // documents are specified, the test runner MUST create the collection with a "majority" write concern.
+                // The test runner MUST use the internal MongoClient for these operations.
+                if let initialData = file.initialData {
+                    for collData in initialData {
+                        let db = self.internalClient.db(collData.databaseName)
+                        let coll = db.collection(collData.collectionName)
```
patrickfreed, line 104: it might be easier to just specify the write concern on the collection itself so that if we ever happen to add new statements to this block we won't have to remember to specify it again.

patrickfreed (commented):
overall looks really good, just have some minor comments

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
kmahar: done 👍

Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
patrickfreed: Yeah, I made the same mistake back during the libmongoc upgrade, and it did cause me a lot of confusion lol.

patrickfreed (approved):
LGTM!

PR #544, opened 2020-11-21
SWIFT-913 Start adding support for operations; support non-error result and event matching

kmahar:
This adds support for everything that is needed to execute the [first test in the first file](https://github.com/mongodb/mongo-swift-driver/blob/8bd6f978d56a69a54dfb4c9ce74d8b8fdd78d828/Tests/Specs/unified-test-format/tests/valid-pass/poc-change-streams.json#L92) and perform the required assertions on the result:

- Extends `UnifiedOperationProtocol` to add an `execute` method and adds a general `executeAndCheckResult` method to `UnifiedOperation`
- Implements the `execute` method for operations used in this test (`createChangeStream`, `insertOne`, `iterateUntilDocumentOrError`)
- Matching (non-error) results
- Saving results as entities
- Matching observed events

The code is temporarily modified, for now, to only execute that test.... I could have kept going but I wanted to get your feedback first on what I have so far before this gets to be too massive of a diff 🙂

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Write.swift
+++ b/Sources/MongoSwift/MongoCollection+Write.swift
@@ -359,7 +359,7 @@ public struct DeleteOptions: Codable, BulkWriteOptionsConvertible {
 // Write command results structs
 
 /// The result of an `insertOne` command on a `MongoCollection`.
-public struct InsertOneResult: Decodable {
+public struct InsertOneResult: Codable {
     /// The identifier that was inserted. If the document doesn't have an identifier, this value
     /// will be generated and added to the document before insertion.
     public let insertedID: BSON
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
@@ -202,6 +202,8 @@ enum Entity {
     case database(MongoDatabase)
     case collection(MongoCollection<BSONDocument>)
     case session(ClientSession)
+    case changeStream(ChangeStream<BSONDocument>)
+    case bson(BSON)
 
     func asTestClient() throws -> UnifiedTestClient {
         guard case let .client(client) = self else {
@@ -230,12 +232,28 @@ enum Entity {
         }
         return session
     }
+
+    func asChangeStream() throws -> ChangeStream<BSONDocument> {
+        guard case let .changeStream(cs) = self else {
+            throw TestError(message: "Failed to return entity \(self) as a change stream")
+        }
+        return cs
+    }
+
+    func asBSON() throws -> BSON {
+        guard case let .bson(bson) = self else {
+            throw TestError(message: "Failed to return entity \(self) as a BSON")
+        }
+        return bson
+    }
 }
+
+typealias EntityMap = [String: Entity]
 
 extension Array where Element == EntityDescription {
     /// Converts an array of entity descriptions from a test file into an entity map.
-    func toEntityMap() throws -> [String: Entity] {
-        var map = [String: Entity]()
+    func toEntityMap() throws -> EntityMap {
+        var map = EntityMap()
         for desc in self {
             switch desc {
             case let .client(clientDesc):
@@ -266,5 +284,41 @@ extension Array where Element == EntityDescription {
             }
         }
         return map
+    }
+}
+
+extension EntityMap {
+    func getEntity(from object: UnifiedOperation.Object) throws -> Entity {
+        try self.getEntity(id: object.asEntityId())
+    }
+
+    func getEntity(id: String) throws -> Entity {
+        guard let entity = self[id] else {
+            throw TestError(message: "No entity with id \(id) found in entity map")
+        }
+        return entity
+    }
+
+    func getEntityAsCollection(from object: UnifiedOperation.Object) throws -> MongoCollection<BSONDocument> {
+        try self.getEntity(from: object).asCollection()
+    }
+
+    func getEntityAsChangeStream(from object: UnifiedOperation.Object) throws -> ChangeStream<BSONDocument> {
+        try self.getEntity(from: object).asChangeStream()
+    }
+
+    func getEntityAsBSON(from object: UnifiedOperation.Object) throws -> BSON {
+        try self.getEntity(from: object).asBSON()
+    }
+
+    func getEntityAsSession(from object: UnifiedOperation.Object) throws -> ClientSession {
+        try self.getEntity(from: object).asSession()
+    }
+
+    func resolveSession(id: String?) throws -> ClientSession? {
+        guard let id = id else {
+            return nil
+        }
+        return try self.getEntity(id: id).asSession()
     }
 }
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
@@ -299,22 +299,6 @@ extension EntityMap {
         return entity
     }
 
-    func getEntityAsCollection(from object: UnifiedOperation.Object) throws -> MongoCollection<BSONDocument> {
-        try self.getEntity(from: object).asCollection()
-    }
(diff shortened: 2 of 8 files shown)


Sources/MongoSwift/MongoCollection+Write.swift
```diff
@@ -359,7 +359,7 @@ public struct DeleteOptions: Codable, BulkWriteOptionsConvertible {
 // Write command results structs
 
 /// The result of an `insertOne` command on a `MongoCollection`.
-public struct InsertOneResult: Decodable {
+public struct InsertOneResult: Codable {
```
kmahar, line 362: This was necessary because of the way I'm representing results as `BSON`s. It seemed relatively low risk to make this `Encodable` as well. 

We may run into some future situations where it is preferable to manually convert a result object to a `BSONDocument` and/or add the `Encodable` conformance for tests only.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -202,6 +202,8 @@ enum Entity {
     case database(MongoDatabase)
     case collection(MongoCollection<BSONDocument>)
     case session(ClientSession)
+    case changeStream(ChangeStream<BSONDocument>)
```
kmahar, line 205: (somehow I missed before that both of these could be entity types too)

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -0,0 +1,263 @@
+import MongoSwiftSync
+import TestsCommon
+
```
kmahar, line 3: as I mentioned earlier this file is close to a straight port of @isabelatkinson's matching code in Rust, so hopefully looks familiar 🙂

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
```diff
@@ -6,6 +6,39 @@ import TestsCommon
 protocol UnifiedOperationProtocol: Decodable {
     /// Set of supported arguments for the operation.
     static var knownArguments: Set<String> { get }
+
+    /// Indicates whether this operation returns an array of root documents. This is used when determining how to match
+    /// against the operation's expected result.
+    static var returnsRootDocuments: Bool { get }
```
patrickfreed, line 12: I wonder if it makes more sense to make `UnifiedOperatoinResult` a little more specific and store this on the result itself rather than on the operation.

e.g. 
```swift
enum UnifiedOperationResult {
    case changeStream(...)
    case value(BSON) // this could be broken down further if we want also
    case rootDocumentArray([BSONDocument])
}
```
I'm finding the various booleans passed around in the `matches` function to be a little confusing, and this data-oriented approach makes a little more sense to me. The cost here is introducing more specificity at the operation execution level, which I don't think should be a big deal. What do you think?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -0,0 +1,311 @@
...
+/// Enum representing types that can be matched against expected values.
+enum MatchableResult {
+    /// A root document. i.e. a documents where extra keys are ignored when matching against an expected document.
+    case rootDocument(BSONDocument)
+    /// An array of root documents.
+    case rootDocumentArray([BSONDocument])
+    /// A (non-root) document.
+    case document(BSONDocument)
```
patrickfreed, line 11: what about naming this `subdocument` or `subDocument` to indicate it is non-root?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -0,0 +1,311 @@
...
+        return try matchesInner(
+            expected: expected,
+            actual: actual,
+            entities: entities
+        )
+    }
+}
+
+/// Determines whether `actual` matches `expected`, recursing if needed for nested documents and arrays.
+private func matchesInner(
```
patrickfreed, line 62: Now that we have a `MatchableResult` type, what about adding this as a method to it like we added a method to `UnifiedOperationResult`?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
patrickfreed: I think the new `MatchableResult` looks great! I left a few other comments with minor suggestions about it, but otherwise I think its a nice improvent.

In regards to the new enum, I think it would be mostly useful as a way to distinguish `specialOperator` from the rest of the cases, but I don't necessarily see much value in differentiating from among the various BSON cases. I think the `BSON` enum is sufficient for that.

e.g. something like

```swift
enum ExpectedResult {
    case specialOperator(SpecialOperator)
    case bson(BSON)
}

enum SpecialOperator {
    case exists(key: String)
    case type(type: String)
    case matchesEntity(entityId: String)
    ...
}
```
I think these type(s) could be useful for making the `matchesSpecial` code a little cleaner via `Codable`. I don't think this would be _necssary_, but I do think it would be a minor improvement, especially if the special matches get more complicated / involve more data down the road. What do you think?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -0,0 +1,311 @@
...
+        let id = value.stringValue!
+        let session = try entities.getEntity(id: id).asSession()
+        return actualDoc == session.id
+    default:
+        fatalError("Unrecognized special operator \(op)")
+    }
+}
+
+/// Determines whether `actual` satisfies the $$type operator value `expectedType`.
+func matchesType(expectedType: BSON, actual: MatchableResult) -> Bool {
```
kmahar, line 189: I realized that the value for $$type can be either a single string or an array of strings so I've updated the type matching code to reflect that.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
kmahar: sorry, couldn't reply to this with my previous batch:

I tried out the `ExpectedResult` approach, however I found it rather clunky because I had to explicitly initialize `ExpectedResult`s in the places where we call `matchesInner` (now renamed just `matches`). It also introduced another level of switching in `matches` since we have to first switch on the `ExpectedResult` and then the nested `BSON`.

what I ended up doing is still adding a `SpecialOperator` type, which we just directly initialize and use when we see a special operator doc. I think it makes the special matching code a bit cleaner without making the rest of the code more complicated; let me know what you think.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -0,0 +1,328 @@
...
+            self = .subDocument(doc)
+        case let .array(arr):
+            self = .array(arr)
+        default:
+            self = .scalar(bson)
+        }
+    }
+
+    /// Determines whether `self` matches `expected`, recursing if needed for nested documents and arrays.
+    fileprivate func matches(_ expected: BSON, entities: EntityMap) throws -> Bool {
```
patrickfreed, line 57: We discussed some options for plumbing down a key path or returning a more detailed result here so that the error output is more useful. I think those things would be really nice to have, but in the interests of moving this project along I think they can wait until another PR. Is that the current plan? I honestly don't remember where we left that discussion.

patrickfreed (approved):
LGTM! (besides one question about some follow on work)

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
kmahar: honestly, I don't remember either 🙂  I think I maybe said I would investigate it but then didn't do so. I think tackling it in a follow up PR makes sense

PR #545, opened 2020-11-23
SWIFT-828 Add support for hidden indexes

ncarbon:
Adds support for hidden indexes. [JIRA](https://jira.mongodb.org/browse/SWIFT-828)

### DIFF
--- a/Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
+++ b/Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
@@ -96,6 +96,10 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
             options.bucketSize = 10
         }
 
+        if try _client!.serverVersion() >= ServerVersion(major: 4, minor: 4, patch: 0) {
+            options.hidden = false
+        }
+
         let model = IndexModel(keys: ["cat": 1, "_id": -1], options: options)
         expect(try self.coll.createIndex(model)).to(equal("testOptions"))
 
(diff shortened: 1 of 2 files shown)


patrickfreed:
There are some test failures on the evergreen run. I actually looked into them because I couldn't quite figure out what was going on. I think this _may_ be a libmongoc bug where `mongoc_collection_find_indexes` won't return a `hidden` field if it's set to false, but it will if it's true. I checked and this doesn't seem to be a bug in the Swift driver, and the field also seems to be present server side. We can look into it a bit more tomorrow and try to make a repro for the C driver team.

ncarbon:
@patrickfreed Ohh it looks like it’s probably not a bug after all. I was just testing this incorrectly. 😅 
According to the docs https://docs.mongodb.com/manual/core/index-hidden
> The index option hidden is only returned if the value is true.

patrickfreed:
I think you're right, good catch. I just tested it out and the shell does in fact only return when it's true too, so I don't think there's any bug after all.

Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
```diff
@@ -96,6 +96,10 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
             options.bucketSize = 10
         }
 
+        if try _client!.serverVersion() >= ServerVersion(major: 4, minor: 4, patch: 0) {
+            options.hidden = false
```
kmahar, line 100: given conversation I think we should use `true` to ensure we get the option back and can assert on it, otherwise this PR LGTM and I don't need to review again assuming that makes Evergreen all green. nice job getting right back into things 🥳

patrickfreed (approved):
LGTM!

PR #546, opened 2020-11-25
SWIFT-459 Add a renamed helper

ncarbon:
[JIRA](https://jira.mongodb.org/browse/SWIFT-459)

Adds the renamed helper method.

### DIFF
--- a/Sources/MongoSwift/MongoCollection.swift
+++ b/Sources/MongoSwift/MongoCollection.swift
@@ -12,6 +12,21 @@ public struct DropCollectionOptions: Codable {
     }
 }
 
+/// Options to use when renaming a collection.
+public struct RenamedCollectionOptions: Codable {
+    /// Specifies a collation.
+    public var collation: BSONDocument?
+
+    /// An optional `WriteConcern` to use for the command.
+    public var writeConcern: WriteConcern?
+
+    /// Initializer allowing any/all parameters to be omitted.
+    public init(writeConcern: WriteConcern? = nil, collation: BSONDocument? = nil) {
+        self.collation = collation
+        self.writeConcern = writeConcern
+    }
+}
+
 // sourcery: skipSyncExport
 /// A MongoDB collection.
 public struct MongoCollection<T: Codable> {
@@ -48,6 +63,9 @@ public struct MongoCollection<T: Codable> {
         self.namespace.collection! // swiftlint:disable:this force_unwrapping
     }
 
+    /// The options used to initialize this `MongoCollection`.
+    public let options: MongoCollectionOptions?
+
     /// The `ReadConcern` set on this collection, or `nil` if one is not set.
     public let readConcern: ReadConcern?
 
@@ -62,6 +80,7 @@ public struct MongoCollection<T: Codable> {
     internal init(name: String, database: MongoDatabase, options: MongoCollectionOptions?) {
         self.namespace = MongoNamespace(db: database.name, collection: name)
         self._client = database._client
+        self.options = options
 
         // for both read concern and write concern, we look for a read concern in the following order:
         // 1. options provided for this collection
@@ -100,6 +119,41 @@ public struct MongoCollection<T: Codable> {
      */
     public func drop(options: DropCollectionOptions? = nil, session: ClientSession? = nil) -> EventLoopFuture<Void> {
         let operation = DropCollectionOperation(collection: self, options: options)
+        return self._client.operationExecutor.execute(operation, client: self._client, session: session)
+    }
+
+    /**
+     * Renames this collection with the specified options.
+     *
+     * - Parameters:
+     *   - to: A `String`, the new name for the collection
+     *   - dropTarget: A `Bool`, indicating whether the target collection should be dropped prior to renaming it.
+     *                 The default value is false.
+     *   - options: Optional `RenamedCollectionOptions` to use for the collection
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Returns:
+     *    An `EventLoopFuture<MongoCollection>` evaluating to a copy of the target `MongoCollection` with the new name.
+     *
+     *    If the future evaluates to an error, it is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+    public func renamed(
+        _ to: String,
+        dropTarget: Bool = false,
+        options: RenamedCollectionOptions? = nil,
+        session: ClientSession? = nil
+    ) -> EventLoopFuture<MongoCollection> {
+        let operation = RenamedCollectionOperation(
+            collection: self,
+            to: to,
+            dropTarget: dropTarget,
+            options: options
+        )
         return self._client.operationExecutor.execute(operation, client: self._client, session: session)
     }
 
--- a/Sources/MongoSwift/Operations/RenameCollectionOperation.swift
+++ b/Sources/MongoSwift/Operations/RenameCollectionOperation.swift
@@ -1,0 +1,46 @@
+import CLibMongoC
+
+/// An operation corresponding to a "createIndexes" command.
+internal struct RenamedCollectionOperation<T: Codable>: Operation {
+    private var collection: MongoCollection<T>
+    private let to: String
+    private let dropTarget: Bool
+    private let options: RenamedCollectionOptions?
+
+    internal init(
+        collection: MongoCollection<T>,
+        to: String,
+        dropTarget: Bool,
+        options: RenamedCollectionOptions? = nil
+    ) {
+        self.collection = collection
+        self.to = to
+        self.dropTarget = dropTarget
+        self.options = options
+    }
+
+    internal func execute(using connection: Connection, session: ClientSession?) throws -> MongoCollection<T> {
+        let opts = try encodeOptions(options: options, session: session)
+        var error = bson_error_t()
+        return try self.collection.withMongocCollection(from: connection) { collPtr in
+            try withOptionalBSONPointer(to: opts) { optsPtr in
+                let success = mongoc_collection_rename_with_opts(
+                    collPtr,
+                    nil,
+                    self.to,
+                    self.dropTarget,
+                    optsPtr,
+                    &error
+                )
+
+                guard success else {
+                    throw extractMongoError(error: error)
+                }
+
+                let db = self.collection._client.db(self.collection.namespace.db)
+                let opts = self.collection.options
+                return MongoCollection(name: self.to, database: db, options: opts)
+            }
+        }
+    }
+}
--- a/Sources/MongoSwiftSync/MongoCollection.swift
+++ b/Sources/MongoSwiftSync/MongoCollection.swift
@@ -63,4 +63,40 @@ public struct MongoCollection<T: Codable> {
     public func drop(options: DropCollectionOptions? = nil, session: ClientSession? = nil) throws {
         try self.asyncColl.drop(options: options, session: session?.asyncSession).wait()
     }
+
+    /**
+     * Renames this collection with the specified options.
+     *
+     * - Parameters:
+     *   - to: A `String`, the new name for the collection
+     *   - dropTarget: A `Bool`, indicating whether the target collection should be dropped prior to renaming it.
+     *                 The default value is false.
+     *   - options: Optional `RenamedCollectionOptions` to use for the collection
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Returns:
+     *    An `EventLoopFuture<MongoCollection>` evaluating to a copy of the target `MongoCollection` with the new name.
+     *
+     *    If the future evaluates to an error, it is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+    public func renamed(
+        _ to: String,
+        dropTarget: Bool = false,
+        options: RenamedCollectionOptions? = nil,
+        session: ClientSession? = nil
+    ) throws -> MongoCollection {
+        let newAsyncColl = try self.asyncColl.renamed(
+            to,
+            dropTarget: dropTarget,
+            options: options,
+            session: session?.asyncSession
+        )
+        .wait()
+        return MongoCollection(client: self.client, asyncCollection: newAsyncColl)
+    }
 }
--- a/Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
+++ b/Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
@@ -272,6 +272,28 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(deleteManyResult).to(beNil())
     }
 
+    func testRenamed() throws {
+        let encoder = BSONEncoder()
+        let client = try MongoClient.makeTestClient()
+        let db = client.db("testRenxcgamedDB")
+        defer { try? db.drop() }
+        let coll = try db.createCollection("testRenamesdfdColl")
+        let to = "newName"
+        let expectedWriteConcern: WriteConcern = try WriteConcern(journal: true, w: .number(1))
+
+        let monitor = client.addCommandMonitor()
+        try monitor.captureEvents {
+            let opts = RenamedCollectionOptions(writeConcern: expectedWriteConcern)
+            expect(try coll.renamed(to, dropTarget: true, options: opts)).toNot(throwError())
+        }
+
+        let event = monitor.commandStartedEvents().first
+        expect(event).toNot(beNil())
+        expect(event?.command["renameCollection"]).toNot(beNil())
+        expect(event?.command["writeConcern"]?.documentValue).to(sortedEqual(try? encoder.encode(expectedWriteConcern)))
+        expect(event?.command["dropTarget"]).to(equal(true))
+    }
+
     func testReplaceOne() throws {
         let replaceOneResult = try coll.replaceOne(filter: ["_id": 1], replacement: ["apple": "banana"])
         expect(replaceOneResult?.matchedCount).to(equal(1))
--- a/Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
+++ b/Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
@@ -460,20 +460,14 @@ struct ReplaceOne: TestOperation {
 }
 
 struct RenameCollection: TestOperation {
-    let session: String?
     let to: String
 
     func execute(
         on collection: MongoCollection<BSONDocument>,
-        sessions: [String: ClientSession]
+        sessions _: [String: ClientSession]
     ) throws -> TestOperationResult? {
-        let databaseName = collection.namespace.db
-        let cmd: BSONDocument = [
-            "renameCollection": .string(databaseName + "." + collection.name),
-            "to": .string(databaseName + "." + self.to)
-        ]
-        let reply = try collection._client.db("admin").runCommand(cmd, session: sessions[self.session ?? ""])
-        return TestOperationResult(from: reply)
+        try collection.renamed(self.to)
+        return nil
     }
 }
 
--- a/Sources/MongoSwiftSync/Exports.swift
+++ b/Sources/MongoSwiftSync/Exports.swift
@@ -77,6 +77,7 @@
 @_exported import enum MongoSwift.OperationType
 @_exported import struct MongoSwift.ReadConcern
 @_exported import struct MongoSwift.ReadPreference
+@_exported import struct MongoSwift.RenamedCollectionOptions
 @_exported import struct MongoSwift.ReplaceOneModelOptions
 @_exported import struct MongoSwift.ReplaceOptions
 @_exported import struct MongoSwift.ResumeToken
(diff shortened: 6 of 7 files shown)


Sources/MongoSwift/MongoCollection.swift
```diff
@@ -103,6 +122,41 @@ public struct MongoCollection<T: Codable> {
...
+     *
+     *    If the future evaluates to an error, it is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+    public func renamed(
+        _ to: String,
```
patrickfreed, line 146: I think we want to use this label, since it makes the method invocation slightly more readable:
```swift
coll.renamed(to: "new_name")
// vs
coll.renamed("new_name")
```

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -103,6 +122,41 @@ public struct MongoCollection<T: Codable> {
...
+        dropTarget: Bool = false,
```
patrickfreed, line 147: I think this option can just go in the options struct since it isn't required / used super frequently

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -12,6 +12,21 @@ public struct DropCollectionOptions: Codable {
     }
 }
 
+/// Options to use when renaming a collection.
+public struct RenamedCollectionOptions: Codable {
+    /// Specifies a collation.
+    public var collation: BSONDocument?
```
patrickfreed, line 18: I don't believe the `renameCollection` command supports a collation. (https://docs.mongodb.com/manual/reference/collation/#operations-that-support-collation)

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -12,6 +12,21 @@ public struct DropCollectionOptions: Codable {
     }
 }
 
+/// Options to use when renaming a collection.
+public struct RenamedCollectionOptions: Codable {
```
patrickfreed, line 16: I think this makes more sense as `RenameCollectionOptions` or `RenameOptions` since they're configuring the underlying command, whereas the past tense is mostly for communicating the semantics of the method invocation. What do you think @kmahar ?

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,46 @@
+import CLibMongoC
+
+/// An operation corresponding to a "createIndexes" command.
+internal struct RenamedCollectionOperation<T: Codable>: Operation {
```
patrickfreed, line 4: ditto this operation can just be `RenameCollectionOperation`

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    internal func execute(using connection: Connection, session: ClientSession?) throws -> MongoCollection<T> {
+        let opts = try encodeOptions(options: options, session: session)
+        var error = bson_error_t()
+        return try self.collection.withMongocCollection(from: connection) { collPtr in
+            try withOptionalBSONPointer(to: opts) { optsPtr in
+                let success = mongoc_collection_rename_with_opts(
+                    collPtr,
+                    nil,
+                    self.to,
```
patrickfreed, line 30: since we're only allowing to rename a collection within a single database, we need to prepend "\<dbname\>." to `to` here to ensure we use full namespace notation (e.g. "db.coll"). See the `renameCollection` [docs](https://docs.mongodb.com/manual/reference/command/renameCollection/#dbcmd.renameCollection) for an example.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
```diff
@@ -460,20 +460,14 @@ struct ReplaceOne: TestOperation {
...
+        sessions _: [String: ClientSession]
     ) throws -> TestOperationResult? {
-        let databaseName = collection.namespace.db
-        let cmd: BSONDocument = [
-            "renameCollection": .string(databaseName + "." + collection.name),
-            "to": .string(databaseName + "." + self.to)
-        ]
-        let reply = try collection._client.db("admin").runCommand(cmd, session: sessions[self.session ?? ""])
-        return TestOperationResult(from: reply)
+        try collection.renamed(self.to)
```
patrickfreed, line 469: we still need to pass in the session to this method

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -272,6 +272,28 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(deleteManyResult).to(beNil())
     }
 
+    func testRenamed() throws {
```
patrickfreed, line 275: I think we should have three tests to ensure adequate coverage: one that uses `dropTarget: true`, one without it set, and one with an invalid collection name.

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -272,6 +272,28 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+        let encoder = BSONEncoder()
+        let client = try MongoClient.makeTestClient()
+        let db = client.db("testRenamedDB")
+        defer { try? db.drop() }
+        let coll = try db.createCollection("testRenamedColl")
+        let to = "newName"
+        let expectedWriteConcern: WriteConcern = try WriteConcern(journal: true, w: .number(1))
```
patrickfreed, line 282: this `: WriteConcern` type annotation isn't necessary. The compiler should be able to infer the type via the `WriteConcern` initializer.

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -272,6 +272,28 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(deleteManyResult).to(beNil())
     }
 
+    func testRenamed() throws {
+        let encoder = BSONEncoder()
+        let client = try MongoClient.makeTestClient()
+        let db = client.db("testRenamedDB")
+        defer { try? db.drop() }
+        let coll = try db.createCollection("testRenamedColl")
```
patrickfreed, line 280: to ensure the rename actually worked, we should insert some stuff into this collection first, then rename it, then do a count on the new collection and ensure the documents moved.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -103,6 +122,41 @@ public struct MongoCollection<T: Codable> {
         return self._client.operationExecutor.execute(operation, client: self._client, session: session)
     }
 
+    /**
+     * Renames this collection with the specified options.
```
patrickfreed, line 126: I think we need to include a bit more detail here to cover some of the confusion users might have with this method. Specifically, we need to emphasize that this method renames the collection server side and returns a handle to the renamed collection. The current handle (which the method is invoked on) will continue to refer to the old collection, which will now be empty.

Another important point to mention is that this method does not work on sharded collections.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -103,6 +122,41 @@ public struct MongoCollection<T: Codable> {
...
+     *
+     * - Parameters:
+     *   - to: A `String`, the new name for the collection
+     *   - dropTarget: A `Bool`, indicating whether the target collection should be dropped prior to renaming it.
```
patrickfreed, line 130: This option means to drop the destination collection before renaming the current one to it, ensuring the command will work even if the new collection name is already taken by an existing collection. The current docstring is a little unclear in how it expresses that, could it be updated to be more specific?

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -48,6 +63,9 @@ public struct MongoCollection<T: Codable> {
         self.namespace.collection! // swiftlint:disable:this force_unwrapping
     }
 
+    /// The options used to initialize this `MongoCollection`.
+    public let options: MongoCollectionOptions?
```
patrickfreed, line 67: can we make this `private`? Otherwise I worry the duplication with the other fields (e.g. `readConcern` and `writeConcern`) may confuse users.

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,46 @@
...
+                    &error
+                )
+
+                guard success else {
+                    throw extractMongoError(error: error)
+                }
+
+                let db = self.collection._client.db(self.collection.namespace.db)
+                let opts = self.collection.options
+                return MongoCollection(name: self.to, database: db, options: opts)
```
kmahar, line 42: our API is designed such that `MongoDatabase`s inherit various options (e.g. read concern) from their parent `MongoClient`, and so that `MongoCollection`s inherit those same options from their parent `MongoDatabase`. 

a user can override this by passing in a different setting to the `db` or `collection` method. right now I think we have a problem where database options might not be properly inherited. consider the following:
```swift
let db = client.db("db1", options: DatabaseOptions(readPreference: .secondary))
let coll = db.collection("coll1") // self.options will be nil, but `self.readPreference` will be `.secondary`, inherited rom db
let coll2 = coll.renamed(to: "coll2") // since we just copied what was in `self.options` and created a new `MongoDatabase` from scratch, we didn't copy the read preference
```

I think probably the easiest way to work around this is to introduce a new internal `copying` init to `MongoCollection` that takes in another collection and just copies over all of its properties but sets the namespace to the new namespace, e.g.

```swift
init(copying other: MongoCollection<T>, name: String) {
    self.namespace = MongoNamespace(db: other.namespace.db, collection: name)
    self.readPreference = other.readPreference
    // ...etc
}
```

and then you could use that here instead of creating the `db` and using it to call the old initializer. that way we know all the properties except the namespace will be exactly the same.

I think this would also eliminate the need to store the `options` property on `MongoCollection`

kmahar: alternatively you could update the existing initializer to update the passed-in options object to reflect inherited options, e.g. store `readPreference` on that so `coll1`'s options would reflect the actual read preference, but I'd lean toward option 1 since it cuts down on storing duplicate information

Sources/MongoSwiftSync/MongoCollection.swift
```diff
@@ -63,4 +63,41 @@ public struct MongoCollection<T: Codable> {
     public func drop(options: DropCollectionOptions? = nil, session: ClientSession? = nil) throws {
         try self.asyncColl.drop(options: options, session: session?.asyncSession).wait()
     }
+
+    /**
+     * Renames this collection with the specified options on the server. This method will return a handle to the
```
patrickfreed, line 68: The "with the specified options" sort of sounds like it will be used as a filter to decide which collection to rename. I think we can just drop that part from the docstring and it should still be clear enough. The listing of the parameters below should be enough to indicate the command is configurable.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -12,6 +12,33 @@ public struct DropCollectionOptions: Codable {
     }
 }
 
+/// Options to use when renaming a collection.
+public struct RenameCollectionOptions: Codable {
```
patrickfreed, line 16: I see that `DropCollectionOptions` were also included in this file, but I think that was a mistake. We generally include the options for a given operation in the operations file. So in this case we'd have these options in `RenameCollectionOperation.swift`.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -12,6 +12,33 @@ public struct DropCollectionOptions: Codable {
     }
 }
 
+/// Options to use when renaming a collection.
```
patrickfreed, line 15: Given the difference in names between the options and the method they're used for, can we add another sentence to this docstring mentioning the options will be used for `MongoCollection.renamed`? If you include it in backticks `\`\`` like that, the docs generator should automatically linkify that text when it gets rendered.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
patrickfreed: this still needs to be fixed

patrickfreed (commented):
Looking good! I just have some minor comments.

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -103,6 +113,42 @@ public struct MongoCollection<T: Codable> {
...
+     *
+     *    If the future evaluates to an error, it is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+    public func renamed(
+        to: String,
```
kmahar, line 140: small nit: Swift has both "argument labels" and "parameter names". in this case, since you only provide one value `to`, "to" is used as _both_ the label and the name. 

(previously, when you had `_ to: String` here, `_` was the label (i.e. it had no label) and `to` was the name.)

in cases where the label is a preposition (like `to`) it's idiomatic to provide a separate name for the parameter, e.g. `to newName: String`. Then users get the benefit of calling it with the label (`rename(to: "abc")`) but in the method body it has a clearer name.

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,72 @@
+import CLibMongoC
+
+/// Options to use when renaming a collection. These options will be used for `MongoCollection.renamed`.
+public struct RenameCollectionOptions: Codable {
+    /// Specifies whether an existing collection matching the new name should be dropped before the rename.
+    /// If this is not set to true and a collection with the new collection name exists, the server will throw an error.
+    public let dropTarget: Bool?
```
kmahar, line 7: this should be a `var`, not a `let`, so users can mutate this property on an existing instance of `RenameCollectionOptions`.

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,72 @@
...
+    /// Initializer allowing any/all parameters to be omitted.
+    public init(dropTarget: Bool = false, writeConcern: WriteConcern? = nil) {
+        self.dropTarget = dropTarget
+        self.writeConcern = writeConcern
+    }
+}
+
+/// An operation corresponding to a "createIndexes" command.
+internal struct RenameCollectionOperation<T: Codable>: Operation {
+    private var collection: MongoCollection<T>
```
kmahar, line 31: I think this can be a `let` since we don't need to mutate it.

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        self.collection = collection
+        self.to = to
+        self.options = options
+    }
+
+    internal func execute(using connection: Connection, session: ClientSession?) throws -> MongoCollection<T> {
+        let opts = try encodeOptions(options: options, session: session)
+        var error = bson_error_t()
+        var dropTarget = false
```
kmahar, line 48: you can one-line the logic here like: `let dropTarget = options?.dropTarget ?? false`

Sources/MongoSwift/Operations/RenameCollectionOperation.swift
```diff
@@ -0,0 +1,72 @@
...
+
+/// An operation corresponding to a "createIndexes" command.
+internal struct RenameCollectionOperation<T: Codable>: Operation {
+    private var collection: MongoCollection<T>
+    private let to: String
+    private let options: RenameCollectionOptions?
+
+    internal init(
+        collection: MongoCollection<T>,
+        to: String,
```
kmahar, line 37: you might consider making similar naming changes here to those I suggested for the API method, e.g. `newName` or something instead of just `to`

Sources/MongoSwiftSync/MongoCollection.swift
```diff
@@ -63,4 +63,41 @@ public struct MongoCollection<T: Codable> {
...
+     *
+     *    Throws:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     */
+    public func renamed(
+        to: String,
```
kmahar, line 91: (same comment applies as for async version)

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -272,6 +272,55 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+            try coll.insertMany([["doc1": "test1"], ["doc2": "test2"], ["doc1": "test3"]])
+
+            let monitor = client.addCommandMonitor()
+            try monitor.captureEvents {
+                let opts = RenameCollectionOptions(writeConcern: expectedWriteConcern)
+                expect(try coll.renamed(to: to, options: opts)).toNot(throwError())
+            }
+
+            // ensure a collection with the new name exists with the documents
+            expect(try db.collection(to).countDocuments()).to(equal(3))
```
kmahar, line 291: might be worth checking `coll` is now empty.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/CollectionOperations.swift
```diff
@@ -460,20 +460,14 @@ struct ReplaceOne: TestOperation {
...
+        sessions _: [String: ClientSession]
     ) throws -> TestOperationResult? {
-        let databaseName = collection.namespace.db
-        let cmd: BSONDocument = [
-            "renameCollection": .string(databaseName + "." + collection.name),
-            "to": .string(databaseName + "." + self.to)
-        ]
-        let reply = try collection._client.db("admin").runCommand(cmd, session: sessions[self.session ?? ""])
-        return TestOperationResult(from: reply)
+        _ = try collection.renamed(to: self.to)
```
kmahar, line 469: we should be passing in `sessions[self.session ?? ""]` for the session here as we were before when running this via `runCommand`

Sources/MongoSwiftSync/MongoCollection.swift
```diff
@@ -63,4 +63,41 @@ public struct MongoCollection<T: Codable> {
...
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Returns:
+     *    A copy of the target `MongoCollection` with the new name.
+     *
+     *    Throws:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
```
kmahar, line 87:
```suggestion
     *    - `MongoError.LogicError` if this collection's parent client has already been closed.
```

Sources/MongoSwift/MongoCollection.swift
```diff
@@ -103,6 +113,42 @@ public struct MongoCollection<T: Codable> {
...
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Returns:
+     *    An `EventLoopFuture<MongoCollection>` evaluating to a copy of the target `MongoCollection` with the new name.
+     *
+     *    If the future evaluates to an error, it is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this databases's parent client has already been closed.
```
kmahar, line 136:
```suggestion
     *    - `MongoError.LogicError` if this collection's parent client has already been closed.
```

kmahar (approved):
LGTM! nice work

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -273,54 +273,64 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
-        let collName2 = "testRenamedColl2"
-        let coll2 = try db.createCollection(collName2)
-        let opts2 = RenameCollectionOptions(dropTarget: false)
-        // test with drop target set to false and new collection name to existing collection throws an error
-        expect(try coll2.renamed(to: to, options: opts2)).to(throwError(errorType: MongoError.CommandError.self))
+        guard MongoSwiftTestCase.topologyType != .sharded else {
+            print("Skipping test case \(self.name) because renameCollection is not compatible with sharded collections")
+            return
+        }
+        try self.withTestNamespace { _, db, _ in
```
patrickfreed, line 314: `withTestNamespace` actually creates the collection for us, so we dont need to create `coll1` and can instead just use the collection provided as the third parameter of the closure.

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -273,54 +273,64 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+            let collName1 = self.getCollectionName(suffix: "renamed1")
+            let coll1 = try db.createCollection(collName1)
+            let to = "existingColl"
```
patrickfreed, line 317: to ensure this is unique, can we use `self.getCollectionName` here?

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -273,54 +273,64 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+        guard MongoSwiftTestCase.topologyType != .sharded else {
+            print("Skipping test case \(self.name) because renameCollection is not compatible with sharded collections")
+            return
+        }
 
-        // insert before rename
-        try coll.insertMany([["doc1": "test1"], ["doc2": "test2"], ["doc1": "test3"]])
+        try self.withTestNamespace { client, db, coll in
+            let encoder = BSONEncoder()
+            let to = "newName"
```
patrickfreed, line 283: ditto use `getCollectionName` to get a unique name

patrickfreed (commented):
just a few more test nits

Sources/MongoSwift/MongoCollection.swift
ncarbon: That's right, collation is not an option expected by the server. https://jira.mongodb.org/browse/CDRIVER-3819
The C Driver will update their documentation.

patrickfreed (approved):
LGTM pending green CI run. great job with this!

PR #547, opened 2020-12-03
SWIFT-913 Add context passing in tests

kmahar:
This introduces a simple Context object (which currently just contains a path, but could have more stuff on it later if we want) which gets passed around throughout all of the matching assertions and is used to construct error messages on failure.

Rather than returning context objects from the matching methods, it seemed simpler to convert all of them to throw errors in-place when something doesn't match, rather than just indicating it with a boolean.

Some sample output (which I got by messing around with the expectations in the first change stream tests JSON to make them fail intentionally):
```
[MongoSwiftSyncTests.UnifiedRunnerTests testSampleUnifiedTests] : failed: caught error: "NonMatchingError(expected: "MongoSwift.BSON.int32(2)", actual: "MongoSwiftSyncTests.MatchableResult.scalar(MongoSwift.BSON.int32(1))", path: ["Operation 4", "expectResult", "fullDocument", "x"])"

[MongoSwiftSyncTests.UnifiedRunnerTests testSampleUnifiedTests] : failed: caught error: "NonMatchingError(expected: "\"element to have one of the following types: [\\\"string\\\"]\"", actual: "MongoSwiftSyncTests.MatchableResult.scalar(MongoSwift.BSON.objectID(5fc844cb5a6e322d3c0ae887))", path: ["Operation 4", "expectResult", "fullDocument", "_id"])"

[MongoSwiftSyncTests.UnifiedRunnerTests testSampleUnifiedTests] : failed: caught error: "NonMatchingError(expected: "\"foo\"", actual: "\"aggregate\"", path: ["Expected events for client client0", "0", "commandName"])"
```

It's not super elegant but I think it definitely improves the debugging situation a lot, let me know what you think.

### DIFF
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
@@ -97,7 +97,7 @@ enum MatchableResult {
                 for k in actualDoc.keys {
                     try context.withPushedElt(k) {
                         guard expectedDoc.keys.contains(k) else {
-                            throw NonMatchingError(expected: "nil", actual: actualDoc[k], context: context)
+                            throw NonMatchingError(expected: nil, actual: actualDoc[k], context: context)
                         }
                     }
                 }
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
@@ -183,9 +183,7 @@ enum MatchableResult {
                 )
             }
             let session = try entities.getEntity(id: id).asSession()
-            guard actualDoc == session.id else {
-                throw NonMatchingError(expected: session.id, actual: self, context: context)
-            }
+            try equals(expected: session.id, actual: actualDoc, context: context)
         }
     }
 
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
@@ -1,9 +1,28 @@
+import Foundation
 import MongoSwiftSync
 import TestsCommon
 
+/// Generic error thrown when matching fails, containing the expected and actual values as well as the path taken to
+/// get to them for nested assertions.
+struct NonMatchingError: LocalizedError {
+    let expected: String
+    let actual: String
+    let path: [String]
+
+    public var errorDescription: String? {
+        "Element at path \(self.path) did not match: expected \(self.expected), actual: \(self.actual)"
+    }
+
+    init(expected: Any?, actual: Any?, context: Context) {
+        self.expected = expected == nil ? "nil" : String(describing: expected!)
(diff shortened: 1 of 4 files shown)


Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
```diff
@@ -94,8 +94,8 @@ struct UnifiedOperation: Decodable {
                 entities[entityId] = try actualResult.asEntity()
```
kmahar, line 94: can't comment on the exact line, but in the next PR we'll start doing a try/catch when executing the operation since some ops expect errors. I'm thinking that for both expected errors (and their assertions) we can pass this context down in, and for unexpected ones we can put it in the message, so it's easier to know which particular operation unexpectedly failed.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -126,8 +126,9 @@ struct UnifiedTestRunner {
                     }
                 }
 
-                for operation in test.operations {
-                    try operation.executeAndCheckResult(entities: &entityMap)
+                for (i, operation) in test.operations.enumerated() {
+                    let context = Context(["Operation \(i)"])
```
patrickfreed, line 130: I think it might be useful to include the operation name here in addition to its index.

patrickfreed (commented):
Looks really good! I just have a few very minor suggestions

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -54,35 +73,36 @@ enum MatchableResult {
...
             }
 
-            // Documents that are not the root-level document should not contain extra keys.
             if case let .subDocument(actualDoc) = self {
                 for k in actualDoc.keys {
-                    guard expectedDoc.keys.contains(k) else {
-                        return false
+                    try context.withPushedElt(k) {
+                        guard expectedDoc.keys.contains(k) else {
+                            throw NonMatchingError(expected: "nil", actual: actualDoc[k], context: context)
```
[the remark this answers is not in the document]
kmahar: yes we can, good catch. an earlier version of this work I had the error initializer accept non-optionals so this was needed and I never fixed this spot.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
kmahar: done. I ended up just storing the operation name on the `UnifiedOperation` object now, since I thought that would be a little clearer and easier to match up with the yaml/JSON than using the name of the Swift type.

patrickfreed (approved):
LGTM! Nice to finally have a solution to hard-to-debug error assertions once and for all. I think we should consider adopting a similar approach in the Rust driver too (filed https://jira.mongodb.org/browse/RUST-612)

PR #548, opened 2020-12-04
SWIFT-577: Add the db.aggregate helper

ncarbon:
Adds the database aggregation helper. I modified the existing `AggragateOperation` file to also execute an aggregate on a `MongoDatabase`.

### DIFF
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -483,7 +483,7 @@ public struct MongoDatabase {
      * underlying collection, such as `$currentOp` and `$listLocalSessions`.
      *
      * - Parameters:
-     *   - pipeline: an `[Document]` containing the pipeline of aggregation operations to perform.
+     *   - pipeline: an `[BSONDocument]` containing the pipeline of aggregation operations to perform.
      *   - options: Optional `AggregateOptions` to use when executing the command
      *   - session: Optional `ClientSession` to use when executing this command
      *
@@ -495,12 +495,13 @@ public struct MongoDatabase {
      *    An `EventLoopFuture<MongoCursor>`. On success, contains a cursor over the resulting documents.
      *
      *    If the future fails, the error is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs on the server while executing the aggregation
      *    - `MongoError.InvalidArgumentError` if the options passed are an invalid combination.
      *    - `MongoError.LogicError` if the provided session is inactive.
      *    - `MongoError.LogicError` if this database's parent client has already been closed.
      *    - `EncodingError` if an error occurs while encoding the options to BSON.
      *
-     * - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.aggregate/
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/#db-aggregate-stages
      */
     public func aggregate(
         _ pipeline: [BSONDocument],
@@ -516,7 +517,7 @@ public struct MongoDatabase {
      * Associates the specified `Codable` type `OutputType` with the returned `MongoCursor`
      *
      * - Parameters:
-     *   - pipeline: an `[Document]` containing the pipeline of aggregation operations to perform.
+     *   - pipeline: an `[BSONDocument]` containing the pipeline of aggregation operations to perform.
      *   - options: Optional `AggregateOptions` to use when executing the command
      *   - session: Optional `ClientSession` to use when executing this command
      *   - withOutputType: the type that each resulting document of the output
@@ -530,12 +531,13 @@ public struct MongoDatabase {
      *    An `EventLoopFuture<MongoCursor>`over the resulting `OutputType`s
      *
      *    If the future fails, the error is likely one of the following:
+     *    - `MongoError.CommandError` if an error occurs on the server while executing the aggregation
      *    - `MongoError.InvalidArgumentError` if the options passed are an invalid combination.
      *    - `MongoError.LogicError` if the provided session is inactive.
      *    - `MongoError.LogicError` if this database's parent client has already been closed.
      *    - `EncodingError` if an error occurs while encoding the options to BSON.
      *
-     * - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.aggregate/
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/#db-aggregate-stages
      */
     public func aggregate<OutputType: Codable>(
         _ pipeline: [BSONDocument],
--- a/Sources/MongoSwift/MongoDatabase.swift
+++ b/Sources/MongoSwift/MongoDatabase.swift
@@ -478,6 +478,79 @@ public struct MongoDatabase {
         return self._client.operationExecutor.execute(operation, client: self._client, session: session)
     }
 
+    /**
+     * Runs an aggregation framework pipeline against this database for pipeline stages that do not require an
+     * underlying collection, such as `$currentOp` and `$listLocalSessions`.
+     *
+     * - Parameters:
+     *   - pipeline: an `[Document]` containing the pipeline of aggregation operations to perform.
+     *   - options: Optional `AggregateOptions` to use when executing the command
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Warning:
+     *    If the returned cursor is alive when it goes out of scope, it will leak resources. To ensure the cursor
+     *    is dead before it leaves scope, invoke `MongoCursor.kill(...)` on it.
+     *
+     * - Returns:
+     *    An `EventLoopFuture<MongoCursor>`. On success, contains a cursor over the resulting documents.
+     *
+     *    If the future fails, the error is likely one of the following:
+     *    - `MongoError.InvalidArgumentError` if the options passed are an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this database's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     *
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.aggregate/
+     */
+    public func aggregate(
+        _ pipeline: [BSONDocument],
+        options: AggregateOptions? = nil,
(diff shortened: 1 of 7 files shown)


Sources/MongoSwift/Operations/AggregateOperation.swift
```diff
@@ -85,33 +94,56 @@ internal struct AggregateOperation<CollectionType: Codable, OutputType: Codable>
...
+                                rpPtr
+                            ) else {
+                                fatalError(failedToRetrieveCursorMessage)
+                            }
+                            return result
                         }
-                        return result
                     }
+
+                    // since mongoc_collection_aggregate and mongoc_database_aggregate don't do any I/O, use forceIO
```
patrickfreed, line 136: I think this comment is out of date and can be removed.

Sources/MongoSwift/Operations/AggregateOperation.swift
```diff
@@ -85,33 +94,56 @@ internal struct AggregateOperation<CollectionType: Codable, OutputType: Codable>
...
-                            MONGOC_QUERY_NONE,
-                            pipelinePtr,
-                            optsPtr,
-                            rpPtr
-                        ) else {
-                            fatalError(failedToRetrieveCursorMessage)
+        return try pipeline.withBSONPointer { pipelinePtr in
+            try withOptionalBSONPointer(to: opts) { optsPtr in
+                try ReadPreference.withOptionalMongocReadPreference(from: self.options?.readPreference) { rpPtr ->
+                    MongoCursor<OutputType> in
```
patrickfreed, line 100: is this `rpPtr -> MongoCursor<OutputType>` annotation necessary? The formatting is kinda funky since it lines up with the rest of the statements in the closure, so it'd be nice if we could just delete it. If it is necessary, then can we add a line break before the call to `withOptionalMongocReadPreference` to try to get the entire closure type annotation on a single line?

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -478,6 +478,79 @@ public struct MongoDatabase {
...
+     * - Returns:
+     *    An `EventLoopFuture<MongoCursor>`over the resulting `OutputType`s
+     *
+     *    If the future fails, the error is likely one of the following:
+     *    - `MongoError.InvalidArgumentError` if the options passed are an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
+     *    - `MongoError.LogicError` if this database's parent client has already been closed.
+     *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     *
+     * - SeeAlso: https://docs.mongodb.com/manual/reference/method/db.aggregate/
```
patrickfreed, line 538: We try to avoid linking directly to the shell's documentation in drivers docs since it can confuse users when the shell adopts an API different from the one adopted by the drivers. I think we could instead link to https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/#db-aggregate-stages to list the stages supported by db-level aggregation. Unfortunately this page is littered with links to the shell documentation too, but I think it's still an improvement over us linking directly to shell documentation.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -478,6 +478,79 @@ public struct MongoDatabase {
...
+     *     of the aggregation operation will be decoded to
+     *
+     * - Warning:
+     *    If the returned cursor is alive when it goes out of scope, it will leak resources. To ensure the cursor
+     *    is dead before it leaves scope, invoke `MongoCursor.kill(...)` on it.
+     *
+     * - Returns:
+     *    An `EventLoopFuture<MongoCursor>`over the resulting `OutputType`s
+     *
+     *    If the future fails, the error is likely one of the following:
```
patrickfreed, line 532: This made me realize that we are omitting `MongoError.CommandError` as a possible error for aggregate. Could you add it here, something like:
```
- `MongoError.CommandError` if an error occurs on the server while executing the aggregation
```

Tests/MongoSwiftSyncTests/MongoDatabaseTests.swift
```diff
@@ -175,6 +175,75 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
         expect(listNamesEvents[2].command["nameOnly"]).to(equal(false))
         expect(listNamesEvents[3].command["nameOnly"]).to(equal(false))
     }
+
+    func testAggregate() throws {
+        let client = try MongoClient.makeTestClient()
+        // $currentOp must be run on the admin database
+        let db = client.db("admin")
+        let result = try db.aggregate([["$currentOp": ["allUsers": true]]]).all()
+        let op = result.filter {
```
patrickfreed, line 184: instead of `filter { ... } [0]`, you can use [`first(where:)`](https://developer.apple.com/documentation/swift/array/1848165-first) which performs the same purpose in a single method.

Tests/MongoSwiftSyncTests/MongoDatabaseTests.swift
```diff
@@ -175,6 +175,75 @@ final class MongoDatabaseTests: MongoSwiftTestCase {
         expect(listNamesEvents[2].command["nameOnly"]).to(equal(false))
         expect(listNamesEvents[3].command["nameOnly"]).to(equal(false))
     }
+
```
patrickfreed, line 178: @kmahar can the unified tests for this be unskipped without any other changes?

patrickfreed (commented):
Looks great! I just have a few minor comments on documentation / style.

Tests/MongoSwiftSyncTests/MongoDatabaseTests.swift
kmahar: I don't think there are any other changes to make here. The only things to change would be code that is introduced in the PR I just opened. Since I assume this PR is probably going to get merged before mine, I will rebase on top of this and update accordingly once this goes in. It should just require adding support for a database target to the `UnifiedAggregate` operation, and unskipping a test.

Sources/MongoSwift/Operations/AggregateOperation.swift
ncarbon: Ooh oops, @patrickfreed what was it doing before? what is forceIO?

Sources/MongoSwift/Operations/AggregateOperation.swift
```diff
@@ -85,33 +94,56 @@ internal struct AggregateOperation<CollectionType: Codable, OutputType: Codable>
...
+                    var result: OpaquePointer
```
kmahar, line 101: I think `result`, `client`, and `decoder` can all be `let`s.

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -478,6 +478,79 @@ public struct MongoDatabase {
         return self._client.operationExecutor.execute(operation, client: self._client, session: session)
     }
 
+    /**
+     * Runs an aggregation framework pipeline against this database for pipeline stages that do not require an
+     * underlying collection, such as `$currentOp` and `$listLocalSessions`.
+     *
+     * - Parameters:
+     *   - pipeline: an `[Document]` containing the pipeline of aggregation operations to perform.
```
kmahar, line 486:
```suggestion
     *   - pipeline: an `[BSONDocument]` containing the pipeline of aggregation operations to perform.
```

same comment applies to other docstrings you've added.

I think this is a copy paste from the docs from the `MongoCollection` `aggregate` methods which I guess we forgot to update back when we renamed this type. would you mind fixing those ones too?

Sources/MongoSwift/Operations/AggregateOperation.swift
patrickfreed: A lot of mongoc functions that return cursors are lazy and only actually perform any I/O once `mongoc_cursor_next` is called on the cursor. We wanted our API to not be lazy and perform the commands when their respective helpers were invoked instead, since we figured this was what users were likely expecting to happen. `forceIO` was a parameter you could specify to our internal `MongoCursor` initializer which would force it reach out to the server, but we communicate that now differently via the `MongocCursorWrapper` protocol, removing the need for the parameter. It looks like we just forgot to remove this comment back when those changes were made.

ncarbon: Gotcha, thanks!

Sources/MongoSwiftSync/MongoDatabase.swift
```diff
@@ -358,4 +358,78 @@ public struct MongoDatabase {
...
+    /**
+     * Runs an aggregation framework pipeline against this database for pipeline stages that do not require an
+     * underlying collection, such as `$currentOp` and `$listLocalSessions`.
+     *
+     * - Parameters:
+     *   - pipeline: an `[Document]` containing the pipeline of aggregation operations to perform
+     *   - options: Optional `AggregateOptions` to use when executing the command
+     *   - session: Optional `ClientSession` to use when executing this command
+     *
+     * - Warning:
```
kmahar, line 371: this actually isn't a problem in the sync API, so we can get rid of the warning. 

for synchronous cursors, we automatically call `kill` for you in the `deinit` method of the type: https://github.com/mongodb/mongo-swift-driver/blob/43ca721232954904f1d5ae754e8fa6fd56c6db6c/Sources/MongoSwiftSync/MongoCursor.swift#L138-L140

so users don't have to worry about it.

In the async API, users unfortunately have to call the method manually. That is required because of the non-deterministic way in which asynchronous tasks get scheduled and completed. You can write 
```
asyncFunctionA()
asyncFunctionB()
```
but `B` might actually execute/finish before `A` does.

 A cursor being properly killed and cleaned up requires that its parent `MongoClient` is still alive, since that client maintains e.g. a pool of connections that are needed to talk to the server. 
If we just fired off an asynchronous call to `kill` from the asynchronous cursor's `deinit`, there is no way for a user to guarantee that it got cleaned up properly before they close the parent client. Having an explicit method makes it so users can sync up the order these things happen in.

Sources/MongoSwift/Operations/AggregateOperation.swift
patrickfreed: This comment still needs to be removed.

patrickfreed (commented):
Just need to remove that one old comment, otherwise LGTM!

kmahar (approved):
lgtm!

patrickfreed (approved):
LGTM!

PR #549, opened 2020-12-04
SWIFT-913 Implement support for remaining operations and error matching

kmahar:
As I mentioned in the team channel there seem to be some sporadic failures on evg along the lines of the transactions tests ones we've been seeing lately. I'm going to investigate that now but I don't think it needs to block starting review on this.

Sorry that this is a bit larger than past PRs, but a lot of it is just operation execution boilerplate so hopefully should be easy to review.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -144,10 +144,10 @@ public final class ClientSession {
     }
 
     /// The transaction state of this session.
-    internal var transactionState: TransactionState? {
+    internal var transactionState: TransactionState {
         switch self.state {
         case .notStarted, .ended:
-            return nil
+            return .none
         case let .started(session, _):
             return TransactionState(mongocTransactionState: mongoc_client_session_get_transaction_state(session))
         }
@@ -155,10 +155,7 @@ public final class ClientSession {
 
     /// Indicates whether or not the session is in a transaction.
     internal var inTransaction: Bool {
-        if let transactionState = self.transactionState {
-            return transactionState != .none
-        }
-        return false
+        self.transactionState != .none
     }
 
     /// The most recent cluster time seen by this session. This value will be nil if either of the following are true:
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -316,7 +316,7 @@ public struct BulkWriteOptions: Codable {
 }
 
 /// The result of a bulk write operation on a `MongoCollection`.
-public struct BulkWriteResult: Decodable {
+public struct BulkWriteResult: Codable {
     /// Number of documents deleted.
     public let deletedCount: Int
 
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
@@ -8,7 +8,7 @@ class Context {
     /// Entities created for the test.
     var entities: EntityMap
 
-    /// Fail points enabled during test execution.
+    /// List of (fail point name, server address fail point was set on).
     var enabledFailPoints: [(String, ServerAddress?)] = []
 
     let internalClient: MongoClient
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
@@ -8,8 +8,8 @@ class Context {
     /// Entities created for the test.
     var entities: EntityMap
 
-    /// List of (fail point name, server address fail point was set on).
-    var enabledFailPoints: [(String, ServerAddress?)] = []
+    /// Fail points that have been set during test execution and should be disabled on completion.
+    var enabledFailPoints: [FailPointGuard] = []
 
     let internalClient: MongoClient
 
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
@@ -1,10 +1,22 @@
+import MongoSwiftSync
+
 /// Context passed around while executing tests.
 class Context {
     /// Path taken to get to the current state.
     var path: [String]
 
-    init(_ path: [String]) {
+    /// Entities created for the test.
+    var entities: EntityMap
+
+    /// Fail points enabled during test execution.
+    var enabledFailPoints: [(String, ServerAddress?)] = []
+
+    let internalClient: MongoClient
+
+    init(path: [String], entities: EntityMap, internalClient: MongoClient) {
         self.path = path
+        self.entities = entities
+        self.internalClient = internalClient
     }
 
     /// Executes a closure with the given path element added to the path, removing it after the closure is complete.
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
@@ -122,7 +122,7 @@ enum EntityDescription: Decodable {
 struct UnifiedTestClient {
     let client: MongoClient
 
-    fileprivate let commandMonitor: UnifiedTestCommandMonitor
+    let commandMonitor: UnifiedTestCommandMonitor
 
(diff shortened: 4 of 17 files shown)


Sources/MongoSwift/ClientSession.swift
```diff
@@ -144,21 +144,18 @@ public final class ClientSession {
     }
 
     /// The transaction state of this session.
-    internal var transactionState: TransactionState? {
+    internal var transactionState: TransactionState {
```
kmahar, line 147: I don't think we really need the distinction we had here; if we have not started any libmongoc session or have ended the current one then the transaction state for our Swift session should be none.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -316,7 +316,7 @@ public struct BulkWriteOptions: Codable {
 }
 
 /// The result of a bulk write operation on a `MongoCollection`.
-public struct BulkWriteResult: Decodable {
+public struct BulkWriteResult: Codable {
```
kmahar, line 319: I ended up just making all of these write result types Codable for the purpose of matching results. I don't see why it would be an issue since they all conform in a standard manner, and I highly doubt users will rely on the conformance anyway.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Context.swift
```diff
@@ -1,10 +1,22 @@
+import MongoSwiftSync
+
 /// Context passed around while executing tests.
 class Context {
     /// Path taken to get to the current state.
     var path: [String]
 
-    init(_ path: [String]) {
+    /// Entities created for the test.
+    var entities: EntityMap
```
kmahar, line 9: everywhere we pass a context we also pass the entity map, and in order to support some of the newer operations we needed to start passing around the internal client and have some way of recording set fail points too, so it seemed to work nicely to just use `Context` for all of that.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -122,7 +122,7 @@ enum EntityDescription: Decodable {
 struct UnifiedTestClient {
     let client: MongoClient
 
-    fileprivate let commandMonitor: UnifiedTestCommandMonitor
+    let commandMonitor: UnifiedTestCommandMonitor
```
kmahar, line 125: (changes in this file are because the `assertSame`/`assertDifferentLsidsOnLastTwoCommands` operations needs to be able to access these events)

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -95,10 +95,12 @@ enum MatchableResult {
 
             if case let .subDocument(actualDoc) = self {
                 for k in actualDoc.keys {
-                    try context.withPushedElt(k) {
-                        guard expectedDoc.keys.contains(k) else {
-                            throw NonMatchingError(expected: nil, actual: actualDoc[k], context: context)
-                        }
+                    guard expectedDoc.keys.contains(k) else {
```
kmahar, line 98: so I got a failure here in implementing the rest of the operations and I realized that the output was confusing, because the path contained a bunch of elements from the test file e.g. but then ended with a key that was only in the "actual" output, and it wasn't obvious that the problem was we saw the key in "actual". So I updated this to just more explicitly state what the problem is

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -361,6 +363,226 @@ func matchesEvents(expected: [ExpectedEvent], actual: [CommandEvent], entities:
...
+        if let codeName = expected.errorCodeName {
+            try context.withPushedElt("errorCodeName") {
+                guard let actualWithCodeNames = self as? HasErrorCodes else {
+                    throw NonMatchingError(
+                        expected: "error to have error code(s)",
+                        actual: self,
+                        context: context
+                    )
+                }
+                // TODO: SWIFT-1022: Due to CDRIVER-3147 many of our errors are currently missing code names, so we
```
kmahar, line 489: this is kind of not great in the case of a bulk error since you could mistakenly take the presence of an empty string that is really present because of some different error code/name pair to mean we had the right code name here. that said, I'm not too worried about it because we do make the error code number assertions correctly.

CDRIVER-3147 has also been slated to be fixed in 1.18 so I don't think we'll have this issue for very long.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -148,13 +204,30 @@ struct UnifiedTestRunner {
...
+                                expected: expectedEventList.events,
+                                actual: actualEvents,
+                                context: context
+                            )
+                        }
+                    }
+                }
+
+                if let expectedOutcome = test.outcome {
+                    for cd in expectedOutcome {
```
patrickfreed, line 218: what does `cd` stand for here?

Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
```diff
@@ -143,4 +267,23 @@ struct UnifiedTargetedFailPoint: UnifiedOperationProtocol {
     static var knownArguments: Set<String> {
         ["failPoint", "session"]
     }
+
+    func execute(on _: UnifiedOperation.Object, context: Context) throws -> UnifiedOperationResult {
```
patrickfreed, line 271: could we use the `FailPoint` type here instead of manual runCommand? It has a `Decodable` implementation so we can swap out the `BSONDocument` directly. As another improvement, we can add a new `FailPointGuard` type that is returned from `FailPoint.enable` which has a `deinit` that disables the enabled failpoint, and we can store those instead of the command name / address tuple. We do something similar in Rust.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -361,6 +363,226 @@ func matchesEvents(expected: [ExpectedEvent], actual: [CommandEvent], entities:
...
+            codeNames.append(wcCodeName)
+        }
+        return codeNames
+    }
+}
+
+extension Error {
+    func matches(_ expected: ExpectedError, context: Context) throws {
+        if let isClientError = expected.isClientError {
+            let actualIsClientError = self is MongoUserError || self is MongoRuntimeError
```
patrickfreed, line 419: could we add this as an extension of `MongoErrorProtocol`? That way if we need to change / amend it, it'll be easier to find in the future.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
kmahar: ✅ ✅

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -361,6 +363,208 @@ func matchesEvents(expected: [ExpectedEvent], actual: [CommandEvent], entities:
...
+    var errorCodeNames: [String] {
+        var codeNames = self.writeFailures?.map { $0.codeName } ?? []
+        if let wcCodeName = self.writeConcernFailure?.codeName {
+            codeNames.append(wcCodeName)
+        }
+        return codeNames
+    }
+}
+
+extension MongoErrorProtocol {
```
kmahar, line 416: I ended up just making this a method on `MongoErrorProtocol` instead; we only need to match on our own errors and it removes the need to do some casting

Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
kmahar: so this actually led a bunch of other tests that use the failpoint type to start failing, because I was not using the return value of `enable` so the failpoints were just being immediately deactivated. 
I don't think it's worth the time to go through and update those runners to pass around the `FailPointGuard`s and keep them alive, so I'm just going to switch to manually constructing the `FailPointGuard`s in the new operations.

patrickfreed (approved):
LGTM!

PR #550, opened 2020-12-08
SWIFT-519 Make startAfter public

ncarbon:
Makes `startAfter` public now that libmongoc 4.2 change streams has been released.

### DIFF
--- a/Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
+++ b/Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
@@ -957,6 +957,17 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
 
         // expect this change stream to have more events after resuming
         expect(try changeStream2.nextWithTimeout()).toNot(beNil())
+
+        // test with startAfter
+        let newColl = try coll.renamed(to: self.getCollectionName(suffix: "2"))
+
+        _ = try changeStream2.nextWithTimeout()
+        expect(try changeStream2.nextWithTimeout()?.operationType).to(equal(.invalidate))
+
+        let opts = ChangeStreamOptions(startAfter: changeStream2.resumeToken)
+
+        // resuming (with startAfter) after an invalidate event should work
+        expect(try newColl.watch(options: opts)).toNot(throwError())
     }
 
     struct MyFullDocumentType: Codable, Equatable {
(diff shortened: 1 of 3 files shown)


Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -957,6 +957,22 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
...
+        // test with startAfter
+        let newColl = try coll.renamed(to: self.getCollectionName(suffix: "2"))
+
+        _ = try changeStream2.nextWithTimeout()
+        expect(try changeStream2.nextWithTimeout()?.operationType).to(equal(.invalidate))
+
+        let opts = ChangeStreamOptions(startAfter: changeStream2.resumeToken)
+
+        // resuming (with startAfter) after an invalidate event should work
+        expect(try newColl.watch(options: opts)).toNot(throwError())
```
patrickfreed, line 975: I think `startAfter` is used to essentially resume a change stream on the original collection, even after that collection may have been invalidated. So I think we want to use `coll.watch` here. Also, I think we should perform an insert or something into `coll` and ensure the new change stream resumed via `startAfter` picks up the event.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -957,6 +957,27 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
...
+        // test with startAfter
+        _ = try coll.renamed(to: self.getCollectionName(suffix: "2"))
+
+        _ = try changeStream2.nextWithTimeout()
+        expect(try changeStream2.nextWithTimeout()?.operationType).to(equal(.invalidate))
+
+        let opts = ChangeStreamOptions(startAfter: changeStream2.resumeToken)
+
+        // resuming (with startAfter) after an invalidate event should work
+        expect(try coll.watch(options: opts)).toNot(throwError())
```
patrickfreed, line 975: rather than creating multiple change streams here, I think it's okay to just rely on `try` to report any test failures and not do the `expect(...).toNot(throwError())`.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -957,6 +957,26 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
...
+
+        _ = try changeStream2.nextWithTimeout()
+        expect(try changeStream2.nextWithTimeout()?.operationType).to(equal(.invalidate))
+
+        let opts = ChangeStreamOptions(startAfter: changeStream2.resumeToken)
+
+        // resuming (with startAfter) after an invalidate event should work
+        let changeStream3 = try coll.watch(options: opts)
+
+        try coll.insertOne(["kitty": "cat"])
```
patrickfreed, line 977: can we also add an insert before the call to `watch` and verify they're both seen? This tests the picking-up-where-you-left-off feature of `resumeAfter`.

patrickfreed (approved):

kmahar (approved):

PR #552, opened 2020-12-09
SWIFT-481 Support index all paths

ncarbon:
adds the `wildcardProjection` option to `IndexOptions`

### DIFF
--- a/Sources/MongoSwift/MongoCollection+Indexes.swift
+++ b/Sources/MongoSwift/MongoCollection+Indexes.swift
@@ -111,6 +111,9 @@ public struct IndexOptions: Codable {
     /// Optionally specifies fields in the index and their corresponding weight values.
     public var weights: BSONDocument?
 
+    /// Optionally specifies the wildcard projection of a wildcard index.
+    public var wildcardProjection: BSONDocument?
+
     /// Convenience initializer allowing any/all parameters to be omitted.
     public init(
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -111,6 +111,9 @@ public struct IndexOptions: Codable {
     /// Optionally specifies fields in the index and their corresponding weight values.
     public var weights: BSONDocument?
 
+    /// Optionally specifies the wildcard projection of a wildcard index.
+    public var wildcardProjection: BSONDocument?
```
patrickfreed, line 115: Can we update this to be more more specific? Something like "Optionally specifies a projection document used to determine which fields are indexed by a wildcard index."

Can specify that this feature is only supported in MongoDB 4.2+? 

Also, can we add a link / direct users to dochub.mongodb.org/core/index-wildcard for more information about wild card indexes.

patrickfreed (commented):
Code changes LGTM. Just some suggestions for documentation

patrickfreed:
Also, looks like the server changed something such that the wildcardProjects use `true` and `false` instead of 1 and 0.

patrickfreed (approved):
LGTM pending green ci!

kmahar (approved):

PR #554, opened 2020-12-10
SWIFT-1028 Create EventLoopBoundMongoClient type

ncarbon:


Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -150,3 +166,15 @@ final class MongoClientTests: MongoSwiftTestCase {
         try closeFuture.wait()
     }
 }
+
+class TestEventLoop: Equatable {
```
ncarbon, line 170: This wrapper is just a helper for comparing event loops in the tests since `EventLoop` is a protocol. Is there a another way to be able to check the event loop used in the operation is the same as the one specified in `bound()`? Or is this fine?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -655,6 +655,20 @@ public class MongoClient {
         return self.operationExecutor.execute(operation, client: self, session: session)
     }
 
+    /**
+     * Returns an `EventLoopBoundMongoClient`, a wrapper around this `MongoClient` that will use only the specified
```
ncarbon, line 659: Do we want to say 'wrapper' in the docstrings?

Sources/MongoSwift/EventLoopBoundMongoClient.swift
```diff
@@ -0,0 +1,98 @@
+import NIO
+
+// sourcery: skipSyncExport
+/// An event loop bound MongoDB Client.
+public struct EventLoopBoundMongoClient {
+    /// The underlying `MongoClient`.
+    internal let client: MongoClient
+
+    /// The `EventLoop` bound to this `EventLoopBoundMongoClient`.
```
patrickfreed, line 9: The "binding" is actually happening the other way around: this is the `EventLoop` that this client is bound to.

Sources/MongoSwift/EventLoopBoundMongoClient.swift
```diff
@@ -0,0 +1,98 @@
...
+    public func listDatabases(
+        _ filter: BSONDocument? = nil,
+        options: ListDatabasesOptions? = nil,
+        session: ClientSession? = nil
+    ) -> EventLoopFuture<[DatabaseSpecification]> {
+        let operation = ListDatabasesOperation(client: self.client, filter: filter, nameOnly: nil, options: options)
+        return self.client.operationExecutor.execute(
+            operation,
+            client: self.client,
+            on: self.eventLoop,
```
patrickfreed, line 50: I think we were planning on using helpers here to ensure the event loop was passed in properly, but perhaps the helper isn't actually necessary if the event loop argument is non-optional.

Sources/MongoSwift/MongoClient.swift
patrickfreed: I think its fine to say wrapper. It indicates that the thing returned from this method should have the same behavior as a regular client.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -655,6 +655,20 @@ public class MongoClient {
...
+     * `EventLoop` for executing commands.
+     *
+     * - Parameters:
+     *   - eventLoop: An `EventLoop` which the returned `EventLoopBoundMongoClient` will be bound to.
+     *
+     * - Returns:
+     *    An `EventLoopBoundMongoClient` with the specified `EventLoop` bound to it.
```
patrickfreed, line 666: ditto flip ordering here: the client is bound to the event loop, not the other way around.

Sources/MongoSwift/EventLoopBoundMongoClient.swift
```diff
@@ -0,0 +1,98 @@
+import NIO
+
+// sourcery: skipSyncExport
+/// An event loop bound MongoDB Client.
```
patrickfreed, line 4: I think we should rewrite this docstring to use different words than are in the type itself to explain what being bound means, similar to the docstring on the `bound` method.

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -82,6 +82,22 @@ final class MongoClientTests: MongoSwiftTestCase {
         expect(ids[2]).to(equal(ids[1] + 1))
     }
 
+    func testBound() throws {
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
patrickfreed, line 86: can we increase the `numberOfThreads` here? Otherwise the event loop group will only have a single event loop in it, and this test would pass every time.

Tests/MongoSwiftTests/MongoClientTests.swift
patrickfreed: Nimble supplies matchers for using the `===` operator directly, so you can just use those in the test.

https://github.com/Quick/Nimble#identity

Sources/MongoSwift/EventLoopBoundMongoClient.swift
ncarbon: Oh yeah, I actually meant to ask clarification on the helpers. I will ask more in detail in the slack channel.

Tests/MongoSwiftTests/MongoClientTests.swift
ncarbon: thanks!

Tests/MongoSwiftTests/MongoClientTests.swift
```diff
@@ -82,6 +82,22 @@ final class MongoClientTests: MongoSwiftTestCase {
         expect(ids[2]).to(equal(ids[1] + 1))
     }
 
+    func testBound() throws {
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
+        let expectedEventLoop = elg.next()
+        try self.withTestClient { client in
```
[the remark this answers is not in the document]
kmahar: I don't see a way to directly check whether an event loop is a member of a group. however, I'm not sure that is a problem, can you think of any issues you might run into allowing an arbitrary event loop?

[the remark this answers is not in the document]
patrickfreed: I wasn't sure if there were any NIO invariants that would be violated by hopping between loops in different groups, though I don't really have anything in mind / can't think of anything.

Tests/MongoSwiftTests/AsyncTestUtils.swift
```diff
@@ -73,9 +73,10 @@ extension MongoSwiftTestCase {
     internal func withTestClient<T>(
         _ uri: String = MongoSwiftTestCase.getConnectionString().toString(),
         options: MongoClientOptions? = nil,
+        eventLoopGroup: MultiThreadedEventLoopGroup? = nil,
```
ncarbon, line 76: It accepts a `MultiThreadedEventLoopGroup` since this method already uses one. I am not sure why `makeTestClient` just accepts an `EventLoopGroup` though. I wonder if we want to do that here too.

patrickfreed: `EventLoopGroup` is the protocol that `MultiThreadedEventLoopGroup` conforms to. I think we use it in `makeTestClient` in case we want to try mock an event loop or something. I think we ought to just use the protocol here too, though I don't anticipate using a non-`MutliThreadedEventLoopGroup` for a while.

patrickfreed (approved):
LGTM mod the update to using `EventLoopGroup` instead of the specific implementation.

Sources/MongoSwift/EventLoopBoundMongoClient.swift
```diff
@@ -0,0 +1,98 @@
...
+        .flatMapThrowing { result in
+            guard case let .specs(dbs) = result else {
+                throw MongoError.InternalError(message: "Invalid result")
+            }
+            return dbs
+        }
+    }
+
+    /**
+     * Gets the names of databases in this client's MongoDB deployment. The returned future will be bound to the
```
kmahar, line 62: (same comments on bound apply here and under `Returns`)

kmahar (commented):
just some nits about documentation, otherwise lgtm

Tests/MongoSwiftTests/MongoClientTests.swift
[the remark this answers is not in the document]
kmahar: I'm not aware of any potential issues with that. I think the ELG is just a convenience for multiplexing across event loops and there's no relationship between the individual event loops otherwise.

Sources/MongoSwift/EventLoopBoundMongoClient.swift
```diff
@@ -0,0 +1,98 @@
+import NIO
+
+// sourcery: skipSyncExport
+/// A wrapper around a `MongoClient` that will use only the specified `EventLoop` for executing commands.
+public struct EventLoopBoundMongoClient {
+    /// The underlying `MongoClient`.
+    internal let client: MongoClient
+
+    /// The `EventLoop` this `EventLoopBoundMongoClient` will be bound to.
+    internal let eventLoop: EventLoop
```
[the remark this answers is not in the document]
patrickfreed: Oh yeah, good catch.

kmahar (approved):
LGTM! nice work

PR #555, opened 2020-12-16
SWIFT-1029 Add support for EventLoop binding for MongoDatabase and MongoCollection

ncarbon:
- Adds a method `EventLoopBoundMongoClient.db` that returns a bound `MongoDatabase`.
- Updates `MongoDatabase.collection` to return a bound `MongoCollection`.
- Updates the calls to `operationExecutor` to pass in the `EventLoop` (except for the ones where `MongoCursor` and `ChangeStreams` are returned, those will be modified in the next PR).
- Adds tests to verify that the futures returned by the methods are on the bound event loop.

### DIFF
--- a/Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
+++ b/Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
@@ -8,22 +8,18 @@ final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
         let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
         let expectedEventLoop = elg.next()
 
-        try self.withTestClient { client in
+        try self.withTestClient(eventLoopGroup: elg) { client in
+            // TODO: SWIFT-1030 - add tests for database operations that return ChangeStream and MongoCursor
             let elBoundClient = EventLoopBoundMongoClient(client: client, eventLoop: expectedEventLoop)
-            expect(elBoundClient.db(Self.testDatabase).eventLoop) === expectedEventLoop
-        }
-    }
-
-    func testCollection() throws {
-        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
-        let expectedEventLoop = elg.next()
-
-        try self.withTestClient { client in
-            let elBoundClient = EventLoopBoundMongoClient(client: client, eventLoop: expectedEventLoop)
-
             let db = elBoundClient.db(Self.testDatabase)
-            defer { try? db.drop().wait() }
+            expect(db.eventLoop) === expectedEventLoop
+
+            // test the MongoDatabase operations return futures on the expected event loop
+            expect(db.createCollection("test").eventLoop) === expectedEventLoop
             expect(db.collection("test").eventLoop) === expectedEventLoop
+            expect(db.listMongoCollections().eventLoop) === expectedEventLoop
+            expect(db.listCollectionNames().eventLoop) === expectedEventLoop
+            expect(db.runCommand(["insert": "coll", "documents": [["foo": "bar"]]]).eventLoop) === expectedEventLoop
         }
     }
 
@@ -31,7 +27,7 @@ final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
         let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
         let expectedEventLoop = elg.next()
 
-        try self.withTestClient { client in
+        try self.withTestClient(eventLoopGroup: elg) { client in
             let elBoundClient = EventLoopBoundMongoClient(client: client, eventLoop: expectedEventLoop)
 
             let db = elBoundClient.db(Self.testDatabase)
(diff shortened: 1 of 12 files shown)


Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -0,0 +1,140 @@
+@testable import MongoSwift
+import Nimble
+import NIO
+import TestsCommon
+
+final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
```
patrickfreed, line 6: each of these tests need to pass in the `elg` to `withTestClient` for automatic cleanup, since otherwise we'll be leaking resources.

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -0,0 +1,140 @@
...
+    func testEventLoopBoundDb() throws {
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 4)
+        let expectedEventLoop = elg.next()
+
+        try self.withTestClient { client in
+            let elBoundClient = EventLoopBoundMongoClient(client: client, eventLoop: expectedEventLoop)
+            expect(elBoundClient.db(Self.testDatabase).eventLoop) === expectedEventLoop
```
patrickfreed, line 13: can we add a few tests for `MongoDatabase` operations here? As part of that, one that ensures the `MongoCollection` returned from `createCollection` has the right event loop.

patrickfreed (commented):
Looks pretty good, just have a few testing suggestions

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
ncarbon: oops, yes, I forgot to update these after `withTestClient` was changed.

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -8,53 +8,49 @@ final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
...
-
-        try self.withTestClient { client in
-            let elBoundClient = EventLoopBoundMongoClient(client: client, eventLoop: expectedEventLoop)
-
             let db = elBoundClient.db(Self.testDatabase)
-            defer { try? db.drop().wait() }
+            expect(db.eventLoop) === expectedEventLoop
+
+            // test the MongoDatabase operations return futures on the expected event loop
+            expect(db.createCollection("test").eventLoop) === expectedEventLoop
```
patrickfreed, line 18: can we ensure the collection this future evaluates to has the proper event loop too?

patrickfreed (commented):
I think sourcery needs to be run again, it looks like the tests aren't compiling as a result

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
ncarbon: Oh yeah, I also just added checks for the collections returned by `listMongoCollections`, but I wonder if that's redundant?

ncarbon: I forgot to run `make linuxmain`

mongo-swift-driver.xcodeproj/project.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
patrickfreed, line 1: I think this file was accidentally committed

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
patrickfreed: No I don't think it's redundant, its a good test case :+1:

patrickfreed (approved):
LGTM!

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -0,0 +1,149 @@
+@testable import MongoSwift
+import Nimble
+import NIO
+import TestsCommon
+
+final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
```
kmahar, line 6: throughout these tests we probably want to `wait()` on any future we create before returning, even if we don't care about the result or whether it actually executes. 

otherwise, I think we may run into race conditions around the futures completing and the client being closed by `withTestClient`, which may cause some sporadic errors.

kmahar (commented):
looks great! I just have one concern about the tests as written

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -0,0 +1,171 @@
...
+        try self.withTestClient(eventLoopGroup: elg) { client in
+            // TODO: SWIFT-1030 - add tests for database operations that return ChangeStream and MongoCursor
+            let elBoundClient = EventLoopBoundMongoClient(client: client, eventLoop: expectedEventLoop)
+            let db = elBoundClient.db(Self.testDatabase)
+            defer { try? db.drop().wait() }
+            expect(db.eventLoop) === expectedEventLoop
+
+            // test the MongoDatabase operations return futures on the expected event loop
+            expect(db.collection("test").eventLoop) === expectedEventLoop
+            expect(db.listCollectionNames().eventLoop) === expectedEventLoop
```
kmahar, line 20: I think we need to wait on this and the next line

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -0,0 +1,171 @@
...
+            let coll1 = try db.createCollection(self.getCollectionName(suffix: "1")).wait()
+
+            // test renamed
+            let res = coll1.renamed(to: self.getCollectionName(suffix: "2"))
+            let coll2 = try res.wait()
+            expect(res.eventLoop) === expectedEventLoop
+            expect(coll2.eventLoop) === expectedEventLoop
+
+            // test drop
+            expect(coll2.drop().eventLoop) === expectedEventLoop
```
kmahar, line 56: need to wait on this as well

kmahar (commented):
couple more places to `wait()`

kmahar (approved):

PR #556, opened 2020-12-18
SWIFT-1030 Implement EventLoop binding support for change streams and cursors

ncarbon:


Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -90,11 +167,30 @@ final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
             let res3 = coll1.distinct(fieldName: "foo", filter: [:])
             expect(res3.eventLoop) === expectedEventLoop
             _ = try res3.wait()
+
+            // test find
+            let res4 = coll1.find([:])
```
patrickfreed, line 172: can we rename these variables to be a bit more unique? i.e. instead of `res4` we can call it `findFuture` or something like that. That way it's easier to identify what each assertion is doing, and it's harder to accidentally reuse a variable from one set of assertions to the next. Ditto for the cursors and for the other tests that have these numbered variables.

patrickfreed (commented):
Changes look good, just have a minor comment on variable names in the tests.

patrickfreed (approved):

kmahar (commented):
lgtm! nice work

PR #594, opened 2021-01-26
SWIFT-936 Use new BSON library

patrickfreed:
SWIFT-936

This PR updates the driver to use [`swift-bson`](https://github.com/mongodb/swift-bson), the new pure-Swift BSON implementation. It consists of mostly testing changes and should have no effect on the public API. I think the CRUD tests were synced due to a bug in them or something. I did it a while ago though so I'm not exactly sure.


Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -338,7 +333,7 @@ public func cleanEqual(_ expectedValue: String?) -> Predicate<String> {
 
 // Adds a custom "sortedEqual" predicate that compares two `Document`s and returns true if they
 // have the same key/value pairs in them
-public func sortedEqual(_ expectedValue: BSONDocument?) -> Predicate<BSONDocument> {
+public func sortedEqual<T: SortedEquatable>(_ expectedValue: T?) -> Predicate<T> {
```
patrickfreed, line 336: In order to extend this matcher to `BSON` (which may contain arrays with documents or documents themselves), I introduced a protocol and made the matcher generic over it. This became more useful as more things became unordered due to the usage of ExtrasJSON instead of Foundation.

Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
```diff
@@ -70,7 +70,22 @@ internal struct FailPoint: Decodable {
     }
 
     public init(from decoder: Decoder) throws {
-        self.failPoint = try BSONDocument(from: decoder)
+        let container = try decoder.singleValueContainer()
```
patrickfreed, line 73: This is required due to the ordering of the command getting messed up during decoding from JSON.

Tests/MongoSwiftSyncTests/SpecTestRunner/TestOperations/DatabaseOperations.swift
```diff
@@ -44,6 +44,22 @@ struct RunCommand: TestOperation {
     let command: BSONDocument
     let readPreference: ReadPreference?
 
+    /// Return a new `RunCommand` with the command document ordered such that the provided command name
+    /// is the first key.
+    func withCommandName(_ name: String) -> RunCommand {
```
patrickfreed, line 49: ditto required to reorder the command properly

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -265,41 +265,36 @@ public struct TestRequirement: Decodable {
     }
 }
 
-extension BSONDocument {
+public protocol SortedEquatable {
+    func sortedEquals(_ other: Self) -> Bool
+}
+
+extension BSONDocument: SortedEquatable {
     public func sortedEquals(_ other: BSONDocument) -> Bool {
```
patrickfreed, line 273: We have an implementation of this in `swift-bson` now (called `equalsIgnoreKeyOrder`), but it currently has a bug where it doesn't compare documents contained in an array via `equalsIgnoreKeyOrder`. I filed SWIFT-1086 to cover fixing that. Once that's done we can remove this implementation of `sortedEquals` and just call through to `equalsIgnoreKeyOrder`.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -67,8 +58,8 @@ internal struct TestCommandStartedEvent: Decodable, Matchable {
                 return false
             }
         }
-        return self.commandName.matches(expected: expected.commandName)
-            && self.command.matches(expected: expected.command)
+
```
kmahar, line 61: was this just a redundant check? I assume the command name is always in the document already?

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -220,6 +212,16 @@ extension SpecTestFile {
                 continue
             }
 
+            // abort any open transactions from previous test runs
```
kmahar, line 215: if this is working maybe it doesn't matter, but the unified test runner has you run this with primary read preference in repl sets, and run it against every mongos individually. if only the existing spec tests had all been ported to the new format, so you wouldn't have to update this runner...

https://github.com/mongodb/mongo-swift-driver/blob/8e7a0ff92f5dee4de5c5fb365eaacd7ab8206ebf/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift#L24

Sources/MongoSwift/BSON/BSONUtil.swift
```diff
@@ -8,14 +10,81 @@ internal func convertingBSONErrors<T>(_ body: () throws -> T) rethrows -> T {
...
+    /// Initializes a new `Date` representing the instance `msSinceEpoch` milliseconds
+    /// since the Unix epoch.
+    internal init(msSinceEpoch: Int64) {
+        self.init(timeIntervalSince1970: TimeInterval(msSinceEpoch) / 1000.0)
+    }
+}
+
+extension BSONDocument {
+    /**
+     * Initializes a `BSONDocument` using an array where the values are optional
```
kmahar, line 74: nit: they are not optionals

kmahar (commented):
looks good and  it's really exciting to finally be doing this 🎉 I just have some small questions/nits

Sources/MongoSwift/BSON/BSONUtil.swift
patrickfreed: updated the whole docstring to be a little clearer

Sources/MongoSwift/MongoCollection+Indexes.swift
```diff
@@ -17,7 +17,7 @@ public struct IndexModel: Codable {
 
     /// Gets the default name for this index.
     internal var defaultName: String {
-        self.keys.map { k, v in "\(k)_\(v.bsonValue)" }.joined(separator: "_")
+        self.keys.map { k, v in "\(k)_\(v.toInt()?.description ?? "invalid")" }.joined(separator: "_")
```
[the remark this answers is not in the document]
patrickfreed: good point, updated to throw instead of generating a bad name

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
patrickfreed: Yeah, the command name was being derived from the command document, so checking twice wasn't necessary. Also, we can no longer derive the command name from the document because of the ordering issues from JSON, so `self.commandName` is now optional, with it only being non-nil when initializing a `TestCommandStartedEvent` from an actual `CommandStartedEvent`.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
patrickfreed: Since this runner will be deleted eventually anyways, I just copied the implementation over from the unified runner.

kmahar (approved):
LGTM!

PR #596, opened 2021-01-28
SWIFT-1045 Ensure MongoClient initializer does not perform blocking DNS request

patrickfreed:
SWIFT-1045

The `MongoClient` initializer currently performs a blocking DNS request. In the event of a DNS outage, this could block an event loop indefinitely. This PR updates the `MongoClient` initializer to construct the `mongoc_client_pool_t` on the thread pool instead, ensuring the initializer does not block.

I tested this by configuring an invalid DNS server and trying to create a client. After the changes it still happens immediately, prior to the changes it hangs until a backup DNS server is tried (~10s).

### DIFF
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -255,20 +255,26 @@ internal class ConnectionPool {
     /// after it is passed to this method.
     internal func setAPMCallbacks(callbacks: OpaquePointer, client: MongoClient) {
         // lock isn't needed as this is called before pool is in use.
-        switch self.state {
-        case let .opening(future):
-            future.whenSuccess { pool in
-                mongoc_client_pool_set_apm_callbacks(pool, callbacks, Unmanaged.passUnretained(client).toOpaque())
-                mongoc_apm_callbacks_destroy(callbacks)
-            }
-            // this shouldn't be reached but just in case we still free the memory
-            future.whenFailure { _ in
-                mongoc_apm_callbacks_destroy(callbacks)
-            }
-        case .closing, .closed, .open:
+        guard case let .opening(future) = self.state else {
             // this method is called via `initializeMonitoring()`, which is called from `MongoClient.init`.
             // unless we have a bug it's impossible that the pool is already closed.
             fatalError("ConnectionPool in unexpected state \(self.state) while setting APM callbacks")
         }
+
+        // to ensure the callbacks get set before any waiting on this future becomes unblocked, use
+        // `map` to create a new future instead of .whenSuccess
+        let newFut = future.map { pool -> OpaquePointer in
+            mongoc_client_pool_set_apm_callbacks(pool, callbacks, Unmanaged.passUnretained(client).toOpaque())
+            mongoc_apm_callbacks_destroy(callbacks)
+            return pool
+        }
+
+        // this shouldn't be reached but just in case we still free the memory
+        newFut.whenFailure { _ in
+            mongoc_apm_callbacks_destroy(callbacks)
+        }
+
+        // update the stored future to refer to the one that sets the callbacks
+        self.state = .opening(future: newFut)
     }
 }
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -1,5 +1,6 @@
 import CLibMongoC
 import Foundation
+import NIO
 import NIOConcurrencyHelpers
 
 /// A connection to the database.
@@ -39,6 +40,8 @@ extension NSCondition {
 internal class ConnectionPool {
     /// Represents the state of a `ConnectionPool`.
     private enum State {
+        /// Indicates the `ConnectionPool` is still starting up in the background.
+        case opening(future: EventLoopFuture<OpaquePointer>)
         /// Indicates that the `ConnectionPool` is open and using the associated pointer to a `mongoc_client_pool_t`.
         case open(pool: OpaquePointer)
         /// Indicates that the `ConnectionPool` is in the process of closing. Connections can be checked back in, but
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -75,23 +78,25 @@ internal class ConnectionPool {
...
 
     /// Initializes the pool using the provided `ConnectionString`.
-    internal init(from connString: ConnectionString) throws {
-        let pool: OpaquePointer = try connString.withMongocURI { uriPtr in
-            guard let pool = mongoc_client_pool_new(uriPtr) else {
-                throw MongoError.InternalError(message: "Failed to initialize libmongoc client pool")
-            }
-            return pool
-        }
+    internal init(from connString: ConnectionString, executor: OperationExecutor) throws {
```
patrickfreed, line 81: It feels kinda weird to pass the executor in like this, but I wasn't exactly sure how to do this otherwise while still keeping the the pool creation being blocking encapsulated within `ConnectionPool`.

kmahar: yeah, I was going to say make it an `Operation` but then it would need a `Connection` to be executed so of course that doesn't work. it's weird but I'm not sure what else to do really.

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -101,13 +106,36 @@ internal class ConnectionPool {
...
+    /// closure MUST NOT attempt to acquire the lock.
+    ///
+    /// Throws an error if the pool is closed or closing.
+    private func withResolvedPool<T>(body: (OpaquePointer) throws -> T) throws -> T {
+        try self.stateLock.withLock {
+            switch self.state {
+            case let .open(pool):
+                return try body(pool)
+            case let .opening(future):
+                let pool = try future.wait()
```
kmahar, line 120: is it possible that we encounter a race condition where a user creates a client and immediately fires off [threadPoolSize] operations, those somehow manage to get scheduled on the thread pool before the pool creation does, and then all of those are stuck blocking on this future which can't complete? 🤔

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -230,12 +250,22 @@ internal class ConnectionPool {
...
+    /// This method takes ownership of the provided `mongoc_apm_callbacks_t`, so the pointer MUST NOT be used or freed
+    /// after it is passed to this method.
     internal func setAPMCallbacks(callbacks: OpaquePointer, client: MongoClient) {
         // lock isn't needed as this is called before pool is in use.
         switch self.state {
-        case let .open(pool):
-            mongoc_client_pool_set_apm_callbacks(pool, callbacks, Unmanaged.passUnretained(client).toOpaque())
-        case .closing, .closed:
+        case let .opening(future):
+            future.whenSuccess { pool in
```
kmahar, line 260: I'm a little worried about this. could we end up with an interleaving like
1) `future` finishes
2) `future.wait()` is unblocked, so a connection can be checked out above, and is immediately
3) the event loop gets around to executing this callback, but the checked-out connection doesn't properly get the APM callbacks set on it

?

Sources/MongoSwift/ConnectionPool.swift
patrickfreed: Assuming the thread pool operates on a first-come-first-serve basis, the pool creation should always get scheduled first since it happens before the `ConnectionPool` initializer even returns. If there aren't any ordering guarantees then it seems that type of race could be possible.

From looking at the implementation of `runIfActive`, the pool acquires a lock and then appends the work item to a queue and then returns. It looks like the actual worker threads then take work items from the front of the queue, so it seems the first-come-first-serve ordering is preserved here. Good idea though, this could have been a serious issue if the pool wasn't ordered.

Sources/MongoSwift/ConnectionPool.swift
patrickfreed: Great catch, this definitely seems like it could be an issue. I updated `setAPMCallbacks` to use `map` instead of `whenSuccess` to return a new future. Now the calls to `wait` will be on the mapped future and therefore should complete after the callbacks are actually set on the pool.

kmahar (approved):
nit about comment but LGTM otherwise

Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -230,15 +250,31 @@ internal class ConnectionPool {
...
+
+        // to ensure the callbacks get set before any waiting on this future becomes unblocked, use
+        // `map` to create a new future instead of .whenSuccess
+        let newFut = future.map { pool -> OpaquePointer in
+            mongoc_client_pool_set_apm_callbacks(pool, callbacks, Unmanaged.passUnretained(client).toOpaque())
+            mongoc_apm_callbacks_destroy(callbacks)
+            return pool
+        }
+
+        // this shouldn't be reached but just in case we still free the memory
```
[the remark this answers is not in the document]
patrickfreed: We do currently throw an `InternalError` when `mongoc_client_pool_new` returns nil, but I don't think that function is actually failable bar an out-of-memory scenario. I wonder if we should just instead `fatalError` there so the initializer isn't `throws` and this confusion could be avoided.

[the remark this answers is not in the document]
kmahar: oh you're right; I was thinking that the connection pool initializer handled URI parsing, but that is done by `ConnectionString`. 

it does looks like `mongoc_client_pool_new` can also return nil if the URI has TLS specified but libmongoc was not built with SSL support: https://github.com/mongodb/mongo-c-driver/blob/107d094ba28ae793308e7490f2cbc1eef9eb3419/src/libmongoc/src/mongoc/mongoc-client-pool.c#L111

 that said... we appear to hard-code `MONGOC_ENABLE_SSL` to 1:
https://github.com/mongodb/mongo-swift-driver/blob/858b6d08078deaeb0b2b6c17889514a71a1dc061/etc/generated_headers/mongoc-config.h#L122

so I guess that check would never be hit even if e.g. OpenSSL is missing. that might be problematic itself but I guess does mean we could assume there are no nulls returned.... on the other hand it's possible this method could start returning null in other error cases so perhaps it's better to remain defensive.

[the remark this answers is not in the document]
patrickfreed: Ok sounds good, I'll leave the `throw` in there then.

PR #597, opened 2021-02-02
SWIFT-1056 Update Vapor example to use new library

kmahar:
Putting up for discussion/feedback, but this shouldn't actually be merged until we've released the new library.

Sorry the README diff is a little weird, it's comparing it to the original README rather than the original state of the complex example's README


Examples/VaporExample/Sources/Run/main.swift
```diff
@@ -1,16 +1,16 @@
...
-import MongoSwift
+import MongoDBVapor
 import Vapor
 
 var env = try Environment.detect()
 try LoggingSystem.bootstrap(from: &env)
 
 let app = Application(env)
 try configure(app)
+try app.mongoDB.configure("mongodb://localhost:27017")
```
kmahar, line 10: while the parameter isn't required, it seemed useful to me to show it here to make it more clear to a user what they should change if they aren't using the default host/port.

Examples/VaporExample/README.md
```diff
@@ -1,13 +1,70 @@
...
+with the `kittenCollection` property:
+```swift
+extension Request {
+    /// Convenience extension for obtaining a collection which uses the same event loop as a request.
+    var kittenCollection: MongoCollection<Kitten> {
+        self.mongoDB.client.db("home").collection("kittens", withType: Kitten.self)
+    }
+}
+```
+### Codable Usage
```
kmahar, line 54: one thing to note is that our existing example doesn't make use of the ExtendedJSONEncoder/ExtendedJSONDecoder as ContentEncoder/ContentDecoder feature. 

It seems good to demo it though it's not something all users actually need to do, e.g. this app works fine without it, so I wasn't totally sure whether to do it. We could make this app need the feature by adding some field to `Kitten` that is not serializable by `JSONEncoder` if we want.

patrickfreed: Yeah I think that's a good idea. I think it's important we indicate these encoders / decoders exist, what they're used for, and how to enable them (both in example and readme), since otherwise I imagine most users will just not notice or use them at all.

Examples/VaporExample/README.md
```diff
@@ -1,13 +1,70 @@
-# VaporExample
```
patrickfreed, line 1: I think a lot of this text would be really good to include in the Vapor extension README too, since that's likely another visible place for it. Maybe we ought to have an abbreviated version of this README with small code snippets instead of a full example and then link to this one for the more detailed example?

kmahar (commented):
made various updates per our conversations, there are now separate web and REST endpoints and extJSON is used everywhere

Examples/VaporExample/Resources/Views/index.leaf
```diff
@@ -55,8 +57,38 @@
...
+  $(document).ready(() => {
+    $("\#add-btn").click(() => {
+      var name = $("\#name").val();
+      var color = $("\#color").val();
+      var food = $("\#favoriteFood").val();
+      // Serialize request data to extended JSON.
+      var data = BSON.EJSON.stringify({
+        'name': name,
+        'color': color,
+        'favoriteFood': food
```
patrickfreed, line 74: Would it be worth including like a "last updated date" or "birth date" or something here to demonstrate why the extended json is useful? these types are all strings so I think the output is pretty much the same with either encoder. If we do go that route, it might make sense to call it out specifically above that ext JSON is what ensures our dates get moved from JS frontend <-> Swift backend <-> MongoDB seamlessly.

Examples/VaporExample/Resources/Views/index.leaf
```diff
@@ -14,6 +14,8 @@
       font-family: "Trebuchet MS", Helvetica, sans-serif;
     }
   </style>
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+  <script src="../bson.bundle.js"></script>
```
patrickfreed, line 18: I assume there's nowhere this is hosted publicly like jquery is to avoid having to check it into the repo?

patrickfreed (commented):
example looks good, tried it out locally and seems to work perfectly. Just have one suggestion and one question

Examples/VaporExample/README.md
```diff
@@ -1,13 +1,107 @@
...
+The application has both web and REST API interfaces, which support storing a list of kittens and details about them.
+
+The server will handle the following types of web requests:
+1. A GET request at the root URL `/` loads the main index page containing a list of kittens.
+1. A GET request at the URL `/kittens/{name}` loads a web page with information about the kitten with the specified name.
+
+And the following types of API requests:
+1. A GET request at the URL `/rest` returns a list of kittens.
+1. A POST request at the URL `/rest` adds a new kitten.
+1. A GET request at the URL `rest/kittens/{name}` returns information about the kitten with the specified name.
```
patrickfreed, line 35: a few of these are missing the leading `/`

patrickfreed (commented):
Looks really good! I only have two really minor documentation nits but otherwise LGTM

Examples/VaporExample/Resources/Views/kitten.leaf
```diff
@@ -74,8 +75,18 @@ $(document).ready(() => {
     $.ajax({
         type:'PATCH',
         contentType : 'application/json',
-        url:'/kittens/#(name)',
-        data: JSON.stringify({"favoriteFood": newFood}),
+        url:'/rest/kittens/#(name)',
+        // Serialize request data to extended JSON. We use extended JSON
+        // both here, and for serializing/deserializing request data on
+        // the frontend, in order to ensure all MongoDB type information
```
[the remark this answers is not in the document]
kmahar: good catch, I'd copy pasted from the corresponding backend comment, fixed here and in index.leaf

Examples/VaporExample/README.md
kmahar: fixed

patrickfreed (approved):
LGTM!

kmahar:
(going to wait to merge until we tag a release so I can update Package.swift here accordingly)

kmahar:
@patrickfreed FYI updated this example to use js-bson from jsdelivr, and also in developing the template I realized having an `IndexContext` type was unnecessary so I removed that

PR #598, opened 2021-02-03
SWIFT-1086 Update sortedEquals impl to use equalsIgnoreKeyOrder

patrickfreed:
SWIFT-1086 (driver portion)

### DIFF
--- a/Sources/TestsCommon/CommonTestUtils.swift
+++ b/Sources/TestsCommon/CommonTestUtils.swift
@@ -271,15 +271,7 @@ public protocol SortedEquatable {
 
 extension BSONDocument: SortedEquatable {
     public func sortedEquals(_ other: BSONDocument) -> Bool {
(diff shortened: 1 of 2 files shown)


Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -271,15 +271,7 @@ public protocol SortedEquatable {
...
-        }
-        for (k, v) in self {
-            guard let otherValue = other[k], v.sortedEquals(otherValue) else {
-                return false
-            }
-        }
-        return true
+        self.equalsIgnoreKeyOrder(other)
     }
 }
```
patrickfreed, line 277: the `BSON` portion (just below this line) still needs to be duplicated since we don't export that from the BSON library. I think that's fine since it's simple enough, and I don't think the method makes all that much sense on `BSON` outside of the contexts we use it in to include in the public API.

kmahar (approved):

PR #599, opened 2021-02-04
Fix txn tests

kmahar:
Fixes the transactions test failures by using separate clients for running specific commands against each mongos. As discussed in #driver-devs, this accounts for cases where the setup client has not yet reconnected to one of the mongoses after a test.

I thought about porting this over to the unified runner, but that one is fairly specific that the internal client should be used for terminating open transactions and doesn't appear to be suffering from this issue on its own. It may become necessary in the future if any similarly racy tests start using that runner.

Evg patch against all sharded variants (realized those < 4.2 were useless too late): https://evergreen.mongodb.com/version/601c817c850e612ab835c3c9

no macOS results in yet but as you can see this has fixed all the Ubuntu failures anyway.

### DIFF
--- a/Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
+++ b/Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
@@ -182,7 +182,10 @@ extension SpecTestFile {
         }
     }
 
-    func terminateOpenTransactions(using client: MongoSwiftSync.MongoClient, mongosClients: [MongoSwiftSync.MongoClient]?) throws {
+    func terminateOpenTransactions(
+        using client: MongoSwiftSync.MongoClient,
+        mongosClients: [MongoSwiftSync.MongoClient]?
+    ) throws {
         // if connected to a replica set, use the provided client to execute killAllSessions on the primary.
         // if connected to a sharded cluster, use the per-mongos clients to execute killAllSessions on each mongos.
         switch try client.topologyType() {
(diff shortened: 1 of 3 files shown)


kmahar:
update: there were a few macOS failures but they are all due to the libmongoc crashing on sigpipe issue so those will be fixed by vendoring in 1.17.4.

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -229,25 +231,36 @@ extension SpecTestFile {
...
             }
-
-            try self.terminateOpenTransactions(using: setupClient)
+            try self.terminateOpenTransactions(using: setupClient, mongosClients: mongosClients)
             try self.populateData(using: setupClient)
 
             // Due to strange behavior in mongos, a "distinct" command needs to be run against each mongos
             // before the tests run to prevent certain errors from ocurring. (SERVER-39704)
-            if MongoSwiftTestCase.topologyType == .sharded,
+            if [.sharded, .shardedReplicaSet].contains(topologyType),
```
patrickfreed, line 258: Since this will likely be quite common, could we add an `isSharded` (or something similar) method to `TopologyType`?

patrickfreed (approved):
one nit/suggestion but otherwise LGTM (don't need to review again).

great job hunting down these failures!

Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
kmahar: nice idea, done ✅

PR #600, opened 2021-02-05
SWIFT-734 Update gh-pages to versioned setup

patrickfreed:
SWIFT-734

This PR updates the `gh-pages` branch based on the changes to the documentation generation and release scripts. The setup is modeled after [SwiftNIO's documentation layout](https://github.com/apple/swift-nio/tree/gh-pages), with the addition of an index that links to all the available versions. Once this PR goes in, we need to update the GitHub pages path to just be the root, and going forward `/docs` will be where the index lives. `/` will still redirect to the current latest release docs so our old link won't break.

The diff in this PR isn't going to be super useful, instead I'd recommend just looking at the [branch on my fork](https://github.com/patrickfreed/mongo-swift-driver/tree/SWIFT-734/gh-pages-update).


kmahar:
I'm probably missing something / I don't quite remember how the pages stuff works, but does the markdown index.md just automatically get turned into a web page somehow by pages? 
if yes this LGTM

patrickfreed:
Yep! It uses something called jekyll to do it. https://docs.github.com/en/github/working-with-github-pages/about-github-pages-and-jekyll

patrickfreed:
I realized I had a mistake in #601 where I had the python script to update the index on the main branch, but it needs to be in `gh-pages` since it builds the index based on the directory layout. I've committed it here under `_scripts/update-index.py` (starts with underscore so jekyll doesn't pick it up). Will follow up with a commit to fix master afterwards.

PR #601, opened 2021-02-05
SWIFT-734 Update scripts to maintain multiple versions of the documentation

patrickfreed:
SWIFT-734

This PR updates the release and documentation scripts to generate the docs in a way that works for multiple versions of the driver, based on [SwiftNIO's setup](https://github.com/apple/swift-nio/tree/gh-pages). See #600 for the changes it makes to the `gh-pages` branch (or rather, [check out the branch of my fork](https://github.com/patrickfreed/mongo-swift-driver/tree/SWIFT-734/gh-pages-update), since that diff is crazy).

### DIFF
--- a/etc/generate-docs.sh
+++ b/etc/generate-docs.sh
@@ -23,14 +23,15 @@ jazzy_args=(--clean
             --module-version "${version}")
 
 # Generate MongoSwift docs
+sourcekitten doc --spm --module-name MongoSwift > mongoswift-docs.json
 args=("${jazzy_args[@]}"  --output "docs-temp/MongoSwift" --module "MongoSwift" --config ".jazzy.yml" 
+        --sourcekitten-sourcefile mongoswift-docs.json
         --root-url "https://mongodb.github.io/mongo-swift-driver/docs/MongoSwift/")
 jazzy "${args[@]}"
 
 # Generate MongoSwiftSync docs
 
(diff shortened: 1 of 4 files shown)


etc/generate-docs.sh
```diff
@@ -23,14 +23,15 @@ jazzy_args=(--clean
             --module-version "${version}")
 
 # Generate MongoSwift docs
+sourcekitten doc --spm --module-name MongoSwift > mongoswift-docs.json
```
patrickfreed, line 26: this was required to make the script work on Linux.

etc/update-index.py
```diff
@@ -0,0 +1,17 @@
+import os
```
patrickfreed, line 1: I really loathe bash scripting (and I can't write it either) so I just made another python script.

kmahar (approved):
nice, LGTM! 

this got me thinking about getting the BSON lib symbols to show up in these docs going forward... I think we discussed this but don't think we have a ticket about it 🤔 opened [SWIFT-1113](https://jira.mongodb.org/browse/SWIFT-1113)

patrickfreed:
Good point, also I need to port this setup over to the BSON library too, since it probably makes sense for it to host its own docs as well. I'll do a follow up PR in `swift-bson` after this.

PR #602, opened 2021-02-05
SWIFT-1092 Vendor in libmongoc 1.17.4

kmahar:
Vendors in the latest libmongoc version.

all green minus the random failure due to what we discussed in the channel. I'm discussing now with folks in the unified test runner channel how to fix that, but I don't think we need to block the upgrade on that

https://evergreen.mongodb.com/version/601cc5843066152b49a5bc72

### DIFF
--- a/Sources/CLibMongoC/mongoc/mongoc-socket.c
+++ b/Sources/CLibMongoC/mongoc/mongoc-socket.c
@@ -993,6 +993,9 @@ mongoc_socket_new (int domain,   /* IN */
 #else
    int sd;
 #endif
+#ifdef SO_NOSIGPIPE
+   int on = 1;
+#endif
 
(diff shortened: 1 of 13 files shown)


Sources/CLibMongoC/mongoc/mongoc-socket.c
```diff
@@ -1017,6 +1020,12 @@ mongoc_socket_new (int domain,   /* IN */
       _mongoc_socket_setkeepalive (sd);
    }
 
+   /* Set SO_NOSIGPIPE, to ignore SIGPIPE on writes for platforms where
```
kmahar, line 1023: this is the fix for the macOS crashes

patrickfreed (approved):

PR #604, opened 2021-02-10
SWIFT-1061 Update r1.0 branch to support matrix testing

patrickfreed:
SWIFT-1061

This PR makes the following changes to the r1.0 branch to support continuous matrix testing:
- Backport the xunit test results parser for pretty test output (considered a SHOULD in the matrix project)
- Allow optional installation of dependencies in `run-tests.sh` via the `INSTALL_DEPS` variable
- Introduce a workaround to `install-dependencies.sh` to avoid a clash between matrix testsing's `$PLATFORM` and [the `$PLATFORM` used in swiftenv](https://github.com/mbroadst/swiftenv/blob/osx-install-path/libexec/swiftenv-install#L44)
- complete SWIFT-763 to fix failing test on latest (will need to be "forward"-ported to `main` as well)

Example matrix patch: https://spruce.mongodb.com/version/60246becd6d80a75c95f89f1/tasks
Example r1.0 patch: https://spruce.mongodb.com/version/602475c03627e06ab8f78190/tasks

### DIFF
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -9,6 +9,10 @@ INSTALL_DIR="${PROJECT_DIRECTORY}/opt"
 BUILD_DIR="${PROJECT_DIRECTORY}/libmongoc-build"
 EVG_DIR=$(dirname $0)
 
+# this is set by drivers-matrix-testing, and it's a special variable used in swiftenv
+# leaving it set messes with the installation
+unset PLATFORM
+
 export SWIFTENV_ROOT="${INSTALL_DIR}/swiftenv"
 export PATH=/opt/cmake/bin:${SWIFTENV_ROOT}/bin:$PATH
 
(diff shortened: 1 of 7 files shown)


.evergreen/install-dependencies.sh
```diff
@@ -9,6 +9,10 @@ INSTALL_DIR="${PROJECT_DIRECTORY}/opt"
 BUILD_DIR="${PROJECT_DIRECTORY}/libmongoc-build"
 EVG_DIR=$(dirname $0)
 
+# this is set by drivers-matrix-testing, and it's a special variable used in swiftenv
+# leaving it set messes with the installation
+unset PLATFORM
```
patrickfreed, line 14: this shouldn't affect the test suite if it needs access to `$PLATFORM`, since we run `install-dependencies.sh` in its own shell

kmahar (approved):

PR #605, opened 2021-02-11
SWIFT-1034 Add sorted find example to README

patrickfreed:
SWIFT-1034

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,9 @@ print(result?.insertedID ?? "") // prints `.int64(100)`
 **Async**:
 ```swift
 let query: BSONDocument = ["a": 1]
+// The `sort` option specifies the order in which results are returned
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,7 +137,8 @@ print(result?.insertedID ?? "") // prints `.int64(100)`
 **Async**:
 ```swift
 let query: BSONDocument = ["a": 1]
-let result = collection.find(query).flatMap { cursor in
+let options = FindOptions(sort: ["_id": -1])
```
kmahar, line 140: I don't think it would be obvious from context what this is doing. maybe we could add a comment or something?

patrickfreed: done. I figure these examples will largely be used for copy/paste, so I included the comment in both examples to ensure it gets picked up regardless of how a user tries out the driver.

kmahar (approved):

PR #607, opened 2021-02-15
SWIFT-1113 Include BSON symbols in driver docs

kmahar:
This prepares us for the 1.1.0 release where the BSON symbols are no longer available. You can try this out locally with the docs generation script.

### DIFF
--- a/etc/generate-docs.sh
+++ b/etc/generate-docs.sh
@@ -18,6 +18,12 @@ version=${1}
 # Ensure version is non-empty
 [ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
 
+swift build
+working_dir=${PWD}
+cd .build/checkouts/swift-bson
+sourcekitten doc --spm --module-name SwiftBSON > ${working_dir}/bson-docs.json
+cd $working_dir
+
 jazzy_args=(--clean
             --github-file-prefix https://github.com/mongodb/mongo-swift-driver/tree/v${version} 
             --module-version "${version}")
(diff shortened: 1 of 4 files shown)


etc/get_bson_version.py
```diff
@@ -0,0 +1,6 @@
+import json
+
+with open('Package.resolved', 'r') as f:
```
kmahar, line 3: JSON parsing is where I draw the line for what I will try to do in a bash script lol

etc/generate-docs.sh
```diff
@@ -18,14 +18,20 @@ version=${1}
 # Ensure version is non-empty
 [ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
 
+swift build # build to ensure we have a local copy of swift-bson available
+working_dir=${PWD}
+cd .build/checkouts/swift-bson
+sourcekitten doc --spm --module-name SwiftBSON > ${working_dir}/bson-docs.json
```
patrickfreed, line 24: So at least on Linux, `sourcekitten` already performs a `swift build` as part of documenting, and in fact, this step actually fails for me during the `swift build` stage saying "invalidAccess" to the `Package.resolved` of the the `swift-bson` in checkouts. I moved this step until after the `sourcekitten` invocation below for `MongoSwift` and removed the extra `swift build` from above, and then it seemed to work okay.

kmahar: so I got that error one time during development but figured it was just like a weird transient git issue or something as I didn't see it again. I just tried your suggestion to move this step down and ran the script right after deleting `.build`, and I get the error still. if I re-run, it works.... maybe that's why it worked for you, since you had already tried it once? 

using the version in `.build` is a bit of a hack anyway. I went with that approach before realizing I was going to need to do some work to figure out the pinned-to version anyway below. so maybe I should just clone the BSON repo, checkout that tag and run the command fresh in there.

patrickfreed: yeah the cloning seems like its probably more straightforward.

etc/generate-docs.sh
```diff
@@ -18,14 +18,20 @@ version=${1}
 # Ensure version is non-empty
 [ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
 
+swift build # build to ensure we have a local copy of swift-bson available
```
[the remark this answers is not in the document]
kmahar: updated to do so at the start of the script

etc/generate-docs.sh
kmahar: switched to that approach. seems to work ok for me, let me know if you run into any issues

patrickfreed (approved):
LGTM!

PR #611, opened 2021-03-04
Guide reorganization

kmahar:
Doing the Vapor integration docs work I realized that right now we don't properly include guides from the BSON library in the driver docs.
That seems useful, so I updated the script and Jazzy config to use the guides from both repos together.

As part of this I also split the driver's own BSON guide into two separate smaller guides.

(a follow up PR in the Vapor repo will use this same approach to bring in guides from both driver and BSON)


patrickfreed (approved):
Great idea, LGTM!

PR #612, opened 2021-03-15
SWIFT-904 Add connectTimeoutMS option to MongoClientOptions

bynn:
SWIFT-904 Add connectTimeoutMS option to MongoClientOptions

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -98,7 +98,7 @@ internal class ConnectionString {
                 message: "Invalid \(MONGOC_URI_CONNECTTIMEOUTMS): must be between 1 and \(Int32.max)"
             )
         }
-        if let socketTimeoutMS = self.options?[MONGOC_URI_SOCKETTIMEOUTMS]?.int32Value, socketTimeoutMS < 0 {
+        if let socketTimeoutMS = self.options?[MONGOC_URI_SOCKETTIMEOUTMS]?.int32Value, socketTimeoutMS < 1 {
             throw MongoError.InvalidArgumentError(
                 message: "Invalid \(MONGOC_URI_SOCKETTIMEOUTMS): must be between 1 and \(Int32.max)"
             )
@@ -295,6 +295,19 @@ internal class ConnectionString {
             }
 
             try self.setInt32Option(MONGOC_URI_MAXPOOLSIZE, to: value)
+        }
+
+        if let connectTimeoutMS = options?.connectTimeoutMS {
+            guard let value = Int32(exactly: connectTimeoutMS), value > 0 else {
+                throw self.int32OutOfRangeError(
+                    option: MONGOC_URI_CONNECTTIMEOUTMS,
+                    value: connectTimeoutMS,
+                    min: 1,
+                    max: Int32.max
+                )
+            }
+
+            try self.setInt32Option(MONGOC_URI_CONNECTTIMEOUTMS, to: value)
         }
 
         let unsupportedOptions = [
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -14,6 +14,10 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.compressors
     public var compressors: [Compressor]?
 
+    /// Specifies the maximum time, in milliseconds, for an individual connection to establish a TCP
+    /// connection to a MongoDB server before timing out.
+    public var connectTimeoutMS: Int?
+
     /// Specifies authentication options for use with the client.
     public var credential: MongoCredential?
 
@@ -131,6 +135,7 @@ public struct MongoClientOptions: CodingStrategyProvider {
     public init(
         appName: String? = nil,
         compressors: [Compressor]? = nil,
+        connectTimeoutMS: Int? = nil,
         credential: MongoCredential? = nil,
         dataCodingStrategy: DataCodingStrategy? = nil,
         dateCodingStrategy: DateCodingStrategy? = nil,
@@ -157,6 +162,7 @@ public struct MongoClientOptions: CodingStrategyProvider {
     ) {
         self.appName = appName
         self.compressors = compressors
+        self.connectTimeoutMS = connectTimeoutMS
         self.credential = credential
         self.dataCodingStrategy = dataCodingStrategy
         self.dateCodingStrategy = dateCodingStrategy
--- a/Sources/TestsCommon/CommonTestUtils.swift
+++ b/Sources/TestsCommon/CommonTestUtils.swift
@@ -216,6 +216,13 @@ public struct TestRequirement: Decodable {
         )
     ]
 
+    public static let blockTimeSupport: [TestRequirement] = [
+        TestRequirement(
+            minServerVersion: ServerVersion.mongodBlockTimeSupport,
+            acceptableTopologies: [.single, .replicaSet]
+        )
+    ]
+
     public init(
         minServerVersion: ServerVersion? = nil,
         maxServerVersion: ServerVersion? = nil,
--- a/Sources/TestsCommon/CommonTestUtils.swift
+++ b/Sources/TestsCommon/CommonTestUtils.swift
@@ -218,8 +218,7 @@ public struct TestRequirement: Decodable {
 
     public static let blockTimeSupport: [TestRequirement] = [
         TestRequirement(
-            minServerVersion: ServerVersion.mongodBlockTimeSupport,
-            acceptableTopologies: [.single, .replicaSet]
+            minServerVersion: ServerVersion.mongodBlockTimeSupport
         )
     ]
 
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -72,6 +72,7 @@ extension ConnectionStringTests {
         ("testServerSelectionTimeoutMS", testServerSelectionTimeoutMS),
         ("testServerSelectionTimeoutMSWithCommand", testServerSelectionTimeoutMSWithCommand),
         ("testLocalThresholdMSOption", testLocalThresholdMSOption),
+        ("testConnectTimeoutMSOption", testConnectTimeoutMSOption),
         ("testUnsupportedOptions", testUnsupportedOptions),
         ("testCompressionOptions", testCompressionOptions),
         ("testInvalidOptionsCombinations", testInvalidOptionsCombinations),
@@ -342,6 +343,7 @@ extension SyncMongoClientTests {
         ("testClientLifetimeManagement", testClientLifetimeManagement),
         ("testAPMCallbacks", testAPMCallbacks),
         ("testCertificateVerificationOptions", testCertificateVerificationOptions),
+        ("testConnectionTimeout", testConnectionTimeout),
     ]
 }
 
--- a/Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
+++ b/Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
@@ -166,6 +166,7 @@ internal struct FailPoint: Decodable {
     public static func failCommand(
         failCommands: [String],
         mode: Mode,
+        blockTimeMS: Int? = nil,
         closeConnection: Bool? = nil,
         errorCode: Int? = nil,
         errorLabels: [String]? = nil,
@@ -174,6 +175,10 @@ internal struct FailPoint: Decodable {
         var data: BSONDocument = [
             "failCommands": .array(failCommands.map { .string($0) })
         ]
+        if let blockTime = blockTimeMS {
+            data["blockTimeMS"] = BSON(blockTime)
+            data["blockConnection"] = true
+        }
         if let close = closeConnection {
             data["closeConnection"] = .bool(close)
         }
--- a/Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
+++ b/Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
@@ -320,4 +320,30 @@ final class SyncMongoClientTests: MongoSwiftTestCase {
         expect(try MongoClient("mongodb://localhost:12345", options: options))
             .to(throwError(errorType: MongoError.InvalidArgumentError.self))
     }
+
+    func testConnectionTimeout() throws {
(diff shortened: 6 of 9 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -14,6 +14,10 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.compressors
     public var compressors: [Compressor]?
 
+    /// Specifies the time, in milliseconds, for an individual connection from your connection pool to establish a TCP
+    /// connection to the MongoDB server before timing out.
+    public var connectTimeoutMS: ChannelOptions.Types.ConnectTimeoutOption?
```
patrickfreed, line 19: I like the idea of using a NIO type for this, but I think for consistency with our other timeout options we'll need to continue to use just `Int` here. In 2.0 (the pure swift version), I think we'd probably want to update these various timeouts to use such a type, though I think we might opt for the more general `TimeAmount` instead, which `ChannelOptions.Types.ConnectTimeoutOption` is basically an alias for. Could you file a ticket to remind us to do this in 2.0?

BTW, the reason we have to wait for 2.0 to update the types is that the driver follows [Semantic Versioning](https://semver.org/), and since changing the type of existing options would constitute a breaking change, we'd only be able to introduce it in a new major version. We're currently not planning on doing another major version release until the new driver is done, though.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -14,6 +14,10 @@ public struct MongoClientOptions: CodingStrategyProvider {
...
+
```
patrickfreed, line 20: In order for this option to be used by the driver, we'll need to pass it in to libmongoc. The way we do that is via the various `applyAndValidateXOptions` methods in `ConnectionString.swift`. For this option in particular, we'll want to add logic into `applyAndValidateConnectionPoolOptions`.

Once that's done, can we add a test to verify that the setting works? We'll probably want to put it into `SyncMongoClientTests.swift` (so that it will test both sync + async API).

An example test could be something like:
1. enable a [failCommand failpoint](https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point) on the `isMaster` command with blockTimeMS = 500
    - This will ensure that a connection attempt takes at least 500ms, since the `isMaster` command is always run as part of the handshake for a new connection
2. create a client with `connectTimeoutMS` = 100
3. try to run a `ping` command and assert that a failure occurs
4. create a client with `connectTimeoutMS` = 1000
5. try to run a `ping` command and assert that it succeeds

For an example of a test that makes use of a failpoint, check out [`testCursorTerminatesOnError`](https://github.com/mongodb/mongo-swift-driver/blob/main/Tests/MongoSwiftSyncTests/MongoCursorTests.swift#L243).

In order to set failpoints, be sure to start your `mongod` locally with the `--setParameter enableTestCommands=1` option.

Sources/MongoSwift/MongoClient.swift
```diff
@@ -14,6 +14,10 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.compressors
     public var compressors: [Compressor]?
 
+    /// Specifies the time, in milliseconds, for an individual connection from your connection pool to establish a TCP
+    /// connection to the MongoDB server before timing out.
```
patrickfreed, line 18: Some docs nits:
- can we add "maximum" before "time" here?
- can we change "the" (in "the MongoDB server") to "a"?
- this value actually also applies to monitoring connections (i.e. connections not included in the pools), so can we drop the "from your connection pool"?

patrickfreed (commented):
Looks like a good start so far!

A few git related comments:
- you can delete the default checklist in the PR description. We have that here to ensure community members take a few steps before opening PRs, but we don't usually include it ourselves. Instead you can just have the ticket number and a brief description of the changes in the PR, if necessary.
- [Make sure to update the email in your local git client](https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git) to be the same as the one in your GitHub account so GitHub can associate the commits with you properly.

Sources/MongoSwift/MongoClient.swift
bynn: done

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -297,6 +297,19 @@ internal class ConnectionString {
             try self.setInt32Option(MONGOC_URI_MAXPOOLSIZE, to: value)
         }
 
+        if let connectTimeoutMS = options?.connectTimeoutMS {
+            guard let value = Int32(exactly: connectTimeoutMS), value > 0 else {
```
kmahar, line 301: I think 0 is actually a valid value for this option, according to the [URI options spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#list-of-specified-options):
> non-negative integer; 0 means "no timeout"

relatedly, for a lot of these options we have various tests in `ConnectionStringTests.swift` which verify that we e.g. error on invalid values, correctly prioritize options provided via `MongoClientOptions` over those provided via the connection string itself. I think it would be good to add some of those for this option; you might take some inspiration from e.g. `testLocalThresholdMSOption`

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
```diff
@@ -320,4 +320,35 @@ final class SyncMongoClientTests: MongoSwiftTestCase {
...
+
+        var succeedOptions = MongoClientOptions()
+        succeedOptions.connectTimeoutMS = 1000
+        succeedOptions.serverSelectionTimeoutMS = 1000
+
+        let setupClient = try MongoClient.makeTestClient()
+        let timeoutClient = try MongoClient.makeTestClient(options: failOptions)
+        let succeedClient = try MongoClient.makeTestClient(options: succeedOptions)
+
+        guard try setupClient.supportsFailCommand() else {
```
kmahar, line 337: nit: I would put the setupClient initialization and check for failCommand support first in this test before the rest of the code, so we can "fail fast"/exit quickly and the exit conditions are easier to spot at a quick glance.

kmahar (changes requested):
looks great! some minor comments around an edge case and testing.

kmahar:
Oh, also BTW since you added a new test, you need to run `make linuxmain` from the root directory to re-generate the list of tests in `LinuxMain.swift`, which is required for testing on Linux.

Tests/MongoSwiftSyncTests/SpecTestRunner/FailPoint.swift
```diff
@@ -166,6 +166,8 @@ internal struct FailPoint: Decodable {
     public static func failCommand(
         failCommands: [String],
         mode: Mode,
+        blockConnection: Bool? = nil,
+        blockTimeMS: Int? = nil,
```
patrickfreed, line 170: Given that we'll always want `blockConnection` to be true if we've specified a non-nil `blockTimeMS`, can we drop the `blockConnection` argument to this method and just set it true if `blockTimeMS` is specified?

Sources/MongoSwift/MongoClient.swift
```diff
@@ -14,6 +14,10 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/#urioption.compressors
     public var compressors: [Compressor]?
 
+    /// Specifies the maximum time, in milliseconds, for an individual connection  to establish a TCP
```
patrickfreed, line 17:
```suggestion
    /// Specifies the maximum time, in milliseconds, for an individual connection to establish a TCP
```

patrickfreed (commented):
Looking good, just a few minor suggestions. BTW, it seems like your email still isn't set. Once it is, your profile picture / GitHub should start showing up next to your commits instead of the default one.

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
```diff
@@ -320,4 +320,30 @@ final class SyncMongoClientTests: MongoSwiftTestCase {
...
+        var failOptions = MongoClientOptions()
+        failOptions.connectTimeoutMS = 100
+        failOptions.serverSelectionTimeoutMS = 1000
+        let timeoutClient = try MongoClient.makeTestClient(options: failOptions)
+
+        let fp = FailPoint.failCommand(failCommands: ["isMaster"], mode: .alwaysOn, blockTimeMS: 500)
+        try fp.enable()
+        defer { fp.disable() }
+
+        expect(try timeoutClient.db("admin").runCommand(["ping": 1])).to(throwError())
```
kmahar, line 340: there looks to be an issue on 4.0 on Evergreen. it looks like blockConnection actually isn't supported on server versions < 4.2, so we should skip this test on those:
https://jira.mongodb.org/browse/SERVER-41070

Tests/LinuxMain.swift
```diff
@@ -1,7 +1,6 @@
-// Generated using Sourcery 1.0.0 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 1.3.4 — https://github.com/krzysztofzablocki/Sourcery
```
kmahar, line 1: the sourcery task is going to fail on evg because of this; once my PR #614 is merged in you can rebase on top of it and that should fix it.

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -93,14 +93,14 @@ internal class ConnectionString {
         }
 
         // libmongoc does not validate negative timeouts in the input string, so do it ourselves here.
-        if let connectTimeoutMS = self.options?[MONGOC_URI_CONNECTTIMEOUTMS]?.int32Value, connectTimeoutMS < 1 {
+        if let connectTimeoutMS = self.options?[MONGOC_URI_CONNECTTIMEOUTMS]?.int32Value, connectTimeoutMS < 0 {
```
kmahar, line 96: oh no.... I'm sorry to go back and forth on this, but your change here made me look into why I'd done this in the first place.

 libmongoc does not comply with the spec: it does _not_ treat a value of 0 for any timeout option as "no timeout", but instead treats it as "use the default".

see [here](http://mongoc.org/libmongoc/current/mongoc_uri_t.html#connection-options):
> Setting any of the *TimeoutMS options above to 0 will be interpreted as “use the default value”.

since its behavior when 0 is specified is non-spec compliant, I think it would be better for us to not allow a value of `0` for these options after all, and go back to your original code.

in 2.0 we can change our behavior to be in line with the spec.

I should have opened a ticket about this but I don't think I did. Could you please open a ticket along the lines of "allow users to specify 0 for timeout values"?

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
patrickfreed: Specifically, it looks like 4.2.9 was the first server version to support it.

.evergreen/config.yml
```diff
@@ -347,9 +347,10 @@ functions:
           SWIFT_VERSION=${SWIFT_VERSION} \
             bash ${PROJECT_DIRECTORY}/.evergreen/install-tools.sh sourcery
           make linuxmain SOURCERY=${PROJECT_DIRECTORY}/opt/sourcery/bin/sourcery
-          git diff --exit-code Tests/LinuxMain.swift
+          # use regex to ignore lines starting with a /
+          git diff -G "^[^\/].*\n" --exit-code Tests/LinuxMain.swift
           make exports SOURCERY=${PROJECT_DIRECTORY}/opt/sourcery/bin/sourcery
-          git diff --exit-code Sources/MongoSwiftSync/Exports.swift
+          git diff -G "^[^\/].*\n" --exit-code Sources/MongoSwiftSync/Exports.swift
```
patrickfreed, line 353: In order to prevent this diff from @kmahar 's PR from showing up here too, can we rebase off of `main` instead of doing a merge commit?

Sources/MongoSwift/ConnectionString.swift
patrickfreed: I think this still needs to be changed back.

Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -216,6 +216,13 @@ public struct TestRequirement: Decodable {
         )
     ]
 
+    public static let blockTimeSupport: [TestRequirement] = [
+        TestRequirement(
+            minServerVersion: ServerVersion.mongodBlockTimeSupport,
+            acceptableTopologies: [.single, .replicaSet]
```
patrickfreed, line 222: This option should be fine to work on any topology with that version, so I think we can just omit `acceptableTopologies` here.

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
```diff
@@ -320,4 +320,34 @@ final class SyncMongoClientTests: MongoSwiftTestCase {
...
             .to(throwError(errorType: MongoError.InvalidArgumentError.self))
     }
+
+    func testConnectionTimeout() throws {
+        let setupClient = try MongoClient.makeTestClient()
+        guard try setupClient.supportsBlockTime() else {
+            printSkipMessage(testName: self.name, reason: "blockTime not supported")
+            return
+        }
+        guard try setupClient.supportsFailCommand() else {
```
patrickfreed, line 330: Supporting blockTime on fail command implies support for failCommand, so this check can be removed now.

Sources/TestsCommon/CommonTestUtils.swift
bynn: done

Sources/MongoSwift/ConnectionString.swift
bynn: changed back

.evergreen/config.yml
bynn: done

Tests/MongoSwiftSyncTests/SyncMongoClientTests.swift
bynn: done

kmahar (approved):
lgtm assuming Evergreen goes green! nice work, this ended up being a tricky one 🙂

patrickfreed (approved):
LGTM pending green evg run, great job with this!

patrickfreed (approved):

PR #613, opened 2021-03-15
SWIFT-1100, SWIFT-722, SWIFT-798 Catch up on recent change streams spec updates 

kmahar:
This PR is a hodgepodge of changes I'll attempt to explain the thread between here...

Sorry for the large diff, it's mostly due to syncing JSON spec test files; you do not need to review those.

The main purpose of this ticket was to add a new property `truncatedArrays` to the `UpdateDescription` type returned in `ChangeStreamEvent`s. (SWIFT-1100). This was added to the spec via SPEC-1779 (corresponding [diff](https://github.com/mongodb/specifications/commit/07091ae92853530fddad2ac04f5f2641e78b27df)).

I synced the spec tests with the specs repo to bring in the new tests, which ended up bringing in a bunch of old tests we'd never synced either, in particular a bunch that were added via [SPEC-1505](https://jira.mongodb.org/browse/SPEC-1505) (for which our corresponding ticket is SWIFT-798). Getting all of those to pass required some minor changes to the runner which I will call out below. At this point we were also fine to stop skipping various tests we were skipping while we waited on results from SPEC-1505 (SWIFT-722). 

After all of that I returned to my original task, testing `truncatedArrays`. As it turns out, the new test for that actually uses an update pipeline, and we don't yet support passing a pipeline to our update methods. So I ended up writing a prose test that replicates the behavior of the spec test and uses `runCommand` to do an update with a pipeline, and we just skip the spec test for now. So in the end I guess there was no real point in syncing the spec test files but someone was going to have to do all of this eventually anyway so I figured I shouldn't let it go to waste. 🙂

### DIFF
--- a/Sources/MongoSwift/ChangeStreamEvent.swift
+++ b/Sources/MongoSwift/ChangeStreamEvent.swift
@@ -14,18 +14,18 @@ public struct UpdateDescription: Codable {
      *   updated field:
      *     "arrayField": ["foo", {"a": "bar", "b": 3}]
      *   a potential corresponding UpdateDescription:
-     *     {
-     *       "updatedFields": {
+     *     UpdateDescription(
+     *       updatedFields: {
      *         "arrayField.1.b": 3
      *       },
-     *       "removedFields": [],
-     *       "truncatedArrays": [
-     *         {
-     *           "field": "arrayField",
-     *           "newSize": 2
-     *         }
+     *       removedFields: [],
+     *       truncatedArrays: [
+     *         TruncatedArrayDescription(
+     *           field: "arrayField",
+     *           newSize: 2
+     *         )
      *       ]
-     *     }
+     *     )
      *
      * Modifications to array elements are expressed via dot notation.
      * Example: an `update` which sets the element with index 0 in the array field named arrayField to 7 is reported as
@@ -38,23 +38,24 @@ public struct UpdateDescription: Codable {
     /// An array of field names that were removed from the document.
     public let removedFields: [String]
 
+    public struct TruncatedArrayDescription: Codable {
+        /// The name of the array field which was truncated.
+        public let field: String
+        /// The new size of the array.
+        public let newSize: Int
+    }
+
     /**
      * Truncations of arrays may be reported either via this field or via the ‘updatedFields’ field. In the latter
      * case, the entire array is considered to be replaced. The method used to report a truncation is a server
      * implementation detail.
      *
-     * The structure of documents in this field is
-     *   {
-     *      "field": <string>,
-     *      "newSize": <int>
-     *   }
-     *
      * Example: an `update` which shrinks the array `arrayField.0.nestedArrayField` from size 8 to 5 may be reported
-     * via this field as [{"field": "arrayField.0.nestedArrayField", "newSize": 5}]
+     * via this field as [TruncatedArrayDescription(field: "arrayField.0.nestedArrayField", newSize: 5)].
      *
      * This property will only ever be present on MongoDB server versions >= 5.0.
      */
-    public let truncatedArrays: [BSONDocument]?
+    public let truncatedArrays: [TruncatedArrayDescription]?
 }
 
 /// An enum representing the type of operation for this change event.
--- a/Sources/MongoSwift/ChangeStreamEvent.swift
+++ b/Sources/MongoSwift/ChangeStreamEvent.swift
@@ -1,12 +1,60 @@
 /// An `UpdateDescription` containing fields that will be present in the change stream document for
 /// operations of type `update`.
 public struct UpdateDescription: Codable {
-    /// A document containing key:value pairs of names of the fields that were changed, and the new
-    /// value for those fields.
+    /**
+     * A document containing key:value pairs of names of the fields that were changed (excluding the fields reported
+     * via `truncatedArrays`), and the new value for those fields.
+     *
+     * Despite array fields reported via `truncatedArrays` being excluded from this field, changes to fields of the
+     * elements of the array values may be reported via this field.
+     *
+     * Example:
+     *   original field:
+     *     "arrayField": ["foo", {"a": "bar"}, 1, 2, 3]
+     *   updated field:
+     *     "arrayField": ["foo", {"a": "bar", "b": 3}]
+     *   a potential corresponding UpdateDescription:
+     *     {
+     *       "updatedFields": {
+     *         "arrayField.1.b": 3
+     *       },
+     *       "removedFields": [],
+     *       "truncatedArrays": [
+     *         {
+     *           "field": "arrayField",
+     *           "newSize": 2
+     *         }
+     *       ]
+     *     }
+     *
+     * Modifications to array elements are expressed via dot notation.
+     * Example: an `update` which sets the element with index 0 in the array field named arrayField to 7 is reported as
+     *   "updatedFields": {"arrayField.0": 7}
+     *
+     * - SeeAlso: https://docs.mongodb.com/manual/core/document/#document-dot-notation
+     */
     public let updatedFields: BSONDocument
 
     /// An array of field names that were removed from the document.
     public let removedFields: [String]
+
+    /**
+     * Truncations of arrays may be reported either via this field or via the ‘updatedFields’ field. In the latter
+     * case, the entire array is considered to be replaced. The method used to report a truncation is a server
+     * implementation detail.
+     *
+     * The structure of documents in this field is
+     *   {
+     *      "field": <string>,
+     *      "newSize": <int>
+     *   }
+     *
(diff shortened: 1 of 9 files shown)


Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -1,12 +1,60 @@
...
+     * - SeeAlso: https://docs.mongodb.com/manual/core/document/#document-dot-notation
+     */
     public let updatedFields: BSONDocument
 
     /// An array of field names that were removed from the document.
     public let removedFields: [String]
+
+    /**
+     * Truncations of arrays may be reported either via this field or via the ‘updatedFields’ field. In the latter
+     * case, the entire array is considered to be replaced. The method used to report a truncation is a server
```
kmahar, line 43: as an aside, it seems bad to me that the server does not report such changes consistently...

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -132,6 +130,9 @@ internal struct ChangeStreamTest: Decodable, FailPointConfigured {
     /// The minimum server version that this test can be run against.
     let minServerVersion: ServerVersion
 
+    /// The maximum server version that this test can be run against.
+    let maxServerVersion: ServerVersion?
```
kmahar, line 134: some of the newer spec tests I pulled in use a max version.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -190,9 +191,11 @@ internal struct ChangeStreamTest: Decodable, FailPointConfigured {
                 case let .success(events):
                     var seenEvents: [BSONDocument] = []
                     for _ in 0..<events.count {
-                        let event = try changeStream.tryNext()?.get()
-                        expect(event).toNot(beNil(), description: self.description)
-                        seenEvents.append(event!)
+                        guard let event = try changeStream.nextWithTimeout() else {
```
kmahar, line 194: some of the newer tests were not receiving change stream events quickly enough on latest + sharded clusters, so I switched to the timeout method to ensure we actually get them.

conveniently, this fixes one occurrence of the `continueAfterFailure` not working on Linux problem you mentioned earlier @patrickfreed.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -203,7 +206,7 @@ internal struct ChangeStreamTest: Decodable, FailPointConfigured {
 
         if let expectations = self.expectations {
             let commandEvents = monitor.commandStartedEvents()
-                .filter { $0.commandName != "isMaster" }
+                .filter { !["isMaster", "killCursors"].contains($0.commandName) }
```
kmahar, line 209: SPEC-1505 added language to the spec indicating `killCursors` events should be ignored:
> For each `(expected, idx)` in expectations - If `actual[idx]` is a `killCursors` event, skip it and move to `actual[idx+1]`. - Else assert that `actual[idx]` MATCHES `expected`

(in https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/README.rst#spec-test-runner)

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -227,10 +230,10 @@ private struct ChangeStreamTestFile: Decodable {
     let collectionName: String
 
     /// Secondary database.
-    let database2Name: String
+    let database2Name: String?
```
kmahar, line 233: some newer tests do not use the second database/collection.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -239,23 +242,32 @@ private struct ChangeStreamTestFile: Decodable {
 /// Class covering the JSON spec tests associated with change streams.
 final class ChangeStreamSpecTests: MongoSwiftTestCase {
     func testChangeStreamSpec() throws {
-        let tests = try retrieveSpecTestFiles(specName: "change-streams", asType: ChangeStreamTestFile.self)
+        let tests = try retrieveSpecTestFiles(
+            specName: "change-streams",
+            subdirectory: "legacy",
```
kmahar, line 247: tests in the old format were moved into the legacy directory.

@bynn, for some context, we used to have custom YAML/JSON formats for a number of specifications' tests. this was not great because new test runners take a lot of work to implement. we recently created a "unified" format that we are moving toward using across all specs so it is easier for us to test that drivers across languages have consistent behavior.

however, not all tests have been ported to the new format yet so we still have to run some of them.

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -501,73 +553,6 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
         }
     }
 
-    /**
-     * Prose test 5 of change stream spec.
-     *
-     * `ChangeStream` will not attempt to resume after encountering error code 11601 (Interrupted),
-     * 136 (CappedPositionLost), or 237 (CursorKilled) while executing a getMore command.
-     */
-    func testChangeStreamDoesntResume() throws {
```
kmahar, line 510: this test was removed from the change streams test plan as part of SPEC-1505.

Sources/MongoSwift/ChangeStreamEvent.swift
patrickfreed: Yeah that seems really strange. In general this field seems to be really strange. It's not immediately clear what information a user is expected to gain from knowing that an array was truncated, since it doesn't mention which items / indexes were removed. :shrug:

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -1,12 +1,60 @@
...
+     *      "field": <string>,
+     *      "newSize": <int>
+     *   }
+     *
+     * Example: an `update` which shrinks the array `arrayField.0.nestedArrayField` from size 8 to 5 may be reported
+     * via this field as [{"field": "arrayField.0.nestedArrayField", "newSize": 5}]
+     *
+     * This property will only ever be present on MongoDB server versions >= 5.0.
+     */
+    public let truncatedArrays: [BSONDocument]?
```
patrickfreed, line 57: Is there any particular reason for using just a document here instead of a `Codable` type?

kmahar: nope, good point. I remember having this thought at some point but I never came back to it. I added a new `TruncatedArrayDescription` type.

as an aside, I nested that type in `UpdateDescription`, but it got me thinking we don't seem to nest any other change stream related types that probably could/should be nested, such as `OperationType`.
I opened [SWIFT-1154](https://jira.mongodb.org/browse/SWIFT-1154) about considering that for 2.0.

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -1,12 +1,61 @@
...
+     *   "updatedFields": {"arrayField.0": 7}
+     *
+     * - SeeAlso: https://docs.mongodb.com/manual/core/document/#document-dot-notation
+     */
     public let updatedFields: BSONDocument
 
     /// An array of field names that were removed from the document.
     public let removedFields: [String]
+
+    public struct TruncatedArrayDescription: Codable {
```
patrickfreed, line 41: Can we add a docstring to this? Interesting that swiftlint doesn't automatically flag it, maybe because it's a nested type? I tried it out in a basic file and it seems to catch it though, and I am using the same version of swiftlint as evg :thinking:

Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -106,8 +106,6 @@ internal enum ChangeStreamTestResult: Decodable {
         if let labels = labels {
             expect(seenError.errorLabels).toNot(beNil(), description: description)
             expect(seenError.errorLabels).to(equal(labels), description: description)
-        } else {
```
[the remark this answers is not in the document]
patrickfreed: Yep that seems like the right interpretation to me too.

patrickfreed (approved):
One doc nit otherwise good to merge without review once fixed.

Sources/MongoSwift/ChangeStreamEvent.swift
kmahar: huh yeah that is weird, was definitely relying too much on swiftlint here to remind me to document it. good catch

kmahar:
@bynn , do you have any comments/questions on this one?

bynn (approved):
thanks for leaving the helpful explanations! I don't think I was confused about anything in the ticket

PR #614, opened 2021-03-17
Bump sourcery version, and ignore comments when checking Sourcery diffs going forward

kmahar:
Updates us to the latest version of Sourcery, and removes the need for us all to be on the same exact version of Sourcery going forward by ignoring any lines starting with a "/" when checking the diff.

Here's a sample patch with my first commit only, where I updated the version being used on Evergreen but did not actually regenerate the files using that version string: https://spruce.mongodb.com/version/605255a61e2d171f05624717


bynn (approved):

patrickfreed (approved):
LGTM! so glad to see this resolved at long last.

PR #615, opened 2021-03-17
SWIFT-604 delete (uint32_t)

bynn:


patrickfreed:
So the original goal of this ticket was to remove some special handling we had to do when enabling failpoints in Swift (we generally try not to make any direct changes to the vendored C driver), but it looks like it has already been removed! So sorry that you had to investigate this, I'm sure it didn't really make sense as there isn't anywhere in the driver that we are still doing this.

For reference, here is the change: https://github.com/mongodb/mongo-swift-driver/commit/10b375226ee0d6fbf8fe01ac3e566abd95b958c7#diff-9ec869d1bd31666b53e58a8b56e2115d720940520ac456e1f8da5c4a749f27b4L56

Again, sorry about the wild goose chase!

PR #616, opened 2021-03-18
SWIFT-788, SWIFT-1157 Allow passing hint to findAndModify update and replace operations and implement expresibleByXLiteral for indexHint

bynn:
SWIFT-788, SWIFT-1157 Allow passing hint to findAndModify update and replace operations and implement expresibleByXLiteral for indexHint

### DIFF
--- a/Sources/MongoSwift/Hint.swift
+++ b/Sources/MongoSwift/Hint.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// An index to "hint" or force MongoDB to use when performing a query.
-public enum IndexHint: Codable {
+public enum IndexHint: Codable, ExpressibleByStringLiteral, ExpressibleByDictionaryLiteral {
     /// Specifies an index to use by its name.
     case indexName(String)
     /// Specifies an index to use by a specification `BSONDocument` containing the index key(s).
@@ -15,6 +15,18 @@ public enum IndexHint: Codable {
         case let .indexSpec(doc):
             try container.encode(doc)
         }
+    }
+
+    public init(stringLiteral value: StringLiteralType) {
+        self = .indexName(value)
+    }
+
+    public init(dictionaryLiteral keyValuePairs: (String, BSON)...) {
+        var docs = BSONDocument()
+        for (k, v) in keyValuePairs {
+            docs[k] = v
+        }
+        self = .indexSpec(docs)
     }
 
     public init(from decoder: Decoder) throws {
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -199,7 +199,7 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     public var collation: BSONDocument?
 
     /// A document or string that specifies the index to use to support the query.
-    public var hint: BSONDocument?
+    public var hint: IndexHint?
 
     /// The maximum amount of time to allow the query to run.
     public var maxTimeMS: Int?
@@ -237,7 +237,7 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     public init(
         bypassDocumentValidation: Bool? = nil,
         collation: BSONDocument? = nil,
-        hint: BSONDocument? = nil,
+        hint: IndexHint? = nil,
         maxTimeMS: Int? = nil,
         projection: BSONDocument? = nil,
         returnDocument: ReturnDocument? = nil,
@@ -269,7 +269,7 @@ public struct FindOneAndUpdateOptions: FindAndModifyOptionsConvertible, Decodabl
     public var collation: BSONDocument?
 
     /// A document or string that specifies the index to use to support the query.
-    public var hint: BSONDocument?
+    public var hint: IndexHint?
 
     /// The maximum amount of time to allow the query to run.
     public var maxTimeMS: Int?
@@ -309,7 +309,7 @@ public struct FindOneAndUpdateOptions: FindAndModifyOptionsConvertible, Decodabl
         arrayFilters: [BSONDocument]? = nil,
         bypassDocumentValidation: Bool? = nil,
         collation: BSONDocument? = nil,
-        hint: BSONDocument? = nil,
+        hint: IndexHint? = nil,
         maxTimeMS: Int? = nil,
         projection: BSONDocument? = nil,
         returnDocument: ReturnDocument? = nil,
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -198,7 +198,7 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     /// Specifies a collation to use.
     public var collation: BSONDocument?
 
-    /// A document or string that specifies the index to use to support the query. Only supported in server 4.4+
+    /// A document or string that specifies the index to use to support the query. Only supported in server 4.4+.
     public var hint: IndexHint?
 
     /// The maximum amount of time to allow the query to run.
--- a/Sources/MongoSwift/MongoCollection+FindAndModify.swift
+++ b/Sources/MongoSwift/MongoCollection+FindAndModify.swift
@@ -198,6 +198,9 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     /// Specifies a collation to use.
     public var collation: BSONDocument?
 
+    /// A document or string that specifies the index to use to support the query.
+    public var hint: BSONDocument?
+
     /// The maximum amount of time to allow the query to run.
     public var maxTimeMS: Int?
 
@@ -220,6 +223,7 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
         try FindAndModifyOptions(
             bypassDocumentValidation: self.bypassDocumentValidation,
             collation: self.collation,
+            hint: self.hint,
             maxTimeMS: self.maxTimeMS,
             projection: self.projection,
             returnDocument: self.returnDocument,
@@ -233,6 +237,7 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     public init(
         bypassDocumentValidation: Bool? = nil,
         collation: BSONDocument? = nil,
+        hint: BSONDocument? = nil,
         maxTimeMS: Int? = nil,
         projection: BSONDocument? = nil,
         returnDocument: ReturnDocument? = nil,
@@ -242,6 +247,7 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     ) {
         self.bypassDocumentValidation = bypassDocumentValidation
         self.collation = collation
+        self.hint = hint
         self.maxTimeMS = maxTimeMS
         self.projection = projection
         self.returnDocument = returnDocument
@@ -261,6 +267,9 @@ public struct FindOneAndUpdateOptions: FindAndModifyOptionsConvertible, Decodabl
 
     /// Specifies a collation to use.
     public var collation: BSONDocument?
+
+    /// A document or string that specifies the index to use to support the query.
+    public var hint: BSONDocument?
 
     /// The maximum amount of time to allow the query to run.
     public var maxTimeMS: Int?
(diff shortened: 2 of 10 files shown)


Sources/MongoSwift/MongoCollection+FindAndModify.swift
```diff
@@ -198,6 +198,9 @@ public struct FindOneAndReplaceOptions: FindAndModifyOptionsConvertible, Decodab
     /// Specifies a collation to use.
     public var collation: BSONDocument?
 
+    /// A document or string that specifies the index to use to support the query.
+    public var hint: BSONDocument?
```
patrickfreed, line 202: We actually have a already `Hint` Swift type defined for similar use-cases elsewhere in the driver. Can this be updated to use that instead?

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -540,13 +540,28 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(result3).to(equal(["_id": 2, "cat": "blah"]))
         expect(try self.coll.countDocuments()).to(equal(2))
 
-        // test invalid maxTimeMS throws error
+        // test using a hint
+        let model = IndexModel(keys: ["cat": 1])
+        expect(try self.coll.createIndex(model)).to(equal("cat_1"))
+        let opts4 = FindOneAndReplaceOptions(hint: ["cat": 1])
```
patrickfreed, line 546: can we add a test case that also uses the name of the index too?

Sources/MongoSwift/MongoCollection+FindAndModify.swift
patrickfreed, line 202: this option is only supported in MongoDB 4.4+, can we mention so in the docstring? ditto for the option in the other docstring

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -540,13 +540,28 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(result3).to(equal(["_id": 2, "cat": "blah"]))
         expect(try self.coll.countDocuments()).to(equal(2))
 
-        // test invalid maxTimeMS throws error
+        // test using a hint
```
patrickfreed, line 543: since hint is only supported in MongoDB 4.4+, we'll need to skip this test on earlier versions to ensure the tests pass. Because of this, it probably makes sense to break the hint stuff out to its own test case too.

Sources/MongoSwift/MongoCollection+FindAndModify.swift
kmahar: (it's named `IndexHint`)

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -589,6 +589,66 @@ final class MongoCollectionTests: MongoSwiftTestCase {
             .to(throwError(errorType: MongoError.InvalidArgumentError.self))
     }
 
+    func testFindAndModifyWithHint() throws {
+        guard let client = _client else {
```
patrickfreed, line 593: This file is kind of old, so it uses some of our older testing mechanisms (e.g. shared client / collection on `self`). These days we've moved away from this and have instead opted to create new clients / dbs / collections per test to avoid sharing state and to keep things encapsulated. To that end, could we update this test to make use of `withTestNamespace` instead of `_client` and `self.coll`?

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
```diff
@@ -131,6 +131,10 @@ extension MongoClient {
     internal func supportsBlockTime() throws -> Bool {
         try self.meetsAnyRequirement(in: TestRequirement.blockTimeSupport)
     }
+
+    internal func supportsHint() throws -> Bool {
```
patrickfreed, line 135: While this made sense for `supportsfailCommand` and `supportsblockTime`, since those methods will be reused in many different tests, I'm not sure if that will necessarily be the case for this one. Instead, I think we should just test this version directly in `testFindAndModifyWithHint` itself, and if we need it again we can consider abstracting out to a method.

Sources/TestsCommon/ServerVersion.swift
```diff
@@ -5,6 +5,7 @@ public struct ServerVersion: Comparable, Decodable, CustomStringConvertible {
     public static let mongodFailCommandSupport = ServerVersion(major: 4, minor: 0)
     public static let mongosFailCommandSupport = ServerVersion(major: 4, minor: 1, patch: 5)
     public static let mongodBlockTimeSupport = ServerVersion(major: 4, minor: 2, patch: 9)
+    public static let mongodHintSupport = ServerVersion(major: 4, minor: 4)
```
patrickfreed, line 8: see my comment on `supportsHint`, but I think we'll want to just use this directly in the test without defining a constant for it. Also note that hints have had gained support in different methods over a few different versions of MongoDB--4.4 was just the one for `findAndModify`.

patrickfreed (commented):
Looks good! I just have some minor testing suggestions.

Also, can the title of the PR be updated to include "findAndModify" as per the ticket title? As-is, it could be confused with `updateOne` and `replaceOne` etc.

Sources/TestsCommon/ServerVersion.swift
bynn: removed

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
bynn: removed

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
bynn: done

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
```diff
@@ -46,6 +46,7 @@ extension MongoSwiftTestCase {
             try database.collection(collName).drop()
             collection = try database.createCollection(collName, options: options)
         }
+        try collection.deleteMany([:])
```
kmahar, line 49: I would think this is unnecessary given that we either just create the collection, or just dropped and recreated it. can you explain why this was added?

kmahar (changes requested):
lookin good! just some tiny nits and a question

Sources/MongoSwift/Hint.swift
```diff
@@ -17,6 +17,18 @@ public enum IndexHint: Codable {
         }
     }
 
+    public init(stringLiteral value: StringLiteralType) {
+        self = .indexName(value)
+    }
+
+    public init(dictionaryLiteral keyValuePairs: (String, BSON)...) {
+        var docs = BSONDocument()
```
patrickfreed, line 25: So as to not repeat `BSONDocument`'s implementation of this protocol, we can just call straight through to it (e.g. `BSONDocument(dictionaryLiteral: keyValuePairs)`)

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
bynn: I was trying to figure out how to fix this behavior of collection drop only dropping from primaries and sharded replica sets. Although this fix seemed to break everything instead which I also don't understand why because the collection should hold nothing at this point. 

https://docs.mongodb.com/manual/reference/method/db.collection.drop/

patrickfreed: So the issue is that in sharded clusters, each mongos retains an internal cache of the available collections, and apparently re-creating a dropped collection may retain some of the previous one's documents (at least on that mongos?). This deleteMany ensures those documents are actually cleared out correctly. The linked documentation in @bynn's comment includes a small tidbit about having to do this, but doesn't really explain why its necessary.

This is important for this test in particular because the `setUp` inserts a few documents which need to be cleared out.

Sources/MongoSwift/Hint.swift
bynn: if I understand correctly I would be doing something like `self = .indexSpec(BSONDocument(dictionaryLiteral: keyValuePairs))` but that gives compiler errors because the types

patrickfreed: Oh hm, good point. It looks like we'd need https://bugs.swift.org/browse/SR-128 to be implemented before we could do something like that unfortunately. Anyways, sorry for sending you down this dead end, this looks good as-is.

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
patrickfreed: It's probably worth including a comment inline here explaining this for future readers.

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -589,6 +589,78 @@ final class MongoCollectionTests: MongoSwiftTestCase {
             .to(throwError(errorType: MongoError.InvalidArgumentError.self))
     }
 
+    func testFindAndModifyWithHint() throws {
+        try withTestNamespace { client, _, collection in
+            let requirements = TestRequirement(
+                minServerVersion: ServerVersion(major: 4, minor: 4)
```
patrickfreed, line 595: According to the [spec change](https://github.com/mongodb/specifications/commit/0f803725487cc951877f6975061dd1e24056a045#diff-2fe314a3da48fca10d3a170b09a5a393a93dbb4e9369d6741c99656a95dd7034R1609), drivers are actually in charge of returning errors if the server is < 4.2. Could we add a separate test case that verifies we return such an error?

bynn: done

Tests/MongoSwiftSyncTests/SyncTestUtils.swift
bynn: done

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -589,6 +589,105 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+                return
+            }
+
+            // Drivers must raise an error for servers < 4.2.
+            let invalidOpts1 = FindOneAndReplaceOptions(hint: ["nonexistant_index": 1])
+            let invalidOpts2 = FindOneAndReplaceOptions(hint: "nonexistant_index")
+            let invalidOpts3 = FindOneAndUpdateOptions(hint: ["nonexistant_index": 1])
+            let invalidOpts4 = FindOneAndUpdateOptions(hint: "nonexistant_index")
+            expect(try collection.findOneAndReplace(filter: [:], replacement: [:], options: invalidOpts1))
+                .to(throwError(errorType: MongoError.InternalError.self))
```
patrickfreed, line 609: Generally speaking, we shouldn't be `InternalError`s, since those are reserved for buggy behavior / system failure. We'll likely need to add some more support for parsing the error that libmongoc is returning here to return a more specific one.

From looking at the libmongoc source for this, it seems to yield a `bson_error_t` with domain `MONGOC_ERROR_COMMAND` and code `MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION` on < 4.2 when a hint is spcified. Looking at `parseMongocError` in `MongoError.swift`, it looks like we don't handle this particular combination, hence why this is currently throwing `InternalError`. We should update that method to return a `CompatibilityError`, like we do for the`(MONGOC_ERROR_PROTOCOL, MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION)` combination, and then these failures should be more accurate.

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -589,6 +589,105 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+                maxServerVersion: ServerVersion(major: 4, minor: 2)
+            )
+            let unmetRequirement = try client.getUnmetRequirement(requirements)
+            guard unmetRequirement == nil else {
+                printSkipMessage(testName: self.name, unmetRequirement: unmetRequirement!)
+                return
+            }
+
+            // Drivers must raise an error for servers < 4.2.
+            let invalidOpts1 = FindOneAndReplaceOptions(hint: ["nonexistant_index": 1])
```
patrickfreed, line 604: Because this test would also currently generate errors even on 4.2+ due  to the index not existing, it makes it slightly unclear exactly what is being verified on < 4.2. If we instead modify the case such that it'd cause errors _only_ on 4.2, then it becomes more clear what is being evaluated. To that end, I think it would be marginally better to run this test with a valid hint specified (i.e. an index that does exist) so that we can ensure the driver is failing for the right reason.

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
bynn: done

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
bynn: done

patrickfreed (approved):
LGTM pending green CI, great job with this!

kmahar (approved):
lgtm! great job

PR #617, opened 2021-03-24
SWIFT-1054 Add public API for versioned API

kmahar:
Adds the public API required for versioned API support as defined [here](https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.rst#mongoclient-changes).

In a future PR, `MongoClientOptions.ServerAPI` will gain a `withMongocServerAPI` method that will allow it to interoperate with libmongoc.

### DIFF
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -152,7 +152,7 @@ public struct MongoClientOptions: CodingStrategyProvider {
         replicaSet: String? = nil,
         retryReads: Bool? = nil,
         retryWrites: Bool? = nil,
-        serverAPI: ServerAPI? = nil,
+        serverAPI: MongoServerAPI? = nil,
         serverSelectionTimeoutMS: Int? = nil,
         threadPoolSize: Int? = nil,
         tls: Bool? = nil,
--- a/Sources/MongoSwift/MongoClient.swift
+++ b/Sources/MongoSwift/MongoClient.swift
@@ -66,6 +66,77 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// Determines whether the client should retry supported write operations (on by default).
     public var retryWrites: Bool?
 
+    /// A type containing options for specifying a MongoDB server API version and related behavior.
+    public struct ServerAPI: Codable {
+        /// Represents a server API version.
+        public struct Version: Codable, Equatable, LosslessStringConvertible {
+            /// MongoDB API version 1.
+            public static let v1 = Version(.v1)
+
+            private enum _Version: String {
+                case v1 = "1"
+            }
+
+            private let _version: _Version
+
+            private init(_ version: _Version) {
+                self._version = version
+            }
+
+            /// `LosslessStringConvertible` conformance
+
+            public init?(_ description: String) {
+                guard let _version = _Version(rawValue: description) else {
+                    return nil
+                }
+                self.init(_version)
+            }
+
+            public var description: String {
+                self._version.rawValue
+            }
+
+            /// `Codable` conformance
+
+            public func encode(to encoder: Encoder) throws {
+                var container = encoder.singleValueContainer()
+                try container.encode(self._version.rawValue)
+            }
+
+            public init(from decoder: Decoder) throws {
+                let container = try decoder.singleValueContainer()
+                let stringValue = try container.decode(String.self)
+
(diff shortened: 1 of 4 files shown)


Sources/MongoSwift/MongoClient.swift
```diff
@@ -66,6 +66,77 @@ public struct MongoClientOptions: CodingStrategyProvider {
     /// Determines whether the client should retry supported write operations (on by default).
     public var retryWrites: Bool?
 
+    /// A type containing options for specifying a MongoDB server API version and related behavior.
+    public struct ServerAPI: Codable {
```
kmahar, line 70: so I made this a nested type in line with our recent conversation about nesting types when they're only used in the context of others, but just now I noticed [this section](https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.rst#overriding-the-declared-api-version) in the spec:

> In the future, we may want to allow users to override the declared API version on a `MongoDatabase`, `MongoCollection`, or individual operation level. However, this is not necessary until there is a different API version and we have data on why and how users would want to override the declared API version.

given that, it seems like maybe it would be a little more future-proof to make `ServerAPI` a standalone type (while still nesting `Version`) and I'm leaning toward doing so.

what do you think?

Sources/MongoSwift/MongoDatabase.swift
```diff
@@ -357,6 +357,9 @@ public struct MongoDatabase {
      *    - `MongoError.WriteError` if any error occurs while the command was performing a write.
      *    - `MongoError.CommandError` if an error occurs that prevents the command from being performed.
      *    - `EncodingError` if an error occurs while encoding the options to BSON.
+     *
+     * - Note: Attempting to specify an API version in this command is considered undefined behavior. API version may
```
kmahar, line 361: (spec requires we document this)

Sources/MongoSwift/MongoClient.swift
bynn: could there be other api settings other than version in the future?

bynn (approved):

Sources/MongoSwift/MongoClient.swift
patrickfreed: Given that warning yeah it probably makes sense to bump it out to the global namespace. We'd need to rename it to `MongoServerAPI` as part of that.

As a side note: my guess is that it will always remain a client option, but you never know how users will want to use things :slightly_smiling_face:

Sources/MongoSwift/MongoClient.swift
```diff
@@ -66,6 +66,77 @@ public struct MongoClientOptions: CodingStrategyProvider {
...
+
+        /// Convenience initializer allowing optional parameters to be optional or omitted.
+        public init(version: Version, strict: Bool? = nil, deprecationErrors: Bool? = nil) {
+            self.version = version
+            self.strict = strict
+            self.deprecationErrors = deprecationErrors
+        }
+    }
+
+    /// Specifies a MongoDB server API version and related options.
```
patrickfreed, line 137: Is there documentation we can link to for this yet?

Sources/MongoSwift/MongoClient.swift
kmahar: > could there be other api settings other than version in the future?
There are already a couple of others I included below as properties on `ServerAPI`, `strict`, and `deprecationErrors`. There could definitely be more settings added to this struct in the future!

I agree it's probably unlikely but better to be safe I think. I have moved it out to a top level type, in its own file, and renamed it `MongoServerAPI`.

Sources/MongoSwift/MongoClient.swift
kmahar: I don't think so. The relevant [DOCSP epic](https://jira.mongodb.org/browse/DOCSP-14217) is all in unstarted or in progress.
Opened [SWIFT-1159](https://jira.mongodb.org/browse/SWIFT-1159) and added a couple TODOs about updating later on once there are official docs.

patrickfreed (approved):

patrickfreed:
Actually, it looks like a reference to `ServerAPI` needs to be updated, but other than that LGTM (don't need to review the fix).

kmahar:
🤦‍♀️ I swore I had made sure this built before pushing but I guess not. sigh. fixed 🙂

PR #618, opened 2021-03-24
SWIFT-797 Allow hinting the delete command

bynn:
SWIFT-797 Modified `deleteOne` `deleteMany` and `findOneAndDelete`


Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -256,6 +256,31 @@ final class MongoCollectionTests: MongoSwiftTestCase {
...
+            let doc2: BSONDocument = ["_id": 2, "cat": "nya", "dog": "woof"]
+            try collection.insertMany([doc1, doc2])
+
+            let model = IndexModel(keys: ["cat": 1, "dog": 1])
+            expect(try collection.createIndex(model)).to(equal("cat_1_dog_1"))
+
+            let opts1 = DeleteOptions(hint: ["cat": 1, "dog": 1])
+            let opts2 = DeleteOptions(hint: "cat_1_dog_1")
+            expect(try collection.deleteOne(["cat": "meow"], options: opts1)?.deletedCount).to(equal(1))
+            expect(try collection.deleteOne(["cat": "nya"], options: opts2)?.deletedCount).to(equal(1))
```
patrickfreed, line 280: can we add an assertion that a bad hint fails for each of these?

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
```diff
@@ -256,6 +256,31 @@ final class MongoCollectionTests: MongoSwiftTestCase {
         expect(try self.coll.deleteOne(["cat": "cat"])?.deletedCount).to(equal(1))
```
patrickfreed, line 256: can we amend the `testHintInPreviousVersion` test to include `findOneAndDelete`? From looking at the spec, it seems it is only required for `findAndModify` based commands and isn't required for the other ones though.

As a side note: can we update that test name to include a reference to `findAndModify` to clarify that?

patrickfreed (commented):
Looks good! I just have a few minor testing suggestions

Tests/MongoSwiftSyncTests/MongoCollectionTests.swift
bynn: added for `deleteOne` and `deleteMany`

patrickfreed (approved):
LGTM!

kmahar (approved):
lgtm!

PR #621, opened 2021-03-25
SWIFT-1054 Unified test runner updates for versioned API

kmahar:
Makes some updates to the runner that are prerequisites for decoding and running the versioned API tests:

* Support for decoding more types of operations we hadn't added before
* Support for a new type of test requirement, required server parameters

### DIFF
--- a/Sources/MongoSwift/MongoCollection+BulkWrite.swift
+++ b/Sources/MongoSwift/MongoCollection+BulkWrite.swift
@@ -325,6 +325,17 @@ public struct BulkWriteOptions: Codable {
         self.ordered = options.ordered
         self.writeConcern = options.writeConcern
     }
+
+    private enum CodingKeys: String, CodingKey {
+        case bypassDocumentValidation, ordered, writeConcern
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.bypassDocumentValidation = try container.decodeIfPresent(Bool.self, forKey: .bypassDocumentValidation)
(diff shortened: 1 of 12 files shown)


Sources/MongoSwift/MongoCollection+BulkWrite.swift
```diff
@@ -325,6 +325,18 @@ public struct BulkWriteOptions: Codable {
...
+
+    private enum CodingKeys: String, CodingKey {
+        case bypassDocumentValidation, ordered, writeConcern
+    }
+
+    // A manual implementation is required to enforce a default value for ordered without making it optional.
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.bypassDocumentValidation = try container.decodeIfPresent(Bool.self, forKey: .bypassDocumentValidation)
+        self.ordered = try container.decodeIfPresent(Bool.self, forKey: .ordered) ?? true
```
patrickfreed, line 337: if we provide a default value like `public var ordered = true` does the synthesized implementation not use that when ordered is missing?

patrickfreed (commented):
All looks good, just have one minor question about a `Codable` conformance.

patrickfreed:
Oh, also looks like sourcery needs to be run.

Sources/MongoSwift/MongoCollection+BulkWrite.swift
kmahar: I had the same thought. unfortunately no, that default value does not get propagated to the Codable conformance.

kmahar:
whoops, updated linuxmain.

patrickfreed (approved):

bynn (approved):
everything looks good

PR #622, opened 2021-03-25
Vendor libmongoc @ 203e6b9e26bfce6ca76f1d087470c3840318afdd

kmahar:
This is a temporary vendoring as of a recent libmongoc commit to allow work on versioned API to progress. 

I did this by editing the vendoring script a bit to support vendoring in a branch rather than a version. I did not commit that  but you can see what the modified script to do so looks like here: https://github.com/mongodb/mongo-swift-driver/blob/vendor-libmongoc-versioned-api/etc/vendor-libmongoc.sh

Once there is a tag next week I will re-vendor from that.

@bynn, any changes within the `Sources/CLibMongoc` directory are just code copied directly from libmongoc and can be ignored. 

All that really changed here:
* some updates to the patches we apply as part of the vendoring script; since the corresponding libmongoc files are a bit different from the previous release, the line numbers are surrounding lines are too
* temporarily skipping some outdated retryable reads tests that no longer work on 4.9+ due to the switch to using $collStats for estimatedDocumentCount. I will sync those tests and unskip after this is merged. technically we only need to skip them on 4.9+ but I didn't think it really mattered given it will only stay this way for a day or two
* unskipped change stream retryable reads tests while I was editing the skip list anyway, as the problem has been long fixed in libmongoc.
* libmongoc added a new generated header file, common-config.h.in, which would normally be generated by CMake. since SwiftPM doesn't support generated header files, we have to have hard-coded copies of these, so I've added one of those.

running a more thorough evg patch here: https://spruce.mongodb.com/version/605d0d76850e6110f9aea9b3


kmahar:
(I forgot to run swiftformat so that failed in the linked patch because I forgot a colon in my "TODO". fixed)

patrickfreed (approved):
LGTM

Tests/MongoSwiftSyncTests/RetryableReadsTests.swift
```diff
@@ -42,11 +42,11 @@ private struct RetryableReadsTestFile: Decodable, SpecTestFile {
     let tests: [RetryableReadsTest]
 
     static let skippedTestFileNameKeywords = [
-        "changeStream", // TODO: SWIFT-648: Unskip this test
```
patrickfreed, line 45: I went ahead and moved this ticket into code review so we remember to close it once this PR is merged.

kmahar: thanks! I forgot to go do that.

bynn (approved):

PR #623, opened 2021-03-29
SWIFT-1160 Introduce new MongoConnectionString type

bynn:
SWIFT-1160 created MongoConnectionString.swift 

I just created tempOptions so the linter would stop complaining about unused variables

### DIFF
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -48,12 +48,8 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         let identifiersAndOptions = schemeAndRest[1].components(separatedBy: "/")
         let userAndHost = identifiersAndOptions[0].components(separatedBy: "@")
         let userInfo = userAndHost.count == 2 ? userAndHost[0].components(separatedBy: ":") : nil
-        guard let hosts = userInfo != nil ?
-            try? userAndHost[1].components(separatedBy: ",").map(ServerAddress.init) :
-            try? userAndHost[0].components(separatedBy: ",").map(ServerAddress.init)
-        else {
-            throw MongoError.InvalidArgumentError(message: "Invalid connection string host")
-        }
+        let hostString = userInfo != nil ? userAndHost[1] : userAndHost[0]
+        let hosts = try hostString.components(separatedBy: ",").map(ServerAddress.init)
         self.scheme = scheme
         self.hosts = hosts
     }
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -34,14 +34,11 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         public var description: String { self._scheme.rawValue }
     }
 
-    /// A struct representing a network address, consisting of a host and port.
+    /// A struct representing a host identifier, consisting of a host and an optional port.
+    /// In standard connection strings, this describes the address of a mongod or mongos to connect to.
+    /// In mongodb+srv connection strings, this describes a DNS name to be queried for SRV and TXT records.
     public struct HostIdentifier: Equatable, CustomStringConvertible {
-        private static func verifyPort(from: String, scheme: Scheme) throws -> UInt16? {
-            guard scheme != .srv else {
-                throw MongoError.InvalidArgumentError(
-                    message: "mongodb+srv with port number"
-                )
-            }
+        private static func parsePort(from: String) throws -> UInt16? {
             guard let port = UInt16(from), port > 0 else {
                 throw MongoError.InvalidArgumentError(
                     message: "port must be a valid, positive unsigned 16 bit integer"
@@ -66,7 +63,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         private let type: HostType
 
         /// Initializes a ServerAddress, using the default localhost:27017 if a host/port is not provided.
-        internal init(_ hostAndPort: String = "localhost:27017", scheme: Scheme) throws {
+        internal init(_ hostAndPort: String = "localhost:27017") throws {
             guard !hostAndPort.contains("?") else {
                 throw MongoError.InvalidArgumentError(message: "\(hostAndPort) contains invalid characters")
             }
@@ -88,9 +85,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                 }
                 host = String(hostAndPort[hostRange])
                 if let portRange = Range(match.range(at: 2), in: hostAndPort) {
-                    port = try HostIdentifier.verifyPort(from: String(hostAndPort[portRange]), scheme: scheme)
-                } else {
-                    port = scheme == .srv ? nil : 27017
+                    port = try HostIdentifier.parsePort(from: String(hostAndPort[portRange]))
                 }
                 self.type = .ipLiteral
             } else {
@@ -102,9 +97,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                     )
                 }
                 if parts.count > 1 {
-                    port = try HostIdentifier.verifyPort(from: parts[1], scheme: scheme)
-                } else {
-                    port = scheme == .srv ? nil : 27017
+                    port = try HostIdentifier.parsePort(from: parts[1])
                 }
                 self.type = .hostname
             }
@@ -113,20 +106,17 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         }
 
         public var description: String {
+            var ret = ""
             switch self.type {
             case .ipLiteral:
-                if let prt = port {
-                    return "[\(self.host)]:\(prt)"
-                } else {
-                    return "[\(self.host)]"
-                }
+                ret += "[\(self.host)]"
             default:
-                if let prt = port {
-                    return "\(self.host):\(prt)"
-                } else {
-                    return "\(self.host)"
-                }
+                ret += "\(self.host)"
             }
+            if let port = self.port {
+                ret += ":\(port)"
+            }
+            return ret
         }
     }
 
@@ -144,8 +134,11 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
             throw MongoError.InvalidArgumentError(message: "Invalid connection string")
         }
         let identifiersAndOptions = schemeAndRest[1].components(separatedBy: "/")
+        // TODO: SWIFT-1174: handle unescaped slashes in unix domain sockets.
         guard identifiersAndOptions.count <= 2 else {
-            throw MongoError.InvalidArgumentError(message: "Host with unescaped slash")
+            throw MongoError.InvalidArgumentError(
+                message: "Connection string contrains an unescaped slash"
+            )
         }
         let userAndHost = identifiersAndOptions[0].components(separatedBy: "@")
         guard userAndHost.count <= 2 else {
@@ -153,12 +146,15 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         }
         let userInfo = userAndHost.count == 2 ? userAndHost[0].components(separatedBy: ":") : nil
         let hostString = userInfo != nil ? userAndHost[1] : userAndHost[0]
-        let hosts = try hostString.components(separatedBy: ",").map {
-            try HostIdentifier($0, scheme: scheme)
-        }
+        let hosts = try hostString.components(separatedBy: ",").map(HostIdentifier.init)
         if case .srv = scheme {
-            guard hosts.count == 1 else {
-                throw MongoError.InvalidArgumentError(message: "mongodb+srv with multiple service names")
+            guard hosts.count == 1, hosts[0].port == nil else {
+                throw MongoError.InvalidArgumentError(
+                    message: "Only a single host identifier may be specified in a mongodb+srv connection string")
+            }
+            guard hosts[0].port == nil else {
+                throw MongoError.InvalidArgumentError(
+                    message: "A port cannot be specified in a mongodb+srv connection string")
             }
         }
         self.scheme = scheme
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// Represents a MongoDB connection string.
 /// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/
 public struct MongoConnectionString: Codable, LosslessStringConvertible {
@@ -32,6 +34,102 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         public var description: String { self._scheme.rawValue }
     }
 
+    /// A struct representing a network address, consisting of a host and port.
+    public struct HostIdentifier: Equatable, CustomStringConvertible {
+        private static func verifyPort(from: String, scheme: Scheme) throws -> UInt16? {
+            guard scheme != .srv else {
+                throw MongoError.InvalidArgumentError(
+                    message: "mongodb+srv with port number"
+                )
+            }
+            guard let port = UInt16(from), port > 0 else {
+                throw MongoError.InvalidArgumentError(
+                    message: "port must be a valid, positive unsigned 16 bit integer"
+                )
+            }
+            return port
+        }
+
+        private enum HostType: String {
+            case ipv4
+            case ipLiteral = "ip_literal"
+            case hostname
+            case unixDomainSocket
+        }
+
+        /// The hostname or IP address.
+        public let host: String
+
+        /// The port number.
+        public let port: UInt16?
+
+        private let type: HostType
+
+        /// Initializes a ServerAddress, using the default localhost:27017 if a host/port is not provided.
+        internal init(_ hostAndPort: String = "localhost:27017", scheme: Scheme) throws {
+            guard !hostAndPort.contains("?") else {
+                throw MongoError.InvalidArgumentError(message: "\(hostAndPort) contains invalid characters")
+            }
+
+            var host: String
+            var port: UInt16?
+
+            // Check if host is an IPv6 literal.
+            if hostAndPort.first == "[" {
+                let ipLiteralRegex = try NSRegularExpression(pattern: #"^\[(.*)\](?::([0-9]+))?$"#)
+                guard
+                    let match = ipLiteralRegex.firstMatch(
+                        in: hostAndPort,
+                        range: NSRange(hostAndPort.startIndex..<hostAndPort.endIndex, in: hostAndPort)
+                    ),
+                    let hostRange = Range(match.range(at: 1), in: hostAndPort)
+                else {
+                    throw MongoError.InvalidArgumentError(message: "couldn't parse address from \(hostAndPort)")
+                }
+                host = String(hostAndPort[hostRange])
+                if let portRange = Range(match.range(at: 2), in: hostAndPort) {
+                    port = try HostIdentifier.verifyPort(from: String(hostAndPort[portRange]), scheme: scheme)
+                } else {
+                    port = scheme == .srv ? nil : 27017
+                }
+                self.type = .ipLiteral
+            } else {
+                let parts = hostAndPort.components(separatedBy: ":")
+                host = String(parts[0])
+                guard parts.count <= 2 else {
+                    throw MongoError.InvalidArgumentError(
+                        message: "expected only a single port delimiter ':' in \(hostAndPort)"
+                    )
+                }
+                if parts.count > 1 {
+                    port = try HostIdentifier.verifyPort(from: parts[1], scheme: scheme)
+                } else {
+                    port = scheme == .srv ? nil : 27017
+                }
+                self.type = .hostname
+            }
+            self.host = host
+            self.port = port
+        }
+
+        public var description: String {
+            switch self.type {
+            case .ipLiteral:
+                if let prt = port {
+                    return "[\(self.host)]:\(prt)"
+                } else {
+                    return "[\(self.host)]"
+                }
+            default:
+                if let prt = port {
+                    return "\(self.host):\(prt)"
+                } else {
+                    return "\(self.host)"
+                }
+            }
+        }
+    }
+
     /// Parses a new `MongoConnectionString` instance from the provided string.
     /// - Throws:
     ///   - `MongoError.InvalidArgumentError` if the input is invalid.
@@ -46,13 +144,23 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
             throw MongoError.InvalidArgumentError(message: "Invalid connection string")
         }
         let identifiersAndOptions = schemeAndRest[1].components(separatedBy: "/")
+        guard identifiersAndOptions.count <= 2 else {
+            throw MongoError.InvalidArgumentError(message: "Host with unescaped slash")
+        }
         let userAndHost = identifiersAndOptions[0].components(separatedBy: "@")
         guard userAndHost.count <= 2 else {
             throw MongoError.InvalidArgumentError(message: "Invalid user information")
         }
         let userInfo = userAndHost.count == 2 ? userAndHost[0].components(separatedBy: ":") : nil
         let hostString = userInfo != nil ? userAndHost[1] : userAndHost[0]
-        let hosts = try hostString.components(separatedBy: ",").map(ServerAddress.init)
+        let hosts = try hostString.components(separatedBy: ",").map {
+            try HostIdentifier($0, scheme: scheme)
+        }
+        if case .srv = scheme {
+            guard hosts.count == 1 else {
+                throw MongoError.InvalidArgumentError(message: "mongodb+srv with multiple service names")
+            }
+        }
         self.scheme = scheme
         self.hosts = hosts
     }
@@ -85,5 +193,5 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
     public var scheme: Scheme
 
     /// Specifies one or more host/ports to connect to.
-    public var hosts: [ServerAddress]
+    public var hosts: [HostIdentifier]
 }
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -32,7 +32,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         public var description: String { self._scheme.rawValue }
     }
 
-    /// Parses a new `MongoConnectionString` instance from the provided string
+    /// Parses a new `MongoConnectionString` instance from the provided string.
     /// - Throws:
     ///   - `MongoError.InvalidArgumentError` if the input is invalid.
     public init(throwsIfInvalid input: String) throws {
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -36,30 +36,26 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
     /// - Throws:
     ///   - `MongoError.InvalidArgumentError` if the input is invalid.
     public init(throwsIfInvalid input: String) throws {
-        let getScheme = input.components(separatedBy: "://")
-        guard getScheme.count > 1, let scheme = Scheme(getScheme[0]) else {
+        let schemeAndRest = input.components(separatedBy: "://")
+        guard schemeAndRest.count == 2, let scheme = Scheme(schemeAndRest[0]) else {
             throw MongoError.InvalidArgumentError(
                 message: "Invalid connection string scheme, expecting \'mongodb\' or \'mongodb+srv\'"
             )
         }
-        let getUser = getScheme[1].components(separatedBy: "@")
-        let userInfo = getUser.count > 1 ? getUser[0].components(separatedBy: ":") : nil
-        let getHost = getUser.count > 1 ? getUser[1].components(separatedBy: "/") :
-            getUser[0].components(separatedBy: "/")
-        guard let hosts = try? getHost[0].components(separatedBy: ",").map(ServerAddress.init) else {
-            throw MongoError.InvalidArgumentError(message: "Invalid URI Host")
+        guard !schemeAndRest[1].isEmpty else {
+            throw MongoError.InvalidArgumentError(message: "Invalid connection string")
         }
-        let getAuth = getHost.count > 1 ? getHost[1].components(separatedBy: "?") : nil
-        let source = (getAuth?[0] ?? "").isEmpty ? nil : getAuth?[0]
-        let cred = userInfo != nil || getAuth != nil ?
-            MongoCredential(username: userInfo?[0], password: userInfo?[1], source: source) : nil
-        let getOptions = getAuth != nil ? getAuth?[1] : nil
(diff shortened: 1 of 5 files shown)


Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,107 @@
...
+    public init(throwsIfInvalid input: String) throws {
+        let getScheme = input.components(separatedBy: "://")
+        guard getScheme.count > 1, let scheme = Scheme(getScheme[0]) else {
+            throw MongoError.InvalidArgumentError(
+                message: "Invalid URI Schema, expecting \'mongodb://\' or \'mongodb+srv://\'"
+            )
+        }
+        let getUser = getScheme[1].components(separatedBy: "@")
+        let userInfo = getUser.count > 1 ?  getUser[0].components(separatedBy: ":") : nil
+        let getHost = getUser.count > 1 ? getUser[1].components(separatedBy: "/") :
```
bynn, line 47: feels like im repeating some code here `getUser.count > 1`

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,107 @@
...
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// Parses `input` into its constituent parts.
+    /// - Throws:
+    ///   - `MongoError.InvalidArgumentError` if the input is invalid.
+    public init(throwsIfInvalid input: String) throws {
+        let getScheme = input.components(separatedBy: "://")
+        guard getScheme.count > 1, let scheme = Scheme(getScheme[0]) else {
```
patrickfreed, line 40: We actually want to verify that the count is exactly 2, otherwise that means the user included multiple `://`, which I do not believe is allowed. We probably want a separate `guard` for that to give a more accurate error message though. Once we make  that assertion we can also assign a more meaningful variable name to `getScheme[1]`.

Sources/MongoSwift/MongoConnectionString.swift
patrickfreed: for now, let's not worry about userinfo parsing, as that will be handled in the next PR for SWIFT-1161. For this PR, we can assume that any connection strings won't have any userinfo in them.

BTW: the specification includes a reference implementation, which can help guide on the order of splitting and what not: https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#reference-implementation

For example, after we perform step 2, we can assume only "User information and host identifiers" part will be present for the moment, and after we perform step 3 we can assume only the "host identifiers" part will be present and skip step 4 altogether. In later PRs, we can incrementally add logic in for each of the steps to build up the functionality step by step.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /// Specifies the format this connection string is in.
+    public var scheme: Scheme
+
+    /// Specifies one or more host/ports to connect to.
+    public var hosts: [ServerAddress]
+
+    /// Authentication credentials for the `MongoClient`.
+    public var credential: MongoCredential?
```
patrickfreed, line 104: for now, let's omit these last two fields, they'll be addressed in later PRs. The scope of this one is just the basic API and host information.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,107 @@
+/// Represents a MongoDB connection string.
+/// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/
+public struct MongoConnectionString: Codable, LosslessStringConvertible {
```
patrickfreed, line 3: As we build up functionality for this type, we'll want to continually enable more of the connection string spec tests. For this PR in particular, we should be able to pass the subset of tests that only verify host and scheme information (I think this is only `valid-host_identifiers.json`). As part of that, we'll need to update the runner in `ConnectionStringTests.swift` to use `MongoConnectionString` instead of the old `ConnectionString` and comment out the assertions that we're not able to pass yet (e.g. anything checking the options or authentication information). We'll also need to update `shouldSkip` to skip anything we can't pass yet. As we implement more PRs, we can uncomment more assertions and unskip more tests.

bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: deleted

bynn (commented):
i had to add some validation for ServerAddress

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +24,41 @@ public struct ServerAddress: Equatable {
 
     /// Initializes a ServerAddress, using the default localhost:27017 if a host/port is not provided.
     internal init(_ hostAndPort: String = "localhost:27017") throws {
```
patrickfreed, line 26: We'll also want the ability to parse unix domain sockets, but I think we should just punt that to a later PR if we have time at the end of the rotation (they're not super high priority). I've filed SWIFT-1174 to track the work for that.

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +24,41 @@ public struct ServerAddress: Equatable {
...
+            host = String(hostAndPort[swiftRange])
+            let rest = hostAndPort[swiftRange.upperBound...].split(separator: ":")
+            guard let portOptional = rest.count > 1 ? UInt16(rest[1]) : 27017, portOptional > 0 else {
+                throw MongoError.InternalError(message: "couldn't parse address from \(hostAndPort)")
+            }
+            port = portOptional
+        } else {
+            let parts = hostAndPort.components(separatedBy: ":")
+            host = String(parts[0])
+            guard parts.count <= 2,
```
patrickfreed, line 53: I think this ought to be split up to allow us to throw more accurate errors:
- There are multiple `:` (i.e. `parts.count > 2`)
  - We'd want to throw an error indicating that we expected only a single port delimiter / colon 
- A port was provided, but we could not convert it to a port number (i.e. `UInt16(parts[1])` is nil)
  - We'd want to throw an error indicating the port was invalid
- A port was provided but it was 0

We could probably consolidate the second two by saying something like `"port must be a valid, positive unsigned 16 bit integer"`. The first one would need to be on its own though.

patrickfreed (commented):
I have a bunch of mostly minor or stylistic suggestions, but the overall functionality looks really good!

Sources/MongoSwift/SDAM.swift
bynn: done

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -140,12 +148,13 @@ final class ConnectionStringTests: MongoSwiftTestCase {
                 guard !shouldSkip(file: filename, test: testCase.description) else {
                     continue
                 }
+                print(testCase.uri, testCase.valid)
```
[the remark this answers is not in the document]
bynn: deleted

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -203,383 +212,383 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         try self.runTests("connection-string")
     }
 
-    func testAppNameOption() throws {
```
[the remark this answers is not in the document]
bynn: uncommented

bynn (commented):
ServerAddress type is always nill for now. Can we add it to the scope of SWIFT-1174

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +33,45 @@ public struct ServerAddress: Equatable {
...
+            throw MongoError.InvalidArgumentError(message: "\(hostAndPort) contains invalid characters")
         }
+
+        var host: String
+        var port: UInt16
+
+        // Check if host is an IPv6 literal.
+        let ipLiteralRegex = try NSRegularExpression(pattern: #"^\[(.*)\](?::([0-9]+))?$"#)
+        if hostAndPort.first == "[" {
+            guard let match = ipLiteralRegex.firstMatch(
```
patrickfreed, line 46: I think this might be a little easier to read if this `let` was moved to the line below to line up with the next one.
e.g.
```swift
guard
    let match = ...,
    let hostRange = ...
else {
    // ...
}
```

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +33,45 @@ public struct ServerAddress: Equatable {
...
+                in: hostAndPort,
+                range: NSRange(location: 0, length: hostAndPort.utf16.count)
```
patrickfreed, line 48: Strings in Swift are pretty complicated, since they can represent a lot more than just ASCII characters and can be viewed in multiple different encodings. In this case, we're counting how long the UTF-16 encoded version of the string is, but this is most of the time different than how many characters there are in the string. The count of the UTF-8 representation often does reflect the number of characters, but sometimes that is also off if the string contains any multibyte characters. To handle this situation, Swift Strings use abstract character indexes which can be used independent of any encoding of the underlying data. We can use such indexes to define a range that properly spans the entirety of the string, regardless of any weird characters in it.

e.g.
```swift
NSRange(data.startIndex..<data.endIndex, in: data)
```
For more information on this, the documentation of the `String` type has a good overview: https://developer.apple.com/documentation/swift/string

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +33,45 @@ public struct ServerAddress: Equatable {
...
-            throw MongoError.InternalError(message: "couldn't parse address from \(hostAndPort)")
+        guard !hostAndPort.contains("?") else {
+            throw MongoError.InvalidArgumentError(message: "\(hostAndPort) contains invalid characters")
         }
+
+        var host: String
+        var port: UInt16
+
+        // Check if host is an IPv6 literal.
+        let ipLiteralRegex = try NSRegularExpression(pattern: #"^\[(.*)\](?::([0-9]+))?$"#)
```
patrickfreed, line 44: We can move the construction of this regex into the body of the conditional, since that's the only place it'll be used.

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +33,45 @@ public struct ServerAddress: Equatable {
...
+            guard let match = ipLiteralRegex.firstMatch(
+                in: hostAndPort,
+                range: NSRange(location: 0, length: hostAndPort.utf16.count)
+            ),
+                let hostRange = Range(match.range(at: 1), in: hostAndPort)
+            else {
+                throw MongoError.InvalidArgumentError(message: "couldn't parse address from \(hostAndPort)")
+            }
+            host = String(hostAndPort[hostRange])
+            let rest = hostAndPort[hostRange.upperBound...].split(separator: ":")
```
patrickfreed, line 55: The new regex includes the port number in the second capture group, so we can use `match.range(at: 2)` to access it instead of doing further String parsing operations.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -203,6 +211,11 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         try self.runTests("connection-string")
     }
 
+//    func testA() throws {
```
patrickfreed, line 214: I think this was accidentally committed.

Sources/MongoSwift/SDAM.swift
```diff
@@ -3,12 +3,21 @@ import Foundation
...
+        case unix
+    }
+
     /// The hostname or IP address.
     public let host: String
 
     /// The port number.
     public let port: UInt16
 
+    public var type: HostType?
```
patrickfreed, line 19: I'm not sure if this is something we want to expose to users, so can we keep it `private` for now? We can always expose it later if we want to.

Sources/MongoSwift/SDAM.swift
```diff
@@ -3,12 +3,21 @@ import Foundation
 
 /// A struct representing a network address, consisting of a host and port.
 public struct ServerAddress: Equatable {
+    public enum HostType: String {
+        case ipv4
+        case ip_literal
```
patrickfreed, line 8: in Swift, the style for enum cases is generally camelCase. If you need to specify the string version separately for the sake of `Codable` conformance, then you can do so inline:
```swift
enum HostType: String {
    case ipLiteral = "ip_literal"
}
```

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +33,45 @@ public struct ServerAddress: Equatable {
...
+            else {
+                throw MongoError.InvalidArgumentError(message: "couldn't parse address from \(hostAndPort)")
+            }
+            host = String(hostAndPort[hostRange])
+            let rest = hostAndPort[hostRange.upperBound...].split(separator: ":")
+            guard let portOptional = rest.count > 1 ? UInt16(rest[1]) : 27017, portOptional > 0 else {
+                throw MongoError.InvalidArgumentError(message: "couldn't parse address from \(hostAndPort)")
+            }
+            port = portOptional
+            self.type = HostType.ip_literal
```
patrickfreed, line 60: `HostType` can be omitted here due to type inference.

Sources/MongoSwift/SDAM.swift
```diff
@@ -3,12 +3,21 @@ import Foundation
 
 /// A struct representing a network address, consisting of a host and port.
 public struct ServerAddress: Equatable {
+    public enum HostType: String {
+        case ipv4
+        case ip_literal
+        case hostname
+        case unix
```
patrickfreed, line 10: Can this be expanded to `unixDomainSocket` to be more specific?

Sources/MongoSwift/SDAM.swift
```diff
@@ -40,7 +83,11 @@ public struct ServerAddress: Equatable {
 
 extension ServerAddress: CustomStringConvertible {
     public var description: String {
-        "\(self.host):\(self.port)"
+        if self.type == HostType.ip_literal {
```
patrickfreed, line 86: Since we have an enum, it's more common to use a `switch` statement to handle the various cases:
```swift
switch self.type {
case .ipLiteral:
    // do stuff
default:
   // default case
}
```

Sources/MongoSwift/SDAM.swift
patrickfreed, line 19: rather than having this be optional, can we just assign everything that isn't an IP literal to `.hostname`?

patrickfreed (commented):
Looking good! I mostly have minor stylistic comments.

Sources/MongoSwift/SDAM.swift
bynn: done

kmahar (changes requested):
have to run to the doctor so didn't quite make it all the way through, but here are my comments on the first half! looks good, just some minor points

Sources/MongoSwift/SDAM.swift
patrickfreed: This still needs to be updated from the UTF-16 count to use the indexes.

Sources/MongoSwift/SDAM.swift
```diff
@@ -24,11 +33,51 @@ public struct ServerAddress: Equatable {
...
+                    in: hostAndPort,
+                    range: NSRange(location: 0, length: hostAndPort.utf16.count)
+                ),
+                let hostRange = Range(match.range(at: 1), in: hostAndPort)
+            else {
+                throw MongoError.InvalidArgumentError(message: "couldn't parse address from \(hostAndPort)")
+            }
+            host = String(hostAndPort[hostRange])
+            if let portRange = Range(match.range(at: 2), in: hostAndPort) {
+                port = UInt16(hostAndPort[portRange]) ?? 27017
```
patrickfreed, line 57: Similar to in the other branch, we need to be sure to throw errors here instead of defaulting to 27017 when `UInt16(hostAndPort[portRange])` returns nil. Since we'll be doing the same logic in both branches, we may want to consider defining a function for it to reduce code duplication.

Sources/MongoSwift/SDAM.swift
```diff
@@ -3,12 +3,21 @@ import Foundation
...
+        case unixDomainSocket
+    }
+
     /// The hostname or IP address.
     public let host: String
 
     /// The port number.
     public let port: UInt16
 
+    private var type: HostType = .hostname
```
patrickfreed, line 19: Rather than having this be mutable with a default type, let's have it be immutable and just set it in the initializers.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,86 @@
...
+        let hostString = userInfo != nil ? userAndHost[1] : userAndHost[0]
+        let hosts = try hostString.components(separatedBy: ",").map(ServerAddress.init)
+        self.scheme = scheme
+        self.hosts = hosts
+    }
+
+    /// `Codable` conformance
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self)
```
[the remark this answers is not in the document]
bynn: encodes description string and added testCodable

Sources/MongoSwift/SDAM.swift
bynn: created an private static func verifyPort

Sources/MongoSwift/SDAM.swift
```diff
@@ -19,28 +37,72 @@ public struct ServerAddress: Equatable {
...
+        } else {
+            let parts = hostAndPort.components(separatedBy: ":")
+            host = String(parts[0])
+            guard parts.count <= 2 else {
+                throw MongoError.InvalidArgumentError(
+                    message: "expected only a single port delimiter ':' in \(hostAndPort)"
+                )
+            }
+            if parts.count > 1 {
+                port = try ServerAddress.verifyPort(portOptional: UInt16(parts[1]))
```
patrickfreed, line 81: instead of having this function just verify the port, let's have it do both the parsing and verification of the port. That way the `UInt16` portion won't have to be repeated. We could then call it `parsePort(from: String)` or something along those lines.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -203,6 +211,13 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         try self.runTests("connection-string")
     }
 
+    func testCodable() throws {
+        let connStr = try MongoConnectionString(throwsIfInvalid: "mongodb://localhost:27017")
+        let encodedData = try JSONEncoder().encode(connStr)
+        let decodedResult = try JSONDecoder().decode(MongoConnectionString.self, from: encodedData)
+        expect(connStr.description).to(equal(decodedResult.description))
```
patrickfreed, line 218: we also want to add an additional assertion here to ensure the `.description` is working correctly and evaluates to `"mongodb://localhost:27017"`. For example, if `.description` just returned the empty string, this test would currently pass as written.

Sources/MongoSwift/SDAM.swift
bynn: Had to use Substring

Tests/MongoSwiftTests/ConnectionStringTests.swift
bynn: Done

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -110,17 +110,25 @@ let shouldWarnButLibmongocErrors: [String: [String]] = [
...
-        "Valid connection and timeout options are parsed correctly"
-    ],
-    "compression-options.json": ["Multiple compressors are parsed correctly"], // requires Snappy, see SWIFT-894
-    "valid-db-with-dotted-name.json": ["*"] // libmongoc doesn't allow db names in dotted form in the URI
+    "connection-options.json": ["*"],
+    "compression-options.json": ["*"], // requires Snappy, see SWIFT-894
+    "valid-db-with-dotted-name.json": ["*"], // libmongoc doesn't allow db names in dotted form in the URI
+
+    // Disabled for MongoConnectionString
+    "invalid-uris.json": ["*"],
```
patrickfreed, line 120: There are some valid test cases in this file actually, instead of `"*"` could we specify only the tests that verify options / auth / other unsupported behavior? Feel free to update `shouldSkip` to check for substring equality instead of exact equality if you think it'd make defining this skip list a little easier. (e.g. `"invalid-uris.json": [ "option", "username", "password"]`)

Sources/MongoSwift/SDAM.swift
patrickfreed: Just to ping on this again, it looks like this still needs to be updated.

Tests/MongoSwiftTests/ConnectionStringTests.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+            }
+            self.init(_scheme)
+        }
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// A struct representing a network address, consisting of a host and port.
+    public struct HostIdentifier: Equatable, CustomStringConvertible {
+        private static func verifyPort(from: String, scheme: Scheme) throws -> UInt16? {
```
patrickfreed, line 39: I think it's probably simpler if we constrain the functionality of this method to just parsing the port from a string (and we can call it something to reflect that e.g. `parsePort`). As it is now, it's hard to infer from the name of the method and the arguments exactly what this method does.

Similarly, I think we can simplify things by not passing the scheme to the `HostIdentifier` initializer that parses the string, and instead have that initializer only focus on parsing strings. We can then enforce the SRV scheme constraints (i.e. only a single host with no port) in the `MongoConnectionString` initializer, since that's its area of concern.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+            guard scheme != .srv else {
+                throw MongoError.InvalidArgumentError(
+                    message: "mongodb+srv with port number"
```
patrickfreed, line 42: Can we update this error message to be a bit more specific, e.g. `"a port cannot be specified in a mongodb+srv connection string"`

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+            throw MongoError.InvalidArgumentError(message: "Invalid user information")
+        }
+        let userInfo = userAndHost.count == 2 ? userAndHost[0].components(separatedBy: ":") : nil
+        let hostString = userInfo != nil ? userAndHost[1] : userAndHost[0]
+        let hosts = try hostString.components(separatedBy: ",").map {
+            try HostIdentifier($0, scheme: scheme)
+        }
+        if case .srv = scheme {
+            guard hosts.count == 1 else {
+                throw MongoError.InvalidArgumentError(message: "mongodb+srv with multiple service names")
```
patrickfreed, line 161: Ditto updating this message to be more specific. Also, I don't think "service names" is the right term here (I think the description for the test in `invalid-uris.yml` is incorrect). Something like `"only a single host identifier may be specified in a mongodb+srv connection string"` seems good.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+                return nil
+            }
+            self.init(_scheme)
+        }
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// A struct representing a network address, consisting of a host and port.
+    public struct HostIdentifier: Equatable, CustomStringConvertible {
```
patrickfreed, line 38: FYI @kmahar, @bynn and I discussed this offline, but `ServerAddress` currently requires both a host and a port, but in `mongodb+srv` addresses a port is actually forbidden, so we needed to introduce a new type specific to the connection string for representing host identifiers.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+            guard let _scheme = _Scheme(rawValue: description) else {
+                return nil
+            }
+            self.init(_scheme)
+        }
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// A struct representing a network address, consisting of a host and port.
```
patrickfreed, line 37: Now that the port is optional and this struct is a bit more broad in what it represents, we should update the docstring accordingly. It could say something along the lines of 
```
A struct representing a host identifier, consisting of a host and an optional port.
In standard connection strings, this describes the address of a mongod or mongos to connect to.
In mongodb+srv connection strings, this describes a DNS name to be queried for SRV and TXT records.
```

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+                }
+                self.type = .hostname
+            }
+            self.host = host
+            self.port = port
+        }
+
+        public var description: String {
+            switch self.type {
+            case .ipLiteral:
```
patrickfreed, line 117: This could be condensed a little bit to have the `switch` construct the host portion, and then at the end have a single conditional that handles appending the port.

```swift
// because host will be initialized in all branches of the switch, we can leave it uninitialized here
// the Swift compiler will complain if there is a branch that doesn't initialize it
let host: String
switch self.type {
// ...
}
if let port = self.port {
    // ..
}
// ...
```

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,197 @@
...
+            throw MongoError.InvalidArgumentError(
+                message: "Invalid connection string scheme, expecting \'mongodb\' or \'mongodb+srv\'"
+            )
+        }
+        guard !schemeAndRest[1].isEmpty else {
+            throw MongoError.InvalidArgumentError(message: "Invalid connection string")
+        }
+        let identifiersAndOptions = schemeAndRest[1].components(separatedBy: "/")
+        guard identifiersAndOptions.count <= 2 else {
+            throw MongoError.InvalidArgumentError(message: "Host with unescaped slash")
```
patrickfreed, line 148: I don't think this error message quite lines up with the condition it's testing. For instance, the following connection string will yield the error, but the problem isn't really with the host identifier:
```
mongodb://localhost:27017/?hello=world/dog=cat
```
That being said, I think it's more the scope of SWIFT-1174 to improve this, so I'm fine with just leaving it and adding a `TODO` with the ticket number so we remember to improve it.

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// A struct representing a host identifier, consisting of a host and an optional port.
+    /// In standard connection strings, this describes the address of a mongod or mongos to connect to.
+    /// In mongodb+srv connection strings, this describes a DNS name to be queried for SRV and TXT records.
+    public struct HostIdentifier: Equatable, CustomStringConvertible {
+        private static func parsePort(from: String) throws -> UInt16? {
```
patrickfreed, line 41: This never actually returns nil, so it can just return `UInt16` instead of `UInt16?`.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,193 @@
...
+
+        private let type: HostType
+
+        /// Initializes a ServerAddress, using the default localhost:27017 if a host/port is not provided.
+        internal init(_ hostAndPort: String = "localhost:27017") throws {
+            guard !hostAndPort.contains("?") else {
+                throw MongoError.InvalidArgumentError(message: "\(hostAndPort) contains invalid characters")
+            }
+
+            var host: String
```
patrickfreed, line 71: do we need these two variables? Or can we just assign to `self.host` and `self.port` directly?

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,193 @@
...
+        }
+        let userAndHost = identifiersAndOptions[0].components(separatedBy: "@")
+        guard userAndHost.count <= 2 else {
+            throw MongoError.InvalidArgumentError(message: "Invalid user information")
+        }
+        let userInfo = userAndHost.count == 2 ? userAndHost[0].components(separatedBy: ":") : nil
+        let hostString = userInfo != nil ? userAndHost[1] : userAndHost[0]
+        let hosts = try hostString.components(separatedBy: ",").map(HostIdentifier.init)
+        if case .srv = scheme {
+            guard hosts.count == 1, hosts[0].port == nil else {
```
patrickfreed, line 151: since we're checking the port is nil below, we can remove it from this `guard`.

patrickfreed (commented):
Looking really good! I just have a few more little things.

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
bynn: we can assign it directly

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

patrickfreed (approved):
LGTM! Great job with this, especially with handling all the different things that came up post-design.

kmahar (approved):
one suggestion about testing down the line, but this LGTM! great job, and thanks to both you and @patrickfreed for diligently working through all the twists that came up throughout implementing this!

kmahar:
BTW, the failures on the latest Evergreen patch are caused by what I describe/fix in #628. since they're unrelated to your work you're good to merge this

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -203,6 +211,13 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         try self.runTests("connection-string")
     }
 
+    func testCodable() throws {
```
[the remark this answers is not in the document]
bynn: I didn't add a specific ticket because I think all the tests will be able to use MongoConnectionString at different times

PR #625, opened 2021-04-02
SWIFT-1025 Integrate with libmongoc versioned API and add test code

kmahar:
This PR starts actually running all of the versioned API tests, and additionally updates our Evergreen configuration to add two new tasks around versioned API. 

[Here](https://spruce.mongodb.com/version/60666759d6d80a29b099a909/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) is a sample patch. 

Caveat: that patch was also run with changes with mongodb/mongo-c-driver#770 vendored in. However, I am not bringing those into our codebase yet since they are still going through code review.

That said, I figured I would go ahead and put this up for review so it is ready to go once the libmongoc work is all in.


.evergreen/config.yml
```diff
@@ -109,6 +109,7 @@ functions:
             TOPOLOGY=${TOPOLOGY} \
             SSL=${SSL} \
             AUTH=${AUTH} \
+            REQUIRE_API_VERSION=${REQUIRE_API_VERSION} \
```
kmahar, line 112: these changes were heavily cribbed from the Python driver's Evergreen config.

.evergreen/config.yml
```diff
@@ -157,6 +158,8 @@ functions:
           SSL=${SSL} \
           AUTH=${AUTH} \
           SWIFT_VERSION=${SWIFT_VERSION} \
+          MONGODB_API_VERSION=${MONGODB_API_VERSION} \
+          TEST_FILTER=${TEST_FILTER} \
```
kmahar, line 162: this wasn't strictly necessary, as we could have just run the whole test suite with `acceptAPIVersion2` on, but it seemed redundant to do so.

.evergreen/config.yml
```diff
@@ -1055,6 +1090,17 @@ buildvariants:
       # macOS MongoDB servers do not staple OCSP responses and only support RSA.
       - name: ".ocsp-rsa !.ocsp-staple"
 
+- matrix_name: "versioned-api-tests"
+  matrix_spec:
+    os-fully-featured: ubuntu-18.04
```
kmahar, line 1095: since the point is just to test versioned API, it seemed probably fine to me to limit this to a single platform and Swift version. what do you think?

etc/convert-test-results.swift
```diff
@@ -85,7 +85,8 @@ let ignoreSuites = [
     "mongo-swift-driverPackageTests.xctest",
     // linux
     "All tests",
-    "debug.xctest"
+    "debug.xctest",
+    "Selected tests" // this shows us when --filter is used
```
kmahar, line 89: this accounts for the test output being slightly different when the `--filter` parameter is used and ensures we still parse the results into XML correctly

.evergreen/config.yml
patrickfreed: Seems fine to me

.evergreen/run-tests.sh
```diff
@@ -47,7 +48,14 @@ swift build
...
 set -o pipefail # propagate error codes in the following pipes
 
-MONGODB_TOPOLOGY=${TOPOLOGY} MONGODB_URI=$MONGODB_URI swift test 2>&1 | tee ${RAW_TEST_RESULTS}
+# construct an optional filter statement for tests
+FILTER_STATEMENT=""
+if [ "$TEST_FILTER" != "NO_FILTER" ]; then
+    FILTER_STATEMENT="--filter ${TEST_FILTER}"
+fi
+
+echo $FILTER_STATEMENT
```
patrickfreed, line 57: leftover debug print?

patrickfreed (approved):
LGTM bar one debug echo

bynn (approved):
lgtm

.evergreen/run-tests.sh
kmahar: good catch, removed

kmahar:
red on evg, but that's because testing "accept API version 2" is currently broken for all drivers as the server team renamed the parameter from `acceptAPIVersion2` to `acceptApiVersion2`.

PR #626, opened 2021-04-02
SWIFT-1099 Sync tests for changing estimatedDocumentCount to use the $collStats aggregation stage

kmahar:
Syncs new CRUD and retryable reads test files and makes some small changes to support running them in the test runner.


patrickfreed (approved):

Sources/MongoSwift/Operations/EstimatedDocumentCountOperation.swift
```diff
@@ -24,7 +24,7 @@ public struct EstimatedDocumentCountOptions: Codable {
         self.readPreference = readPreference
     }
 
-    private enum CodingKeys: String, CodingKey {
+    internal enum CodingKeys: String, CodingKey, CaseIterable {
```
bynn, line 27: Is there a reason we want it to be internal now?

bynn (approved):

Sources/MongoSwift/Operations/EstimatedDocumentCountOperation.swift
kmahar: good question. I am using this type to generate a set of "known arguments" for estimatedDocumentCount. the unified test runner requires we error if any arguments specified for an operation.
to access this type from the tests, `internal` is necessary (when `private`, it was only accessible within `EstimatedDocumentCountOptions` itself)

PR #629, opened 2021-04-08
SWIFT-1090 Add code examples for versioned API docs

kmahar:
https://jira.mongodb.org/browse/SWIFT-1090

Using my other PR as the base branch since I developed/tested these atop that.

### DIFF
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -225,3 +225,51 @@ private func transactions() throws {
     }.wait()
     // End Transactions Retry Example 3
 }
+
+private func versionedAPI() throws {
+    let uri = "mongodb://localhost:27017"
+    let myEventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    defer { try? myEventLoopGroup.syncShutdownGracefully() }
+
+    do {
+        // Start Versioned API Example 1
+
+        // Declare API version "1" for the client
+        var clientOpts = MongoClientOptions()
+        clientOpts.serverAPI = MongoServerAPI(version: .v1)
+        let client = try MongoClient(uri, using: myEventLoopGroup, options: clientOpts)
+        // End Versioned API Example 1
+
+        try client.syncClose()
+    }
+
+    do {
+        // Start Versioned API Example 2
+
+        // Use the `strict` option
+        var opts = MongoClientOptions()
+        opts.serverAPI = MongoServerAPI(version: .v1, strict: true)
+        let client = try MongoClient(uri, using: myEventLoopGroup, options: opts)
+
+        var findOpts = FindOptions()
+        findOpts.cursorType = .tailable
+        // Fails with an error because `tailable` is not part of version 1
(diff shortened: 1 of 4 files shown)


Sources/MongoSwiftSync/MongoDatabase.swift
```diff
@@ -231,6 +231,9 @@ public struct MongoDatabase {
      *   - `MongoError.WriteError` if any error occurs while the command was performing a write.
      *   - `MongoError.CommandError` if an error occurs that prevents the command from being performed.
      *   - `EncodingError` if an error occurs while encoding the options to BSON.
+     *
+     * - Note: Attempting to specify an API version in this command is considered undefined behavior. API version may
```
kmahar, line 235: I already added this in the async API, but I forgot to copy it to the sync

Examples/Docs/Sources/AsyncExamples/main.swift
```diff
@@ -225,3 +225,54 @@ private func transactions() throws {
...
+        // Use the `strict` option
+        var opts = MongoClientOptions()
+        opts.serverAPI = MongoServerAPI(version: .v1, strict: true)
+        let client = try MongoClient(uri, using: myEventLoopGroup, options: opts)
+
+        var findOpts = FindOptions()
+        findOpts.cursorType = .tailable
+        // Fails with an error because `tailable` is not part of version 1
+        do {
+            let cursor = try client.db("db").collection("coll").find(options: findOpts).wait()
```
patrickfreed, line 258: would it be better to write this using a `.whenFailure` callback to more closely simulate async error handling?

Examples/Docs/Sources/SyncExamples/main.swift
```diff
@@ -165,3 +165,42 @@ private func transactions() throws {
...
+
+        // Use the `strict` option
+        var opts = MongoClientOptions()
+        opts.serverAPI = MongoServerAPI(version: .v1, strict: true)
+        let client = try MongoClient(uri, options: opts)
+
+        var findOpts = FindOptions()
+        findOpts.cursorType = .tailable
+        // Fails with an error because `tailable` is not part of version 1
+        let cursor = try client.db("db").collection("coll").find(options: findOpts)
```
patrickfreed, line 193: `try`/`catch` here to match the previous example?

patrickfreed (commented):
Just have a few comments on the examples, and this should be good to rebase off of `main` now.

Examples/Docs/Sources/SyncExamples/main.swift
kmahar: updated (ironically, I removed the try catch from the other example because of your comment there)

Examples/Docs/Sources/AsyncExamples/main.swift
kmahar: updated. I added a `try? cursor.wait()` down after the example to ensure the future actually finishes before we close the client. I debated putting it in the example, but it didn't seem particularly relevant

patrickfreed (approved):
examples LGTM, looks like the linter is complaining though (no need to review those fixes)

kmahar:
sigh... my hubris not re-running the linter... fixed

PR #630, opened 2021-04-09
SWIFT-1168 Vendor libmongoc 1.18.0-alpha

kmahar:
Vendors in libmongoc's 1.18.0 alpha.

In my [patch](https://evergreen.mongodb.com/version/606f84d430661542ec102870) I encountered some random timeout errors in various tests where we were using maxTimeMS on macOS.... Some of these tests I see also failing occasionally prior to the upgrade on the waterfall.  
The maxTimeMS value choices seemed arbitrary in all of these tests so I just increased them in an attempt to make them less flaky. In any case, since maxTimeMS is purely a server-side option I don't think this has anything to do with the upgrade.


etc/lower-minheartbeatfrequencyms.diff
```diff
@@ -1,5 +1,5 @@
 diff --git a/Sources/CLibMongoC/include/CLibMongoC_mongoc-client-pool.h b/Sources/CLibMongoC/include/CLibMongoC_mongoc-client-pool.h
-index 16d7eb1..bcf1cde 100644
+index 16d7eb1a..bcf1cdee 100644
```
kmahar, line 2: this patch needed to be regenerated due to changes in `mongoc-client-pool.{h, c}`

patrickfreed (approved):

PR #631, opened 2021-04-10
SWIFT-1161 MongoConnectionString authentication options support

bynn:
This currently doesn't have the accepted behavior for specifying the database and the authSource and I wanted to get some feedback on how I can improve the design


Tests/MongoSwiftTests/AuthTests.swift
```diff
@@ -27,18 +27,18 @@ final class AuthTests: MongoSwiftTestCase {
 
         for (_, file) in testFiles {
             for testCase in file.tests {
+                print(testCase.uri, testCase.valid)
```
bynn, line 30:
```suggestion
```

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -119,6 +119,31 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         }
     }
 
+    private static func percentDecode(from: String?) throws -> String? {
+        if from == nil {
```
kmahar, line 123: style nit: I think using `guard let` to unwrap `from` would be more idiomatic here:
```swift
guard let from = from else {
   return nil
}
// use non-optional `from` down here...
```

generally, when you are checking some precondition for a parameter that is necessary for the rest of the method to execute, `guard` is appropriate

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -119,6 +119,31 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+            return nil
+        }
+        let decodedOptional = from?.removingPercentEncoding
+        guard let decoded = decodedOptional else {
+            throw MongoError.InvalidArgumentError(
+                message: "Contains an unescaped percent sign"
```
kmahar, line 129:
```suggestion
                message: "Connection string contains an unescaped percent sign"
```

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -119,6 +119,31 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         }
     }
 
+    private static func percentDecode(from: String?) throws -> String? {
+        if from == nil {
+            return nil
+        }
+        let decodedOptional = from?.removingPercentEncoding
```
kmahar, line 126: anther nit - you can one-line this, too, and save a variable:
`guard let decoded = from.removingPercentEncoding else {`

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -189,4 +358,9 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
 
     /// Specifies one or more host/ports to connect to.
     public var hosts: [HostIdentifier]
+
+    /// Returns the credential configured on this URI. Will be empty if no options are set.
+    public var credential: MongoCredential?
+
+    public var database: String?
```
kmahar, line 365: this property needs a docstring

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -156,8 +181,109 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+                guard keyValue.count == 2 else {
+                    throw MongoError.InvalidArgumentError(
+                        message: "Invalid key value pair for option \(keyValue)"
+                    )
+                }
+                let key = keyValue[0]
+                let value = keyValue[1]
+                switch key {
+                // authentication options.
+                case "authSource":
```
kmahar, line 220: related to my comment above, perhaps you could break out parsing auth options into a separate helper method, and figure out a way to consolidate the repeated check for whether `self.credential != nil`. 

I think what you're doing here to avoid multiple linear scans through the array is a good idea, so maybe you could continue supporting that by e.g. parsing out options into a `[String: String]` that you pass around into helpers that parse specific options?

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -180,7 +306,50 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
         des += "\(self.scheme)://"
+        if let username = self.credential?.username {
+            des += username
+            if let password = self.credential?.password {
+                des += ":\(password)"
+            }
+            des += "@"
+        }
         des += self.hosts.map { $0.description }.joined(separator: ",")
+//        if let database = self.database {
```
kmahar, line 317: should this be uncommented?

kmahar (changes requested):

Sources/MongoSwift/MongoConnectionString.swift
bynn: authOptions is just a temporary variable but is accessible if needed 
options will not be applied and verified in separate functions

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -156,8 +183,115 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                     message: "A port cannot be specified in a mongodb+srv connection string")
             }
         }
+        let authAndOptions = identifiersAndOptions.count == 2 ?
```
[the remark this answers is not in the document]
bynn: applyAuthOptions created

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -119,6 +119,82 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+                    guard propertyKeyValue.count == 2 else {
+                        throw MongoError.InvalidArgumentError(
+                            message: "Invalid key value pair for authMechanismProperties \(propertyKeyValue)"
+                        )
+                    }
+                    let propertyKey = propertyKeyValue[0]
+                    let propertyValue = propertyKeyValue[1]
+                    switch propertyKey {
+                    case "SERVICE_NAME", "SERVICE_REALM":
+                        if let cred = self.credential?.mechanism {
```
kmahar, line 157: what if we process the auth mechanism properties before we process the mechanism option? could we run into an issue?

kmahar (changes requested):
thanks for making those changes! my next batch of comments is mainly around trying to make the logic flow a little more clear for future readers of the code, and ensuring we're handling edge cases

Sources/MongoSwift/MongoConnectionString.swift
bynn: done

Tests/MongoSwiftTests/AuthTests.swift
```diff
@@ -21,37 +21,65 @@ struct AuthTestCase: Decodable {
     let credential: MongoCredential?
 }
 
+extension MongoCredential {
+    func matches(testCredential: MongoCredential, description: String) throws {
+        var connStringCredential = self
```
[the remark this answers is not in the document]
bynn: done

Tests/MongoSwiftTests/AuthTests.swift
```diff
@@ -21,37 +21,65 @@ struct AuthTestCase: Decodable {
...
+        var credential = testCredential
+
+        expect(connStringCredential).toNot(beNil(), description: description)
+
+        // Ignore default properties for now.
```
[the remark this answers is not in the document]
bynn: done

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -180,7 +308,45 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+            var properties: [String] = []
+            // TODO: SWIFT-1177 Remove after BSON conforms to CustomStringConvertible
+            for (k, v) in mechanismProperty {
+                if let value = v.stringValue {
+                    properties.append("\(k):\(value)")
+                } else if let value = v.boolValue {
+                    properties.append("\(k):\(String(value))")
+                }
+            }
+            options.append("authMechanismProperties=\(properties.joined(separator: ","))")
```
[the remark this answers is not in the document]
bynn: now checks for empty list

kmahar:
we're closing this PR now as it now longer merges cleanly, however we will keep this branch around for when we have time to revisit the project, so we can use Bynn's work.

PR #632, opened 2021-04-12
Use ExtendedJSONEncoder in MongoConnectionString.testCodable

bynn:
https://spruce.mongodb.com/version/607064580305b974a7e7bb3e/tasks

this is the evergreen patch and there still seems to be an issue on oscp but I think patrick said it was okay


kmahar (approved):
lgtm! yeah those OCSP tests are known failures. there's also a weird sharded cluster timeout one on that patch but it seems unrelated to this change.

patrickfreed (approved):
Changes LGTM. though could you update the PR title / commit message to more specifically describe the fix/issue? 

e.g. PR title could be something like "Use ExtendedJSONEncoder in MongoConnectionString.testCodable" and then in the commit message explain that the switch was made because JSONEncoder doesn't support encoding fragments in older Swift versions.

PR #633, opened 2021-04-30
SWIFT-477: standardize copyright formatting

agolin95:
Recently did PHPC-1792 so thought I'd keep on the roll and try and close out this old ticket. Standardized everything to <year>-present and cleaned up a few miscellaneous formatting inconsistencies. 

Lots of files but all super tiny changes, hoping it will be _fairly_ painless to do a visual scroll through.

### DIFF
--- a/Sources/CLibMongoC/bson/bcon.c
+++ b/Sources/CLibMongoC/bson/bcon.c
@@ -3,7 +3,7 @@
  * @brief BCON (BSON C Object Notation) Implementation
  */
 
-/*    Copyright 2009-2013 MongoDB, Inc.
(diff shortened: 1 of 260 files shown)


kmahar:
Hey @agolin95 ! Thanks for taking this on. 

Unfortunately, all of the files you've updated are libmongoc files. Any time we vendor in a new version of the C driver, these changes are going to get overwritten. I'd argue probably the right thing to do here is to make this PR to libmongoc. Then next time we upgrade we will pull in the files with the correct copyright formatting. WDYT? 

Arguably we could add the copyright notices to our own files, but I'm inclined not to as the license is already in the repository itself.

PR #635, opened 2021-05-14
SWIFT-1116 Sync in some unified runner tests

kmahar:
This brings in the test added by DRIVERS-1538 as well as a couple other newish test files/test file changes. I did not bother syncing anything that our runner isn't able to run yet due to schema version, or any of the "invalid" test files since we don't actually do anything with those. (Arguably we should delete those altogether since they exist for testing the schema, not testing runners?)

### DIFF
--- a/Tests/Specs/unified-test-format/tests/valid-fail/entity-client-apiVersion-unsupported.json
+++ b/Tests/Specs/unified-test-format/tests/valid-fail/entity-client-apiVersion-unsupported.json
@@ -1,0 +1,20 @@
+{
+  "description": "entity-client-apiVersion-unsupported",
+  "schemaVersion": "1.1",
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "serverApi": {
+          "version": "server_will_never_support_this_api_version"
+        }
+      }
+    }
+  ],
+  "tests": [
+    {
+      "description": "foo",
+      "operations": []
+    }
+  ]
+}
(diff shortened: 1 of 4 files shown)


Tests/Specs/unified-test-format/tests/valid-fail/entity-client-apiVersion-unsupported.json
```diff
@@ -0,0 +1,20 @@
+{
```
kmahar, line 1: I should have synced this file back when implementing versioned API but I neglected to ( we can't actually run this anyway, but it seemed good to catch up)

Tests/Specs/unified-test-format/tests/valid-pass/poc-retryable-writes.json
```diff
@@ -297,13 +297,15 @@
...
           },
           "expectResult": {
-            "insertedCount": {
-              "$$unsetOrMatches": 2
-            },
-            "insertedIds": {
-              "$$unsetOrMatches": {
-                "0": 3,
-                "1": 4
+            "$$unsetOrMatches": {
```
kmahar, line 300: this seems to be a change that was made for CSHARP purposes that never had a corresponding DRIVERS ticket for everyone to sync
(https://github.com/mongodb/specifications/commit/ec947b3700c38c2de80d6dff79921416983200fe)

Tests/Specs/unified-test-format/tests/valid-fail/operation-failure.json
```diff
@@ -0,0 +1,56 @@
+{
```
kmahar, line 1: this is the test that this ticket was actually asking us to sync

kmahar:
@patrickfreed I think this might have slipped through the cracks

patrickfreed (approved):

PR #637, opened 2021-06-07
SWIFT-1208, SWIFT-1209, SWIFT-1221, SWIFT-1005 Various Evergreen configuration updates 

kmahar:
This PR addresses a backlog of necessary Evergreen changes that have piled up recently:
- Since MongoDB 5.0 is no longer “latest” we need to test it separately
- Swift 5.4 has been released, so test against that and start testing against 5.5 development snapshots
- Stop testing on Ubuntu 16.04 and start testing on Ubuntu 20.04 (note that as discussed we don't test on < 4.4 on 20.04 now due to lack of specific builds and the end of generic builds w/ 4.0)
- Skip OCSP stapling tests on Ubuntu 18.04
- Use macOS 10.15 for all tests since it supports all of our Swift versions

complete patch: https://evergreen.mongodb.com/version/60baef80d1fe070beb0641c8

It looks kind of bad, but (I think) all the failures are expected:
- All tasks testing against 5.0 and latest, and the versioned API-specific tests, are red, which is also true on the waterfall already and is because we are missing recent C driver work to address late-breaking changes around versioned API, and also because https://jira.mongodb.org/browse/DRIVERS-1781 is still being worked on to change now-failing command monitoring tests.
- OCSP on macOS
    - The "soft fail" tests have frequently failed on macOS (SWIFT-1134)
    - The other failures are OCSP test 3, "valid cert + server that does not staple", which CDRIVER-3759 also notes to be flaky on macOS (though maybe it's not merely flaky and is just straight up broken on 10.15?)

Let me know what you think about the OCSP ones as I'm not super familiar with those tests.

As an aside, we probably want to vendor in libmongoc changes sooner rather than later to get things more green.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1087,16 +1087,16 @@ buildvariants:
      - ".4.0"
      - ".3.6"
   rules:
-  # pre 4.0 we are just using legacy linux mongoDB since ubuntu 18.04 didn't
-  # exist when 3.6 came out. legacy doesn't link to OpenSSL so we can't run
-  # SSL tests there.
+  # MongoDB 3.6 does not have Ubuntu 18.04-specific Linux releases, so we just use
+  # generic Linux there. This do not link to OpenSSL, so we have to skip SSL tests.
   - if:
       os-fully-featured:
         - "ubuntu-18.04"
       swift-version: "*"
       ssl-auth: "ssl-auth"
     then:
-      remove_tasks: ".3.6"
+      remove_tasks:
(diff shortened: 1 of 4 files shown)


patrickfreed (approved):
Your evaluation of the macOS OCSP failures seems reasonable. Might be worth investigating further on a GBF now that the failures are more consistent.

kmahar:
I updated https://jira.mongodb.org/browse/SWIFT-1134 to discuss test 3 as well and added the greenerbuild label.

PR #638, opened 2021-06-15
SWIFT-1232 Vendor libmongoc 1.18.0-alpha2

kmahar:
As usual, you can ignore all the `.c` and `.h` files. Some minor test changes were required, I'll elaborate inline 

glorious fully green patch: https://evergreen.mongodb.com/version/60c7f98d2fbabe562ad9b25c

this will allow us to close out the following SWIFT tickets:
-  SWIFT-1173: Use "hello" command when API Version is declared 
- SWIFT-1192: Require passing Versioned API options to getMore and transaction-continuing commands
- SWIFT-1204: Remove slaveOK connection string option from URI parsing


Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -189,7 +189,7 @@ extension SpecTestFile {
         // if connected to a replica set, use the provided client to execute killAllSessions on the primary.
         // if connected to a sharded cluster, use the per-mongos clients to execute killAllSessions on each mongos.
         switch try client.topologyType() {
-        case .single:
+        case .single, .loadBalancer:
```
kmahar, line 192: it's not clear to me what the right behavior would be to terminate all open transactions when testing against a load balancer, and isn't mentioned in the unified test runner spec. so for now I'm assuming this is a no-op, but maybe when implementing LB support we'll need to do something else here

Tests/Specs/versioned-api/tests/crud-api-version-1-strict.json
```diff
@@ -651,7 +651,7 @@
       ]
     },
     {
-      "description": "find command with declared API version appends to the command, but getMore does not",
```
kmahar, line 654: the old version of this test was failing, so syncing was required. however, the test file as written was schema version 1.4 and required support for your serverless runOn requirement. I didn't want to re-implement and conflict with whatever you've done for that, so I just manually edited this file to only include the change to this test case. I assume as part of the serverless PR you'll need to bring in the updated versions of these files anyway(?)

Tests/Specs/versioned-api/tests/crud-api-version-1.json
```diff
@@ -643,7 +643,7 @@
       ]
     },
     {
-      "description": "find command with declared API version appends to the command, but getMore does not",
+      "description": "find and getMore append API version",
```
kmahar, line 646: ditto this file with the manual edits

Tests/Specs/versioned-api/tests/transaction-handling.json
```diff
@@ -6,7 +6,8 @@
       "minServerVersion": "4.9",
       "topologies": [
         "replicaset",
-        "sharded-replicaset"
```
kmahar, line 9: curiously, this file did not have its schema version bumped, even though I think support for load-balanced topology probably would have warranted that? in any case, this file was just synced directly.

Tests/Specs/versioned-api/tests/crud-api-version-1-strict.json
patrickfreed: Yep, after I rebase off of this patch I'll just make sure these tests are synced up.

Tests/Specs/versioned-api/tests/transaction-handling.json
patrickfreed: Yeah this should be bumped to 1.3 for load balancer I believe. Probably worth filing a PR/DRIVERS ticket for that.

patrickfreed (approved):
LGTM!

Tests/Specs/versioned-api/tests/transaction-handling.json
kmahar: opened mongodb/specifications#1023

PR #644, opened 2021-07-09
SWIFT-1217 Revert "SWIFT-1160 Introduce new MongoConnectionString type (#623)"

kmahar:
Since there was a single commit which cleanly reverted, that seemed like the most straightforward way to accomplish SWIFT-1217, which needs to be done before we can tag a release with any 5.0 features.

We could arguably leave `MongoConnectionString.swift` in the codebase and make the type private but I'm not sure if it makes much of a difference since we can always just copy that file back in whenever we actually need it.


kmahar:
One of the change streams tests failed against a sharded cluster + latest. I am unsure of the cause but I am able to reproduce it locally on `main` so I don't believe it has any relation to this and is llkely the result of some server change.

patrickfreed (approved):

PR #645, opened 2021-07-13
Skip testChangeStreamAutomaticResume against latest servers

kmahar:
This is failing due to CDRIVER-4077. I could have changed the test to test this in a different way e.g. explicitly killing the cursor via runCommand and ensuring a resume happens, however Kevin said he thinks his current work for load balancer (specifically CDRIVER-3653) may fix this issue, so I figured we can wait and try that first before rewriting the test.

### DIFF
--- a/Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
+++ b/Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
@@ -434,6 +434,8 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
      */
     func testChangeStreamAutomaticResume() throws {
         let testRequirements = TestRequirement(
+            // TODO SWIFT-1257: remove server version requirement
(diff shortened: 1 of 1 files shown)


Tests/MongoSwiftSyncTests/SyncChangeStreamTests.swift
```diff
@@ -434,6 +434,8 @@ final class SyncChangeStreamTests: MongoSwiftTestCase {
      */
     func testChangeStreamAutomaticResume() throws {
         let testRequirements = TestRequirement(
+            // TODO SWIFT-1257: remove server version requirement
+            maxServerVersion: ServerVersion(major: 4, minor: 9, patch: 0),
```
kmahar, line 438: technically 5.0 is fine, the problem is just on latest, however as you recently noted in driver-devs the version numbers are currently wrong for the server's master branch so I think we just have to skip on 5.0 too

patrickfreed (approved):
LGTM, but looks like swiftformat needs to be run again

PR #649, opened 2021-07-30
SWIFT-1249 / SWIFT-1269 Skip serverless tests that use disabled features

patrickfreed:
SWIFT-1249 / SWIFT-1269

Serverless removed support for some features (`$out`, collation), so we needed to skip some tests and sync some spec tests.

### DIFF
--- a/Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
+++ b/Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
@@ -96,6 +96,13 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
             options.hidden = true
         }
 
+        // some options not supported by serverless
+        if MongoSwiftTestCase.serverless {
+            options.storageEngine = nil
+            options.collation = nil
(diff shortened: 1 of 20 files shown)


Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
```diff
@@ -113,6 +120,7 @@ final class MongoCollection_IndexTests: MongoSwiftTestCase {
         // testOptions index
         var expectedTestOptions = options
         expectedTestOptions.name = "testOptions"
+        expectedTestOptions.version = 2
```
kmahar, line 123: so serverless won't let you specify an index version but returns one either way?

kmahar (approved):
one question about changes, lgtm though!

Tests/MongoSwiftSyncTests/MongoCollection+IndexTests.swift
patrickfreed: Yeah that appears to be the case.

PR #650, opened 2021-07-31
SWIFT-1107 Sync "Mitigate pain of using field names with dots and dollars" tests

kmahar:
I already synced some of these tests as part of vendoring in libmongoc 1.18; this syncs the rest and makes some related test runner changes.

### DIFF
--- a/Tests/MongoSwiftSyncTests/CrudTests.swift
+++ b/Tests/MongoSwiftSyncTests/CrudTests.swift
@@ -75,6 +75,9 @@ final class CrudTests: MongoSwiftTestCase {
                 // TODO: SWIFT-560 unskip these files
                 "bulkWrite-updateMany-dots_and_dollars.json",
                 "bulkWrite-updateOne-dots_and_dollars.json",
+                "updateMany-dots_and_dollars.json",
+                "updateOne-dots_and_dollars.json",
+                "findOneAndUpdate-dots_and_dollars.json",
                 "updateWithPipelines.json"
             ],
             asType: UnifiedTestFile.self
--- a/Tests/MongoSwiftSyncTests/CrudTests.swift
+++ b/Tests/MongoSwiftSyncTests/CrudTests.swift
@@ -68,11 +68,19 @@ final class CrudTests: MongoSwiftTestCase {
     }
 
     func testCrudUnified() throws {
-        let files = try retrieveSpecTestFiles(specName: "crud", subdirectory: "unified", asType: UnifiedTestFile.self)
+        let files = try retrieveSpecTestFiles(
+            specName: "crud",
+            subdirectory: "unified",
+            excludeFiles: [
+                // TODO: SWIFT-560 unskip these files
+                "bulkWrite-updateMany-dots_and_dollars.json",
+                "bulkWrite-updateOne-dots_and_dollars.json",
+                "updateWithPipelines.json"
+            ],
+            asType: UnifiedTestFile.self
+        )
         let runner = try UnifiedTestRunner()
         let skipList: [String: [String]] = [
-            // TODO: SWIFT-560 unskip this test
-            "updateWithPipelines": ["*"],
             // libmongoc chose not to implement CDRIVER-3630 in anticipation of DRIVERS-1340,
             // so we cannot pass these tests for now.
             // TODO: DRIVERS-1340 unskip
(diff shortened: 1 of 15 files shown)


Tests/MongoSwiftSyncTests/CrudTests.swift
```diff
@@ -68,11 +68,22 @@ final class CrudTests: MongoSwiftTestCase {
     }
 
     func testCrudUnified() throws {
-        let files = try retrieveSpecTestFiles(specName: "crud", subdirectory: "unified", asType: UnifiedTestFile.self)
+        let files = try retrieveSpecTestFiles(
+            specName: "crud",
+            subdirectory: "unified",
+            excludeFiles: [
```
kmahar, line 74: when implementing the unified runner, I chose to just silently ignore any cases where update pipelines were used while decoding, setting them to an empty document (see changes in files below), and rely on the skip list below to not actually run the tests.

when syncing in a number of new tests using update pipelines as part of this work, this made it very challenging to debug the various tests that were failing, as update commands were being sent to the server but with an empty update document, which led to a variety of different failures in those tests.

I've now switched to erroring if an array pipeline is encountered, and just skipping all of the relevant files here so we don't bother trying to decode them. every file utilizing update pipelines is exclusively composed of tests using update pipelines, so we aren't losing any coverage by taking the "skip the whole file" approach.

relatedly, I'm wondering if we want to think about prioritizing SWIFT-560 since update pipelines have been coming up in spec tests a lot lately and we're the only driver to not yet support them, so I've moved it back into needs triage

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedOperation.swift
```diff
@@ -18,6 +18,11 @@ enum UnifiedOperationResult {
     case rootDocumentArray([BSONDocument])
     case none
 
+    /// Some tests for unacknowledged writes assume that the result is non-nil with an "acknowledged" field
```
kmahar, line 21: I considered adding special matching logic for when the "acknowledged" field is expected, but that seemed messier. I also considered just returning empty documents in those cases, since that would also satisfy the "unset or matches" operator, but it seemed a little nicer to use a document like this so it's clear what the return value is indicating.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCollectionOperations.swift
```diff
@@ -108,10 +108,10 @@ struct UnifiedBulkWrite: UnifiedOperationProtocol {
         let collection = try context.entities.getEntity(from: object).asCollection()
         let session = try context.entities.resolveSession(id: self.session)
         guard let result = try collection.bulkWrite(self.requests, options: self.options, session: session) else {
-            return .none
+            return .unacknowledgedWrite
         }
         let encodedResult = try BSONEncoder().encode(result)
-        return .bson(.document(encodedResult))
+        return .rootDocument(encodedResult)
```
kmahar, line 114: rereading the spec, it turns out a number of operations should have had their results considered root documents. I think I may have missed the "including, but not limited to" part [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#allowing-extra-fields-in-root-level-documents). 

a couple of those cases led to failures with the new tests, so I went ahead and updated all of these.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCollectionOperations.swift
```diff
@@ -108,10 +108,10 @@ struct UnifiedBulkWrite: UnifiedOperationProtocol {
         let collection = try context.entities.getEntity(from: object).asCollection()
         let session = try context.entities.resolveSession(id: self.session)
         guard let result = try collection.bulkWrite(self.requests, options: self.options, session: session) else {
-            return .none
+            return .unacknowledgedWrite
```
kmahar, line 111: see comment in file below for explanation of this new value.

Tests/MongoSwiftSyncTests/CrudTests.swift
patrickfreed: This approach to incompatible formats seems like a nice improvement. And sounds good about the update pipeline work.

patrickfreed (approved):

PR #652, opened 2021-08-04
SWIFT-1206, SWIFT-1266: Update and expand versioned API examples

kmahar:
We already had some versioned API examples from back around the time of the serverless beta. This updates them to match the form requested in SWIFT-1206, as well as adds a new example requested by SWIFT-1266.

### DIFF
--- a/Examples/Docs/Sources/AsyncExamples/main.swift
+++ b/Examples/Docs/Sources/AsyncExamples/main.swift
@@ -312,8 +312,8 @@ private func versionedAPIMigrationExample() throws {
 
     try insertResult.wait()
 
-    // 2. Run "count" using a strict client, observe error
-    let countResult = client.db("test").runCommand(["count": "sales"])
+    // 2. After insert is complete, run "count" using a strict client, observe error
+    let countResult = insertResult.flatMap { _ in client.db("test").runCommand(["count": "sales"]) }
 
     countResult.whenFailure { error in
         print(error) // prints:
@@ -328,8 +328,8 @@ private func versionedAPIMigrationExample() throws {
 
     try? countResult.wait()
 
-    // 3. New way to count documents
-    let newCountResult = client.db("test").collection("sales").countDocuments()
+    // 3. New way to count documents: after insert is complete, use countDocuments()
+    let newCountResult = insertResult.flatMap { _ in client.db("test").collection("sales").countDocuments() }
     newCountResult.whenSuccess { result in
         print(result) // prints: 8
     }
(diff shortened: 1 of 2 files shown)


Examples/Docs/Sources/AsyncExamples/main.swift
```diff
@@ -246,34 +247,94 @@ private func versionedAPI() throws {
...
+        )
         let client = try MongoClient(uri, using: myEventLoopGroup, options: opts)
+
         // End Versioned API Example 3
 
         try client.syncClose()
     }
 }
+
+private func versionedAPIMigrationExample() throws {
```
kmahar, line 287: this is a weird one... basically they want driver code equivalent to what is on this page: https://docs.mongodb.com/v5.0/reference/versioned-api/#example---migration

Examples/Docs/Sources/AsyncExamples/main.swift
```diff
@@ -246,34 +247,94 @@ private func versionedAPI() throws {
...
+    ])
+    // End step 1
+    // swiftlint:enable line_length
+
+    try insertResult.wait()
+
+    // 2. Run "count" using a strict client, observe error
+    let countResult = client.db("test").runCommand(["count": "sales"])
+
+    countResult.whenFailure { error in
```
kmahar, line 318: writing these examples is going to be so much nicer with async await....

Examples/Docs/Sources/AsyncExamples/main.swift
```diff
@@ -246,34 +247,94 @@ private func versionedAPI() throws {
...
+        ["_id": 7, "item": "xyz", "price": 5, "quantity": 10, "date": .datetime(dateFormatter.date(from: "2021-02-15T14:12:12Z")!)],
+        ["_id": 8, "item": "abc", "price": 10, "quantity": 5, "date": .datetime(dateFormatter.date(from: "2021-03-16T20:20:13Z")!)]
+    ])
+    // End step 1
+    // swiftlint:enable line_length
+
+    try insertResult.wait()
+
+    // 2. Run "count" using a strict client, observe error
+    let countResult = client.db("test").runCommand(["count": "sales"])
```
patrickfreed, line 316: do we want to construct this as a callback on `insertResult`? I worry that a user might copy/paste the examples and have this and the insert running concurrently. ditto for the `countDocuments` one.

kmahar: good point; I updated both calls to wait on the insert

patrickfreed (approved):

PR #653, opened 2021-08-04
SWIFT-1212, SWIFT-1213: Remove oppressive language

kmahar:
I looked into this while waiting around for Evergreen patches and it ended up being super small so figured we should just go ahead with it and close out the epic. 

Patch with added Atlas connectivity tasks since I changed the code in that target:  https://spruce.mongodb.com/version/610aed5dd6d80a1eba4406e9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/Sources/AtlasConnectivity/main.swift
+++ b/Sources/AtlasConnectivity/main.swift
@@ -3,9 +3,9 @@ import MongoSwiftSync
 
 private let configs = ["ATLAS_REPL", "ATLAS_SHRD", "ATLAS_FREE", "ATLAS_TLS11", "ATLAS_TLS12"]
 private let srvConfigs = configs.map { $0 + "_SRV" }
-/// Currently, almost all of the Atlas test instances are running server versions that do not yet support the new
-/// "hello" command.
+/// Currently, almost all of the Atlas test instances are running 3.4 which does not support the new "hello" command.
 private let legacyHello = "ismaster"
+private let supportsHello = ["ATLAS_FREE", "ATLAS_FREE_SRV"]
 
 for config in configs + srvConfigs {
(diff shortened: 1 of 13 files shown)


Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -4,6 +4,8 @@ import Foundation
 import Nimble
 import XCTest
 
+public let LEGACY_HELLO = "ismaster"
```
kmahar, line 7: because hello was backported all the way to 3.6 and we test against relatively recent patch versions on Evergreen, we are safe to just always use "hello" when we need to run the command manually in tests to figure out topology info.

however, we still need this constant in some cases, as pre-4.2 servers will continue to use legacy hello for monitoring since helloOk was not backported to those versions.

Sources/AtlasConnectivity/main.swift
```diff
@@ -3,6 +3,9 @@ import MongoSwiftSync
 
 private let configs = ["ATLAS_REPL", "ATLAS_SHRD", "ATLAS_FREE", "ATLAS_TLS11", "ATLAS_TLS12"]
 private let srvConfigs = configs.map { $0 + "_SRV" }
+/// Currently, almost all of the Atlas test instances are running 3.4 which does not support the new "hello" command.
```
kmahar, line 6: with the current work being scoped to set all drivers' minWireVersions to 6, I suspect these clusters will end up being upgraded before too long

kmahar:
I realized there are also some spec test files that need syncing but I will handle those in a follow up PR.

patrickfreed (approved):

PR #655, opened 2021-08-04
SWIFT-1122 Keep libmongoc version up-to-date in generated headers

kmahar:
This includes some changes to the vendoring script as well as files in the `generated_headers/` directory, along with the resulting diff in `Sources/CLibMongoC/` from running the updating vendoring script.

### DIFF
--- a/etc/vendor-libmongoc.sh
+++ b/etc/vendor-libmongoc.sh
@@ -3,8 +3,14 @@
 set -eou pipefail
 
 PWD=`pwd`
-LIBMONGOC_VERSION=1.18.0
-TARBALL_URL=https://github.com/mongodb/mongo-c-driver/releases/download/$LIBMONGOC_VERSION/mongo-c-driver-$LIBMONGOC_VERSION.tar.gz
+LIBMONGOC_MAJOR_VERSION=1
+LIBMONGOC_MINOR_VERSION=18
+LIBMONGOC_PATCH_VERSION=0
+LIBMONGOC_PRERELEASE_VERSION=
+
+LIBMONGOC_FULL_VERSION=${LIBMONGOC_MAJOR_VERSION}.${LIBMONGOC_MINOR_VERSION}.${LIBMONGOC_PATCH_VERSION}${LIBMONGOC_PRERELEASE_VERSION:+-$LIBMONGOC_PRERELEASE_VERSION}
+
+TARBALL_URL=https://github.com/mongodb/mongo-c-driver/releases/download/$LIBMONGOC_FULL_VERSION/mongo-c-driver-$LIBMONGOC_FULL_VERSION.tar.gz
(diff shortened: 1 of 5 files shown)


etc/vendor-libmongoc.sh
```diff
@@ -3,8 +3,14 @@
 set -eou pipefail
 
 PWD=`pwd`
-LIBMONGOC_VERSION=1.18.0
-TARBALL_URL=https://github.com/mongodb/mongo-c-driver/releases/download/$LIBMONGOC_VERSION/mongo-c-driver-$LIBMONGOC_VERSION.tar.gz
+LIBMONGOC_MAJOR_VERSION=1
```
kmahar, line 6: I didn't even want to try to think about how to do string parsing in a bash script to separate this into components...

etc/vendor-libmongoc.sh
```diff
@@ -3,8 +3,14 @@
...
+LIBMONGOC_MINOR_VERSION=18
+LIBMONGOC_PATCH_VERSION=0
+LIBMONGOC_PRERELEASE_VERSION=
```
kmahar, line 9: I tested the logic for when a prerelease version is specified here by setting it to "alpha2", as 1.18.0-alpha2 was a recent prerelease.
In that case script fails toward the end at the stage where we apply the various .diff files due to incompatibilities with those (when we vendored in 1.18.0 I had to update the patch files a bit), but the new logic added here works anyway.

etc/vendor-libmongoc.sh
```diff
@@ -3,8 +3,14 @@
...
+
+LIBMONGOC_FULL_VERSION=${LIBMONGOC_MAJOR_VERSION}.${LIBMONGOC_MINOR_VERSION}.${LIBMONGOC_PATCH_VERSION}${LIBMONGOC_PRERELEASE_VERSION:+-$LIBMONGOC_PRERELEASE_VERSION}
```
patrickfreed, line 11: what does the `:+-` syntax do here?

kmahar: if the value before the colon is set, it expands to whatever comes after the colon. if the value before the colon is unset, it expands to nothing.

so basically, if the prerelease version is set, the string becomes "-[prerelease version here]" e.g. "-alpha2", but if the prerelease version is not set this is just empty. 

this was necessary so we only append a dash if there is a prerelease version 

(thank you stack overflow for this: https://stackoverflow.com/a/40771884)

kmahar: I guess you could also think of it like a ternary statement: `${a:+b}` is equivalent to `a != nil ? b : nil`

patrickfreed: ah makes sense, thanks for the explanation :+1:

patrickfreed (approved):

PR #659, opened 2021-08-10
Run killAllSessions less frequently in unified test runner

kmahar:
Decreases the frequency with which we run `killAllSessions` to just after failed tests, rather than at the start of every suite and after every suite.

Sample patch here: https://evergreen.mongodb.com/version/6111b473c9ec44484753fc97

There is likely still some possibility of flakiness here, but given the failures did not show up on any of the tasks in that patch I think this has hopefully at least decreased the frequency.

I left the transactions test runner alone for now, since those issues have not been coming up as much, and hopefully DRIVERS-1539 will be done fairly soon to cut down on how much we need to run it there too.

### DIFF
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
@@ -17,10 +17,6 @@ struct UnifiedTestRunner {
         self.serverVersion = try self.internalClient.serverVersion()
         self.topologyType = try self.internalClient.topologyType()
         self.serverParameters = try self.internalClient.serverParameters()
-
-        // The test runner SHOULD terminate any open transactions using the internal MongoClient before executing any
-        // tests.
-        try self.terminateOpenTransactions()
     }
 
     func terminateOpenTransactions() throws {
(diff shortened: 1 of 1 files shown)


Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -57,17 +53,6 @@ struct UnifiedTestRunner {
     /// strings indicating cases to skip. If the array contains a single string "*" all tests in the file will be
     /// skipped.
     func runFiles(_ files: [UnifiedTestFile], skipTests: [String: [String]] = [:]) throws {
```
kmahar, line 55: I highly recommend reviewing this with "hide whitespace changes" on, as most of the diff here is from indenting a bunch of this code to wrap it in a try/catch.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -17,10 +17,6 @@ struct UnifiedTestRunner {
         self.serverVersion = try self.internalClient.serverVersion()
         self.topologyType = try self.internalClient.topologyType()
         self.serverParameters = try self.internalClient.serverParameters()
-
-        // The test runner SHOULD terminate any open transactions using the internal MongoClient before executing any
-        // tests.
-        try self.terminateOpenTransactions()
```
kmahar, line 23: this only seems useful if there was a test that ran beforehand which left open transactions. I think the only case that might happen is right after the transactions tests run (unless if there are others leaving txns open that I'm not aware of). so it doesn't seem worth running this every time

PR #660, opened 2021-08-11
SWIFT-1293 Add support for loadBalanced option

kmahar:
This is atop the `vendor-libmongoc`branch, will change base to `main` after that PR gets merged.


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -76,10 +76,6 @@ internal class ConnectionString {
             self.readConcern = rc
         }
 
-        if let replicaSet = options?.replicaSet {
```
kmahar, line 79: I needed to move this since some of the validation in `applyAndValidateSDAMOptions` relies on the value of this. it also seems like this probably should have been in that method anyway since it's an option related to SDAM

PR #661, opened 2021-08-16
SWIFT-1256 Pull in final changes for libmongoc 1.19

kmahar:
Now that the release has been tagged, this pulls in the final changes and gets the version numbers updated correctly.


Sources/MongoSwift/ConnectionString.swift
```diff
@@ -242,38 +242,47 @@ internal class ConnectionString {
 
     /// Sets and validates authentication-related on the underlying `mongoc_uri_t`.
     private func applyAndValidateAuthOptions(_ options: MongoClientOptions?) throws {
-        guard let credential = options?.credential else {
```
kmahar, line 245: this file is mostly whitespace changes due to some indenting, so suggest the "hide whitespace changes" option to cut down on the diff

patrickfreed (approved):

PR #664, opened 2021-08-19
SWIFT-1318 Expose libmongoc support for mocking serviceIds for load balancer testing 

kmahar:
This adds a small patch to the libmongoc vendoring process which makes it possible for us to configure this global variable, by moving it out of the private header we do not include into one that we do.

I did not really know anything about `extern` until today but hopefully all this makes sense:

Per conversation with Kevin, if we had a way to declare an extern variable in Swift, we could do so, set its value, and it would set the value in libmongoc as well. For example, a C program like
```c
#include <mongoc/mongoc.h>
extern bool mongoc_global_mock_service_id;
int main () {
   mongoc_global_mock_service_id = true;
}
```
Would control the value of the global defined in libmongoc itself.

However, apparently this is [not possible in Swift](https://twitter.com/jckarter/status/1157428699123212293?s=20):
>  the only way to declare extern C stuff in Swift is to literally write it in C and import it

And doing what seemed like "the closest" thing (declaring my own global with the same name and setting its value) had no effect on the libmongoc value.

I also could have just included the private header instead, but that seemed like an equal amount of hackiness (as it would require special-casing in the vendoring script to move that header elsewhere) for about the same benefit.

I tested this works locally by running haproxy in front of a sharded cluster (the Evg LB test config) and writing a small test that sends a command to the server. The initial handshake succeeds IFF `setMockServiceId(to: true)` is called before the initial handshake occurs/the command is run.

### DIFF
--- a/Sources/CLibMongoC/include/CLibMongoC_mongoc-server-description.h
+++ b/Sources/CLibMongoC/include/CLibMongoC_mongoc-server-description.h
@@ -66,6 +66,12 @@ MONGOC_EXPORT (int32_t)
 mongoc_server_description_compressor_id (
    const mongoc_server_description_t *description);
 
+/* mongoc_global_mock_service_id is only used for testing. The test runner sets
+ * this to true when testing against a load balanced deployment to mock the
+ * presence of a serviceId field in the "hello" response. The purpose of this is
+ * further described in the Load Balancer test README. */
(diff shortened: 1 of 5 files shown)


kmahar:
closing in favor of #666 which groups this with some additional changes that rely upon the work here.

PR #665, opened 2021-08-19
SWIFT-1309 Update ServerType and TopologyType for new load balancer types

kmahar:
Switch these types to structs wrapping enums for their internal representations, and add new "load balanced" topology type as well as "load balancer" server type.


kmahar:
closing in favor of #666 which groups this with some additional changes that rely upon the work here.

PR #666, opened 2021-08-23
SWIFT-1309, SWIFT-1318, SWIFT-1241, SWIFT-1323, SWIFT-1324 Support testing against a load balancer locally

kmahar:
This contains a hodgepodge of changes which, taken together, are all that is necessary to run our existing test suite against local load balancers. Future PRs will bring in the new load balancer-specific tests and support running load balancer tests in Evergreen.

### DIFF
--- a/Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
+++ b/Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
@@ -190,7 +190,7 @@ extension SpecTestFile {
         // if connected to a sharded cluster, use the per-mongos clients to execute killAllSessions on each mongos.
         switch try client.topologyType() {
         case .single,
-             .loadBalancer,
+             .loadBalanced,
              _ where MongoSwiftTestCase.serverless:
             return
         case .replicaSet:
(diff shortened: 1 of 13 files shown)


Tests/MongoSwiftSyncTests/SpecTestRunner/SpecTest.swift
```diff
@@ -184,13 +184,13 @@ extension SpecTestFile {
 
     func terminateOpenTransactions(
         using client: MongoSwiftSync.MongoClient,
-        mongosClients: [MongoSwiftSync.MongoClient]?
+        mongosClients: [ServerAddress: MongoSwiftSync.MongoClient]
```
kmahar, line 187: starting here and in the files below, I make some changes I laid out in the ticket description for SWIFT-1324: https://jira.mongodb.org/browse/SWIFT-1324

in summary, we now use the per-mongos clients (which we already had around for running `killAllSessions` and the "distinct hack") for targeted failpoints, which means the setup client can now just connect to a single mongos. 

we need the setup client to connect to a single mongos when in LB mode because if we use the LB connected to multiple mongoses then we cannot control which mongos a failpoint is set on. I considered a less invasive approach, where when connected to a load balancer we would just always use a single mongos for the setup client and use multiple mongoses for other topologies, but this seemed like a better solution overall as it helps with test flakiness issues too.

patrickfreed (approved):
LGTM!

patrickfreed (commented):
Oh one thing I just thought of, can we remove the `runCommand` logic for targeting a specific server now that we do the client-per-mongos strategy?

PR #667, opened 2021-08-29
SWIFT-1329 Use a client per mongos in unified test runner

kmahar:
This removes our (internal) support for targeting a particular host in `runCommand`, given that its bypassing of the server selection loop makes it susceptible to failure, and switches the unified test runner to use a client-per-mongos approach.


patrickfreed (approved):
LGTM! I really like the idea of having a new type to encapsulate this.

PR #669, opened 2021-09-01
SWIFT-1326 Run load balancer spec tests 

kmahar:
[this is currently based on the Evergreen testing PR, I think GitHub will automatically update the base once that goes in]

This adds code to actually start running the new tests, as well as making some various unified runner / testing infrastructure changes to support it.

[patch](https://spruce.mongodb.com/version/612f9ed7850e612d5d269c78/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/Tests/MongoSwiftSyncTests/LoadBalancerTests.swift
+++ b/Tests/MongoSwiftSyncTests/LoadBalancerTests.swift
@@ -13,7 +13,7 @@ final class LoadBalancerTests: MongoSwiftTestCase {
             specName: "load-balancers",
             excludeFiles: skipFiles,
             asType: UnifiedTestFile.self
-        ).filter { $0.0 == "cursors.json" }.map { $0.1 }
+        ).map { $0.1 }
 
         let skipTests = [
             // The sessions spec requires that sessions can only be used with the MongoClient that created them.
--- a/Tests/MongoSwiftSyncTests/LoadBalancerTests.swift
+++ b/Tests/MongoSwiftSyncTests/LoadBalancerTests.swift
@@ -1,0 +1,65 @@
+import MongoSwiftSync
+import Nimble
+import TestsCommon
+
+final class LoadBalancerTests: MongoSwiftTestCase {
+    let skipFiles: [String] = [
+        // We don't support this option.
+        "wait-queue-timeouts.json"
+    ]
+
+    func testLoadBalancers() throws {
+        let tests = try retrieveSpecTestFiles(
+            specName: "load-balancers",
+            excludeFiles: skipFiles,
+            asType: UnifiedTestFile.self
+        ).filter { $0.0 == "cursors.json" }.map { $0.1 }
+
+        let skipTests = [
+            // The sessions spec requires that sessions can only be used with the MongoClient that created them.
+            // Consequently, libmongoc enforces that a `mongoc_client_session_t` may only be used with the
+            // `mongoc_client_t` that created it. In Swift, this translates to a requirement that we always pin
+            // `Connection`s to `ClientSession`s from the time the session is first used until it is closed/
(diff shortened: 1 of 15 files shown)


Tests/MongoSwiftSyncTests/LoadBalancerTests.swift
```diff
@@ -0,0 +1,65 @@
...
+    ]
+
+    func testLoadBalancers() throws {
+        let tests = try retrieveSpecTestFiles(
+            specName: "load-balancers",
+            excludeFiles: skipFiles,
+            asType: UnifiedTestFile.self
+        ).filter { $0.0 == "cursors.json" }.map { $0.1 }
+
+        let skipTests = [
```
kmahar, line 18: unfortunately a lot of caveats here, but we still do get to run a decent number of the tests, so it seems better than nothing. let me know if all the reasoning below makes sense.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/ExpectedEventsForClient.swift
```diff
@@ -6,10 +6,35 @@ struct ExpectedEventsForClient: Decodable {
...
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.client = try container.decode(String.self, forKey: .client)
+        // Defaults to command if omitted.
+        self.eventType = try container.decodeIfPresent(EventType.self, forKey: .eventType) ?? .command
+        switch self.eventType {
+        case .command:
+            self.events = try container.decode([ExpectedEvent].self, forKey: .events)
+        case .cmap:
+            // TODO: SWIFT-1321 actually parse these out.
```
kmahar, line 34: for now, there is just always an empty array, and in the runner below if `ExpectedEventsForClient.type` is `cmap` we skip doing assertions as we don't capture any events.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCollectionOperations.swift
```diff
@@ -75,6 +75,37 @@ struct UnifiedCreateIndex: UnifiedOperationProtocol {
     }
 }
 
+struct UnifiedListIndexes: UnifiedOperationProtocol {
```
kmahar, line 78: we don't actually end up using this operation since we skip the listIndexes cursor test, but I figured we would need it eventually and adding it seemed like a reasonable way to allow us to decode `cursors.json`

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedDatabaseOperations.swift
```diff
@@ -63,3 +64,38 @@ struct UnifiedRunCommand: UnifiedOperationProtocol {
         return .none
     }
 }
+
+struct UnifiedListCollections: UnifiedOperationProtocol {
```
kmahar, line 68: similarly, this operation ends up going unused but I added it for completeness.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCollectionOperations.swift
```diff
@@ -75,6 +75,44 @@ struct UnifiedCreateIndex: UnifiedOperationProtocol {
...
+    /// Optional identifier for a session entity to use.
+    let session: String?
+
+    /// We consider this a known argument and decode ite ven though we don't support it, because a load balancer test
```
patrickfreed, line 82:
```suggestion
    /// We consider this a known argument and decode it even though we don't support it, because a load balancer test
```

patrickfreed (approved):
one typo fix but otherwise LGTM

PR #673, opened 2021-09-03
SWIFT-1322 Support batchSize option for listCollections

kmahar:
Reimplements listCollections by running the command and creating a cursor from it. This is necessary because the libmongoc helper ignores the batchSize option.

### DIFF
--- a/Sources/MongoSwift/MongoError.swift
+++ b/Sources/MongoSwift/MongoError.swift
@@ -522,7 +522,7 @@ internal func toErrorString(_ error: bson_error_t) -> String {
     }
 }
 
-internal let failedToRetrieveCursorMessage = "Couldn't get cursor from the server"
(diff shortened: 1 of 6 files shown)


Sources/MongoSwift/MongoError.swift
```diff
@@ -522,7 +522,7 @@ internal func toErrorString(_ error: bson_error_t) -> String {
     }
 }
 
-internal let failedToRetrieveCursorMessage = "Couldn't get cursor from the server"
+internal let failedToRetrieveCursorMessage = "Expected libmongoc to return a cursor, unexpectedly got nil"
```
kmahar, line 525: this seemed inaccurate, since if this happens it's not necessarily a server issue and is more likely a bug in libmongoc, which is not supposed to ever return nil from the corresponding methods

patrickfreed (approved):

PR #675, opened 2021-09-13
SWIFT-1347 Fix memory leak in tests caused by ClientSession.pinnedServerAddress

kmahar:
As discovered via adding `leaks` to CI in my other PR. After this gets merged in I will rebase over there to get the new tasks passing.

### DIFF
--- a/Sources/MongoSwift/ClientSession.swift
+++ b/Sources/MongoSwift/ClientSession.swift
@@ -94,8 +94,11 @@ public final class ClientSession {
             return nil
         }
         return connection.withMongocConnection { client in
-            let serverDescription =
-                ServerDescription(mongoc_client_get_server_description(client, serverID))
+            guard let sd = mongoc_client_get_server_description(client, serverID) else {
+                return nil
+            }
+            defer { mongoc_server_description_destroy(sd) }
+            let serverDescription = ServerDescription(sd)
             return serverDescription.address
         }
     }


Sources/MongoSwift/ClientSession.swift
```diff
@@ -94,8 +94,11 @@ public final class ClientSession {
             return nil
         }
         return connection.withMongocConnection { client in
-            let serverDescription =
-                ServerDescription(mongoc_client_get_server_description(client, serverID))
+            guard let sd = mongoc_client_get_server_description(client, serverID) else {
```
kmahar, line 97: I don't fully understand why, but if the result of `mongoc_client_get_server_description` is passed directly into the `ServerDescription` initializer it doesn't need to be unwrapped.  But if I assign it to a variable like this then I have to unwrap it to be able to pass it in. 

[This](https://github.com/apple/swift-evolution/blob/master/proposals/0055-optional-unsafe-pointers.md) evolution proposal explains why the pointer is optional in the first place, and makes sense, but it doesn't make much sense to me that it gets coerced into a non-optional by passing it directly to the initializer... 🤷‍♀️

patrickfreed: Huh, yeah not sure what's going on there either.

patrickfreed (approved):

kmahar:
(test failure is an unrelated issue caused by a server change, see my message in driver-devs)

PR #676, opened 2021-09-27
Skip failing test against latest sharded clusters

kmahar:
Unclear when the server team will fix this issue, so skipping for now to make the waterfall less red.

### DIFF
--- a/Tests/MongoSwiftSyncTests/CrudTests.swift
+++ b/Tests/MongoSwiftSyncTests/CrudTests.swift
@@ -90,7 +90,7 @@ final class CrudTests: MongoSwiftTestCase {
             asType: UnifiedTestFile.self
         )
         let runner = try UnifiedTestRunner()
-        let skipList: [String: [String]] = [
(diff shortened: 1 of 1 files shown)


Tests/MongoSwiftSyncTests/CrudTests.swift
```diff
@@ -106,6 +106,13 @@ final class CrudTests: MongoSwiftTestCase {
             "unacknowledged-bulkWrite-update-hint-clientError": ["*"],
             "unacknowledged-bulkWrite-replace-hint-clientError": ["*"]
         ]
+        // Skipping due to a bug in server latest. TODO: SWIFT-1359 unskip
+        let client = try MongoClient.makeTestClient()
```
kmahar, line 110: we don't really have a way in the runner itself to do conditional skips based on server version/topology, it seems, besides editing changes into the JSON files which didn't seem like a great approach.

patrickfreed (approved):

PR #679, opened 2021-10-05
SWIFT-1313 Test serverless behind a load balancer

kmahar:
Updates our tests and Evergreen config to work with serverless behind a load balancer.

### DIFF
--- a/Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
+++ b/Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
@@ -4,6 +4,7 @@ import MongoSwiftSync
 @testable import class MongoSwiftSync.ClientSession
 import Nimble
 import TestsCommon
+import XCTest
 
 struct UnifiedFailPoint: UnifiedOperationProtocol {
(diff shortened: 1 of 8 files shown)


Tests/MongoSwiftSyncTests/UnifiedTestRunner/SpecialTestOperations.swift
```diff
@@ -272,11 +273,10 @@ struct UnifiedTargetedFailPoint: UnifiedOperationProtocol {
         // The mongos on which to set the fail point is determined by the session argument (after resolution to a
         // session entity). Test runners MUST error if the session is not pinned to a mongos server at the time this
         // operation is executed.
-        expect(session.pinnedServerAddress)
-            .toNot(
-                beNil(),
-                description: "Session \(self.session) unexpectedly not pinned to a mongos. Path: \(context.path)"
-            )
+        guard session.pinnedServerAddress != nil else {
```
kmahar, line 276: switching to this approach as we otherwise crash on line 281 on Linux when the pinnedServerAddress is nil due to SWIFT-1146.
I ran into this when I was accidentally running a test that uses this assertion against serverless and it failed

PR #680, opened 2021-10-06
SWIFT-560 Allow specifying pipelines to update commands

isabelatkinson:


Tests/MongoSwiftSyncTests/UnifiedTestRunner/TestWriteModel.swift
```diff
@@ -57,15 +57,13 @@ enum TestWriteModel: Decodable {
             matchedKey = .replaceOne
         } else if let nested = try? container.nestedContainer(keyedBy: UpdateKeys.self, forKey: .updateOne) {
             let filter = try nested.decode(BSONDocument.self, forKey: .filter)
-            // TODO: SWIFT-560 handle decoding pipelines properly
-            let update = try nested.decode(BSONDocument.self, forKey: .update)
+            let update = try nested.decode(UpdateModel.self, forKey: .update).asDocument()
```
isabelatkinson, line 60: Because we're not updating the bulk write API to include pipelines, we need to convert pipelines specified in the spec tests to documents.

PR #695, opened 2021-11-01
SWIFT-1161 `MongoConnectionString` authentication options support

isabelatkinson:
Authentication options support for `MongoConnectionString`.

This PR also includes a commit to reapply the commit introducing the `MongoConnectionString` type. When this gets merged I'll keep the commits separate.

### DIFF
--- a/Sources/MongoSwift/ConnectionString.swift
+++ b/Sources/MongoSwift/ConnectionString.swift
@@ -262,7 +262,7 @@ internal class ConnectionString {
             }
 
             if let mechanism = credential.mechanism {
-                guard mongoc_uri_set_auth_mechanism(self._uri, mechanism.name) else {
+                guard mongoc_uri_set_auth_mechanism(self._uri, mechanism.description) else {
                     throw self.failedToSet(MONGOC_URI_AUTHMECHANISM, to: mechanism)
                 }
             }
@@ -595,7 +595,7 @@ internal class ConnectionString {
             return nil
         }
         let str = String(cString: mechanism)
-        return MongoCredential.Mechanism(str)
+        return try? MongoCredential.Mechanism(str)
     }
 
     /// Returns a document containing the auth mechanism properties if any were provided, otherwise nil.
@@ -653,7 +653,7 @@ internal class ConnectionString {
             copy.authsource = .string(authSource)
         }
         if let authMechanism = self.authMechanism {
-            copy.authmechanism = .string(authMechanism.name)
+            copy.authmechanism = .string(authMechanism.description)
         }
         if let authMechanismProperties = self.authMechanismProperties {
             copy.authmechanismproperties = .document(authMechanismProperties)
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -1,0 +1,192 @@
+import Foundation
+
+/// Represents a MongoDB connection string.
+/// - SeeAlso: https://docs.mongodb.com/manual/reference/connection-string/
+public struct MongoConnectionString: Codable, LosslessStringConvertible {
+    /// Represents a connection string scheme.
+    public struct Scheme: LosslessStringConvertible, Equatable {
+        /// Indicates that this connection string uses the scheme `mongodb`.
+        public static let mongodb = Scheme(.mongodb)
+
+        /// Indicates that this connection string uses the scheme `mongodb+srv`.
+        public static let srv = Scheme(.srv)
+
+        /// Internal representation of a scheme.
+        private enum _Scheme: String {
+            case mongodb
+            case srv = "mongodb+srv"
+        }
+
+        private let _scheme: _Scheme
+
+        private init(_ value: _Scheme) {
+            self._scheme = value
+        }
+
+        /// LosslessStringConvertible` protocol requirements
+        public init?(_ description: String) {
+            guard let _scheme = _Scheme(rawValue: description) else {
+                return nil
+            }
+            self.init(_scheme)
+        }
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// A struct representing a host identifier, consisting of a host and an optional port.
+    /// In standard connection strings, this describes the address of a mongod or mongos to connect to.
+    /// In mongodb+srv connection strings, this describes a DNS name to be queried for SRV and TXT records.
+    public struct HostIdentifier: Equatable, CustomStringConvertible {
+        private static func parsePort(from: String) throws -> UInt16 {
+            guard let port = UInt16(from), port > 0 else {
+                throw MongoError.InvalidArgumentError(
+                    message: "port must be a valid, positive unsigned 16 bit integer"
+                )
+            }
+            return port
+        }
+
+        private enum HostType: String {
+            case ipv4
+            case ipLiteral = "ip_literal"
+            case hostname
+            case unixDomainSocket
+        }
+
+        /// The hostname or IP address.
+        public let host: String
+
+        /// The port number.
+        public let port: UInt16?
+
+        private let type: HostType
+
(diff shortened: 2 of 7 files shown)


Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -177,16 +325,38 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+
+    /// The default database to use for authentication if an `authSource` is unspecified in the connection string.
+    /// Defaults to "admin" if unspecified.
+    public var defaultAuthDB: String?
+
+    /// Specifies the authentication credentials.
+    public var credential: MongoCredential?
+}
+
+extension StringProtocol {
```
isabelatkinson, line 344: extending `StringProtocol` rather than `String` here because split returns substrings

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -203,13 +155,49 @@ final class ConnectionStringTests: MongoSwiftTestCase {
...
+                if let expectedAuth = testCase.auth {
+                    let actual = connString.credential ?? MongoCredential()
+                    guard expectedAuth.matches(actual) else {
+                        print(testCase.description)
+                        XCTFail("Expected credentials: \(expectedAuth) do not match parsed credentials: \(actual)")
+                        continue
+                    }
+                }
+
+                if let expectedOptions = testCase.options {
```
isabelatkinson, line 167: I'll probably revisit this implementation once I have a better handle on all of the options that need to be checked/tested.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,362 @@
...
```
kmahar, line 344: do these methods need to be internal or could they be `fileprivate`?

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,362 @@
...
+        let identifiersAndOptions = schemeAndRest[1].components(separatedBy: "/")
+        // TODO: SWIFT-1174: handle unescaped slashes in unix domain sockets.
+        guard identifiersAndOptions.count <= 2 else {
+            throw MongoError.InvalidArgumentError(
+                message: "Connection string contains an unescaped slash"
+            )
+        }
+        let userAndHost = identifiersAndOptions[0].components(separatedBy: "@")
+        guard userAndHost.count <= 2 else {
+            throw MongoError.InvalidArgumentError(message: "Invalid user information")
```
kmahar, line 218: I think the error message here could be a little more specific like the one on 213, since it seems like the problem here would likely be an unescaped `@` symbol

Sources/MongoSwift/MongoCredential.swift
```diff
@@ -71,5 +71,71 @@ public struct MongoCredential: Decodable, Equatable {
...
+
+        internal func validateCredential(credential: MongoCredential) throws {
+            switch self {
+            case .gssAPI:
+                if credential.username == nil {
+                    throw MongoError.InvalidArgumentError(
+                        message: "Username must be provided for GSSAPI authentication"
+                    )
+                }
+                if let source = credential.source {
```
kmahar, line 97: style nit: we can reduce the amount of nesting in some of the checks in this method by combining the `if`s into a single line like
```swift
if let source = credential.source, source != "$external" {
 // ... error
}
```

kmahar (commented):
looking great so far! just some minor comments/questions and a small-ish refactoring suggestion.
(as an aside, should this be a non-draft PR now?)

Sources/MongoSwift/ConnectionString.swift
```diff
@@ -595,7 +595,7 @@ internal class ConnectionString {
             return nil
         }
         let str = String(cString: mechanism)
-        return MongoCredential.Mechanism(str)
+        return try? MongoCredential.Mechanism(str)
```
isabelatkinson, line 598: Constructing a `Mechanism` now possibly throws if the name is invalid, but that shouldn't be an issue here because we're getting the mechanism from a `mongoc_uri`. (We'll also be moving away from this type anyway.)

Tests/MongoSwiftTests/AuthTests.swift
```diff
@@ -2,14 +2,15 @@ import Foundation
 @testable import MongoSwift
 import Nimble
 import TestsCommon
+import XCTest
 
 /// Represents a single file containing auth tests.
-struct AuthTestFile: Decodable {
+private struct AuthTestFile: Decodable {
```
isabelatkinson, line 8: Updated some visibility here so that I could use the name `TestCredential` in multiple places.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,373 @@
...
+            self.init(_scheme)
+        }
+
+        public var description: String { self._scheme.rawValue }
+    }
+
+    /// A struct representing a host identifier, consisting of a host and an optional port.
+    /// In standard connection strings, this describes the address of a mongod or mongos to connect to.
+    /// In mongodb+srv connection strings, this describes a DNS name to be queried for SRV and TXT records.
+    public struct HostIdentifier: Equatable, CustomStringConvertible {
```
patrickfreed, line 45: Why is a new type needed instead of `ServerAddress`?

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,373 @@
...
+    /// `Codable` conformance
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self.description)
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let stringValue = try container.decode(String.self)
+        try self.init(throwsIfInvalid: stringValue)
```
patrickfreed, line 323: We'll want to convert this error to a `DecodingError` to preserve the decoding metadata (e.g. the key path) for debugging. You can see an example of that here: https://github.com/mongodb/mongo-swift-driver/blob/main/Sources/MongoSwift/MongoServerAPI.swift#L46.

patrickfreed (commented):
Looks good! I just have a few minor questions / comments

Sources/MongoSwift/MongoConnectionString.swift
isabelatkinson: Actually not sure, this is part of the first commit which is just a reapply of Bynn's original commit to introduce `MongoConnectionString` -- was this design discussed during the original implementation?

kmahar: I remember we talked about this and ended up concluding we needed a new type but I do not remember why. looking at the differences though, the `port` is optional here but non-optional for `ServerAddress`, and this type also supports a host type whereas the other does not. I think the problem might have been that we need to be able to store unresolved DNS names here, whereas we could only store resolved ones in a `ServerAddress`?

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -0,0 +1,362 @@
...
+            }
+            if let port = self.port {
+                ret += ":\(port)"
+            }
+            return ret
+        }
+    }
+
+    private struct Options {
+        fileprivate var authSource: String?
```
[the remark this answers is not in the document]
kmahar: ah, that makes sense 👍

kmahar (commented):
lgtm mod a couple tiny comments

kmahar (approved):
lgtm!

Sources/MongoSwift/MongoConnectionString.swift
patrickfreed: ah yeah that sounds familiar.

patrickfreed (approved):
LGTM!

PR #696, opened 2021-11-02
SWIFT-1406 Don't use git:// for cloning drivers-evergreen-tools

kmahar:
See DRIVERS-1970 for details.


patrickfreed (approved):

kmahar:
just gonna go ahead and merge to unbreak evg, don't think we need two reviewers on this

PR #701, opened 2021-11-04
SWIFT-1162 `MongoConnectionString` TLS options support

isabelatkinson:


Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -154,6 +168,27 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                     self.authMechanism = try MongoCredential.Mechanism(value)
                 case "authmechanismproperties":
                     self.authMechanismProperties = try self.parseAuthMechanismProperties(properties: value)
+                case "ssl":
```
patrickfreed, line 171: I kind of think we should accept both of these options but only have a `tls` Swift property for it. This would mean that round tripping a non-canonical connection string through `MongoConnectionString` wouldn't work, but IMO that's okay. WDYT? cc @kmahar

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -154,6 +168,27 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+                    self.ssl = try value.getBool(forKey: "ssl")
+                case "tls":
+                    self.tls = try value.getBool(forKey: "tls")
+                case "tlsallowinvalidcertificates":
```
patrickfreed, line 175: since these constants are used in a few places, I think it would be beneficial to define them as Swift constants to help prevent accidental typos.

Sources/MongoSwift/MongoConnectionString.swift
kmahar: yeah I think that makes sense for the public API on `MongoConnectionString` (on this `Options` type is fine to keep both since it's internal and used for validation). and it seems acceptable to me as well to do that normalization and only include `tls` if someone round trips a string that specified both `tls` and `ssl`.

Sources/MongoSwift/MongoConnectionString.swift
kmahar: +1

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -8,6 +9,22 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
     /// - SeeAlso: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2
     fileprivate static let forbiddenUserInfoCharacters = [":", "/", "?", "#", "[", "]", "@"]
 
+    private struct Names {
```
isabelatkinson, line 12: introduced a struct to store the name constants to avoid name collision with the `MongoConnectionString` fields

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -8,6 +9,22 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
     /// - SeeAlso: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2
     fileprivate static let forbiddenUserInfoCharacters = [":", "/", "?", "#", "[", "]", "@"]
 
+    private enum Names {
```
kmahar, line 12: maybe we should consider giving the enum a raw type of `String`, and making these static properties enum cases instead? 

it might make things a little safer in the big switch on option names below, since you could initialize an enum instance from the option name and switch on that. would also I think make the TODO at the bottom to error on unknown options easy

isabelatkinson: Weird, I originally made this a struct but it looks like the formatter changed it into an enum 🤔 Switching to a proper enum worked out nicely though, updated

kmahar (approved):
lgtm!

Sources/MongoSwift/MongoConnectionString.swift
kmahar: ah, yeah generally an empty enum is considered the "right" way to do namespacing: https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#enumnamespaces 

we don't do this much in the driver but we did use it to namespace our error types so they are all prefixed with `MongoError`

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -8,6 +9,22 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
     /// - SeeAlso: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2
     fileprivate static let forbiddenUserInfoCharacters = [":", "/", "?", "#", "[", "]", "@"]
 
+    fileprivate enum Name: String {
```
patrickfreed, line 12: could we name this something a bit more specific, e.g. `OptionName`?

patrickfreed (commented):
LGTM mod one bikeshed

patrickfreed (approved):
LGTM!

PR #704, opened 2021-11-09
SWIFT-1391 Add MongoCollection async/await API

kmahar:
This is a large PR but most of the content is quite straightforward, just more async versions of existing API methods, so hopefully it is not too painful to review.

### DIFF
--- a/Sources/MongoSwift/AsyncAwait/MongoClient+AsyncAwait.swift
+++ b/Sources/MongoSwift/AsyncAwait/MongoClient+AsyncAwait.swift
@@ -48,8 +48,8 @@ extension MongoClient {
      * Run the `listDatabases` command.
      *
      * - Parameters:
-     *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
-     *     on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
+     *   - filter: Optional `BSONDocument` specifying a filter that the listed databases must pass. This filter can be
+     *     based on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
      *   - options: Optional `ListDatabasesOptions` specifying options for listing databases.
      *   - session: Optional `ClientSession` to use when executing this command.
      *
@@ -75,7 +75,7 @@ extension MongoClient {
      * Get a list of `MongoDatabase`s.
      *
      * - Parameters:
-     *   - filter: Optional `Document` specifying a filter on the names of the returned databases.
+     *   - filter: Optional `BSONDocument` specifying a filter on the names of the returned databases.
      *   - options: Optional `ListDatabasesOptions` specifying options for listing databases.
      *   - session: Optional `ClientSession` to use when executing this command.
      *
@@ -98,7 +98,7 @@ extension MongoClient {
      * Get a list of names of databases.
      *
      * - Parameters:
-     *   - filter: Optional `Document` specifying a filter on the names of the returned databases.
+     *   - filter: Optional `BSONDocument` specifying a filter on the names of the returned databases.
      *   - options: Optional `ListDatabasesOptions` specifying options for listing databases.
      *   - session: Optional `ClientSession` to use when executing this command.
      *
--- a/Sources/MongoSwift/AsyncAwait/MongoCollection+AsyncAwait.swift
+++ b/Sources/MongoSwift/AsyncAwait/MongoCollection+AsyncAwait.swift
@@ -1,0 +1,48 @@
+#if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
+/// Extension to `MongoCollection` to support async/await APIs.
+extension MongoCollection {
+    /**
+     *   Drops this collection from its parent database.
+     * - Parameters:
+     *   - options: An optional `DropCollectionOptions` to use when executing this command.
+     *   - session: An optional `ClientSession` to use when executing this command.
+     *
+     * - Throws:
+     *   - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     */
+    public func drop(options: DropCollectionOptions? = nil, session: ClientSession? = nil) async throws {
+        try await self.drop(options: options, session: session).get()
+    }
+
+    /**
+     * Renames this collection on the server. This method will return a handle to the renamed collection. The handle
+     * which this method is invoked on will continue to refer to the old collection, which will then be empty.
+     * The server will throw an error if the new name matches an existing collection unless the `dropTarget` option
+     * is set to true.
+     *
+     * Note: This method is not supported on sharded collections.
+     *
+     * - Parameters:
+     *   - to: A `String`, the new name for the collection.
+     *   - options: Optional `RenameCollectionOptions` to use for the collection.
+     *   - session: Optional `ClientSession` to use when executing this command.
+     *
+     * - Returns:
+     *    A copy of the target `MongoCollection` with the new name.
+     *
+     *    Throws:
+     *    - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *    - `MongoError.InvalidArgumentError` if the options passed in form an invalid combination.
+     *    - `MongoError.LogicError` if the provided session is inactive.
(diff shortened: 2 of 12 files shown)


Sources/MongoSwift/AsyncAwait/MongoClient+AsyncAwait.swift
```diff
@@ -48,8 +48,8 @@ extension MongoClient {
      * Run the `listDatabases` command.
      *
      * - Parameters:
-     *   - filter: Optional `Document` specifying a filter that the listed databases must pass. This filter can be based
-     *     on the "name", "sizeOnDisk", "empty", or "shards" fields of the output.
+     *   - filter: Optional `BSONDocument` specifying a filter that the listed databases must pass. This filter can be
```
kmahar, line 51: I ended up doing some general cleanup of docstrings throughout the new API methods during this. I noticed some remaining instances of `Document` in existing API methods as well which these were copied from but I am going to make a separate PR to fix those up.

Tests/MongoSwiftTests/AsyncAwaitTestUtils.swift
```diff
@@ -0,0 +1,137 @@
+#if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
```
kmahar, line 1: I was starting to add a lot of utility functions so I ended up breaking all this out into a separate file. however, there are no changes in lines 24 through 90 so you do not need to re-review those.

Tests/MongoSwiftTests/AsyncAwaitTestUtils.swift
```diff
@@ -0,0 +1,137 @@
...
+import NIO
+import TestsCommon
+import XCTest
+
+/// Temporary utility function until XCTest supports `async` tests.
+func testAsync(_ block: @escaping () async throws -> Void) {
+    let group = DispatchGroup()
+    group.enter()
+    Task.detached {
+        do {
```
kmahar, line 14: I realized I needed to add a do/catch here, because otherwise if the call to `block` threw an error the Task would finish without ever calling `group.leave()` and so the tests would hang indefinitely.

Tests/MongoSwiftTests/AsyncAwaitTestUtils.swift
```diff
@@ -0,0 +1,141 @@
...
+            throw TestError(message: " reply missing version string: \(reply)")
+        }
+        return try ServerVersion(versionString)
+    }
+
+    internal func serverParameters() async throws -> BSONDocument {
+        try await self.db("admin").runCommand(["getParameter": "*"])
+    }
+
+    internal func topologyType() async throws -> TestTopologyConfiguration {
```
kmahar, line 108: was fun to get to try out some of the new concurrency features in these next few methods and implicitly confirm the driver can handle parallel usage across tasks! (explicit test of that to come soon in another PR)

Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -123,21 +44,74 @@ final class AsyncAwaitTests: MongoSwiftTestCase {
...
+                    expect(doc).to(equal(["_id": 1]))
+                }
             }
         }
     }
 
     func testMongoDatabase() throws {
         testAsync {
             try await self.withTestNamespace { _, db, _ in
+                try await db.drop()
```
kmahar, line 91: I think I didn't notice this issue with my last PR because when you run the whole suite the test DB ends up getting cleaned up somewhere else, but running just the async/await tests multiple times in a row leads to failure here since test.foo wasn't being cleaned up.

Sources/MongoSwift/AsyncAwait/MongoCollection+AsyncAwait.swift
```diff
@@ -0,0 +1,48 @@
+#if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
+/// Extension to `MongoCollection` to support async/await APIs.
+extension MongoCollection {
+    /**
+     *   Drops this collection from its parent database.
```
patrickfreed, line 5:
```suggestion
     * Drops this collection from its parent database.
```

Tests/MongoSwiftTests/AsyncAwaitTestUtils.swift
```diff
@@ -0,0 +1,141 @@
...
+        async let helloReply = try self.db("admin").runCommand(["hello": 1])
+        async let shards = try self.db("config").collection("shards").find().get().toArray().get()
+        return try await TestTopologyConfiguration(helloReply: helloReply, shards: shards)
+    }
+
+    internal func getUnmetRequirement(_ testRequirement: TestRequirement) async throws -> UnmetRequirement? {
+        async let topologyType = try self.topologyType()
```
patrickfreed, line 115: just curious, is `async let` required to make these three evaluate simultaneously? Or would invoking them directly in the argument position of `getUnmetRequirements` behave the same way?

side note: GH doesn't seem able to highlight `async let` properly yet lol

patrickfreed (approved):
LGTM! (mod one space)

Tests/MongoSwiftTests/AsyncAwaitTestUtils.swift
kmahar: it does compile to one-line it like:
```swift
await testRequirement.getUnmetRequirement(givenCurrent: try self.serverVersion(), try self.topologyType(), try self.serverParameters())
```

but the behavior seems different - I roughly tested it by inserting print statements in each method before and after their `async` work like:
```swift
    internal func serverParameters() async throws -> BSONDocument {
        print("getting server parameters")
        let params = try await self.db("admin").runCommand(["getParameter": "*"])
        print("returning server parameters")
        return params
    }
```

as originally written, the interleaving of "getting x" and "returning x" varies on each run, for example one output is:
```
getting topology type
getting server version
getting server parameters
returning server version
returning topology type
returning server parameters
```

and another is
```
getting server version
getting topology type
getting server parameters
returning server parameters
returning server version
returning topology type
```

 which makes since given they are being run in parallel.

but written as you ask about in one line, the interleaving is always the same and suggests serial evaluation of each called function:
```
getting server version
returning server version
getting topology type
returning topology type
getting server parameters
returning server parameters
```

Sources/MongoSwift/AsyncAwait/MongoCollection+Indexes+AsyncAwait.swift
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Retrieves a list of names of the indexes currently on this collection.
+     *
+     * - Parameters:
+     *   - session: Optional `ClientSession` to use when executing this command.
+     *
+     * - Returns: A `MongoCursor` over the index names.
+     *
+     * - Throws: `MongoError.LogicError` if the provided session is inactive.
```
isabelatkinson, line 201: General question, do we call out all possible errors that could be thrown in documentation? This method and the one above could encounter server/network errors e.g., but I'm not sure how that maps to our error types/if that's something we include in docs

isabelatkinson (commented):
lgtm, just one question

Sources/MongoSwift/AsyncAwait/MongoCollection+Indexes+AsyncAwait.swift
kmahar: At some point, @patrickfreed went through the whole API and called out what we thought the most common errors would be. we generally haven't tried to list them exhaustively, I think the goal has generally been to focus on common pitfalls and errors that could be specific to the use of the particular API method.

there is an error handling guide in the docs that talks about some of the general common errors users could see on any method, like a `ConnectionError` or `ServerSelectionError`:
https://mongodb.github.io/mongo-swift-driver/docs/current/MongoSwift/error-handling-guide.html

isabelatkinson: cool, makes sense to focus on errors that are relevant to usage. thanks for the clarification!

isabelatkinson (approved):

PR #705, opened 2021-11-09
SWIFT-1414 Update release script to not only work on main

kmahar:
This script previously assumed you were running it from `main`; this generalizes it to work regardless of branch. 

I'll cherry-pick this onto r1.2 as well.

### DIFF
--- a/etc/release.sh
+++ b/etc/release.sh
@@ -10,7 +10,7 @@ version=${1}
 [ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
 
 # verify that the examples build with the current branch we're releasing on
-current_branch=$(git rev-parse --abbrev-ref HEAD)
+current_branch=$(git branch --show-current)
 ./etc/build-examples.sh ${current_branch}
 
 # regenerate documentation with new version string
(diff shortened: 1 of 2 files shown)


etc/release.sh
```diff
@@ -30,7 +31,7 @@ git commit -m "${version} docs"
 git push
 
 # go back to our original branch
-git checkout -
+git checkout ${current_branch}
```
kmahar, line 34: this would still work, but it seemed better to be explicit.

isabelatkinson (approved):

etc/release.sh
```diff
@@ -9,8 +9,9 @@ version=${1}
 # Ensure version is non-empty
 [ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
 
-# verify that the examples build
-./etc/build-examples.sh
+# verify that the examples build with the current branch we're releasing on
+current_branch=$(git rev-parse --abbrev-ref HEAD)
```
patrickfreed, line 13: can we use `git branch --show-current` here instead? or is this version required for certain cases?

patrickfreed (commented):
one question otherwise LGTM

etc/release.sh
kmahar: that should work, the stack overflow post I got this from seems to predate the existence of `--show-current` 🙂 updated!

patrickfreed (approved):

PR #706, opened 2021-11-11
SWIFT-1392 Add MongoCursor and ChangeStream async/await APIs

kmahar:
Add async versions of the APIs for these types, including conformance to `AsyncSequence`.

### DIFF
--- a/Sources/MongoSwift/ChangeStream.swift
+++ b/Sources/MongoSwift/ChangeStream.swift
@@ -61,7 +61,7 @@ internal let changeStreamNamespaceKey = CodingUserInfoKey(rawValue: "namespace")
 /// A MongoDB change stream.
 /// - SeeAlso: https://docs.mongodb.com/manual/changeStreams/
 public class ChangeStream<T: Codable>: CursorProtocol {
-    internal typealias Element = T
+    public typealias Element = T
 
     /// The client this change stream descended from.
     private let client: MongoClient
@@ -270,8 +270,10 @@ public class ChangeStream<T: Codable>: CursorProtocol {
      * This method MAY be called if the change stream is already dead. It will have no effect.
      *
      * - Warning:
-     *    If this change stream is alive when it goes out of scope, it will leak resources. To ensure it
-     *    is dead before it leaves scope, invoke this method.
+     *    On Swift versions and platforms where structured concurrency is not available, if a change stream is alive
+     *    when it goes out of scope, it will leak resources. On those Swift versions/platforms, you must invoke this
+     *    method to ensure resources are properly cleaned up. If structured concurrency is available, it is not
+     *    necessary to call this method as resources will be cleaned up automatically during deinitialization.
      *
      * - Returns:
      *   An `EventLoopFuture` that evaluates when the change stream has completed closing. This future should not fail.
@@ -281,4 +283,18 @@ public class ChangeStream<T: Codable>: CursorProtocol {
             self.wrappedCursor.kill()
         }
     }
+
+#if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
+    /// When concurrency is available, we can ensure change streams are always cleaned up properly.
+    deinit {
+        let client = self.client
+        let el = self.eventLoop
+        let wrappedCursor = self.wrappedCursor
+        Task {
+            try await client.operationExecutor.execute(on: el) {
+                wrappedCursor.kill()
+            }
+        }
+    }
+#endif
 }
--- a/Sources/TestsCommon/APMUtils.swift
+++ b/Sources/TestsCommon/APMUtils.swift
@@ -60,11 +60,9 @@ public class TestCommandMonitor: CommandEventHandler {
 
     /// Capture events that occur while the the provided closure executes.
     public func captureEvents<T>(_ f: () throws -> T) rethrows -> T {
-        try self.lock.withLock {
-            self.monitoring = true
-            defer { self.monitoring = false }
-            return try f()
-        }
+        self.enable()
+        defer { self.disable() }
+        return try f()
     }
 
     /// Enable monitoring, if it is not enabled already.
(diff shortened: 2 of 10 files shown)


Sources/MongoSwift/ChangeStream.swift
```diff
@@ -281,4 +283,18 @@ public class ChangeStream<T: Codable>: CursorProtocol {
             self.wrappedCursor.kill()
         }
     }
+
+#if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
+    /// When concurrency is available, we can ensure change streams are always cleaned up properly.
+    deinit {
```
kmahar, line 289: I think this is a lot simpler than what I had to do for `ClientSession`s because the internal `Cursor` class already neatly packages up the state and stores a references to the `Connection`. 

`kill()` is defined here: https://github.com/mongodb/mongo-swift-driver/blob/f22e1fd3340982f58f0b87718d9651aa9dd76a43/Sources/MongoSwift/CursorCommon.swift#L354

And calls through to `close()`, where only once destroying the cursor/change stream completes does `self.state` get mutated and the reference dropped: https://github.com/mongodb/mongo-swift-driver/blob/f22e1fd3340982f58f0b87718d9651aa9dd76a43/Sources/MongoSwift/CursorCommon.swift#L219

And since we are calling `kill` on the `Cursor` itself here, the `Cursor` should stay alive until the method completes.

that said, would appreciate if y'all review the logic in that file to check my reasoning and confirm what I have here should work.

Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -114,4 +115,161 @@ final class AsyncAwaitTests: MongoSwiftTestCase {
...
+                let fourth = try await cursor.next()
+                expect(fourth).to(beNil())
+                let isAlive = try await cursor.isAlive()
+                expect(isAlive).to(beFalse())
+            }
+        }
+    }
+
+    // Test that a tailable cursor that is continually polling the server can be killed by cancelling the parent Task.
+    func testTailableCursorHandlesTaskCancellation() throws {
```
kmahar, line 150: I had mentioned on Slack creating the cursor outside of the child task and then asserting on `isAlive()`, however the thread sanitizer did not like that. I think this still covers what we are trying to test though, to your point @patrickfreed about assuming tailable cursors work right.

Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -114,4 +115,168 @@ final class AsyncAwaitTests: MongoSwiftTestCase {
...
+                            _ = eventCount.add(1)
+                        }
+                    } catch {
+                        XCTFail("\(error)")
+                    }
+                    group.leave()
+                }
+
+                // Wait until the change stream has been opened before we start inserting data.
+                try await assertIsEventuallyTrue(description: "change stream should be started") {
```
kmahar, line 253: this was a race condition that for whatever reason only showed up on server versions up to but not including 4.4, where occasionally the docs would get inserted below before we managed to create the change stream and so no events would be observed.

Sources/MongoSwift/ChangeStream.swift
patrickfreed: Yeah I think this looks good. If this were future based I'd be a bit worried that `client` might go out of scope before `kill` completed, but I think the `await` here ensures it stays alive until the future is evaluated.

Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -114,4 +115,168 @@ final class AsyncAwaitTests: MongoSwiftTestCase {
     }
 }
```
patrickfreed, line 117: I think it would be nice to have a test for the `deinit` behavior. Basically something that starts up a cursor, doesn't iterate it to completion, and then asserts that a `killCursors` command was eventually sent or something. Ditto for change streams.

patrickfreed (commented):
API LGTM, just have a few comments/questions and one test suggestion.

Tests/MongoSwiftTests/AsyncAwaitTests.swift
kmahar: nice idea! added tests for both.

Sources/TestsCommon/APMUtils.swift
```diff
@@ -1,22 +1,28 @@
 import Foundation
 import MongoSwift
+import NIOConcurrencyHelpers
 
 /// A command event handler that caches the events it encounters.
 /// Note: it will only cache events that occur while closures passed to `captureEvents` are executing.
 public class TestCommandMonitor: CommandEventHandler {
```
kmahar, line 7: the thread sanitizer was not pleased by the use of this type in the new tests, so I introduced a lock around the state to ensure access is synchronized

patrickfreed: It looks like this lock might be causing a deadlock (see the timed out tests). I'm not exactly sure how to work around this in a way that tsan will like though.

kmahar: the problem here was that `TestCommandMonitor.captureEvents` was attempting to hold the lock for its entire body, which prevented `TestCommandMonitor.handleCommandEvent` from ever being able to acquire the lock and handle an event.

I'd just gone through and added a `withLock` to each method, but holding the lock for the entire time the provided closure executes is not needed. I've changed it now so that we delegate to the (lock acquiring) `enable()`/`disable()` methods for turning monitoring on and off, and then rely on the lock acquisition within `handleCommandEvent` to ensure synchronization of any access to the monitor state that occurs while executing the closure.

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #707, opened 2021-11-12
SWIFT-1163 `MongoConnectionString` options support

isabelatkinson:


Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -531,16 +978,42 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+            self[name.rawValue]
+        }
+        set {
+            self[name.rawValue] = newValue
+        }
+    }
+}
+
+extension Compressor {
+    fileprivate init(name: String, level: Int32?) throws {
```
isabelatkinson, line 1009: Including this here rather than in the implementation for `Compressor` because it relies upon validations made in the parser.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -159,9 +199,14 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
         fileprivate var tlsDisableOCSPEndpointCheck: Bool?
         fileprivate var tlsInsecure: Bool?
+        fileprivate var w: WriteConcern.W?
+        fileprivate var wTimeoutMS: Int?
+        fileprivate var zlibCompressionLevel: Int32?
 
         fileprivate init(_ uriOptions: String) throws {
             let options = uriOptions.components(separatedBy: "&")
+            // tracks options that have already been set to error on duplicates
+            var setOptions: [String] = []
```
patrickfreed, line 209: rather than using an array here, we can use a [`Set`](https://developer.apple.com/documentation/swift/set) since we'll be doing mostly inclusion lookups.

Tests/MongoSwiftTests/ConnectionStringTests.swift
```diff
@@ -63,37 +63,31 @@ private struct TestCredential: Decodable {
...
         "minpoolsize=0 does not error",
-        // We don't allow maxPoolSize=0, see SWIFT-1339.
+        // we don't allow maxPoolSize=0, see SWIFT-1339
         "maxpoolsize=0 does not error"
     ],
-    "connection-options.json": ["*"], // requires maxIdleTimeMS
-    "compression-options.json": ["*"], // requires Snappy, see SWIFT-894
-    "valid-db-with-dotted-name.json": ["*"] // libmongoc doesn't allow db names in dotted form in the URI
+    "connection-options.json": [
+        "valid connection and timeout options are parsed correctly" // we don't support maxIdleTimeMS
```
patrickfreed, line 78: even though we don't support maxIdleTimeMS in the driver, it is a valid connection string option, so do we want to support it in `MongoConnectionString`? Ditto for the other options we don't support.

isabelatkinson: I think it might be confusing/misleading to users if they set an option and receive no error but the option isn't applied anywhere in the driver. Ideally this would be a nice place to log a warning, but absent that ability it could be deceiving to ignore these options silently. This does harm the portability of connection strings, though, which might be more bothersome to users. What do you think @kmahar?

kmahar: @patrickfreed, what are you thinking the use case is for supporting these? portability?

> I think it might be confusing/misleading to users if they set an option and receive no error but the option isn't applied anywhere in the driver

I agree this is definitely undesirable; we need to communicate to the user at some point in some way that we're ignoring these.

to me the options seem to be: 
1) error upon `MongoConnectionString` initialization
2) error upon `MongoClient` initialization `MongoConnectionString` that contains one of these options. this might be technically more correct since they are valid options in general, but is probably equally annoying in practice for users as option 1, and for options we never intend to support (for example single threaded options) it may be confusing to users. 

I'm not sure if this type will have value for users outside of the context of creating a client so I'm not sure if there is a meaningful distinction between where the validation happens, but to avoid confusion I'd probably lean toward option 1

patrickfreed: It's pretty niche but I guess there could be users who want to write a tool for building or parsing connection strings but not actually connecting to a database. 

More generally speaking, I was wondering if we wanted `MongoConnectionString` to be a general purpose connection string model that can be used for `MongoClient` configuration or a configuration tool for `MongoClient` that can be used to model a connection string (basically 2 vs 1).

Option 2 feels "cleaner" to me in an abstract sense, but practically speaking 1 is probably better. Either way they both seem fine to me, so happy to defer to your all's opinions on this. Just figured it was worth discussing.

kmahar: makes sense! yeah, I think for now I prefer option 1, but we could always move toward the general purpose model in the future in a non-breaking way if that comes up as a desired use case.

isabelatkinson: yeah I'm fine with sticking with the current functionality of erroring upon connection string initialization

kmahar (changes requested):
looks good! just some minor questions and nits.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -498,9 +879,75 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+    /// Specifies the maximum time, in milliseconds, for an individual connection to establish a TCP
+    /// connection to a MongoDB server before timing out.
+    public var connectTimeoutMS: Int?
+
     /// Specifies the authentication credentials.
     public var credential: MongoCredential?
 
+    /// Specifies whether the client should connect directly to a single host. When false, the client will attempt to
+    /// automatically discover all replica set members if a replica set name is provided. Defaults to false.
+    /// It is an error to set this option to `true` when used with a mongodb+srv connection string or when multiple
```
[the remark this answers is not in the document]
isabelatkinson: yep, updated

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -448,14 +746,85 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+        }
+        if let authMechanism = self.credential?.mechanism {
+            options[.authMechanism] = .string(authMechanism.description)
+        }
+        if let authMechanismProperties = self.credential?.mechanismProperties {
+            options[.authMechanismProperties] = .document(authMechanismProperties)
+        }
+        if let compressors = self.compressors {
+            var compressorNames: [BSON] = []
+            options[.compressors] = .array([])
```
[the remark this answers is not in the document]
isabelatkinson: ah this was a placeholder, removed

Sources/MongoSwift/MongoConnectionString.swift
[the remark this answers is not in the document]
kmahar: I got to wondering why this wasn't causing any test failures, but I realized it only comes up via the DNS seedlist tests which create a client with the connection string, so we wouldn't see it until we plug this type in to the rest of the driver

kmahar (approved):

patrickfreed (approved):
LGTM!

PR #711, opened 2021-11-18
SWIFT-1407 / SWIFT-1174 ipv4 address and UNIX domain socket support

isabelatkinson:


Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -135,18 +134,29 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
+
+                let host = parts[0]
+                if host.hasSuffix(".sock") {
+                    self.host = try host.getPercentDecoded(forKey: "UNIX domain socket")
+                    self.type = .unixDomainSocket
+                } else if host.isIPV4() {
+                    self.host = host
+                    self.type = .ipv4
+                } else {
+                    self.host = try host.getPercentDecoded(forKey: "hostname")
```
isabelatkinson, line 151: The rule for reg-name (i.e. a non-IPLiteral/IPv4 host) is:
```
reg-name    = *( unreserved / pct-encoded / sub-delims )
```
so we want to support percent decoding here.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -420,7 +430,6 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
             throw MongoError.InvalidArgumentError(message: "Invalid connection string")
         }
         let identifiersAndOptions = schemeAndRest[1].components(separatedBy: "/")
-        // TODO: SWIFT-1174: handle unescaped slashes in unix domain sockets.
```
isabelatkinson, line 423: Not sure why this was here; the spec [indicates](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#host) that hosts cannot contain unescaped slashes and there are no tests that allow for this behavior.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -135,18 +134,29 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
...
                     throw MongoError.InvalidArgumentError(
                         message: "expected only a single port delimiter ':' in \(hostAndPort)"
                     )
                 }
+
+                let host = parts[0]
+                if host.hasSuffix(".sock") {
+                    self.host = try host.getPercentDecoded(forKey: "UNIX domain socket")
+                    self.type = .unixDomainSocket
+                } else if host.isIPv4() {
```
isabelatkinson, line 147: The logic here is based off of this language in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2):

> `host        = IP-literal / IPv4address / reg-name`
>
>The syntax rule for host is ambiguous because it does not completely
distinguish between an IPv4address and a reg-name.  In order to
disambiguate the syntax, we apply the "first-match-wins" algorithm:
If host matches the rule for IPv4address, then it should be
considered an IPv4 address literal and not a reg-name.

Essentially, if the host does not match IPv4, fall back to hostname. The rules for reg-name permit "invalid" IPv4 addresses, so we don't want to error on a string like "1.2.3" as that is a valid hostname.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -1033,6 +1045,19 @@ extension StringProtocol {
         return decoded
     }
 
+    fileprivate func isIPv4() -> Bool {
```
isabelatkinson, line 1048: The regular expression for IPv4 addresses is quite unwieldy ([see](https://devblogs.microsoft.com/oldnewthing/20060522-08/?p=31113)) so I opted for this method for the sake of readability/correctness.

Sources/MongoSwift/MongoConnectionString.swift
kmahar: thanks for the explanation! and I was going to suggest we add a test to make sure we don't break this in the future but I see you already did 🙂

kmahar (approved):
lgtm!

patrickfreed (approved):
LGTM!

PR #712, opened 2021-11-19
SWIFT-1393 Add concurrent MongoClient usage test

kmahar:
Adds a test that uses a `MongoClient` in parallel across a number of tasks to verify that works as we expect it to.

### DIFF
--- a/Tests/MongoSwiftTests/AsyncAwaitTests.swift
+++ b/Tests/MongoSwiftTests/AsyncAwaitTests.swift
@@ -113,6 +113,31 @@ final class AsyncAwaitTests: MongoSwiftTestCase {
             }
         }
     }
+
+    func testConcurrentClientUsage() throws {
+        testAsync {
+            try await self.withTestNamespace { _, db, _ in
+                defer {
+                    db.syncDropOrFail()
+                }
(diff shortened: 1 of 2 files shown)


Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -113,6 +113,31 @@ final class AsyncAwaitTests: MongoSwiftTestCase {
...
     }
+
+    func testConcurrentClientUsage() throws {
+        testAsync {
+            try await self.withTestNamespace { _, db, _ in
+                defer {
+                    db.syncDropOrFail()
+                }
+                try await withThrowingTaskGroup(of: Int.self) { group in
+                    for i in 1...20 {
```
kmahar, line 124: the choice of number of tasks and number of operations was somewhat arbitrary here, and I'm open to any suggestions on adjusting these if they seem unreasonable. this seemed to strike a reasonable balance between "high degree of concurrency" and "test that completes in a reasonable amount of time" (~3 seconds).

isabelatkinson (approved):

patrickfreed (approved):

PR #713, opened 2021-11-19
SWIFT-1408 Remove conditional definition on Linux for async/await APIs

kmahar:
Due to CI issues, I initially defined all the new APIs on Linux only. This removes that code and switches to using `@available` annotations to indicate macOS compatibility instead.
That said, we still are not testing the new code in CI, since we have no macOS 12 hosts to test on. There is back-deployment support coming for concurrency in Xcode 13.2 (which will allow using concurrency features when targeting as low as macOS 10.15), however I'm still not sure if we'll be able to run tests then since I think Xcode 13.2 requires macOS 11.

Maybe I can get testing on those macOS 11 ARM hosts working by then, or alternatively we can look into getting a simple GH actions setup together to actually be able to run these tests on macOS in CI. I filed SWIFT-1421 about this

### DIFF
--- a/Tests/MongoSwiftTests/AsyncAwaitTests.swift
+++ b/Tests/MongoSwiftTests/AsyncAwaitTests.swift
@@ -1,4 +1,6 @@
 #if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
+// TODO: SWIFT-1421 remove this comment and the os(Linux) check above. if the described bug is not fixed, we'll need to
+// adjust our macOS < 12 testing to skip these tests by providing a filter to `swift test`.
 // we shouldn't have to check the operating system here, but there is a bug currently where on older macOS versions
 // the @available checks don't work. since we don't have support for macOS 12+ in CI, we can work around this by just
 // only defining the tests on Linux for now.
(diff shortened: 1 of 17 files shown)


Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -1,4 +1,7 @@
 #if compiler(>=5.5) && canImport(_Concurrency) && os(Linux)
+// we shouldn't have to check the operating system here, but there is a bug currently where on older macOS versions
```
patrickfreed, line 2: Once we do have support for macOS 12+ in CI, how will we handle these availability checks? Will we just drop the affected versions from the matrix? Also, do we want to add a TODO here for the macOS 12 ticket?

patrickfreed (approved):
LGTM, just have a question about the test skipping on macOS

Tests/MongoSwiftTests/AsyncAwaitTests.swift
kmahar: I think the bug fix would come in a Swift 5.5 patch release. 

if it is out by whenever that is, we could just bump our Swift 5.5 version in CI on all macOS versions and these should get properly skipped.

if it is not fixed, then we could use a filter with `swift test` so that we would skip all of these on old macOS versions. this is [what Vapor does](https://github.com/vapor/vapor/blob/6a5a3b5244d39e2614382c77ddf62e63b712ad06/.github/workflows/test.yml#L75). the crashes only happens when the test cases get wrongly executed, not when compiling or running other tests, so the bug doesn't block everything else from executing. the only annoyance there is that for local testing on macOS 11, all of you who have not upgraded would have to remember to always provide a filter to `swift test` (I guess we could add something to the makefile for this to make it slightly easier). 

and yeah good idea on the TODO, added and also added mention of the issue you point out here / workarounds.

isabelatkinson (approved):

PR #718, opened 2021-12-09
SWIFT-1165 Replace `ConnectionString` with `MongoConnectionString`

isabelatkinson:
Removes the old `ConnectionString` type and replaces it with `MongoConnectionString`.


kmahar (changes requested):
looking great! nice refactoring, all seems good to me mod fixing the directConnection issue discussed in comment

Tests/MongoSwiftSyncTests/SDAMTests.swift
```diff
@@ -86,7 +86,7 @@ final class SDAMTests: MongoSwiftTestCase {
         let prevTopology = event4.previousDescription
         let newTopology = event4.newDescription
 
-        expect(prevTopology.type).to(equal(.unknown))
```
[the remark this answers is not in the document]
isabelatkinson: Thanks for doing all of this digging! I updated `MongoConnectionString.validate` to set `directConnection` to `false` when unspecified and reverted the SDAM test.

Tests/MongoSwiftTests/MongoConnectionStringTests.swift
```diff
@@ -171,7 +171,8 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         let encodedData = try JSONEncoder().encode(connStr)
         let decodedResult = try JSONDecoder().decode(MongoConnectionString.self, from: encodedData)
         expect(connStr.description).to(equal(decodedResult.description))
-        expect(connStr.description).to(equal("mongodb://localhost:27017/"))
+        // directConnection is always specified in URIs to override libmongoc behavior
+        expect(connStr.description).to(equal("mongodb://localhost:27017/?directconnection=false"))
```
kmahar, line 175: it's a little funny that it always shows up now when round tripping, though off the top of my head I can't think of any ill effects of that besides possibly a user finding it strange. 

what do you think about moving the defaulting logic into `withMongocURI`, and only setting it there for the purpose of libmongoc interop? then I think we wouldn't run into this issue, and it would be more obvious that we can delete the special logic once we stop interoperating with libmongoc

isabelatkinson: good idea, updated.

patrickfreed (approved):
LGTM! Great job with this!

kmahar (approved):
LGTM! 🎉👏

Issue #720, opened 2021-12-25
SWIFT-1438 ⁃ Change @available attribute on async/await features


atultw:
### Swift version
```
swift-driver version: 1.26.21 Apple Swift version 5.5.2 (swiftlang-1300.0.47.2 clang-1300.0.29.30)
Target: x86_64-apple-macosx11.0
```

### Operating system
```
Darwin Someones-MBP 20.6.0 Darwin Kernel Version 20.6.0: Wed Nov 10 22:23:07 PST 2021; root:xnu-7195.141.14~1/RELEASE_X86_64 x86_64
```

## What is the problem? 

Apple has released an Xcode update enabling async/await on macOS older than 12. 

https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes#Swift

The codebase includes a lot of `@available(macOS 12, *)` attributes on the async features. Given the new update we should change the `12` to `10.15`.

https://github.com/mongodb/mongo-swift-driver/search?q=%40available%28mac

Should I go ahead and make a PR?

atultw:
Related to https://github.com/apple/swift-nio/pull/2004

That issue was resolved in a new version. Updating `swift-nio` dependency to `2.36.0` allows the `@available` here to be updated too.

bajanam:
@atultw  Thanks for the analysis and PR.  The team is off for the holidays but we will review this based our development capacity when they return.

bajanam:
This was fixed PR# 21

PR #721, opened 2021-12-25
SWIFT-1438 Change concurrency minimum macOS version from 12 to 10.15.0

atultw:
See issue #720 

Change concurrency minimum mac version from 12 to 10.15.0
Update swift-nio dependency to 2.36.0

Some things to make sure you have:
- [x] Read `CONTRIBUTING.md`!
- [x] Ran `swiftlint` and ensured that your additions/deletions are not causing
any new warnings or errors.
- [ ] Added new tests that correctly verify the correctness of any new/added
functionality.
- [ ] Ran `make test` and ensured that the driver still passes all of its
pre-existing tests.
- [x] _Not_ included any changes to the generated documentation (we update
the documentation as part of a separate commit).


kmahar:
Hi @atultw, thank you very much for getting in touch and for the pull request, and for your patience as I looked into this matter!

Changing the minimum concurrency version for us has some complicating factors:

1) Bumping the minimum NIO version implicitly drops support for Swift 5.1 from the driver going forward, since NIO 2.36 only works on Swift 5.2+. (We're not necessarily opposed to doing this, but it is a tradeoff to consider.)
2) As written, your PR would break driver compilation for anyone who is on Xcode < 13.2 with Swift 5.5.0 or 5.5.1, since we conditionally define the new APIs on Swift 5.5.0+.  There are a couple workarounds for this...
    * Increase these checks to Swift 5.5.2+ instead. This doesn't seem that unreasonable (we haven't even released this code yet, so requiring the latest patch to use it seems fair) OR
    * Do what NIO did as discussed on apple/swift-nio#2000 and define everything twice, with different availability checks based on compiler version. Given how large our `async` API surface is though I think this is less feasible for us than it was for NIO and would require duplicating an enormous amount of code and documentation. 

Overall, I'd say my feeling is that if this is important to the community and our users it would be worth us doing. But I am not sure how important it is.
I know there is a strong argument for doing this in libraries that are used in apps deployed on Apple platforms like iOS, as such applications often support old OS versions for quite a while. For server-side Swift libraries though, my expectation is most users are developing on macOS and deploying on Linux. So the use case for this seems like it might mainly be to allow people who have not updated the machines they develop on to macOS 12+. I'll also note that e.g. Vapor still has the concurrency checks written with macOS 12+. Beyond NIO and swift-service-lifecycle I haven't found any other libraries that have done this, yet. 

Given that I'm wondering if you could share a little more about your motivation for the PR - do you want/need to use the driver and these new APIs on old macOS versions?

atultw:
Thanks for the response! Re:
> For server-side Swift libraries though, my expectation is most users are developing on macOS and deploying on Linux. So the use case for this seems like it might mainly be to allow people who have not updated the machines they develop on to macOS 12+

This is exactly my use case. My mbp (2014) is too old for Monterey, so the availability check is a roadblock for me. 

As I understand it the choice is between:

* support Monterey only with older Xcode versions
* support older macOS but only running Xcode 13.2

Please correct me if I’m missing something, but given that choice it seems updating Xcode is more convenient than updating macOS (plus macOS upgrade is impossible for some)

kmahar:
Thank you for getting back to me!

After discussing with the team we're happy to support use on older OSes but I think we'll need some modifications to what you've got here before we can merge this - 

1) We should update all of the `if #compiler(>=5.5.0)` check around the new APIs to `if #compiler(>=5.5.2)`, because 
Swift compiler version 5.5.0 and 5.5.1 still think that macOS 12+ is required to use concurrency. This will help us avoid breaking things for users on those versions

2) To avoid having to drop Swift 5.1 support we should only conditionally set the NIO minimum version to 2.36 in the case where we need it. We could do this by adding a function in `Package.swift` like
```swift
func getMinNIOVersion() -> PackageDescription.Package.Dependency.Requirement {
#if compiler(>=5.5.2) && canImport(_Concurrency)
    return .upToNextMajor(from: "2.36.0")
#else
    return .upToNextMajor(from: "2.15.0")
#endif
}
```
and then calling that in the place the min NIO version is specified.

If you are up for making them that would be great but we can also take things from here if you prefer!

atultw:
Great idea! This makes sense. I’ll make the changes in PR by tomorrow

Thanks

atultw:
Made the changes. Builds successfully with Big Sur Mac as target

kmahar (changes requested):
thanks! couple small things but looking good otherwise!

kmahar (approved):

isabelatkinson (approved):
LGTM, thank you for your contribution!

patrickfreed (approved):

PR #723, opened 2022-01-05
Simplify building and testing in Xcode

kmahar:
Prior to Xcode's SwiftPM integration, one needed to generate a `.xcodeproj` to build a Swift package in Xcode.  Nowadays, the preferred way to build a SwiftPM package in Xcode is to simply open the directory there - one can do this from the command line by running `xed .` from the root directory, and then Xcode should automatically download dependencies and build the driver.

To include non-source code (spec test) files in the `.xcodeproj`, we previously used a Ruby script that updated the `.xcodeproj` data to copy the spec files into the project as "resources". This was required because, as part of its build process, Xcode copies only the source code files into another directory for building, rendering the relative paths to non-source we could use from the command line useless.
The preferred way to include non-source code files in a SwiftPM package now is using [Resources](https://developer.apple.com/documentation/swift_packages/bundling_resources_with_a_swift_package) defined in `Package.swift`, which basically has the same effect the ruby script used to on the `.xcodeproj`; however the new approach only works with Swift 5.3+ and we are still supporting older versions than that.

For now, a fairly clean workaround is to obtain an absolute path to the specs directory in the source repository via `#file`. Then, even when we are in whatever directory Xcode is building/running the tests in, we can access the original versions of the spec files.


Makefile
```diff
@@ -5,12 +5,6 @@ else
 	FILTERARG =
 endif
 
-ifdef DOCSVERSION
```
kmahar, line 8: I think we used to have a `make docs` target in here where this was used, but it's irrelevant now

Makefile
```diff
@@ -48,8 +35,6 @@ coverage:
 	xcrun llvm-cov export -format="lcov" .build/debug/mongo-swift-driverPackageTests.xctest/Contents/MacOS/mongo-swift-driverPackageTests -instr-profile .build/debug/codecov/default.profdata > info.lcov
 
 clean:
-	rm -rf Packages
```
kmahar, line 51: this might have been relevant long ago when the driver supported cocoapods, but isn't now

Makefile
```diff
@@ -48,8 +35,6 @@ coverage:
...
 	rm -rf .build
-	rm -rf MongoSwift.xcodeproj
```
kmahar, line 53: this is doubly irrelevant because 1) the generated xcodeproj is actually named `mongo-swift-driver.xcodeproj` ever since we renamed the package, and now also 2) no more `.xcodeproj` needed to build

kmahar:
Oh I meant to say, it would be good if yall can try this out to make sure it works locally for you too. 

This also reminded me that Xcode builds code differently than SwiftPM, via `xcodebuild`, and we should probably do SWIFT-953 soon so we're actually testing in CI that our code builds that way

isabelatkinson (approved):
lgtm, this all works for me locally! nice to be able to run tests in xcode easily, I hadn't bothered to get that working before this

nspektor (approved):
works for me also! Awesome

patrickfreed (approved):

PR #725, opened 2022-01-06
Only check isAlive() for tailable cursors in async MongoCursor.next()

kmahar:
I noticed from the benchmarks that iterating through a cursor's results one by one in the async API was about twice as a slow as doing so in the sync API. From using Xcode + Instruments with the time profiler instrument I was able to see that a lot of time was being spent calling `isAlive()`. 

That method and the looping behavior implemented using it is only useful when cursors are tailable, but as written we are currently paying the price of an extra call to `isAlive()` for every call to `next()`.

This updates the implementation of `MongoCursor.next()` so that in the common, non-tailable case we do not perform the unnecessary check, which saves us a journey across the `async`-futures boundary and a trip into the thread pool.

To improve the tailable cursor case (and the equivalent behavior in `ChangeStream.next()`), we could also consider some sort of refactor that allows bundling the calls to `mongoc_cursor_more` and `mongoc_cursor_next` in a single trip into the thread pool via a method that both checks for liveness and returns a doc if one exists. Then the looping logic would be something like
```swift
while true {
    let next = try await self.client.operationExecutor.execute {
        try self.wrappedCursor.tailableNext()
    }
    switch next {
    case .cursorDied: // i.e. `isAlive()` would have returned false
        return nil
    case .doc(document): // `isAlive()` would have returned true and `tryNext()` would have returned a doc
        return doc
   case .cursorAlive: // `isAlive()` would have returned true and `tryNext()` would have returned nil
        continue
    }
}
```
Thoughts?

### DIFF
--- a/Sources/MongoSwift/AsyncAwait/MongoCursor+AsyncSequence.swift
+++ b/Sources/MongoSwift/AsyncAwait/MongoCursor+AsyncSequence.swift
@@ -50,17 +50,25 @@ extension MongoCursor: AsyncSequence, AsyncIteratorProtocol {
      *   - `DecodingError` if an error occurs decoding the server's response to a `T`.
      */
     public func next() async throws -> T? {
-        while try await self.isAlive() {
-            if Task.isCancelled {
-                return nil
-            }
-            if let doc = try await self.tryNext() {
-                return doc
-            }
-            await Task.yield()
+        guard !Task.isCancelled else {
+            return nil
         }
 
-        return nil
+        switch self.cursorType {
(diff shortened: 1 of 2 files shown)


patrickfreed (approved):
Code changes here LGTM.

Your proposed improvement for tailable cursors also SGTM, though it doesn't seem super high priority given how rarely tailable cursors are used. Maybe just file a ticket for it now?

isabelatkinson (approved):

kmahar:
yeah I agree, I will file a ticket

PR #726, opened 2022-01-11
SWIFT-1396 Update Vapor example to use async/await

kmahar:
Updates our example Vapor app to use the new async/await APIs.

This can be reviewed now but we'll want to hold off on merging until there is an alpha tag to depend on.

I am also going to make some PRs to mongodb-vapor's documentation and mongodb-vapor-template's code to incorporate async/await info.


Examples/VaporExample/Package.swift
```diff
@@ -1,15 +1,15 @@
...
     platforms: [
-        .macOS(.v10_15)
+        .macOS(.v12)
     ],
     dependencies: [
-        .package(url: "https://github.com/vapor/vapor", .upToNextMajor(from: "4.7.0")),
+        .package(url: "https://github.com/vapor/vapor", .upToNextMajor(from: "4.50.0")),
         .package(url: "https://github.com/vapor/leaf", .upToNextMajor(from: "4.0.0")),
-        .package(url: "https://github.com/mongodb/mongodb-vapor", .upToNextMajor(from: "1.0.0"))
+        .package(url: "https://github.com/mongodb/mongodb-vapor", .branch("async-await"))
```
kmahar, line 12: all this branch does is depend on the driver main branch so we get the new APIs. there are no code changes needed

Examples/VaporExample/README.md
```diff
@@ -8,22 +8,22 @@ This application is intended to demonstrate best practices for integrating the d
 
 The application contains both a REST API as well as a minimal frontend built with Vapor's templating language [Leaf](https://github.com/vapor/leaf).
 
-This application require Swift 5.2 and MongoDB 3.6+. It will run on Linux as well as macOS 10.15+.
+This application requires Swift 5.5+ and MongoDB 3.6+. It will run on Linux as well as macOS 12+.
+
+If you are looking for an example application that uses older Swift versions prior to the introduction of structured concurrency, please refer to a previous version of this application and README [here](https://github.com/mongodb/mongo-swift-driver/tree/5d9fad121d7cb3ded61087de300ff007766ccd55/Examples/VaporExample).
```
kmahar, line 13: I considered trying to maintain two versions of the example but I don't know if it is worth the effort, so just linking to the latest commit seemed like a reasonable way to not totally lose the info but strongly push users toward async/await.

Examples/VaporExample/Package.swift
```diff
@@ -1,15 +1,15 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.5
 import PackageDescription
 
 let package = Package(
     name: "VaporExample",
     platforms: [
-        .macOS(.v10_15)
+        .macOS(.v12)
```
kmahar, line 7: Vapor has not added support for concurrrency on old macOS versions (yet anyway, not sure if they plan to), so even if we did support it this would need to be 12+

Examples/VaporExample/Package.swift
```diff
@@ -1,15 +1,15 @@
...
     ],
     dependencies: [
-        .package(url: "https://github.com/vapor/vapor", .upToNextMajor(from: "4.7.0")),
+        .package(url: "https://github.com/vapor/vapor", .upToNextMajor(from: "4.50.0")),
```
kmahar, line 10: min vapor version where async/await APIs exist

Examples/VaporExample/README.md
```diff
@@ -8,22 +8,22 @@ This application is intended to demonstrate best practices for integrating the d
...
 1. Run `./loadData.sh` to load example application data into the database.
-1. Install Swift 5.2+ on your system if you haven't already. You can download Swift and find instructions for installing it [here](https://swift.org/download/).
+1. Install Swift 5.5+ on your system if you haven't already. You can download Swift and find instructions for installing it [here](https://swift.org/download/).
 1. From the root directory of the project, run `swift build`. This will likely take a while the first time you do so.
 1. Once building has completed, run `swift run` from the root directory. You should get a message that the server has started running on `http://127.0.0.1:8080`.
 1. Open up your browser and visit `http://127.0.0.1:8080`. You should see the application and be able to test out adding, deleting, and editing data in the collection.
 
 ## Application Architecture
 
-This is a fully asynchronous application. At its core is [SwiftNIO](https://github.com/apple/swift-nio), which is used to implement both Vapor and the MongoDB driver.
```
kmahar, line 24: users don't really need to think about NIO as much now (and in fact can completely ignore it when using MongoDBVapor since we handle setting up the client and passing in the EventLoopGroup for them), so it didn't seem necessary to still call this out

Examples/VaporExample/README.md
```diff
@@ -43,19 +43,14 @@ If you'd like to point the application to a MongoDB server elsewhere (e.g. on [M
 
 The call to `configure()` initializes a global `MongoClient` to back your application. `MongoClient` is implemented with that approach in mind: it is safe to use across threads, and is backed by a [connection pool](https://en.wikipedia.org/wiki/Connection_pool) which enables sharing resources throughout the application.
 
-Throughout your application, you can access the global client via `app.mongoDB.client`. Note that the global client may return `EventLoopFuture`s on *any* `EventLoop` in the application's `EventLoopGroup`, so if you use this client you will need to ensure you "hop" the futures back to the event loop you are currently on. See the `EventLoopFuture` [documentation](https://apple.github.io/swift-nio/docs/current/NIO/Classes/EventLoopFuture.html) for more details.
```
kmahar, line 46: I suspected, and confirmed with folks on the Vapor discord server, that the whole "stay on the same event loop that the request is executing on" requirement from the future based API is gone when using async/await, since the request handler closures are not running on event loops anymore but instead on `Task`s.

given that, I don't think that there is a reason for async/await users to use the "event loop bound" clients/DBs/collections we added support for last year. I don't think it is particularly harmful for users to do this -- it just means that anytime we cross into EventLoop land from the Request's `Task` we would be on the same `EventLoop` -- but there's also no real benefit, and there might even be some performance impacts by constraining things to particular event loops. So I've updated the example here to just direct users to the global client instead, and for 2.0 of MongoDBVapor we'd maybe either get rid of the extension on `Request` or update it so that it just points to the global client.

I filed SWIFT-1449 about these changes

Examples/VaporExample/README.md
patrickfreed: I think a back link strikes a good compromise, no need to maintain a separate example here for the reasons you list IMO.

patrickfreed (approved):

Examples/VaporExample/README.md
isabelatkinson: I think linking back to the previous version is sufficient for now, as it keeps the ELG examples accessible while making async/await the more discoverable option. If there's significant interest in the ELG examples or we need to make changes to them for some reason we still have the opportunity to reintroduce them down the line, but for now I don't think it's super necessary.

isabelatkinson (approved):
whoops, meant to hit approve

PR #727, opened 2022-01-12
SWIFT-1447 Add filtering by tag sets to finding suitable servers

nspektor:
- Added filtering by tag sets in a helper method
- reformatted the test file and added some tag set tests


Sources/MongoSwift/SDAM.swift
```diff
@@ -427,12 +432,25 @@ extension TopologyDescription {
...
-        // While waiting for the above to be implemented, this helper just returns the servers it was passed
-        servers
+        return matches
+    }
+
+    internal func doTagsMatch(serverTags: [String: String], tagSets: [BSONDocument]) -> Bool {
+        if tagSets.isEmpty {
+            return true
+        }
+        for tagSet in tagSets {
```
kmahar, line 448: I think the logic here needs to be a little different. the spec says:

> If the tag_sets list is not empty, then tag sets are tried in order until a tag set matches at least one candidate server. All candidate servers matching that tag set are eligible servers. Subsequent tag sets in the list are ignored.

currently, I think the list of matches above will contain any server that matches any tag set. if I am understanding the spec wording correctly, instead, we need to find the first tag set that matches any server, and then return all servers matching that tag set. 

here's the equivalent Rust logic for some inspiration: https://github.com/mongodb/mongo-rust-driver/blob/cc7a1c3046e3eac925dcaf85d1e4876bd2a897f8/src/sdam/description/topology/server_selection/mod.rs#L370-L386

it would be good to add a test around this behavior as well

Tests/MongoSwiftTests/ServerSelectionTests.swift
```diff
@@ -6,62 +6,135 @@ import TestsCommon
...
-    func testServerSelection() throws {
-        let standaloneServer = ServerDescription(type: .standalone)
-        let rsPrimaryServer = ServerDescription(type: .rsPrimary)
-        let rsSecondaryServer1 = ServerDescription(type: .rsSecondary)
-        let rsSecondaryServer2 = ServerDescription(type: .rsSecondary)
-        let mongosServer = ServerDescription(type: .mongos)
-
-        // unknown
+    // Servers
+    let standaloneServer = ServerDescription(type: .standalone)
```
kmahar, line 10: nice idea pulling these out into test class properties 👍

Tests/MongoSwiftTests/ServerSelectionTests.swift
```diff
@@ -6,62 +6,135 @@ import TestsCommon
...
+            maxStalenessSeconds: nil
+        )
+
+        let suitable2 = try topology
+            .findSuitableServers(readPreference: secondaryReadPreferenceWithTagSet2)
+        expect(suitable2[0].type).to(equal(.rsSecondary))
+        expect(suitable2).to(haveCount(2))
+
+        // invalid tag set passing
+        let primaryReadPreferenceWithTagSet = try ReadPreference(
```
kmahar, line 112: I was surprised this works but I realized this initializer does not actually validate that tagSets aren't specified when the mode is primary.

could you please add validation for that in the initializer? 

it would prevent us from testing this case but I think that's ok; hitting the new `InternalError` you added is unexpected and points to a bug in our own code somewhere

kmahar (changes requested):
looking good so far! just some comments around edge cases

Sources/MongoSwift/SDAM.swift
```diff
@@ -417,8 +417,13 @@ extension TopologyDescription {
                 }
                 let secondaries = self.servers.filter { $0.type == .rsSecondary }
                 return self.filterReplicaSetServers(readPreference: readPreference, servers: secondaries)
-            default:
+            default: // or .primary
                 // the default mode is 'primary'.
+                guard readPreference?.tagSets == nil || readPreference?.tagSets == [BSONDocument()] else {
+                    throw MongoError.InternalError(
+                        message: "non-empty tag set was given in tag_sets and the mode field was 'primary'"
```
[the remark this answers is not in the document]
nspektor: I think this whole check is not needed anymore right? With the check added to the initializer

Sources/MongoSwift/SDAM.swift
nspektor: Yep, thanks!

Tests/MongoSwiftTests/ServerSelectionTests.swift
nspektor: Yeah it makes more sense in the initializer, I added it. I just modified the test to make sure the error is thrown from the initializer

Sources/MongoSwift/SDAM.swift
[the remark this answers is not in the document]
kmahar: yeah it seems fine to just remove it since we don't allow the invalid ones to be constructed now

kmahar (changes requested):
looking good! just a couple more little things

Sources/MongoSwift/SDAM.swift
```diff
@@ -384,7 +384,7 @@ public struct TopologyDescription: Equatable {
 }
 
 extension TopologyDescription {
-    internal func findSuitableServers(readPreference: ReadPreference? = nil) -> [ServerDescription] {
+    internal func findSuitableServers(readPreference: ReadPreference? = nil) throws -> [ServerDescription] {
```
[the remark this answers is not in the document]
nspektor: oh right, nice! cleans up the tests a lot

kmahar (approved):
lgtm! great job with this

Sources/MongoSwift/ReadPreference.swift
```diff
@@ -202,6 +202,16 @@ public struct ReadPreference: Equatable {
                 )
             }
         }
+        if let tagSets = tagSets {
+            if mode == .primary {
+                guard tagSets == [BSONDocument()] else {
+                    throw MongoError.InvalidArgumentError(
+                        message: "Invalid \(MONGOC_URI_READPREFERENCETAGS) \(tagSets): " +
```
isabelatkinson, line 209: Can we update this to use `MongoConnectionString.OptionName.readPreferenceTags`? This way when we eventually stop using libmongoc we won't need to change this message. The `OptionName` enum is currently `fileprivate` so you'll need to update it to `internal`.

isabelatkinson (commented):
looks good, just one small comment!

Sources/MongoSwift/ReadPreference.swift
nspektor: sure

isabelatkinson (approved):
lgtm! nice job with this project :)

patrickfreed (approved):
LGTM!

PR #729, opened 2022-01-20
Fix `MongoConnectionString` errors

isabelatkinson:
The patch on this PR has been updated to run all build variants.

### DIFF
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -416,6 +416,8 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
             // If no other authentication options or defaultAuthDB were provided, we should use "admin" as the
             // credential source. This will be overwritten later if a defaultAuthDB or an authSource is provided.
             credential.source = "admin"
+            // Overwrite the sourceFromAuthSource field to false as the source is a default.
+            credential.sourceFromAuthSource = false
             self.credential = credential
         }
 
@@ -462,6 +464,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                 self.credential = MongoCredential()
             }
             self.credential?.source = decoded
+            self.credential?.sourceFromAuthSource = false
         }
 
         guard authDatabaseAndOptions.count == 2 else {
@@ -717,6 +720,8 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                     ?? self.defaultAuthDB
                     ?? "admin"
                 self.credential?.source = defaultSource
+                // Overwrite the sourceFromAuthSource field to false as the source is a default.
+                self.credential?.sourceFromAuthSource = false
             }
             if self.credential?.mechanism != nil {
                 // credential cannot be nil within the external conditional
@@ -931,7 +936,9 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
             }
             return property
         }.joined(separator: ","))
-        uri.appendOption(name: .authSource, option: self.credential?.source)
+        if self.credential?.sourceFromAuthSource == true {
+            uri.appendOption(name: .authSource, option: self.credential?.source)
+        }
         uri.appendOption(name: .compressors, option: self.compressors?.map {
             switch $0._compressor {
             case let .zlib(level):
--- a/Sources/MongoSwift/MongoConnectionString.swift
+++ b/Sources/MongoSwift/MongoConnectionString.swift
@@ -120,10 +120,6 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
 
         /// Initializes a ServerAddress, using the default localhost:27017 if a host/port is not provided.
         internal init(_ hostAndPort: String = "localhost:27017") throws {
-            guard !hostAndPort.contains("?") else {
-                throw MongoError.InvalidArgumentError(message: "\(hostAndPort) contains invalid characters")
-            }
-
             // Check if host is an IPv6 literal.
             if hostAndPort.first == "[" {
                 let ipLiteralRegex = try NSRegularExpression(pattern: #"^\[(.*)\](?::([0-9]+))?$"#)
@@ -228,7 +224,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         fileprivate var wTimeoutMS: Int?
         fileprivate var zlibCompressionLevel: Int?
 
-        fileprivate init(_ uriOptions: String) throws {
+        fileprivate init(_ uriOptions: Substring) throws {
             let options = uriOptions.components(separatedBy: "&")
             // tracks options that have already been set to error on duplicates
             var setOptions: Set<String> = []
@@ -384,6 +380,7 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
     /// - Throws:
     ///   - `MongoError.InvalidArgumentError` if the input is invalid.
     public init(string input: String) throws {
+        // Parse the connection string's scheme.
         let schemeAndRest = input.components(separatedBy: "://")
         guard schemeAndRest.count == 2, let scheme = Scheme(schemeAndRest[0]) else {
             throw MongoError.InvalidArgumentError(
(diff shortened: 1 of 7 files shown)


Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -90,6 +90,11 @@ final class DNSSeedlistTests: MongoSwiftTestCase {
 
     func runDNSSeedlistTests(_ tests: [(String, DNSSeedlistTestCase)]) throws {
         for (fileName, testCase) in tests {
+            // TODO: SWIFT-1455: unskip these test
+            if fileName == "loadBalanced-no-results.json" || fileName == "loadBalanced-true-multiple-hosts.json" {
```
isabelatkinson, line 94: These tests need to be skipped because their URIs are considered invalid due to the issue I brought up in the c-driver channel.

The reason these tests are now failing with the introduction of `MongoConnectionString` is related to client cleanup. Prior to the introduction of `MongoConnectionString`, if a URI considered invalid by libmongoc were passed to a `MongoClient`, creating the client would fail because the underlying `ConnectionString` would attempt to create a `mongoc_uri_t`. This means that when `withTestClient` was called on line 113/118 below, client creation would fail [here](https://github.com/mongodb/mongo-swift-driver/blob/79c9683d56f92540f4065f40b9f55e1911a1ff5b/Tests/MongoSwiftTests/AsyncTestUtils.swift#L97) _and the `defer` block to clean up the client wouldn't be set_. (Because of this, the test wasn't really running at all; due to when libmongoc performs SRV lookup, the correct behavior for this test would be for client creation to succeed and the expected error to occur when SRV lookup happens when attempting to select a server.)

However, because `MongoConnectionString` does not create an underlying `mongoc_uri_t`, client creation no longer fails when a URI considered invalid by libmongoc is created. This is generally fine because any error that libmongoc would throw should be caught by our parser instead, but in this case libmongoc was rejecting valid URIs. This means that a `MongoClient` could be created with a URI that libmongoc thought was invalid, and more specifically that client creation would succeed where it previously failed above and the `defer` block to clean up the client would be set. This is problematic specifically when closing the connection pool. The [future created to open the connection pool](https://github.com/mongodb/mongo-swift-driver/blob/79c9683d56f92540f4065f40b9f55e1911a1ff5b/Sources/MongoSwift/ConnectionPool.swift#L87) calls `withMongocURI`, which immediately fails when attempting to create a `mongoc_uri_t`. Effectively this means that the pool is stuck in an `.opening` state, as any attempt to open the pool fails and the state never transitions. Thus, when attempting to close the client within the `defer` block, closing its `ConnectionPool` fails [here](https://github.com/mongodb/mongo-swift-driver/blob/79c9683d56f92540f4065f40b9f55e1911a1ff5b/Sources/MongoSwift/ConnectionPool.swift#L150) because the pool can't be transitioned out of `.opening`.

Anyway, this tangent isn't particularly related to the code changes and should be irrelevant when we vendor in the fix, but I thought it was an interesting bug :)

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -432,45 +454,44 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
                     message: "A port cannot be specified in a mongodb+srv connection string"
                 )
             }
+            guard hosts[0].host.filter({ $0 == "." }).count >= 2 else {
+                throw MongoError.InvalidArgumentError(
+                    message: "The host specified in a mongodb+srv connection string must contain a host name, a domain"
+                        + " name, and a TLD"
```
kmahar, line 460: what's a TLD? (should we spell out whatever it stands for?)

Tests/MongoSwiftTests/DNSSeedlistTests.swift
kmahar: wow, nice job investigating all that!

> Because of this, the test wasn't really running at all; due to when libmongoc performs SRV lookup, the correct behavior for this test would be for client creation to succeed and the expected error to occur when SRV lookup happens when attempting to select a server.

would it make sense to update the runner logic below so we would have caught this and not mistook an error during client creation for an error during SRV lookup? or are there any DNS seedlist tests that should actually throw during client creation?

> Effectively this means that the pool is stuck in an .opening state, as any attempt to open the pool fails and the state never transitions.

I'm also wondering if we ought to do anything about this. on the one hand it seems like it can only result from a bug/mismatch in our connection string parsing and theirs... but on the other hand it seems the more expected/graceful way for us to handle a situation where we fail to open the pool would be to automatically consider the pool "closed" if we fail to open it to begin with. ideally, the test would have failed with some error indicating we failed to open the pool but not led to us hitting the assertion failure in `deinit`.  thoughts? (cc @patrickfreed)

kmahar (commented):
looks good and nice work getting to the bottom of this! just a few questions

Tests/MongoSwiftTests/DNSSeedlistTests.swift
isabelatkinson: [This test](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/tests/load-balanced/srvMaxHosts-conflicts_with_loadBalanced-true.json) should fail during client creation because of an invalid combination of options in the connection string, although there's an argument to be made that it should just be a URI options test because it doesn't involve lookup/resolving options. It seems the usage of `withTestClient` can be tricky in cases like these because it's difficult to disambiguate where an error came from if one occurs -- in the case that client creation should always succeed, it would make sense to include some sort of assertion to avoid a client creation error getting mistaken for an expected failure from the closure provided, but that's not true here. This might just be a pitfall in the way the spec tests are written; without specific filename logic we can't really do much to separate out where the error should be occurring. One thing to note is that if we do lookup during connection string/client creation when we implement it in pure Swift we can simplify this test to expect creation of a `MongoClient` to succeed/fail.

I was also thinking about other cases in which a pool could get stuck in an opening state like this. The `deinit` error was pretty confusing and it would be nice to have a more graceful way to indicate the issue. I do agree it seems more graceful to close the pool if it can't be opened, but I'm not sure if there are situations in which opening could initially fail but then succeed after another attempt -- in that case, it'd be best not to close after a failure. I'm also not sure how we would differentiate there.

kmahar: > This might just be a pitfall in the way the spec tests are written; without specific filename logic we can't really do much to separate out where the error should be occurring.

yeah, it sounds that way, I suppose we should just leave this as-is for now. 

>  I'm not sure if there are situations in which opening could initially fail but then succeed after another attempt -- in that case, it'd be best not to close after a failure.

so I believe we only ever attempt to open the pool once. the logic looks a little funny there since we `wait()` on the future in a few places, but doing so doesn't lead to re-executing the future, it just blocks all those code paths on the future having completed (or failed) its sole execution. 
one solution might be to add an additional case to the pool's state enum like `openingFailed(Error)` which we transition to if opening fails. Then any time we attempt to use the pool after that, we can throw an error indicating the pool failed to open, and in `close()` we can just go ahead and transition to `closed` without erroring.

regardless of it we take that approach or something else I think this is probably out of scope for your work here and we could just handle it in a follow up ticket

kmahar (approved):
LGTM!

Tests/MongoSwiftTests/DNSSeedlistTests.swift
patrickfreed: Yeah I agree that the stuck-opening situation seems like it needs to be addressed, and I like the idea of a separate, terminal error state for the connection pool as a solution. Separate work also sounds good to me too, and let's make sure to file a ticket about this before merging (with some of this context, but you can also just link to this thread) so we don't forget.

patrickfreed (approved):
Awesome job investigating this tricky situation!

patrickfreed:
It looks like TSAN is upset with something, though I'm not sure how to interpret the logs / if it's related to this change.

kmahar:
The TSan failure looks like the false positive with `withUnsafeContinuation` I filed a Swift ticket about a while back and is not our fault: https://bugs.swift.org/browse/SR-15498

We can ignore it, though it has been coming up a lot so hopefully it gets resolved soon. I  just realized we're still testing on Swift 5.5.1 so I think I'll try upgrading us to 5.5.2 on green build Friday and see if that helps. (the Swift ticket is still open but those are often out of sync with their corresponding radar tickets so it may be resolved...)

Tests/MongoSwiftTests/DNSSeedlistTests.swift
isabelatkinson: Filed SWIFT-1459 to address this

PR #730, opened 2022-01-26
SWIFT-1457 Implement server selection logic test runner

isabelatkinson:
spec for tests: https://github.com/mongodb/specifications/tree/master/source/server-selection/tests

### DIFF
--- a/Sources/TestsCommon/CodableExtensions.swift
+++ b/Sources/TestsCommon/CodableExtensions.swift
@@ -145,9 +145,8 @@ extension ReadPreference: StrictDecodable {
     internal typealias CodingKeysType = CodingKeys
 
     public init(from decoder: Decoder) throws {
-        try Self.checkKeys(using: decoder)
-
         if let container = try? decoder.container(keyedBy: CodingKeys.self) {
+            try Self.checkKeys(using: decoder)
             let mode = try container.decode(Mode.self, forKey: .mode)
             let tagSets = try container.decodeIfPresent([BSONDocument].self, forKey: .tagSets)
             try self.init(mode, tagSets: tagSets)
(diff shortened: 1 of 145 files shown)


Sources/TestsCommon/CodableExtensions.swift
```diff
@@ -174,3 +139,78 @@ extension MongoClientOptions: StrictDecodable {
         )
     }
 }
+
+extension ReadPreference: StrictDecodable {
```
kmahar, line 143: `StrictDecodable` was added to satisfy a unified test runner requirement - "the test runner MUST raise an error for an unsupported argument [for an operation]" - since without it we just wouldn't even try to decode a field if it were not something the options type supported. but it is a nice idea to use it generally for test-only decodable conformances to avoid silently ignoring fields 👍

Tests/MongoSwiftTests/ServerSelectionTests.swift
```diff
@@ -5,132 +5,33 @@ import NIO
...
-            .findSuitableServers(readPreference: nil)
-        expect(suitable2).to(haveCount(0))
-
-        let suitable3 = replicaSetNoPrimaryTopology
-            .findSuitableServers(readPreference: self.primaryPrefReadPreferemce)
-        expect(suitable3[0].type).to(equal(.rsSecondary))
-        expect(suitable3).to(haveCount(2))
+private struct ServerSelectionLogicTestFile: Decodable {
+    let topologyDescription: TopologyDescription
+    let operation: String
```
kmahar, line 10: nit: can we make this an enum with read/write cases? it's a little more code, but I think it makes the purpose of the property more obvious

kmahar (commented):
nice! lgtm mod one nit

patrickfreed (approved):
LGTM!

PR #733, opened 2022-02-02
SWIFT-1430 Don't commit Package.resolved

kmahar:


.gitignore
```diff
@@ -1,72 +1,3 @@
-# Xcode
```
kmahar, line 1: basically none of this was relevant at all, this is likely what got generated for new Swift projects back around the time of Swift 4, so I cleaned it up to only include actually relevant things

kmahar:
failure is the now-broken load balancer tests

isabelatkinson (approved):

PR #734, opened 2022-02-02
SWIFT-1383 Test that driver only creates implicit session after connection checkout

kmahar:
This didn't require any driver changes, but there was a prose test added so I went ahead with writing that. 

Prose test description:
```
To confirm that implicit sessions only allocate their server session after a successful connection checkout

    * Create a MongoClient with the following options: ``maxPoolSize=1`` and ``retryWrites=true``
    * Attach a command started listener that collects each command's lsid
    * Initiate the following concurrent operations
      * insertOne
      * deleteOne
      * updateOne
      * bulkWrite ``[ { updateOne } ]``
      * findOneAndDelete
      * findOneAndUpdate
      * findOneAndReplace
      * find
    * Wait for all operations to complete
    * Assert that all commands contain the same lsid
```

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1399,7 +1399,7 @@ buildvariants:
   display_name: "${sanitize} ${os-fully-featured} ${ssl-auth}"
   matrix_spec:
     os-fully-featured: "ubuntu-18.04"
-    swift-version: "5.5"
+    swift-version: "5.4" # TODO SWIFT-1499: update to Swift 5.5
     ssl-auth: "*"
     sanitize: "tsan"
   tasks:
--- a/Tests/MongoSwiftSyncTests/ClientSessionTests.swift
+++ b/Tests/MongoSwiftSyncTests/ClientSessionTests.swift
@@ -320,6 +320,8 @@ final class SyncClientSessionTests: MongoSwiftTestCase {
         }
     }
 
+    /// Sessions spec test 13 was easier to write with concurrency and lives in MongoSwiftTests/AsyncAwaitTests.swift.
+
     /// Test that causal consistency guarantees are met on deployments that support cluster time.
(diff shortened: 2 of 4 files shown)


Tests/MongoSwiftSyncTests/ClientSessionTests.swift
```diff
@@ -512,17 +514,6 @@ final class SyncClientSessionTests: MongoSwiftTestCase {
             expect(startedEvents).to(haveCount(1))
             expect(startedEvents[0].command["readConcern"]?.documentValue?["afterClusterTime"]).to(beNil())
         }
-
-        // Causal consistency spec test 10: When an unacknowledged write is executed in a causally consistent
```
kmahar, line 516: this is that outdated sessions test we were supposed to delete anyway

.evergreen/config.yml
```diff
@@ -1399,7 +1399,7 @@ buildvariants:
   display_name: "${sanitize} ${os-fully-featured} ${ssl-auth}"
   matrix_spec:
     os-fully-featured: "ubuntu-18.04"
-    swift-version: "5.5"
+    swift-version: "5.4" # TODO SWIFT-1499: update to Swift 5.5
```
kmahar, line 1402: filed SWIFT-1499 to track eventually updating this to 5.5 once relevant Swift issues are fixed. we could have stayed with 5.5 and updated the test scripts so that we would run every suite besides the async/await ones, but that would have required either a very long/ugly regex to match every suite name that does not contain "AsyncAwait", or pulling out the async/await tests into their own target, so this seemed like a reasonable solution for now

isabelatkinson (approved):

patrickfreed (approved):
love to see how async/await made this test super easy to read :heart_eyes:

PR #735, opened 2022-02-03
SWIFT-1465 Filter servers by max staleness

isabelatkinson:


Sources/MongoSwift/SDAM.swift
```diff
@@ -444,6 +478,84 @@ extension TopologyDescription {
...
+    /// If staleness cannot be calculated due to an absence of values, `nil` is returned.
+    fileprivate func calculateStalenessSeconds(
+        _ primary: ServerDescription?,
+        _ maxLastWriteDate: Date?,
+        _ heartbeatFrequencyMS: Int
+    ) -> Int? {
+        guard self.type == .rsSecondary else {
+            return 0
+        }
+        guard let lastWriteDate = self.lastWriteDate else {
```
kmahar, line 508: in what circumstances would a server have no last write date? might be in the specs somewhere but I could not find it in the max staleness one

Tests/MongoSwiftTests/ServerSelectionTests.swift
```diff
@@ -23,19 +27,44 @@ private enum OperationType: String, Decodable {
...
+            let selectedServers: [ServerDescription]
+            do {
+                selectedServers = try test.topologyDescription.findSuitableServers(
+                    readPreference: readPreference,
+                    heartbeatFrequencyMS: heartbeatFrequencyMS
+                )
+            } catch where test.error != true {
+                throw error
+            } catch {
+                continue
```
kmahar, line 47: should we assert on the error type here, or does the type thrown vary across tests?

kmahar (commented):
looking great! just a few minor comments/questions

Tests/MongoSwiftTests/ServerSelectionTests.swift
isabelatkinson: this should always be an `IllegalArgumentError` regarding max staleness validity, added an assertion

Sources/MongoSwift/SDAM.swift
isabelatkinson: According to the [spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#lastwritedate-and-optime), this field is always included in hello responses from 3.4+ servers, so theoretically it should always be present on servers the driver interacts with. If that's the case it would probably make sense to make this field non-optional -- I know we have a ticket in Rust for making more of the fields in our hello response struct non-optional, has that been discussed for Swift? If not, it's something to consider for 2.0.

kmahar: ah thanks for the link - yeah it sounds like we can make it non optional in the future then, could you file a ticket please?

kmahar (approved):
lgtm! nice work

patrickfreed (approved):
LGTM!

PR #736, opened 2022-02-03
SWIFT-1466 Implement `updateAverageRoundTripTime` for `ServerDescription`

isabelatkinson:


Sources/MongoSwift/SDAM.swift
```diff
@@ -447,3 +479,14 @@ extension TopologyDescription {
         return []
     }
 }
+
+extension ServerDescription {
+    internal mutating func updateAverageRoundTripTime(roundTripTime: Double) {
+        if let oldAverageRTT = self.averageRoundTripTimeMS {
+            let alpha = 0.2
+            self.averageRoundTripTimeMS = alpha * Double(roundTripTime) + (1 - alpha) * oldAverageRTT
```
kmahar, line 487: looks like `roundTripTime` is already a `Double`, so is casting it a `Double` here needed? (same q for line 489)

kmahar (commented):
LGTM mod one question

Sources/MongoSwift/SDAM.swift
isabelatkinson: oh yeah it was originally an `Int` so it needed to be cast but not anymore, removed

kmahar (approved):

patrickfreed (approved):

PR #738, opened 2022-02-09
SWIFT-1459 Handle errors that occur when opening connection pool

kmahar:
Introduces a new state for a `ConnectionPool` indicating that the asynchronous work to start the pool failed. This can occur if, for example, an invalid option is specified via TXT record.

### DIFF
--- a/Sources/MongoSwift/ConnectionPool.swift
+++ b/Sources/MongoSwift/ConnectionPool.swift
@@ -45,6 +45,8 @@ internal class ConnectionPool {
         case opening(future: EventLoopFuture<OpaquePointer>)
         /// Indicates that the `ConnectionPool` is open and using the associated pointer to a `mongoc_client_pool_t`.
         case open(pool: OpaquePointer)
+        /// Indicates that the `ConnectionPool` failed to open due to the associated error.
+        case failedToOpen(error: Error)
         /// Indicates that the `ConnectionPool` is in the process of closing. Connections can be checked back in, but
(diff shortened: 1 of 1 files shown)


Sources/MongoSwift/ConnectionPool.swift
```diff
@@ -147,10 +171,18 @@ internal class ConnectionPool {
             case let .open(pool):
                 self.state = .closing(pool: pool)
             case let .opening(future):
-                let pool = try future.wait()
-                self.state = .closing(pool: pool)
+                do {
+                    let pool = try future.wait()
+                    self.state = .closing(pool: pool)
+                } catch {
+                    self.state = .closed
```
kmahar, line 178: I wasn't sure whether we should also throw the error here in addition to transitioning the state. currently if you did something like this you would never see the error:
```swift
let client = try MongoClient("uri-triggering-the-problem-here")
try client.syncClose()
```

but if you tried to use the client at all in between creation and close, you would see it.

ultimately a case like this seemed unlikely to me - why create a client to never use it? - and propagating the error would have complicated client shutdown logic (would we consider the client to have failed closing if its pool failed to open?) so I went with this for now, but let me know what y'all think

isabelatkinson: yeah, I agree that a failure to open the pool doesn't necessarily indicate a failure to close the pool (and thereby the client) -- this seems fine to me

isabelatkinson (approved):

PR #739, opened 2022-02-09
Automatically use latest patch for each Swift release in CI

kmahar:
We now specify Swift versions of the form x.y, and at installation time query the Swift GitHub repo to find the latest matching release tag and write the value to a file.
We then add an expansion to set the SWIFT_VERSION variable used throughout the config/test scripts.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -334,6 +334,26 @@ functions:
       params:
         working_dir: "src"
         script: |
+          set -o xtrace
+          set -o errexit
+
+          if [ "${SWIFT_MINOR_VERSION}" = "main-snapshot" ]; then
+              export SWIFT_VERSION="${SWIFT_MINOR_VERSION}"
+          else
+              # otherwise, find the latest patch release for the specified version
+              ${PYTHON} -m virtualenv ./requests-env
+              ./requests-env/${VENV_BIN_DIR}/python3 -m pip install requests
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -330,6 +330,26 @@ functions:
...
 
   "install swift":
+    - command: shell.exec
+      params:
+        working_dir: "src"
+        script: |
+          set -o xtrace
+          set -o errexit
+
+          if [ "${SWIFT_MINOR_VERSION}" = "main-snapshot" ]; then
```
isabelatkinson, line 340: what does "main-snapshot" signify here?

isabelatkinson (approved):
looks good, just a question for my own understanding!

.evergreen/config.yml
kmahar: "main-snapshot" is basically just a placeholder string we use in the Evergreen config to mean "latest development snapshot from the Swift main branch". 
in the Swift installation/configuration scripts, if we see that version, we look at swiftenv's available snapshots to determine the actual version to use, which is usually something like DEVELOPMENT-SNAPSHOT-2022-01-09:

https://github.com/mongodb/mongo-swift-driver/blob/d63c53d4222f0f2277436e77e0d55cac0dec400f/.evergreen/install-swift.sh#L23-L25

patrickfreed (approved):
Looks great! I may need to steal some of this logic for `swiftly`...

PR #743, opened 2022-02-22
Fix Evergreen failures on replica sets and against serverless

kmahar:
* Split tailable cursor tests into their own suites so they aren't run as part of the serverless tests
* Update `testInitialReplicaSetDiscovery` to no longer assert on the error type

### DIFF
--- a/Tests/MongoSwiftSyncTests/MongoCursorTests.swift
+++ b/Tests/MongoSwiftSyncTests/MongoCursorTests.swift
@@ -56,6 +56,111 @@ final class MongoCursorTests: MongoSwiftTestCase {
         }
     }
 
+    func testNext() throws {
+        try self.withTestNamespace { _, _, coll in
+            // query empty collection
+            var cursor = try coll.find()
+            expect(cursor.next()).to(beNil())
+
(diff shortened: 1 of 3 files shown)


Tests/MongoSwiftSyncTests/MongoCursorTests.swift
```diff
@@ -56,6 +56,113 @@ final class MongoCursorTests: MongoSwiftTestCase {
         }
     }
```
kmahar, line 58: the diff is being calculated/displayed weirdly, but all I did in this file was create a new test class and move the tests using tailable cursors down into it.

Tests/MongoSwiftTests/MongoCursorTests.swift
```diff
@@ -53,6 +53,99 @@ final class AsyncMongoCursorTests: MongoSwiftTestCase {
         }
     }
```
kmahar, line 55: same comment re weird diff here, I just moved tests using tailable cursors down into their own class

isabelatkinson (approved):

patrickfreed (approved):

PR #745, opened 2022-02-23
SWIFT-1470 Fix failing load balancer tests

patrickfreed:
SWIFT-1470 / SWIFT-1319


Sources/TestsCommon/CommonTestUtils.swift
```diff
@@ -188,7 +188,6 @@ public enum TestTopologyConfiguration: String, Decodable {
             helloReply["isreplicaset"] != true
         {
             self = .single
-            // TODO: SWIFT-1319: eventually, we should be able to just check for serviceId here.
```
patrickfreed, line 191: as discussed offline, serviceId only shows up in the initial handshake, so we can't check for it here and need to still rely on the environment variable.

etc/vendor-libmongoc.sh
```diff
@@ -153,8 +153,6 @@ echo "RENAMING header files"
 echo "PATCHING libmongoc"
 git apply ${ETC_DIR}/lower-minheartbeatfrequencyms.diff
 git apply ${ETC_DIR}/inttypes-non-modular-header-workaround.diff
-# TODO SWIFT-1319: Remove.
-git apply ${ETC_DIR}/expose-mock-service-id.diff
```
kmahar, line 157: should we delete this diff file now too?

kmahar (approved):
LGTM mod one comment

isabelatkinson (approved):
lgtm mod kaitlin's comment

PR #747, opened 2022-03-09
SWIFT-1510 Lower #available checks to run automatic cleanup in deinits to macOS 10.15+

kmahar:
1) Fixes a bug where automatic cleanup in deinit would not happen for cursors/change streams/sessions on macOS < 12, even when concurrency is available

2) Adds a GitHub action to start running concurrency tests on macOS 11 + Xcode 13.2. Once [BUILD-14630](https://jira.mongodb.org/browse/BUILD-14630) is complete we can do this on Evergreen instead.

3) step 2 required removing the compiler directive to only define our concurrency-specific tests on Linux. doing so led our Evergreen tests to start failing, because despite being on macOS 10.15 where concurrency is now available, those still use a very old Xcode version whose bundled macOS SDK does not include the concurrency library. so while the Swift compiler can _import_ `_Concurrency`, the concurrency library is not actually there to _load_. after doing some research, I realized it is possible to work around the old Xcode version by setting the `DYLD_LIBRARY_PATH` manually -- the concurrency library _does_ ship with the Swift 5.5.3 toolchain, but SwiftPM seems to default to looking at the `DYLD_LIBRARY_PATH` for the system Swift version that ships with Xcode. so I have gone ahead and done that, and it actually enables us to run all of our concurrency tests on macOS on Evergreen now! 🥳

my discovery in step 3 perhaps makes the GitHub action less useful, though arguably it is still somewhat valuable to test that things work in the "normal" way with more recent macOS + Xcode without having to set `DYLD_LIBRARY_PATH`, so I left it in for now... thoughts on whether we should keep it?

### DIFF
--- a/.evergreen/configure-swift.sh
+++ b/.evergreen/configure-swift.sh
@@ -29,6 +29,7 @@ if [ "$OS" == "darwin" ]; then
         sudo xcode-select -s /Applications/Xcode12.app
     fi
 
+    # TODO SWIFT-1421: remove this once we have new Xcode on Evergreen to test with
     export DYLD_LIBRARY_PATH=${SWIFTENV_ROOT}/versions/${SWIFT_VERSION}/usr/lib/swift/macosx/
 fi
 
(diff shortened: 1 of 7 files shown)


patrickfreed (approved):
LGTM. Leaving the GH action test in until we don't have any workarounds on the evergreen version sounds good to me.

isabelatkinson (approved):

PR #748, opened 2022-03-14
SWIFT-1277, SWIFT-1263, SWIFT-1507 Sync latest initial DNS seedlist discovery tests

kmahar:
This sync the latest copy of the initial DNS seedlist discovery tests and makes some modifications to the runner to support running them.

With this PR we can also close out SWIFT-1277 and SWIFT-1263 -- Isabel added Swift support for these options in the new connection string type and our latest vendor of libmongoc brought in the libmongoc support. We already have synced the corresponding URI options tests. I did notice that we didn't add these options to `MongoClientOptions` though so I went ahead with doing that here for completeness.

See mongodb/mongo-rust-driver#595 for equivalent Rust changes.

### DIFF
--- a/Tests/MongoSwiftTests/MongoConnectionStringTests.swift
+++ b/Tests/MongoSwiftTests/MongoConnectionStringTests.swift
@@ -661,4 +661,42 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         connString7.credential?.source = nil
         expect(connString7.description).toNot(contain("authsource"))
     }
+
+    func testSrvMaxHosts() throws {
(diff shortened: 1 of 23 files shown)


Tests/MongoSwiftTests/MongoConnectionStringTests.swift
```diff
@@ -661,4 +661,42 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         connString7.credential?.source = nil
         expect(connString7.description).toNot(contain("authsource"))
     }
+
```
kmahar, line 664: for the connection string only case these already have good coverage from the spec tests, but I figured I would add some additional testing here around setting the options programmatically.

isabelatkinson (approved):

patrickfreed (approved):

PR #750, opened 2022-03-24
Create user so ping succeeds in DNS seedlist tests

kmahar:
It seems like I forgot to re-run these with auth on after adding the ping command. This led tests to start failing because the user we attempt to authenticate as doesn't exist.

Added logic to create the missing user if needed, inspired by corresponding Rust logic: https://github.com/mongodb/mongo-rust-driver/blob/0a50512aa0178f88bec2a1dc651f93e0b08a4f12/src/test/spec/initial_dns_seedlist_discovery.rs#L139-L157


kmahar:
As of a few commits ago we seem to be getting some intermittent leak checker failures including on my patch here. Going to look into it for GBF today

patrickfreed (approved):

isabelatkinson (approved):

PR #754, opened 2022-05-19
SWIFT-1404 Update serverless testing for load balancer fronting single proxy

kmahar:
This implements the required changes for DRIVERS-1967. In summary:
* the `SERVERLESS_DRIVERS_GROUP` variable has been updated on our Evergreen project
* drivers-evergreen-tools only generates a single URI now, `SERVERLESS_URI`
* When testing against serverless, we ignore `useMultipleMongoses` and always use the single LB URI 
* We now run _all_ unified spec tests against serverless 

[successful patch](https://spruce.mongodb.com/version/6286959b3e8e867c9497ebfa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


Tests/Specs/change-streams/tests/unified/change-streams.json
```diff
@@ -1,6 +1,15 @@
 {
   "description": "change-streams",
-  "schemaVersion": "1.0",
+  "schemaVersion": "1.4",
```
kmahar, line 3: this was a manual change to prevent these from being run due to the "unified" addition to the test filter provided in run-serverless-tests.sh.
the latest version of the file has the same runOnRequirements to skip the tests on the serverless, but I can't sync the whole thing without making a number of other changes to skip new tests and support schema version 1.7, so I just copied this part in for now.

patrickfreed (approved):

PR #759, opened 2022-06-16
SWIFT-1533 Add `to` for renames with `ChangeStreamEvents`

rchhaya:
Adding the `to` field that is needed for rename events regarding the `ChangeStreamEvent` struct. Incorporated new unified spec test updates as outlined by `372747fca128f449abbbb01b200dc9a2cabf522d`. To correctly account for these rename events, created a new `UnifiedRename` struct in `UnifiedCollectionOperations`. 
Tests were failing due to lack of comment implementation for various `ChangeStream` helpers, so tests were explicitly omitted in `SyncChangeStreamTests`. The `comment` functionality was partially implemented for `CreateChangeStream` in `UnifiedCursorAndChangeStreamOperations` since the fields were added/initialized but were not implement for the helpers in `execute`


Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCursorAndChangeStreamOperations.swift
```diff
@@ -3,6 +3,10 @@ import MongoSwiftSync
 import TestsCommon
 
 struct CreateChangeStream: UnifiedOperationProtocol {
+    /// Enables users to specify an arbitrary BSON type to help trace the operation through
```
kmahar, line 6: since we don't actually use this property I'd suggest we just add "comment" to `knownArguments` (with a TODO to remove it as part of SWIFT-1385), but don't bother to store it here or add it to `CodingKeys`. I think that should allow be enough to avoid errors while decoding the file.

once we support it, it will be an option on `ChangeStreamOptions` and stored on that type rather than directly on the operation here.

kmahar (changes requested):
looking good! a few minor comments.

Examples/VaporExample/.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
+{
```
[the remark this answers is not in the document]
rchhaya: yep will do

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -100,6 +100,9 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
     /// A document containing the database and collection names in which this change happened.
     public let ns: MongoNamespace
 
+    /// A document containing the new database and collection names for which the `rename` happened.
```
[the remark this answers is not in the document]
rchhaya: yep I just emulated the `ns` comment, Ill go ahead and rephrase both

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -145,6 +148,16 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
             self.ns = ns
         }
 
+        // `to` only exists in `rename` events so similar implementation as `ns`
```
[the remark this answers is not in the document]
rchhaya: Makes sense. I just collapsed the `to` initialization all into one `if (rename) { _ } else { _ }` instead of doing the `do`/`guard`/`catch` implementation of `ns`.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCursorAndChangeStreamOperations.swift
rchhaya: Removed `comment` from everything except `knownArguments` and added `TODO`

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -97,9 +97,12 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
     public let _id: ResumeToken
 
     // TODO: SWIFT-981: Make this field optional.
-    /// A document containing the database and collection names in which this change happened.
+    /// A   `MongoNamespace` containing the database and collection names in which this change happened.
```
isabelatkinson, line 100:
```suggestion
    /// A  `MongoNamespace` containing the database and collection names in which this change happened.
```

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedCursorAndChangeStreamOperations.swift
```diff
@@ -17,10 +17,9 @@ struct CreateChangeStream: UnifiedOperationProtocol {
     }
 
     static var knownArguments: Set<String> {
-        Set(
-            CodingKeys.allCases.map { $0.rawValue } +
-                ChangeStreamOptions().propertyNames
-        )
+        // TODO: SWIFT-1385 Remove `comment` and integrate into ChangeStreamOptions
+        Set(Set(CodingKeys.allCases.map { $0.rawValue }).union(
```
patrickfreed, line 21: This outermost `Set` isn't required, since the return value of `.union` is a `Set` itself. I think the `"comment"` one can just be `["comment"]` due to type inference also.

patrickfreed (commented):
Looks good! just have one minor nit but otherwise good to merge

PR #762, opened 2022-06-22
SWIFT-1553: Always report `wallTime` in the change stream event output

rchhaya:
Adding a field and impl for optional Date `wallTime`. Incorporated updated spec tests as outlined by `e1a513fe027afb7706485a1e79c53a96acc2ec22`. This involved upgrading the `UnifiedTestRunner` from `maxSchema` 1.5 to 1.7
Made associated changes and bumped version up. Excluded several tests that fail because of an old version of `libmongoc`, potentially resolved with beta vendoring

### DIFF
--- a/Sources/MongoSwift/ChangeStreamEvent.swift
+++ b/Sources/MongoSwift/ChangeStreamEvent.swift
@@ -1,5 +1,6 @@
 import Foundation
 import SwiftBSON
+
 /// An `UpdateDescription` containing fields that will be present in the change stream document for
 /// operations of type `update`.
 public struct UpdateDescription: Codable {
--- a/Sources/MongoSwift/ChangeStreamEvent.swift
+++ b/Sources/MongoSwift/ChangeStreamEvent.swift
@@ -1,3 +1,5 @@
+import Foundation
+import SwiftBSON
 /// An `UpdateDescription` containing fields that will be present in the change stream document for
 /// operations of type `update`.
 public struct UpdateDescription: Codable {
@@ -115,6 +117,10 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
     /// type`update`.
     public let updateDescription: UpdateDescription?
 
+    /// The wall time from the `mongod` that the change event originated from.
+    /// Only present for server versions 6.0 and above
+    public let wallTime: Date?
+
     /**
      * Always present for operations of type `insert` and `replace`. Also present for operations of type `update` if
      * the user has specified `.updateLookup` for the `fullDocument` option in the `ChangeStreamOptions` used to create
@@ -129,7 +135,7 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
     public let fullDocument: T?
 
     private enum CodingKeys: String, CodingKey {
-        case operationType, _id, ns, to, documentKey, updateDescription, fullDocument
+        case operationType, _id, ns, to, documentKey, updateDescription, wallTime, fullDocument
     }
 
     // Custom decode method to work around the fact that `invalidate` events do not have an `ns` field in the raw
@@ -156,6 +162,7 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
         }
 
         self.documentKey = try container.decodeIfPresent(BSONDocument.self, forKey: .documentKey)
+        self.wallTime = try container.decodeIfPresent(Date.self, forKey: .wallTime)
         self.updateDescription = try container.decodeIfPresent(UpdateDescription.self, forKey: .updateDescription)
         self.fullDocument = try container.decodeIfPresent(T.self, forKey: .fullDocument)
     }
--- a/Sources/MongoSwift/ChangeStreamEvent.swift
+++ b/Sources/MongoSwift/ChangeStreamEvent.swift
@@ -118,7 +118,7 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
     public let updateDescription: UpdateDescription?
 
     /// The wall time from the `mongod` that the change event originated from.
-    /// Only present for server versions 6.0 and above
+    /// Only present for server versions 6.0 and above.
     public let wallTime: Date?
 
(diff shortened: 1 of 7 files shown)


rchhaya:
The skipped tests might be able to be re-incorporated once `libmongoc` is vendored in correctly

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -115,6 +117,10 @@ public struct ChangeStreamEvent<T: Codable>: Codable {
     /// type`update`.
     public let updateDescription: UpdateDescription?
 
+    /// The wall time from the `mongod` that the change event originated from.
+    /// Only present for server versions 6.0 and above
```
patrickfreed, line 121:
```suggestion
    /// Only present for server versions 6.0 and above.
```

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
```diff
@@ -309,12 +309,18 @@ enum SpecialOperator {
...
-func matchesEvents(expected: [ExpectedEvent], actual: [CommandEvent], context: Context) throws {
-    guard actual.count == expected.count else {
+func matchesEvents(
+    expected: [ExpectedEvent],
+    actual: [CommandEvent],
+    context: Context,
+    ignoreExtraEvents: Bool
+) throws {
+    // (Actual = expected AND the two must match) or (the two don't have to match)
+    guard (actual.count == expected.count && !ignoreExtraEvents) || ignoreExtraEvents else {
```
patrickfreed, line 319: Even if `ignoreExtraEvents` is true, we'll still want to verify that `actual` has at least as many events as expected. That way we won't panic below when when we try to index them both.

Tests/MongoSwiftSyncTests/UnifiedTestRunner/ExpectedEventsForClient.swift
```diff
@@ -100,5 +112,8 @@ enum ExpectedEvent: Decodable {
 
         /// Specifies whether the serviceId field of the event is set.
         let hasServiceId: Bool?
+
+        /// Specifies whether the `serviceConnectionId` field of an event is set.
+        let hasServerConnectionId: Bool?
```
patrickfreed, line 117: Can we add a TODO comment for SWIFT-1262 to update the runner to use this value? That way we'll remember to update this if necessary, and also if someone comes across this field in the future they'll know why it's just here for deserialization.

Sources/MongoSwift/ChangeStreamEvent.swift
rchhaya: done

Tests/MongoSwiftSyncTests/UnifiedTestRunner/Matching.swift
rchhaya: done by adding an extra `guard` statement to prevent an even longer boolean logical expression

Tests/MongoSwiftSyncTests/UnifiedTestRunner/ExpectedEventsForClient.swift
rchhaya: done

Tests/MongoSwiftSyncTests/UnifiedTestRunner/ExpectedEventsForClient.swift
```diff
@@ -100,5 +112,9 @@ enum ExpectedEvent: Decodable {
 
         /// Specifies whether the serviceId field of the event is set.
         let hasServiceId: Bool?
+
+        /// Specifies whether the `serviceConnectionId` field of an event is set.
+        // TODO: SWIFT-1262: update unified test runner to use this value
```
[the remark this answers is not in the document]
rchhaya: committed

Sources/MongoSwift/ChangeStreamEvent.swift
```diff
@@ -1,3 +1,5 @@
+import Foundation
+import SwiftBSON
```
kmahar, line 2: nit: can we add a newline after the imports? we usually do that

Tests/MongoSwiftSyncTests/UnifiedTestRunner/ExpectedEventsForClient.swift
```diff
@@ -18,8 +18,13 @@ struct ExpectedEventsForClient: Decodable {
     /// (excluding ignored events).
     let events: [ExpectedEvent]
 
+    /// Specifies how the `events` array is matched against observed events.  If false, observed events after all
+    /// specified events have matched MUST cause a test failure; if true, observed events after all specified events
+    /// have been matched MUST NOT cause a test failure. Defaults to false.
+    let ignoreExtraEvents: Bool?
```
kmahar, line 24: since we already fill in a default value of `false` for this in the Decodable implementation below, I think we could just make this non-optional here, which would save us from having to do `?? false` when we pass this value into `matchesEvents`.

kmahar (changes requested):
looks great! two minor comments

Sources/MongoSwift/ChangeStreamEvent.swift
rchhaya: done

Tests/MongoSwiftSyncTests/UnifiedTestRunner/ExpectedEventsForClient.swift
rchhaya: Ohh okay yeah that makes sense. got confused when you said "save us having to do `??` " but that was for the `testFiles` function and I thought u meant for the `init` . Made the edits, hopefully it compiles this time

kmahar (approved):
LGTM assuming the remaining evergreen tasks pass!

PR #766, opened 2022-07-12
SWIFT-1600 Update compile script to ensure tests compile on minimum supported Swift version

rchhaya:


rchhaya:
Most similar thing I could find for why `testIPv4AddressParsing()` fails is because of [this](https://www.swiftbysundell.com/tips/same-name-assignments/)

Tests/MongoSwiftTests/MongoConnectionStringTests.swift
```diff
@@ -595,6 +595,7 @@ final class ConnectionStringTests: MongoSwiftTestCase {
         expect(try connStr.applyOptions(opts)).to(throwError(errorType: MongoError.InvalidArgumentError.self))
     }
 
+#if compiler(>=5.4)
```
kmahar, line 598: I think your suspicion around this being caused by lack of shadowing support sounds right. since the changes to make the test compile should be minimal (use a unique variable name each time instead of reusing `host`) I'd suggest we just do that instead rather than conditionally defining the test.

kmahar (approved):
lgtm! thanks for fixing this up

patrickfreed (approved):

isabelatkinson (approved):

PR #769, opened 2022-07-29
SWIFT-1609 Async-ifying the Unified Test Runner and underlying operations

rchhaya:


rchhaya:
Apologies for a ton of minor commits for compiler directive things - tried switching swiftenv but since I cant run on terminal (weird env problems, can only build via xcode), the version switch didnt actully cascade down. Should be all fixed now + working on resolving an issue with sharding

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -72,12 +86,14 @@ struct UnifiedTestRunner {
             // SERVER-38335.
             do {
                 let opts = RunCommandOptions(readPreference: .primary)
-                _ = try self.internalClient.anyClient.db("admin").runCommand(["killAllSessions": []], options: opts)
+                _ = try await self.internalClient.anyClient.db("admin").runCommand(
+                    ["killAllSessions": []], options: opts
```
kmahar, line 90: nit: I'm not sure why this isn't triggering a linter/formatter issue, but typically if the entire function call doesn't fit on one line, we split the arguments so they are one per line, like:
```swift
_ = try await self.internalClient.anyClient.db("admin").runCommand(
    ["killAllSessions": []],
    options: opts
)
```

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -175,46 +188,41 @@ struct UnifiedTestRunner {
...
                 // each mongos server before running any test that might execute distinct within a transaction. To ease
                 // the implementation, test runners MAY execute distinct before every test.
                 if self.topologyType.isSharded && !MongoSwiftTestCase.serverless {
                     let collEntities = context.entities.values.compactMap { try? $0.asCollection() }
-                    for (_, client) in try self.internalClient.asMongosClients() {
+                    let clientMap = try await self.internalClient.asMongosClients()
+                    for (_, client) in clientMap {
                         for entity in collEntities {
-                            _ = try client.db(entity.namespace.db).runCommand(
+                            _ = try? await client.db(entity.namespace.db).runCommand(
```
kmahar, line 199: it seems like a sign to me that something else is amiss if we need to `try?` here. I don't see why the method being async would affect whether or not the call to distinct here succeeds. could you explain why you added this?

Tests/MongoSwiftTests/UnifiedTestRunner/Context.swift
```diff
@@ -19,10 +21,25 @@ class Context {
...
+
+    /// Executes a closure with the given path element added to the path, removing it after closure completes.
     func withPushedElt<T>(_ elt: String, work: () throws -> T) rethrows -> T {
         self.path.append(elt)
         defer { self.path.removeLast() }
         return try work()
     }
+
+    /// Executes an async closure with the given path element added to the path, removing it after closure completes.
+    func withPushedElt<T>(_ elt: String, work: () async throws -> T) async rethrows -> T {
```
kmahar, line 39: aside: in the future with [`reasync`](https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#reasync), we would only need to have one version of this method -- it would allow us to say that the method is only `async` if the provided closure is `async`. similar to how `rethrows` lets us say it only `throws` if the provided closure `throws`

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -175,46 +188,41 @@ struct UnifiedTestRunner {
...
-                            try operation.executeAndCheckResult(context: context)
+                        try await context.withPushedElt("Operation \(i) (\(operation.name))") {
+                            try await operation.executeAndCheckResult(context: context)
                         }
                     }
-
                     var clientEvents = [String: [CommandEvent]]()
                     // If any event listeners were enabled on any client entities, the test runner MUST now disable
                     // those event listeners.
+                    // try await self.terminateOpenTransactions()
```
kmahar, line 215: leftover from debugging?

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -248,17 +254,52 @@ struct UnifiedTestRunner {
...
                     throw testErr
                 }
             }
         }
+
+        try self.internalClient.closeClients()
+    }
+
+    func closeEntities(context: Context) async throws {
+        // Close streams and cursors
```
kmahar, line 281: I think it would be helpful to future expand this comment to clarify that cursors and change streams must be killed before their parent clients, hence the separate loops

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -248,17 +254,52 @@ struct UnifiedTestRunner {
...
+                    try await self.terminateOpenTransactions()
+                    await context.disableFailpoints()
+                    try await self.closeEntities(context: context)
                 } catch let testErr {
                     // Test runners SHOULD terminate all open transactions after each failed test by killing all
                     // sessions in the cluster.
                     do {
-                        try self.terminateOpenTransactions()
+                        // Clean up to prevent a client closing error from obscuring the real error
+                        try? await self.terminateOpenTransactions() // bug
```
kmahar, line 266: what does `// bug` mean, and why do we `try?` here?

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -156,17 +169,17 @@ struct UnifiedTestRunner {
...
-                            _ = try db.createCollection(
+                            _ = try await db.createCollection(
                                 collData.collectionName,
                                 options: CreateCollectionOptions(writeConcern: .majority)
                             )
                             continue
                         }
 
-                        try coll.insertMany(collData.documents)
+                        _ = try await coll.insertMany(collData.documents)
```
kmahar, line 182: this method is marked `@discardableResult`, so I don't think you would get a compiler warning if you remove the `_ =` part and ignore the `InsertManyResult`

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -156,17 +169,17 @@ struct UnifiedTestRunner {
                         let db = self.internalClient.anyClient.db(collData.databaseName)
                         let collOpts = MongoCollectionOptions(writeConcern: .majority)
                         let coll = db.collection(collData.collectionName, options: collOpts)
-                        try coll.drop()
+                        _ = try await coll.drop()
```
kmahar, line 172: since this method doesn't actually return anything I don't think you need the `_ =`

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -22,14 +25,25 @@ struct UnifiedTestRunner {
...
+        func asMongosClients() async throws -> [ServerAddress: MongoClient] {
             guard case let .mongosClients(clientMap) = self else {
                 throw TestError(
                     message: "Runner unexpectedly did not create per-mongos clients"
                 )
             }
             return clientMap
         }
+
+        func closeClients() throws {
```
kmahar, line 37: nit: can we just name this `close()`? this type is meant to abstract over the number of clients so it won't always be plural clients.
also, can we add a docstring please, since all the other methods and properties on this types have one

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -1,8 +1,11 @@
-import MongoSwiftSync
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+import MongoSwift
 import Nimble
+import NIOPosix
```
kmahar, line 4: do we need the `NIOPosix` import?

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -22,14 +25,25 @@ struct UnifiedTestRunner {
         }
 
         /// If the internal client is a map of per-mongos clients, returns that map; otherwise throws an erorr.
-        func asMongosClients() throws -> [ServerAddress: MongoClient] {
+        func asMongosClients() async throws -> [ServerAddress: MongoClient] {
```
kmahar, line 28: I don't think this needs to be `async`, it's not calling any other async code

Tests/MongoSwiftTests/UnifiedTestRunner/Context.swift
```diff
@@ -19,10 +21,25 @@ class Context {
         self.internalClient = internalClient
     }
 
-    /// Executes a closure with the given path element added to the path, removing it after the closure is complete.
+    /// Method to manually close failpoints given the new `async` implementation
+    func disableFailpoints() async {
+        for failpointGuard in self.enabledFailPoints {
+            await failpointGuard.failPoint.disable(using: self.internalClient.anyClient)
```
kmahar, line 27: since we aren't really using a failpoint guard for its intended purpose anymore (i.e. disabling a fail point upon deinit), I would suggest we introduce a new async-specific type `EnabledFailPoint` which stores a failpoint + client used to enable it, and switch the `FailPoint` and `TargetedFailPoint` operations to use that. you could put a `disable` method on the type that disables the failpoint using the stored client, and then here call through to that method for each failpoint.

it's important that we disable failpoints with the clients used to create them, because otherwise, in the case of sharded clusters, the command to disable a failpoint will not necessarily execute on the same mongos as it was enabled on, which means it won't actually get disabled at all, which can pollute future test runs.

rchhaya: Makes sense bc I rmbr that using `anyClient` seemed questionable at best. Made the `EnabledFailpoint` class for async-specific operations and wrote a disable method on it that is called here.

.gitignore
```diff
@@ -3,3 +3,4 @@
 Package.resolved
 **/xcuserdata/
 .DS_Store
+.swift-version
```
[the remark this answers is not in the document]
rchhaya: Sgtm, deleted

Tests/MongoSwiftTests/DNSSeedlistTests.swift
```diff
@@ -33,10 +33,6 @@ struct DNSSeedlistTestCase: Decodable {
 }
 
 final class DNSSeedlistTests: MongoSwiftTestCase {
-    override func setUp() {
```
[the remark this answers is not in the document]
kmahar: to take a step back: I think a general principle we try to adhere to in our work is keeping the code changes in a pull request scoped to the task at hand. this makes it easier to track the history of code over time and reason about why particular changes were made, determine why a test started failing or became flaky at some point, etc. it also makes code review easier since there is less to review and it is easier for reviewers to understand the changes. 

this file contains tests that do not use the unified test runner code in any way, so I do not think making any edits to this file would be considered in scope for this ticket.

from a technical perspective, this change _may_ be fine to make, but I'd rather avoid including unnecessary changes for the reasons I note above. you can see what the property does here: https://developer.apple.com/documentation/xctest/xctestcase/1496260-continueafterfailure. in the past we would use `continueAfterFailure` in tests where it was likely that if one assertion failed every assertion following it would also fail. we decided to stop relying on this in new tests at some point because it didn't work properly on Linux, which is why you likely haven't seen it in other files. I'm not sure if that is still the case re Linux, but at least on macOS it provides some value in cases where a test newly starts failing. in cases where tests are passing you wouldn't see any difference from this being present, but for a failing test it can affect the outcome.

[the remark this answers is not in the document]
rchhaya: Oh yep I think my incorrect assumptions about the purpose of the method led me to believe that it was in scope with unified test runner as part of the conversions. Makes sense that we shouldnt alter this if it isnt actually in scope and will revert this change here. Also thanks for the link, the info helps contextualize a bit.  

All changes related to `setUp()` have been reverted. Heres the other places where the diff takes place and a quick summary of them

`CommandMonitoringTests` - only 1 test method that uses the unified runner. 

`MongoCursorTests` - same situation as DNSSeedlist where no unified tests are run.

`ReadPreferenceTests` - same as above.

`UnifiedTests` - uses the unified runner but also has normal prose tests.

Maybe we can open a ticket to re-assess the need for this method (esp in a file in `CommandMonitoringTests.swift`)?

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -248,17 +254,52 @@ struct UnifiedTestRunner {
...
                 }
             }
         }
+
+        try self.internalClient.close()
+    }
+
+    func closeEntities(context: Context) async throws {
+        // Must close streams and cursors before closing the parent client
+        for entity in context.entities.values {
```
kmahar, line 282: we actually probably should close any cursor entities here too, I think. it might just happen that all tests right now iterate cursors to completion (which leads to them being closed) but that might not always be the case for future tests.

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -248,17 +254,52 @@ struct UnifiedTestRunner {
...
+                        try await closeEntities(context: context)
                     } catch {
                         print("Failed to terminate open transactions: \(error)")
                     }
                     throw testErr
                 }
             }
         }
+
+        try self.internalClient.close()
```
kmahar, line 277: it seems like in any cases where we hit an unexpected error executing the tests, we'll fail to execute this and end up getting the preconditionFailures about not closing clients. 
if we add a `deinit` to the unified runner (like I suggested in my other comment about a shared eventloopgroup) we could move this call to there (we'd have to try/catch or try? but I think that's ok)

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
kmahar: hmmm, I'd still like to understand why this only shows up now and never came up when the runner was sync. it might point to some subtle behavioral difference we're missing that results from this refactor. I'd think that we should be doing the same operation in the same order we were before. could you maybe link me to an example evergreen run where this happens so I can see the output?

kmahar (commented):
thanks for addressing my comments! I have a few more mostly minor ones.

Tests/MongoSwiftTests/Failpoint.swift
```diff
@@ -0,0 +1,49 @@
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+import MongoSwift
+import NIOPosix
+import TestsCommon
+
+@available(macOS 10.15, *)
+/// Convenience class which wraps a `FailPoint` and its underlying client to handle closing.
```
kmahar, line 7:
```suggestion
/// Convenience type which wraps a `FailPoint` and its corresponding client to handle closing.
```

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -248,11 +262,18 @@ struct UnifiedTestRunner {
...
+                    try await self.closeEntities(context: context)
                 } catch let testErr {
                     // Test runners SHOULD terminate all open transactions after each failed test by killing all
                     // sessions in the cluster.
                     do {
-                        try self.terminateOpenTransactions()
+                        // Clean up to prevent a client closing error from obscuring the real error
+                        try? await self.terminateOpenTransactions()
+                        await context.disableFailpoints()
+                        try await closeEntities(context: context)
```
kmahar, line 276:
```suggestion
                        try await self.closeEntities(context: context)
```

kmahar (commented):
couple suggestions; otherwise LGTM mod ongoing thread about needing to `try?` the distinct workaround

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
patrickfreed: If you go back to regular `try` do you still see the error locally and/or on Evergreen? If so, which tests trigger it?

It's possible you were only seeing this locally because you were running a subset of tests that didn't happen to create the database used in the failing test. Or rather, running all the tests only happened to work with `try` previously because the database in question happened to be created in an earlier test.

If this is in fact what is going on, then I think using `try?` here might make sense, because if the database doesn't exist then there's nothing to do. However, we do need to verify that nothing else is amiss here before committing to that.

rchhaya: Replacing with `try` locally and testing on `UnifiedTests` (the problematic file I was dealing with earlier) results in an error for `Unsupported command` as part of `valid-fail`. (running a replica set sharded cluster). The msg is "NameSpaceNotFound: database operation-failure not found"

On Evergreen, the same error occurs with the same test on the sharded clusters only.

It seems like since the errors occur on `valid-fail` which just tests using random operations/entities to ensure they fail that `try?` should be okay here?

Tests/MongoSwiftTests/Failpoint.swift
```diff
@@ -0,0 +1,49 @@
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+import MongoSwift
+import NIOPosix
```
kmahar, line 3: I think we can get rid of the NIOPosix import

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -175,46 +197,47 @@ struct UnifiedTestRunner {
...
-                            _ = try client.db(entity.namespace.db).runCommand(
-                                ["distinct": .string(entity.name), "key": "_id"]
-                            )
+                            do {
+                                _ = try await client.db(entity.namespace.db).runCommand(
+                                    ["distinct": .string(entity.name), "key": "_id"]
+                                )
+                            } catch {
+                                // DB not explicitly initially created for operation-failure test bc no initialData
+                                if !(error is MongoError.CommandError) {
```
kmahar, line 214: we can make this a little more specific actually, so that we don't ignore arbitrary command errors. the [error code](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L63) for `NamespaceNotFound` is 26. 

let's do something like the logic here so that we only catch and ignore command errors with code 26:
https://github.com/mongodb/mongo-swift-driver/blob/2d579e2002e2c80aab8fcdc2159d1ffad79f6147/Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift#L76

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -175,46 +197,47 @@ struct UnifiedTestRunner {
...
                         for entity in collEntities {
-                            _ = try client.db(entity.namespace.db).runCommand(
-                                ["distinct": .string(entity.name), "key": "_id"]
-                            )
+                            do {
+                                _ = try await client.db(entity.namespace.db).runCommand(
+                                    ["distinct": .string(entity.name), "key": "_id"]
+                                )
+                            } catch {
+                                // DB not explicitly initially created for operation-failure test bc no initialData
```
kmahar, line 213: can we make this comment a little mores pecific? something like "this command can fail for tests that have no `initialData` specified and so do not create the namespace, such as the `valid-fail/operation-failure.json` test"

kmahar (approved):

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -248,11 +266,18 @@ struct UnifiedTestRunner {
...
                     // Test runners SHOULD terminate all open transactions after each failed test by killing all
                     // sessions in the cluster.
                     do {
-                        try self.terminateOpenTransactions()
+                        // Clean up to prevent a client closing error from obscuring the real error
+                        try? await self.terminateOpenTransactions()
+                        await context.disableFailpoints()
+                        try await self.closeEntities(context: context)
                     } catch {
                         print("Failed to terminate open transactions: \(error)")
```
patrickfreed, line 282: This print statement seems outdated now, since we run `terminateOpenTransactions` with `try?`.

patrickfreed (commented):
Looks really good! Just have a few minor comments

Tests/MongoSwiftTests/UnifiedTestRunner/EntityDescription.swift
```diff
@@ -147,7 +150,8 @@ struct UnifiedTestClient {
             opts.minHeartbeatFrequencyMS = 50
             opts.heartbeatFrequencyMS = 50
         }
-        self.client = try MongoClient.makeTestClient(connStr, options: opts)
+        self.elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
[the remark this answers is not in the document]
rchhaya: Yep it was originally intentional to coordinate closing properly, but for this case I realize that we can borrow the eventloopgroup from the `UnifiedTestRunner` and pass it in. This ensures all the unified test runner test clients are on the same runtime

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
rchhaya: updated to be `print("Failed to close unified test runner elements: \(error)")`

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
isabelatkinson: doesn't look like this actually got removed

isabelatkinson (approved):
lgtm, just one nit

Tests/MongoSwiftTests/UnifiedTestRunner/UnifiedTestRunner.swift
patrickfreed: This is actually needed in order to use `MultiThreadedEventLoopGroup` in this file.

PR #773, opened 2022-08-29
SWIFT-1337, SWIFT-1629, SWIFT-1557, SWIFT-1566: Various maintenance-related changes 

kmahar:
This is a collection of several minor changes; my original motivation here was dropping Swift 5.1 support and only depending on the necessary NIO modules, but in the process I ended up picking up some various greener build and Evergreen maintenance type changes that we've fallen behind on in recent months.

Included changes:
* SWIFT-1337: Drop Swift 5.1 support
* SWIFT-1629: Only depend on necessary NIO modules in the driver
* SWIFT-1557: Test against MongoDB 6.0 in Evergreen 
* SWIFT-1566: Test MongoDB 5.0 + load balancer support
* Test against latest MongoDB rapid release in Evergreen
* Fix broken testing against development snapshots by updating the Python script to dynamically get the latest dev snapshot from GitHub, similar to how it currently queries for patch releases. (I will port this to BSON after)
* bump SwiftLint and SwiftFormat versions (installed new versions on my new laptop that found some new violations)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,7 +11,7 @@ let package = Package(
         .library(name: "MongoSwiftSync", targets: ["MongoSwiftSync"])
     ],
     dependencies: [
-        .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "8.0.0")),
+        .package(url: "https://github.com/Quick/Nimble", .upToNextMajor(from: "8.0.0")),
         .package(url: "https://github.com/apple/swift-nio", .upToNextMajor(from: "2.36.0")),
         .package(url: "https://github.com/mongodb/swift-bson", .upToNextMajor(from: "3.0.0"))
     ],
@@ -57,6 +57,7 @@ let package = Package(
                 "MongoSwiftSync",
                 "TestsCommon",
                 "MongoSwift",
+                .product(name: "NIOConcurrencyHelpers", package: "swift-nio"),
                 .product(name: "Nimble", package: "Nimble")
             ]
         ),
--- a/Package.swift
+++ b/Package.swift
@@ -1,13 +1,4 @@
-// swift-tools-version:5.1
-
-///  To avoid breaking Swift 5.1 compatibility, only require newer NIO versions when we need NIO's concurrency APIs.
-func getMinNIOVersion() -> PackageDescription.Package.Dependency.Requirement {
-#if compiler(>=5.5.2) && canImport(_Concurrency)
-    return .upToNextMajor(from: "2.36.0")
-#else
-    return .upToNextMajor(from: "2.15.0")
-#endif
-}
+// swift-tools-version:5.2
 
 import PackageDescription
 let package = Package(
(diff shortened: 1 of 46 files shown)


Package.swift
```diff
@@ -1,13 +1,4 @@
-// swift-tools-version:5.1
-
-///  To avoid breaking Swift 5.1 compatibility, only require newer NIO versions when we need NIO's concurrency APIs.
-func getMinNIOVersion() -> PackageDescription.Package.Dependency.Requirement {
-#if compiler(>=5.5.2) && canImport(_Concurrency)
-    return .upToNextMajor(from: "2.36.0")
-#else
-    return .upToNextMajor(from: "2.15.0")
```
kmahar, line 8: I could have left this in and updated this to be NIO 2.32 (the minimum version where the separate modules exist), but there didn't seem to be a huge benefit to doing so as 2.36.0 supports 5.2+, so I thought it was simpler and reasonable to remove the extra complexity here and require everyone to be on 2.36+.

Sources/MongoSwift/MongoConnectionString.swift
```diff
@@ -467,12 +467,10 @@ public struct MongoConnectionString: Codable, LosslessStringConvertible {
         // Parse the defaultAuthDB if present.
         if userHostsAndAuthDB.count == 2 && !userHostsAndAuthDB[1].isEmpty {
             let defaultAuthDB = try userHostsAndAuthDB[1].getPercentDecoded(forKey: "defaultAuthDB")
-            for character in Self.forbiddenDBCharacters {
-                if defaultAuthDB.contains(character) {
-                    throw MongoError.InvalidArgumentError(
-                        message: "defaultAuthDB contains invalid character: \(character)"
-                    )
-                }
+            for character in Self.forbiddenDBCharacters where defaultAuthDB.contains(character) {
```
kmahar, line 470: this and the other change in the file were linter changes due to the version updates. there's a swiftlint rule around preferring "for... where" over a "for" loop with a single nested "if", and I guess swiftlint previously missed these violations

Tests/MongoSwiftSyncTests/UnifiedTestRunner/UnifiedTestRunner.swift
```diff
@@ -183,9 +183,13 @@ struct UnifiedTestRunner {
...
                     for (_, client) in try self.internalClient.asMongosClients() {
                         for entity in collEntities {
-                            _ = try client.db(entity.namespace.db).runCommand(
-                                ["distinct": .string(entity.name), "key": "_id"]
-                            )
+                            do {
+                                _ = try client.db(entity.namespace.db).runCommand(
+                                    ["distinct": .string(entity.name), "key": "_id"]
+                                )
+                            } catch let commandError as MongoError.CommandError where commandError.code == 26 {
```
kmahar, line 190: I noticed that after I synced in the unified change stream spec tests, they started to fail on macOS + sharded clusters, with a similar NamespaceNotFound error problem to that Rohan was encountering. 
I think this was likely also an issue around the exact namespaces that exist / don't exist at the time those tests are run due to test ordering

Tests/MongoSwiftTests/APMTests.swift
```diff
@@ -177,10 +177,8 @@ final class APMTests: MongoSwiftTestCase {
             let sdamTask = Task { () -> [EventType] in
                 var eventTypeSdam: [EventType] = []
                 taskStarted.store(true)
-                for try await event in client.sdamEventStream() {
-                    if !event.isHeartbeatEvent {
-                        eventTypeSdam.append(event.type)
-                    }
+                for try await event in client.sdamEventStream() where !event.isHeartbeatEvent {
```
kmahar, line 180: another linter change

Tests/MongoSwiftTests/EventLoopBoundMongoClientTests.swift
```diff
@@ -320,7 +320,7 @@ final class EventLoopBoundMongoClientTests: MongoSwiftTestCase {
             defer { try? db.drop().wait() }
 
             let wrongEventLoop = MultiThreadedEventLoopGroup(numberOfThreads: 4).next()
-            let withSessionFuture: EventLoopFuture<[String]> = elBoundClient.withSession { session in
+            let withSessionFuture = elBoundClient.withSession { session in
```
kmahar, line 323: the explicit annotation is not needed anymore, and that saves us importing NIOCore here

patrickfreed (approved):
Looks great! awesome job knocking all these different things out

PR #774, opened 2022-09-06
SWIFT-1630 Reduce flakiness of CommandStreamBufferPolicy test

patrickfreed:
SWIFT-1630

This PR doubles the time the test waits for a batch of 60 inserts to complete from 5 seconds to 10 seconds after seeing some flakiness. I will say it is a bit concerning that we can't execute that many in the 5 seconds time, but I think that might have more to do with the hosts at our disposal rather than the driver.


patrickfreed:
Looks like all the timeout failures are related to the zombie xcrun processes.

kmahar (approved):

PR #775, opened 2022-09-12
SWIFT-1610, SWIFT-1378, SWIFT-1632: Spec tests sync + expose clusterTime on ChangeStreamEvent

kmahar:
I picked up SWIFT-1610 which was a test sync; the sync also brought in the files from SWIFT-1378.

I also picked up SWIFT-1632 which I thought was just a test sync but it turns out we didn't have a `clusterTime` on `ChangeStreamEvent`. It seems like we decided against that in SWIFT-428 but now that there is a spec test it seems like we'd want to have one for compliance and parity with other drivers.

### DIFF
--- a/Sources/TestsCommon/APMUtils.swift
+++ b/Sources/TestsCommon/APMUtils.swift
@@ -82,7 +82,7 @@ public class TestCommandMonitor: CommandEventHandler {
 
 public enum EventType: String, Decodable {
     case commandStartedEvent, commandSucceededEvent, commandFailedEvent, connectionCreatedEvent, connectionReadyEvent,
-        connectionClosedEvent, connectionCheckedInEvent, connectionCheckOutStartedEvent, connectionCheckedOutEvent,
+         connectionClosedEvent, connectionCheckedInEvent, connectionCheckOutStartedEvent, connectionCheckedOutEvent,
          connectionCheckOutFailedEvent, poolCreatedEvent, poolReadyEvent, poolClearedEvent, poolClosedEvent,
          topologyDescriptionChanged, topologyOpening, topologyClosed, serverDescriptionChanged, serverOpening,
          serverClosed, serverHeartbeatStarted, serverHeartbeatSucceeded, serverHeartbeatFailed
--- a/Sources/TestsCommon/APMUtils.swift
+++ b/Sources/TestsCommon/APMUtils.swift
@@ -81,13 +81,11 @@ public class TestCommandMonitor: CommandEventHandler {
 }
 
 public enum EventType: String, Decodable {
-    case commandStartedEvent, commandSucceededEvent, commandFailedEvent,
-         connectionCreatedEvent, connectionReadyEvent, connectionClosedEvent,
-         connectionCheckedInEvent, connectionCheckedOutEvent, connectionCheckOutFailedEvent,
-         poolCreatedEvent, poolReadyEvent, poolClearedEvent, poolClosedEvent,
-         topologyDescriptionChanged, topologyOpening, topologyClosed, serverDescriptionChanged,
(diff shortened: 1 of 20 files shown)


Sources/TestsCommon/APMUtils.swift
```diff
@@ -81,13 +81,11 @@ public class TestCommandMonitor: CommandEventHandler {
...
 public enum EventType: String, Decodable {
-    case commandStartedEvent, commandSucceededEvent, commandFailedEvent,
-         connectionCreatedEvent, connectionReadyEvent, connectionClosedEvent,
-         connectionCheckedInEvent, connectionCheckedOutEvent, connectionCheckOutFailedEvent,
-         poolCreatedEvent, poolReadyEvent, poolClearedEvent, poolClosedEvent,
-         topologyDescriptionChanged, topologyOpening, topologyClosed, serverDescriptionChanged,
-         serverOpening, serverClosed, serverHeartbeatStarted, serverHeartbeatSucceeded,
-         serverHeartbeatFailed
+    case commandStartedEvent, commandSucceededEvent, commandFailedEvent, connectionCreatedEvent, connectionReadyEvent,
+         connectionClosedEvent, connectionCheckedInEvent, connectionCheckOutStartedEvent, connectionCheckedOutEvent,
```
kmahar, line 85: only actual change here mod reorganizing lines was adding `connectionCheckOutStartedEvent` which one of the tests I synced uses

Tests/MongoSwiftTests/AsyncAwaitTestUtils.swift
```diff
@@ -165,9 +165,9 @@ extension MongoClient {
     }
 
     internal func getUnmetRequirement(_ testRequirement: TestRequirement) async throws -> UnmetRequirement? {
-        async let topologyType = try self.topologyType()
-        async let serverVersion = try self.serverVersion()
-        async let params = try self.serverParameters()
+        async let topologyType = try await self.topologyType()
```
kmahar, line 168: happened to notice while adding new test using this that this method was actually still calling the blocking methods that `wait()` to get these values. oops

Tests/MongoSwiftTests/AsyncAwaitTests.swift
```diff
@@ -234,7 +234,7 @@ final class MongoCursorAsyncAwaitTests: MongoSwiftTestCase {
         }
 
         testAsync {
-            let opts = CreateCollectionOptions(capped: true, size: 5)
+            let opts = CreateCollectionOptions(capped: true, max: 5, size: 100_000)
```
kmahar, line 237: this is unrelated but this test starting failing against latest once [SERVER-67246](https://jira.mongodb.org/browse/SERVER-67246) went in. `size` is actually supposed to be the max size in bytes, not the number of documents (that's what `max` is for). 
previously the server forced you to have a minimum max size of 4096, so the `size` option here was basically ignored. but that restriction was removed and it is now valid to actually have a 5 byte capped collection. due to this, no documents could actually fit in the collection so this test started to fail.

isabelatkinson: interesting, i'm surprised the previously incorrect value was silently ignored by the server

isabelatkinson (approved):

Tests/MongoSwiftTests/AsyncAwaitTests.swift
kmahar: yeah it's interesting; per the docs [here](https://www.mongodb.com/docs/manual/core/capped-collections/#create-a-capped-collection):

> If the size field is less than or equal to 4096, then the collection will have a cap of 4096 bytes.

patrickfreed (approved):
LGTM! Sorry for the delayed review

PR #776, opened 2022-10-19
clarify in docstrings what happens if you create a collection with the same name

kmahar:
per forums conversation

### DIFF
--- a/Sources/MongoSwift/AsyncAwait/MongoDatabase+AsyncAwait.swift
+++ b/Sources/MongoSwift/AsyncAwait/MongoDatabase+AsyncAwait.swift
@@ -27,6 +27,8 @@ extension MongoDatabase {
      *
      * - Throws:
      *   - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *     If a collection with this name already exists, a `MongoError.CommandError` with code `48`
+     *     will be thrown.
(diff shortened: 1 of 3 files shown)


Sources/MongoSwift/AsyncAwait/MongoDatabase+AsyncAwait.swift
```diff
@@ -27,6 +27,8 @@ extension MongoDatabase {
      *
      * - Throws:
      *   - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *     If a collection with this name already exists, a `MongoError.CommandError` with code `48`
+     *     will be thrown.
```
patrickfreed, line 31: I think it might be good to include the `NamespaceExists` codename in the docstrings too.

Sources/MongoSwift/AsyncAwait/MongoDatabase+AsyncAwait.swift
```diff
@@ -27,6 +27,8 @@ extension MongoDatabase {
      *
      * - Throws:
      *   - `MongoError.CommandError` if an error occurs that prevents the command from executing.
+     *     If a collection with this name already exists, a `MongoError.CommandError` with code `48`
```
patrickfreed, line 30: Unfortunately, due to SERVER-64416, this isn't actually always the case. Not sure if that's worth calling out here or not.

